PR #1, opened 2019-08-12
Initial example for enforcing data submission

nickhurlburt:
This is an initial, mostly clean example of how to enforce data submission to the HRM ("Helpline Relationship Management" system) before closing a task.  The plugin doesn't currently send any data anywhere.

Note: this was converted from flex plugin builder v2 to v3, and there seem to be some issues with running it locally.  When running locally, it errors out on `<!--doctype html>` in the built plugin.  This does not happen when running `npm run deploy` and sending straight to production.  I suspect there's something with the manual upgrade from v2 to v3 that didn't go well.  It may require creating a new, native v3 plugin and copying the code over.

This has no unit testing, but has been tested manually in 'production'.


nickhurlburt:
Also note, this is officially called 'plugin-enforce-order' because there was a problem with plugin names > 32 characters.  This is probably fixed now.  May be another reason to start fresh later.

PR #15, opened 2019-11-20
Add task info to HRM; recent contacts filters by queue

nickhurlburt:
Includes task info such as phone number and queue name in the form data recorded in the HRM, primarily to be able to return it to the recent contacts view.  Some debt-inducing shortcuts like not cloning the form data object before modifying it.


nickhurlburt:
Also, there has to be a better way to be defensive in accessing form data than expressions like `element.FormData && element.FormData.callerInformation && element.FormData.callerInformation.name`.  Yuck.

PR #37, opened 2020-02-20
Set up eslint-config-twilio-react

murilovmachado:
This PR:
- Sets up `eslint-config-twilio-react`
- Fixes all issues in the current code base
- Implements `propTypes` for all components
  - Please take a look at `types/index.js`

The current lint config disables the following rules:
- `no-console`: we use `console.log()` in our code
- `no-warning-comments`: we use `// TODO` in our code
- `no-confusing-arrow`: this rule has some issues with `prettier` rules
- `no-alert`: we use `alert()` in our code
- `import/no-extraneous-dependencies`: we have some dependencies injected by Twilio Flex

The above rules can be enabled later on if needed.


plugin-hrm-form/src/components/TabbedForms.js
```diff
@@ -1,46 +1,53 @@
...
-import FieldText from './FieldText'
-
+import FieldText from './FieldText';
+import BranchingFormIssueCategory from './BranchingFormIssueCategory';
+import { formType, taskType } from '../types';
 
 class TabbedForms extends React.PureComponent {
+  static displayName = 'TabbedForms';
+
+  static propTypes = {
```
nickhurlburt: Is there a rule around propTypes being within the class vs after it as `class.propTypes = { } `?

plugin-hrm-form/src/components/TabbedForms.js
```diff
@@ -1,46 +1,53 @@
...
 
-  curriedHandleChange = (parents, name) =>
-  (e) => this.props.handleChange(
-    this.props.task.taskSid,
-    parents,
-    name,
-    e.target.value || e.currentTarget.value
-  );
+  curriedHandleChange = (parents, name) => e =>
+    this.props.handleChange(this.props.task.taskSid, parents, name, e.target.value || e.currentTarget.value);
```
nickhurlburt: Do you know what the maximum line length for the linter is?

plugin-hrm-form/src/states/ContactState.js
```diff
@@ -1,45 +1,62 @@
...
-  // static handleCallTypeButtonClick = (taskId, value, e) => ({type: HANDLE_CALLTYPE_BUTTON_CLICK, taskId: taskId, value: value});
-  static handleCallTypeButtonClick = (taskId, value, e) => ({type: HANDLE_CHANGE, name: 'callType', taskId: taskId, value: value, parents: []});
-  static initializeContactState = (taskId) => ({type: INITIALIZE_CONTACT_STATE, taskId: taskId});
+  static handleChange = (taskId, parents, name, value) => ({ type: HANDLE_CHANGE, name, taskId, value, parents });
+
+  /*
+   * There has to be a better way to do this rather than a one-off, but MUI does not make it easy
+   * static handleCallTypeButtonClick = (taskId, value, e) => ({type: HANDLE_CALLTYPE_BUTTON_CLICK, taskId: taskId, value: value});
+   */
+  static handleCallTypeButtonClick = (taskId, value, e) => ({
```
nickhurlburt: Why did it list each property in the object on its own line here, but above put it all on the same line?

plugin-hrm-form/src/states/ContactState.js
```diff
@@ -61,75 +78,74 @@ export function reduce(state = initialState, action) {
...
       return {
         ...state,
         tasks: {
           ...state.tasks,
-          [action.taskId]: editNestedField(currentForm,
-                                           action.parents,
-                                           action.name,
-                                           {touched: true})
-        }
+          [action.taskId]: editNestedField(currentForm, action.parents, action.name, { touched: true }),
```
nickhurlburt: Yeah, what are the rules about putting arguments on one line vs separate lines?

plugin-hrm-form/src/types/index.js
```diff
@@ -0,0 +1,88 @@
+import PropTypes from 'prop-types';
+
+export const taskType = PropTypes.shape({
+  taskSid: PropTypes.string,
+});
+
+export const fieldType = PropTypes.shape({
+  value: PropTypes.string,
+  error: PropTypes.string,
+  validation: PropTypes.string,
```
nickhurlburt: The fields in redux can also include `touched`.  But perhaps that does not actually get passed as a prop.

plugin-hrm-form/src/components/TabbedForms.js
murilovmachado: Yes, when defining the propTypes for a class, the linter requires that it should be a `static` property. When it's a functional component, the linter requires the `ComponentName.propTypes = {}` format.

murilovmachado: Later on, I'd like to get rid of all class components and make them all functional components.

plugin-hrm-form/src/components/TabbedForms.js
murilovmachado: It seems it's 120 by default, but we can set it to any other value.

plugin-hrm-form/src/states/ContactState.js
murilovmachado: When it fits the max line length, it prefers to have it all in one line. Otherwise, it splits.

plugin-hrm-form/src/states/ContactState.js
murilovmachado: It's the same rule as the other question: when it fits the max line length, it prefers to have it all in one line. Otherwise, it splits.

PR #38, opened 2020-02-20
Setup CI for plugin-hrm-form

murilovmachado:
This PR:
- Sets up CI with the following steps:
  - Install Dependencies
  - Run Lint
  - Run Tests
- Update tests snapshots
- Remove unnecessary dependencies (redux and react-redux)
- Updates some other dependencies to fix all vulnerabilities


.github/workflows/plugin-hrm-form-ci.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - name: Checkout Branch
+      uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: Create Temp Files
+      run: |
+        touch ./public/appConfig.js
```
nickhurlburt: Do appConfig.js and secret.js get any content inside them?  I guess that may not be needed.

plugin-hrm-form/.gitignore
```diff
@@ -61,6 +61,7 @@ typings/
 .next
 
 # Flex related ignore
+pluginsService.js
```
nickhurlburt: What is this?

plugin-hrm-form/package-lock.json
```diff
@@ -5,27 +5,38 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
-      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
+      "version": "7.8.3",
```
nickhurlburt: I'm curious what prompted such a large amount of change to `package-lock.json`?

plugin-hrm-form/package.json
```diff
@@ -10,7 +10,7 @@
     "build": "flex-plugin build",
     "predeploy": "npm run build",
     "deploy": "flex-plugin deploy",
-    "test": "flex-plugin test --env=jsdom",
+    "test": "craco test --env=jsdom",
```
nickhurlburt: A note to capture for history:  we needed to change this to `craco` because if `craco` returns a non-zero exit code, `flex-plugin` will not pass it through.

.github/workflows/plugin-hrm-form-ci.yml
murilovmachado: No content, just an empty file.

plugin-hrm-form/.gitignore
murilovmachado: after updating the dependencies, flex-plugin is generating this file automatically. Since we won't change it manually, it's better to have it git ignored.

plugin-hrm-form/package-lock.json
nickhurlburt: Offline discussion -- this removed some vulnerabilities and also removing `react-redux` and `redux` from `package.json` caused a lot of change.

PR #42, opened 2020-02-28
(No Review) CI working with flex-plugin

nickhurlburt:
Uses `flex-plugin` for testing rather than going direct to `craco`.

Previously discussed so moving forward without review.


murilovmachado:
@nickhurlburt, as discussed on Slack, when we started working on this `flex-plugin`, was not working properly, so we changed to `craco`. Only after that, we updated the libs versions. I confused myself thinking I had the updated versions previously, but commit history says otherwise. So it makes sense, that `flex-plugin` is now working properly.

PR #43, opened 2020-02-28
(WIP) Add workerSid to Contact record in HRM

nickhurlburt:
In order to include the counselor in search, we need to record the counselor in the HRM when we submit a new contact record.  This WIP PR shows a possible way to do that on the Flex side.  I re-organized code in `HrmFormPlugin.js`, but the most relevant lines are 53-65, especially `const workerSid = manager.workerClient.sid;`, grabbing the current worker/counselor's ID to save it.

This raises a few questions:
1. How do we want to connect a worker/counselor with a contact record?  Is the ID the right thing?  Do we want to also store the person's name, even if the name changes later?
2. How would we get the list of names in the select box when we search by counselor?
3. How do we rightly filter search by helpline?

Connecting Twilio-based ID with our own system will continue to require thought.

FYI @murilovmachado

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -15,47 +15,21 @@ export default class HrmFormPlugin extends FlexPlugin {
   }
 
   /**
-   * This code is run when your plugin is being started
-   * Use this to modify any UI components or attach to the actions framework
+   * This code is run when the plugin is being started
+   * It is ordered into three sections:
+   *  1. Initial startup
+   *  2. Setting up task lifecycle-related items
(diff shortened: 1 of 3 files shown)


murilovmachado:
1. I believe we should store the counselor's id only. When fetching the contacts we can populate the counselor's name. This way, the backend will be always consistent, and the front-end will receive all the information it needs. We can always revisit this if we think performance may be an issue.
2. We can have an API call to populate it, something like `GET /counselors`, that would bring a list of `id` and `name`.
3. Not sure about this one yet.

@nickhurlburt

nickhurlburt:
This is made obsolete by #44.  Closing.

PR #46, opened 2020-03-05
Revert "Send a message to the caller when the counselor has ended the chat"

nickhurlburt:
Reverts tech-matters/flex-plugins#40

FYI @murilovmachado


nickhurlburt:
This is being removed so that we can get only non-customer facing changes on the branch, preparing for a production deploy of non-customer facing changes only.  It will be reinstated later.

PR #49, opened 2020-03-07
[WIP] Example twilio function call

nickhurlburt:
Shows how a Twilio function can be called.  Assumes that the domain name has been added to the Flex configuration.  Will include link to function in Jira ticket.  Though the function uses the Twilio console interface, whereas we should probably use the functions API.  And we should create a new repo for storing all the functions we use.

I don't fully understand or like how I have promises and async/await intertwined here.  This was what worked after multiple tries.


nickhurlburt:
This is obsolete in favor of https://github.com/tech-matters/flex-plugins/pull/58.  Closing.

PR #51, opened 2020-03-11
Gian add duration

GPaoloni:
Added "metadata" property inside the form related with a task, where we track starting and ending times. The conversation time is the difference between them .


plugin-hrm-form/src/components/HrmFormController.js
```diff
@@ -59,6 +59,10 @@ function getNumberFromTask(task) {
 export function saveToHrm(task, form, abortFunction, hrmBaseUrl, workerSid, helpline) {
   // if we got this far, we assume the form is valid and ready to submit
 
+  const { startingTime, endingTime } = form.metadata;
+  const milisecondsElapsed = endingTime - startingTime;
```
nickhurlburt: Please test this by reloading the page -- both before the task goes to 'wrapping', and between 'wrapping' and 'complete'.  Redux gets wiped on page reload.  If the page gets reloaded between 'wrapping' and 'complete', a start time will be created but no end time.  So it will send a large negative number for `conversationDuration` to the backend.

Unfortunately, so long as we implement this using redux, page reloads can cause problems.  But I think we're better off sending a `null` than a negative number.  And it's good to be aware of these edge cases, even if we aren't handling them perfectly right now.

plugin-hrm-form/src/___tests__/ContactFormStateFactory.test.js
```diff
@@ -1,3 +1,5 @@
+import _ from 'lodash';
```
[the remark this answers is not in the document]
GPaoloni: You are right! Changing right away

plugin-hrm-form/src/states/ContactFormStateFactory.js
```diff
@@ -425,5 +426,19 @@ export const generateInitialFormState = (formDefinition = defaultFormDefinition)
...
     });
+
   return initialState;
 };
+
+export const generateInitialFormState = (formDefinition = defaultFormDefinition) => {
+  // get "clean state" with current date as startingTime
+  const initialState = recursivelyGenerateInitialForm(formDefinition);
+  const metadata = {
+    startingTime: new Date(),
```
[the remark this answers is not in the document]
GPaoloni: Actually there's no need for the whole object, so I'm going with your suggestion

plugin-hrm-form/src/components/HrmFormController.js
GPaoloni: Oh nice catch! I'm adding a safe guard check and going to be more careful from now on. Just for the record, can this be mitigated by adding the start and end times as "task attributes"??

plugin-hrm-form/src/components/HrmFormController.js
```diff
@@ -59,6 +59,11 @@ function getNumberFromTask(task) {
 export function saveToHrm(task, form, abortFunction, hrmBaseUrl, workerSid, helpline) {
   // if we got this far, we assume the form is valid and ready to submit
 
+  const { startMillis, endMillis } = form.metadata;
+  const milisecondsElapsed = endMillis - startMillis;
+  const secondsElapsed = Math.floor(milisecondsElapsed / 1000);
+  const conversationDuration = secondsElapsed > 0 ? secondsElapsed : null;
```
nickhurlburt: This captures one case, but still does not address the full issue.  If there was a page refresh while the conversation is in progress (before wrapping), then the start time will be the time of the refresh and the duration will be inaccurate.  I would rather have a null time than an incorrect one.

Notice that the `HANDLE_FOCUS` AND `HANDLE_CHANGE` cases check to see if a form exists in redux, and if not, recreates it.  In the situations where we had to recreate it, then we should pass a null duration.  Another way to say that: we should only send a non-null duration if (1) the form in redux is the one created by `INITIALIZE_CONTACT_STATE` and (2) the form has a non-zero `endTimeMillis` created by the wrapping event.

Later on, we may have to persist data through a refresh, perhaps by writing to the task attributes.  For now, having either correct information or no information is best.

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -25,9 +25,13 @@ export default class HrmFormPlugin extends FlexPlugin {
     console.log(`Welcome to ${PLUGIN_NAME} Version ${PLUGIN_VERSION}`);
     this.registerReducers(manager);
 
-    const onCompleteTask = (sid, task) => {
-      if (task.channelType === 'voice' && task.status !== 'wrapping') {
-        flex.Actions.invokeAction('HangupCall', { sid, task });
+    const onCompleteTask = async (sid, task) => {
```
nickhurlburt: One more thing!  Please increment the version number on Line 10 to 0.3.6.  This helps us keep track of what's been deployed.

plugin-hrm-form/src/states/ContactFormStateFactory.js
```diff
@@ -425,5 +426,34 @@ export const generateInitialFormState = (formDefinition = defaultFormDefinition)
...
+  const metadata = {
+    startMillis: null,
+    endMillis: null,
+    recreated: true,
+  };
+
+  const recreatedForm = { ...initialState, metadata };
+
+  return recreatedForm;
+};
```
nickhurlburt: Seems to me this can be more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  Now that you've factored out `recursivelyGenerateInitialForm()`, I don't think there are any cases where `generateInitialFormState()` ever gets passed the `formDefinition` parameter (but please double check that).  So I think you can do something like:

```
export const generateInitialFormState = (recreated = false) => {
   const initialState = recursivelyGenerateInitialForm(defaultFormDefinition);
   const metadata = {
     startMillis: recreated ? null : new Date().getTime(),
     endMillis: null,
     recreated,
  };

  const initialForm = { ...initialState, metadata };

  return initialForm;
};

export const recreateForm = () => {
  generateInitialForm(true);
}
```

Please double-check the specifics of that code, it's just a rough example.  The point is that we can compose those functions rather than duplicate the logic.

plugin-hrm-form/src/states/ContactState.js
```diff
@@ -145,7 +149,21 @@ export function reduce(state = initialState, action) {
         ...state,
         tasks: {
           ...state.tasks,
-          [action.taskId]: generateInitialFormState(), // taskInitialStateFactory()
+          [action.taskId]: createBlankForm(), // taskInitialStateFactory()
```
nickhurlburt: The `// taskInitialStateFactory()` can be removed, the comment is no longer useful.

nickhurlburt:
And please do 'Squash and merge' to keep the git history clean:
image

PR #56, opened 2020-03-18
Gian CHI-72

GPaoloni:
Fixed bug that broke "end chat" button when reloading page
Added unit testing
Factored out some repetitive code that checks if a form exists in state or should be recreated


plugin-hrm-form/src/states/ContactState.js
```diff
@@ -110,14 +124,8 @@ export function reduce(state = initialState, action) {
...
        */
-      let currentForm;
-      if (state.tasks[action.taskId]) {
-        currentForm = state.tasks[action.taskId];
-      } else {
-        // currentForm = taskInitialStateFactory();
-        currentForm = recreateBlankForm();
-        console.log(`Had to recreate state for taskId ${action.taskId}`);
-      }
+      const currentForm = findOrRecreate(state.tasks, action.taskId);
```
nickhurlburt: Great!  While you're here, can you remove the comments in Lines 117-126?  Happily they are no longer needed!

PR #58, opened 2020-03-19
Gian populate counselors

GPaoloni:


plugin-hrm-form/src/services/fetchProtectedApi.js
```diff
@@ -0,0 +1,35 @@
+import { Manager } from '@twilio/flex-ui';
```
GPaoloni: I know @nickhurlburt don't like this, but there is an explanation!  
JWT token have an expiration date (usually a few hours from generated). If this is added via context (like we are doing with other constant configs), the token that will live in the context will be the one that was valid when Flex plugin was initialized, and might not be the one that is valid at the time we need to access the api (at least that's what I think, correct me if I'm wrong :laughing:).

plugin-hrm-form/src/services/fetchProtectedApi.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Factored out function that handles a protected api call hosted in serverless toolkit.
+ * Will throw Error if server responses with and http error code.
+ * @param {string} url Url to fetch from.
+ * @param {{}} body The body to send via fetch api (expected parameters for the route).
+ * @returns {any}
+ */
+const fetchProtectedApi = async (url, body) => {
+  const { token } = Manager.getInstance().store.getState().flex.session.ssoTokenPayload;
```
GPaoloni: By looking into state we ensure that the token we are using is the last generated one

plugin-hrm-form/src/services/fetchProtectedApi.js
```diff
@@ -0,0 +1,35 @@
+import { Manager } from '@twilio/flex-ui';
+
+/**
+ * Factored out function that handles a protected api call hosted in serverless toolkit.
```
GPaoloni: The idea of factoring this out is having only one file that acces into manager state, and then for every protected api route, we can specify url and body (route params) and use this helper.
If this approach isn't modified, all the protected routes should be addressed this way.

plugin-hrm-form/src/services/fetchProtectedApi.js
nickhurlburt: I see this is a case where we can't just set the `ConfigurationContext` at the beginning, because the value could change.  Two thoughts:

1. I wrote down the underlying principle for my desire to avoid including the `Manager` in more places in [the Abstraction section of our Operating Standards doc](https://benetech.app.box.com/notes/635791611466).  Please take a look.  In short, we may not only ever use this code inside Flex, so we will save our future selves some pain if we are careful about how many places we include Flex-specific things.

2. It looks like the `Manager` is included here so that we can get access to the redux store.  In that case, would it be better to use `mapStateToProps` instead?  Or some other way to get the redux store?  

Ultimately, it does make sense that we have a single place where we handle this.  So whether we use `Manager.getInstance()` or `mapStateToProps` or something else, it's a good idea to have a single place for this.

plugin-hrm-form/src/services/fetchProtectedApi.js
nickhurlburt: Agreed

plugin-hrm-form/src/services/fetchProtectedApi.js
nickhurlburt: Yes.  Per earlier comment, is this the best way to access the redux store, versus `mapStateToProps` or some other method?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -40,6 +40,7 @@ export default class HrmFormPlugin extends FlexPlugin {
     const hrmBaseUrl = manager.serviceConfiguration.attributes.hrm_base_url;
     const workerSid = manager.workerClient.sid;
     const { helpline } = manager.workerClient.attributes;
+    const currentWorkspace = manager.serviceConfiguration.taskrouter_workspace_sid;
```
nickhurlburt: We'll want to add an additional configuration here for the serverless service URL, which we can access from `manager.serviceConfiguration.attributes`, once we do a POST as described in the email I forwarded.

plugin-hrm-form/src/components/SearchForm.js
```diff
@@ -22,27 +24,47 @@ class SearchForm extends Component {
...
   render() {
-    const { firstName, lastName, counselor, phoneNumber, dateFrom, dateTo } = this.state;
+    const { firstName, lastName, counselor, counselors, phoneNumber, dateFrom, dateTo } = this.state;
+    const counselorsOptions = counselors.map(e => ({ label: e.friendlyName, value: e.sid }));
+
     const { helpline } = this.props;
-    const searchParams = { ...this.state, helpline };
+    const searchParams = {
+      ...omit(this.state, 'counselors'),
+      counselor: counselor.value, // backend expects only counselor's SID
```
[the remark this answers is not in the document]
murilovmachado: Right. Can you open a JIRA ticket for that? @nickhurlburt 
@GPaoloni don't worry about doing this in your PR.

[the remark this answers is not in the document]
nickhurlburt: For now, added to https://bugs.benetech.org/browse/CHI-71

plugin-hrm-form/src/services/fetchProtectedApi.js
GPaoloni: If you give a look at the [definitions of connect](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1b0bb5a24177dd622dda4ce0ff4e4a33b4f0c073/types/react-redux/index.d.ts#L188) (function that must be used alongside with mapStateToProps) it has "InferableComponentEnhancerWithProps<>" as return type. I believe using this to wrap a merely function would be hacky. I'll research if there is a better way.

nickhurlburt: Sure, that's reasonable.  This isn't a react component so using `connect` may not make sense.  If you take a few minutes to look at other options and don't see anything, keeping your original method is fine.  I want to minimize use of `Manager` outside `HrmFormPlugin.js`, but sometimes it's the right choice.

plugin-hrm-form/src/components/SearchForm.js
```diff
@@ -22,27 +24,47 @@ class SearchForm extends Component {
...
 
+  async componentDidMount() {
+    try {
+      const { currentWorkspace } = this.props;
+      const counselors = await populateCounselors(currentWorkspace);
+
+      this.setState({ counselors });
+    } catch (err) {
+      // TODO (Gian): probably we need to handle this in a nicer way
+      console.log(err.message);
```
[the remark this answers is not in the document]
GPaoloni: Also would be nice to show some "loading" indicator while waiting for the api results, for example. Do you have a preferred way for that?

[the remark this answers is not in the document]
nickhurlburt: I thought that [this article](https://medium.com/velotio-perspectives/how-to-make-asynchronous-calls-in-redux-without-middlewares-cbf7181020b2) had a decent approach for handling this, though it doesn't get much into the UI design. I created a [JIRA ticket](https://bugs.benetech.org/browse/CHI-74) for it.

plugin-hrm-form/src/services/fetchProtectedApi.js
murilovmachado: I'll add my 2 cents here. We can also pass a function into the context. I can imagine something like this would handle our scenario:
`
const getSsoToken = () => Manager.getInstance().store.getState().flex.session.ssoTokenPayload;
<ConfigurationContext.Provider value={{ hrmBaseUrl, workerSid, helpline, currentWorkspace, getSsoToken }} >
`
and the consumer would call `getSsoToken()` and get an updated token.

nickhurlburt: That sounds like a good idea to me.

plugin-hrm-form/src/components/FieldSelect.js
```diff
@@ -5,46 +5,60 @@ import { ErrorText, StyledLabel, StyledMenuItem, StyledSelect, TextField } from
...
+        return (
+          <StyledMenuItem key={option} value={option}>
+            {option}
+          </StyledMenuItem>
+        );
+      case 'object':
+        return (
+          <StyledMenuItem key={option.value} value={option}>
+            {option.label}
+          </StyledMenuItem>
```
murilovmachado: Shouldn't the `value` property be `value={option.value}>`

GPaoloni:
Still missing some more test files as @nickhurlburt said, but what do you think?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { VERSION } from '@twilio/flex-ui';
+import { VERSION, Manager } from '@twilio/flex-ui';
```
nickhurlburt: Not needed, see below.

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -38,8 +38,11 @@ export default class HrmFormPlugin extends FlexPlugin {
     };
 
     const hrmBaseUrl = manager.serviceConfiguration.attributes.hrm_base_url;
+    const serverlessBaseUrl = manager.serviceConfiguration.attributes.serverless_base_url;
     const workerSid = manager.workerClient.sid;
     const { helpline } = manager.workerClient.attributes;
+    const currentWorkspace = manager.serviceConfiguration.taskrouter_workspace_sid;
+    const getSsoToken = () => Manager.getInstance().store.getState().flex.session.ssoTokenPayload.token;
```
nickhurlburt: The `manager` is passed in as a parameter to the `init()` function.  So you can just use `const getSsoToken = () => manager.store.getState().flex.session.ssoTokenPayload.token;`

PR #59, opened 2020-03-23
Gian fix undefined helpline

GPaoloni:
When helpline is undefined, we want it to default to empty string, (in order to work properly with URLSearchParams)


GPaoloni:
Sorry, wrong base branch!

PR #60, opened 2020-03-23
Fix undefined helpline being sended to serverless endpoint

GPaoloni:
When helpline is undefined, we want it to default to empty string, (in order to work properly with URLSearchParams)

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -40,7 +40,7 @@ export default class HrmFormPlugin extends FlexPlugin {
     const hrmBaseUrl = manager.serviceConfiguration.attributes.hrm_base_url;
     const serverlessBaseUrl = manager.serviceConfiguration.attributes.serverless_base_url;
     const workerSid = manager.workerClient.sid;
-    const { helpline } = manager.workerClient.attributes || ''; // overrides undefined helpline with ''
+    const { helpline } = manager.workerClient.attributes;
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -40,7 +40,7 @@ export default class HrmFormPlugin extends FlexPlugin {
     const hrmBaseUrl = manager.serviceConfiguration.attributes.hrm_base_url;
     const serverlessBaseUrl = manager.serviceConfiguration.attributes.serverless_base_url;
     const workerSid = manager.workerClient.sid;
-    const { helpline } = manager.workerClient.attributes;
+    const { helpline } = manager.workerClient.attributes || ''; // overrides undefined helpline with ''
```
nickhurlburt: This was not what I was thinking.  I was thinking of handling this in `ServerlessService.js` by not sending the parameter at all if the helpline was undefined.  So something like:
```
  const { serverlessBaseUrl, helpline, currentWorkspace, getSsoToken } = context;
  const url = `${serverlessBaseUrl}/populateCounselors`;
  let body = {
    workspaceSID: currentWorkspace,
    Token: getSsoToken(),
  };
  if (helpline !== undefined) {
    body.helpline = helpline;
  };
```
I'm not sure if that's the exact correct syntax, but hopefully gets the point.  It seems like the best way to send `undefined` is to not send the parameter, rather than make the empty string a special code for `undefined`.  Does that make sense?

PR #61, opened 2020-03-23
Tabs visuals rework

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-75

This PR:
- Implements the new visuals for the tabs
- Updates `Container` element to match the new visuals.

Screen Shot 2020-03-23 at 20 31 40
Screen Shot 2020-03-23 at 20 31 49

### DIFF
--- a/plugin-hrm-form/src/Styles/HrmStyles.js
+++ b/plugin-hrm-form/src/Styles/HrmStyles.js
@@ -167,17 +167,29 @@ export const BottomButtonBar = styled('div')`
   justify-content: center;
 `;
 
+export const NameFields = styled('div')`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  > div {
+    flex-grow: 1;
+    flex-basis: 0;
+    margin: 0 10px;
+  }
+`;
+
 export const ColumnarBlock = styled('div')`
   display: flex;
   flex-direction: column;
   flex-grow: 1;
+  flex-basis: 0;
   margin: 0 10px;
 `;
 
 export const TwoColumnLayout = styled('div')`
   display: flex;
   flex-direction: row;
-  justify-content: space-around;
+  justify-content: space-between;
 `;
 
 export const CategoryCheckboxField = styled('div')`
@@ -196,7 +208,7 @@ export const StyledTableCell = styled(TableCell)`
 export const SearchFields = styled('div')`
   display: flex;
   flex-wrap: wrap;
-  div {
+  > div {
     max-width: 160px;
     min-width: 160px;
   }
--- a/plugin-hrm-form/src/Styles/HrmStyles.js
+++ b/plugin-hrm-form/src/Styles/HrmStyles.js
@@ -1,56 +1,25 @@
+import React from 'react';
 import styled, { keyframes } from 'react-emotion';
-import { Input, Select, MenuItem, TableCell } from '@material-ui/core';
+import { Input, Select, MenuItem, TableCell, Tabs, Tab } from '@material-ui/core';
 import { Button, getBackgroundWithHoverCSS } from '@twilio/flex-ui';
 
 export const Container = styled('div')`
   display: flex;
-  padding: 32px 32px 12px;
+  padding: 32px 20px 12px 20px;
   flex-direction: column;
   flex-wrap: wrap;
+  background-color: #ffffff;
+  border-top-left-radius: 4px;
+  border-top-right-radius: 4px;
+  margin: 0 20px;
 `;
 
-/*
- * export const TabContainer = styled('div')`
- *   margin-top: 24px;
- * `;
- */
-
-/*
- * export const ButtonContainer = styled('div')`
- *   display: flex;
- *   margin-top: 24px;
- * `;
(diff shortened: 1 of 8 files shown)


plugin-hrm-form/src/Styles/HrmStyles.js
```diff
@@ -1,56 +1,25 @@
...
+import { Input, Select, MenuItem, TableCell, Tabs, Tab } from '@material-ui/core';
 import { Button, getBackgroundWithHoverCSS } from '@twilio/flex-ui';
 
 export const Container = styled('div')`
   display: flex;
-  padding: 32px 32px 12px;
+  padding: 32px 20px 12px 20px;
   flex-direction: column;
   flex-wrap: wrap;
+  background-color: #ffffff;
```
nickhurlburt: I think this has caused a change to how the call type button background looks:
image

plugin-hrm-form/src/components/TabbedForms.js
```diff
@@ -584,13 +579,13 @@ class TabbedForms extends React.PureComponent {
...
+    tabs.push(<StyledTab searchTab key="Search" icon={<SearchIcon />} />);
     if (isCallerType) {
-      tabs.push(decorateTab('Caller Information', this.props.form.callerInformation));
+      tabs.push(decorateTab('Add Caller Information', this.props.form.callerInformation));
     }
-    tabs.push(decorateTab('Child Information', this.props.form.childInformation));
-    tabs.push(decorateTab('Issue Categorization', this.props.form.caseInformation.categories));
-    tabs.push(<Tab key="Case Information" label="Case Information" />);
+    tabs.push(decorateTab('Add Child Information', this.props.form.childInformation));
+    tabs.push(decorateTab('Categorize Issue', this.props.form.caseInformation.categories));
```
nickhurlburt: The validation error state here now looks a little wonky:
image

I don't have great ideas on what to do about it.

plugin-hrm-form/src/Styles/HrmStyles.js
```diff
@@ -1,56 +1,25 @@
+import React from 'react';
 import styled, { keyframes } from 'react-emotion';
-import { Input, Select, MenuItem, TableCell } from '@material-ui/core';
+import { Input, Select, MenuItem, TableCell, Tabs, Tab } from '@material-ui/core';
 import { Button, getBackgroundWithHoverCSS } from '@twilio/flex-ui';
 
 export const Container = styled('div')`
   display: flex;
-  padding: 32px 32px 12px;
+  padding: 32px 20px 12px 20px;
```
nickhurlburt: It may not be this line causing it, but seems like we want more margin with the button below:
image

plugin-hrm-form/src/components/TabbedForms.js
```diff
@@ -154,6 +146,12 @@ class TabbedForms extends React.PureComponent {
             </ColumnarBlock>
 
             <ColumnarBlock>
+              <FieldText
+                id="CallerInformation_LastName"
```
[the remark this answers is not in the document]
murilovmachado: Right, now I see the real use of `NameFields`. Putting it back.

plugin-hrm-form/src/components/TabbedForms.js
murilovmachado: I'll make this quick fix, so the exclamation mark looks consistent across tabs, but we should ask Jana her ideas for the issue:
Screen Shot 2020-03-24 at 11 44 34

plugin-hrm-form/src/components/TabbedForms.js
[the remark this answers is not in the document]
nickhurlburt: Another alternative is to explicitly set the tab order.  That would allow us to drop `NameFields` and is probably a more robust solution for future customizations.

murilovmachado:
@nickhurlburt thanks, fixed all the points you've mentioned. (Still working on fixing the unit tests)

plugin-hrm-form/src/components/TabbedForms.js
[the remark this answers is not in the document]
murilovmachado: I've tried playing around with `tabindex` attribute, but that would force us to write `tabindex` for all of our fields. So I'm keeping `NameFields` for now.

plugin-hrm-form/src/___tests__/FieldText.test.js
```diff
@@ -48,7 +48,7 @@ test('render required FieldText', () => {
     value: '',
     touched: false,
     error: null,
-    validation: [ValidationType.REQUIRED],
+    validation: ValidationType.REQUIRED,
```
nickhurlburt: I think I missed somewhere that this had turned into a scalar value and not an array.  I had originally made it an array so that we could have multiple validators that run on a field.  But, fine for now and we can revisit that whenever we do more work on validation.

murilovmachado: you're actually right, I think the `types/index` is where our current issue is. I'll revert this to an array, and fix `types/index`.

PR #63, opened 2020-03-25
Gian contact overview

GPaoloni:
Contact preview component has been reworked and separated into smaller components to make easier future changes without requiring major logical changes.

Added folder structure separation in components and styles (starting with the search-related ones).

populateCounselors logic (serverless api call + storing counselors in state) have been moved one level upper in the React tree (from SearchForm.js to Search.js) so we can "populate" both, the search form and each search result from a given query.

Bumped eslint-config-twilio-react version to latest.


plugin-hrm-form/src/components/search/Search.js
```diff
@@ -0,0 +1,65 @@
...
+    counselors: [],
+    counselorsHash: {},
+    results: [],
+  };
+
+  async componentDidMount() {
+    try {
+      const { context } = this.props;
+      const counselors = await populateCounselors(context);
+      const counselorsHash = counselors.reduce(
```
GPaoloni: The idea of this is, that computing this "hash" for our counselors is inexpensive, but saves lots of iterations through the counselors array when adding the counselor's name to each contact (id acts as hash key, and thus retrieving the name becomes a constant operation).

plugin-hrm-form/src/components/search/SearchResults.js
```diff
@@ -0,0 +1,139 @@
...
+  renderSummaryDialog() {
+    const isOpen = Boolean(this.state.selectedSumary);
+
+    return (
+      <Dialog onClose={this.closeDialog} open={isOpen}>
+        <DialogTitle id="simple-dialog-title">Selected Summary</DialogTitle>
+        <DialogContent>{this.state.selectedSumary}</DialogContent>
+      </Dialog>
+    );
+  }
```
GPaoloni: This is acting as some dummy behavior to mimic when clicking a contact preview card (but outside any of the action buttons within it). Is this behavior something we will use or all of this can be removed?

plugin-hrm-form/src/components/search/SearchResults.js
```diff
@@ -0,0 +1,139 @@
...
+    const handleConfirm = () => {
+      this.props.handleSelectSearchResult(this.state.contact);
+    };
+
+    return (
+      <Popover
+        id={id}
+        open={isOpen}
+        anchorEl={this.state.anchorEl}
+        // onClose={handleClose}
```
GPaoloni: Un-commenting this line will allow the user to click outside the "confirm popover" and it will dismiss it. Otherwise, the user's screen is "blocked" (cant click or scroll) until an action is taken (pressing either cancel or confirm button). Wich one is the desired behavior? (I'll rather leave it as it is now, if you ask me).

plugin-hrm-form/src/components/search/SearchResults.js
```diff
@@ -0,0 +1,139 @@
...
+          vertical: 'top',
+          horizontal: 'left',
+        }}
+        transformOrigin={{
+          vertical: 'top',
+          horizontal: 'right',
+        }}
+      >
+        <Paper
+          style={{
```
GPaoloni: Is ok to have inline styling for components we know will be very specific or rather have an isolated styled component even if it's used only here?

plugin-hrm-form/src/Styles/search/index.js
```diff
@@ -0,0 +1,69 @@
...
+const Tag = styled('div')`
+  border-radius: 5%;
+  padding: 5px;
+  padding-left: 10px;
+  padding-right: 10px;
+`;
+
+export const ContactTag = styled(Tag)`
+  margin-left: 10px;
+  background-color: #d8d8d8;
```
nickhurlburt: Where possible, consider using the [Twilio base theme colors](https://www.twilio.com/docs/flex/overriding-themes-branding-and-styling#list-of-base-theme-colors).  You can see an example of how they're accessed on Line 128 of `HrmStyles.js`.  If there's nothing close, then define your own, but where we can use a preset, let's do that.

plugin-hrm-form/src/components/search/Search.js
```diff
@@ -0,0 +1,65 @@
+import React, { Component } from 'react';
```
nickhurlburt: We may want to separate out the move to a separate folder into a separate PR to minimize merge conflicts cc @murilovmachado

plugin-hrm-form/src/components/search/Search.js
nickhurlburt: Yes, sounds reasonable.

plugin-hrm-form/src/components/search/Search.js
```diff
@@ -0,0 +1,65 @@
...
+
+class Search extends Component {
+  static displayName = 'Search';
+
+  static propTypes = {
+    context: contextObject.isRequired,
+    handleSelectSearchResult: PropTypes.func.isRequired,
+  };
+
+  state = {
```
nickhurlburt: Note that this will all have to end up in redux

plugin-hrm-form/src/components/search/SearchResults.js
nickhurlburt: We can put that out for feedback.

plugin-hrm-form/src/components/search/SearchResults.js
nickhurlburt: I think what you have now makes sense.

plugin-hrm-form/src/components/search/SearchResults.js
nickhurlburt: I don't have a firm rule.  We can have a few one-offs, but we have to be careful that they don't add up.  Then we will have a mess on our hands.

plugin-hrm-form/src/Styles/HrmStyles.js
```diff
@@ -244,3 +244,9 @@ export const StyledTab = styled(props => <Tab {...props} classes={{ selected: 's
     }
   }
 `;
+
+export const Row = styled('div')`
```
nickhurlburt: This seems like a very general name.  Would we use this for all or most of the rows in our codebase?

We should probably have a larger conversation as a team about how we can organize our styling.  I am not really sure how to do it and open to ideas.

murilovmachado:
@GPaoloni for the visual changes that @nickhurlburt requested: the `handoff` button on the bottom-left of the screen is now working. After clicking on it, you should be able to see the css for each element. It's not perfect css, but it should help a lot with fonts, sizes, widths, border, shadows etc.

GPaoloni:
> After clicking on 'more notes', there should be a 'less notes'. There should be a 'See full record' that goes to the detail page. See https://marvelapp.com/e1e23j4/screen/67480433.


@nickhurlburt I can remove this button. is no prob, is just that for UX reasons there should be a "collapse again" functionality (in my opinion). Or the idea is that once opened the detailed summary it should stay like that? Can't that be a mess after "showing more" on 3 or 4 different contacts?

A few notes on the new commit:
- I have merged staging into this branch (flex plugin version up)
- The alert that is displayed when the contact is connected (or chained) is tied to the contact its being connected (so it displays "by the contact's side), and its displayed 1 second before actually connecting the contact. This can be changed tho.
- Search.js component returned to parent folder so it will be easier to merge with @murilovmachado 's branch.

plugin-hrm-form/src/Styles/HrmStyles.js
GPaoloni: I intentionally left this as it is for now, because this is one of those "very general styles" we were talking today. There will be very a few times where this will be used just like <Row /> and its actual purpose is to be used like styled(Row)`...` with a more specific name and the specific css properties (like margin, padding, colors, etc), without the need of adding those 3 lines every time we want a new "row-like" div. Does that makes sense? If @murilovmachado knows a better way I'm willing to rework on this as much as necessary!

plugin-hrm-form/src/components/Search.js
```diff
@@ -1,12 +1,62 @@
...
+  const flattened = subcats.flatMap(s => {
+    const cat = s[0];
+    const f = s[1].flatMap(pair => {
+      const [subcat, bool] = pair;
+      if (bool) return cat + subcat; // remove cat if prefix is not needed anymore
+      return null;
+    });
+
+    return f;
+  });
```
GPaoloni: This returns "category-subcategory" string if the nested value is true, or null otherwise, in a flattened array (of 36 elements - 6 categories, 6 subcategories in each). I can rewrite this if its not clear enough tho. If someone comes up with a better solution for retrieving the tags please share it!

plugin-hrm-form/src/components/Search.js
```diff
@@ -17,19 +67,45 @@ class Search extends Component {
...
+      const counselors = await populateCounselors(context);
+      const counselorsHash = cnslrsHash(counselors);
+      this.setState({ counselors, counselorsHash });
+    } catch (err) {
+      // TODO (Gian): probably we need to handle this in a nicer way
+      console.error(err.message);
+    }
+  }
+
+  addDetails = raw => {
```
GPaoloni: Adds counselor's name and tags to a given contact. Used below

plugin-hrm-form/src/components/search/ContactPreview/CardRow3.jsx
```diff
@@ -0,0 +1,56 @@
...
+  handleClick = bool => event => {
+    event.stopPropagation();
+    this.setState({ expanded: bool });
+  };
+
+  render() {
+    return this.state.expanded ? (
+      <div>
+        <Button size="small" color="primary" onClick={this.handleClick(false)}>
+          <NoneTransform>less notes</NoneTransform>
```
nickhurlburt: > > After clicking on 'more notes', there should be a 'less notes'.

(As you figured out, I meant "there should not be...")

> @nickhurlburt I can remove this button, there's no prob, is just that for UX reasons there should be a "collapse again" functionality (in my opinion). Or the idea is that once opened the detailed summary it should stay like that? Can't that be a mess after "showing more" on 3 or 4 different contacts?

I understand where you're coming from, and you have a good point.  Though this was an area that Jana had thought through, so I'd rather go with the original design at first.  Then we can raise those concerns and also be able to demonstrate them.  It's a cleaner workflow to implement designs as given and provide feedback on issues we have, rather than build a different thing.  It's kind of like the one-way data flow in redux :)

plugin-hrm-form/src/components/SearchResults.js
```diff
@@ -1,172 +1,138 @@
...
+    const msg = "Connect current caller's record with this record?";
+
+    const handleClose = () => {
+      this.setState({ anchorEl: null, contact: null });
+    };
+
+    const handleConfirm = () => {
+      const { contact } = this.state;
+      this.setState({ connected: true });
+      setTimeout(() => {
```
nickhurlburt: Yes, I think this is a great plan.  (Showing confirmation and redirecting after 1 second.)

plugin-hrm-form/src/components/search/ContactPreview/ContactPreview.jsx
```diff
@@ -0,0 +1,75 @@
...
+import { searchContactResult } from '../../../types';
+
+const months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+
+/**
+ * Returns first word of the string as uppercase
+ * @param {string} str
+ * @return {string}
+ */
+const fstToUpper = str => {
```
nickhurlburt: I think we need a better mapping here.  We end up with confusing terms like "Hang" or "Wrong".  Let's do a mapping of:
* SOMEONE CALLING ABOUT A CHILD -> CALLER
* CHILD CALLING ABOUT SELF -> SELF
* All others stay as they are

We'll probably revisit these words later.

plugin-hrm-form/src/Styles/HrmStyles.js
```diff
@@ -244,3 +244,15 @@ export const StyledTab = styled(props => <Tab {...props} classes={{ selected: 's
...
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+`;
+
+export const FontOpenSans = styled('p')`
+  color: #000000;
+  font-family: Open Sans;
```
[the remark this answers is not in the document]
GPaoloni: I though the same, but when not specifying any font family the results where those you saw in the first round. Maybe I'm using the themes incorrectly?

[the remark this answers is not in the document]
nickhurlburt: I'm not sure... this is a mystery so far.

plugin-hrm-form/src/components/search/ContactPreview/ContactPreview.jsx
```diff
@@ -0,0 +1,75 @@
...
+  const day = d.getDay();
+  const year = d.getFullYear();
+  const h = d.getHours();
+  const in12 = h % 2;
+  const hours = in12 ? in12 : 12; // hour "00" is displayed as "12"
+  const m = d.getMinutes();
+  const minutes = m < 10 ? `0${m}` : m;
+  const meridiem = h > 11 ? 'pm' : 'am';
+  return `${months[month]} ${day}, ${year} ${hours}:${minutes}${meridiem}`;
+};
```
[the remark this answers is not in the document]
murilovmachado: I'd suggest using `date-fns`

plugin-hrm-form/src/components/search/ContactPreview/ContactPreview.jsx
```diff
@@ -0,0 +1,69 @@
...
+ * @return {string}
+ */
+const mapAndToUpper = str => {
+  const fst = str.split(' ')[0].toUpperCase();
+  const mapped = mapper(fst);
+  return mapped;
+};
+
+const ContactPreview = ({ contact, onClick, handleConnect }) => {
+  const name = (contact.overview.name === ' ' ? 'Unknown' : contact.overview.name).toUpperCase();
```
murilovmachado: Consider using `trim()` here:  `contact.overview.name.trim() === ''`. (and then check equality with empty string)
Currently, It's easy for someone else touching the code accidentally change from `' '` to `''`, and that would cause issues.

plugin-hrm-form/src/Styles/HrmStyles.js
murilovmachado: About styling, I don't want to be imposing too many restrictions at the beginning. Gian's work is a very good starting point. While developing the next components we'll have a better idea of what can be easily reused and what can't.

plugin-hrm-form/src/components/search/ContactPreview/ContactPreview.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+  const dateString = moment(contact.overview.dateTime).format('MMM DD, YYYY HH:mm a');
```
murilovmachado: Minor stuff:  currently, the community prefer to use `date-fns` instead of `moment`. But it's something we can deal with later.

plugin-hrm-form/src/components/search/ContactPreview/ContactPreview.jsx
```diff
@@ -0,0 +1,69 @@
...
+      <Card>
+        <CardContent>
+          {/** child's name and action buttons */}
+          <CardRow1 name={name} onClickChain={handleConnect} onClickFull={mockedAction} onClickMore={mockedAction} />
+          {/** call type and counselor's name */}
+          <CardRow2 callType={mapAndToUpper(contact.overview.callType)} counselor={contact.counselor} />
+          {/** call summary (notes) */}
+          <CardRow3 callSummary={contact.details.caseInformation.callSummary} onClickFull={mockedAction} />
+          {/** date and call tags */}
+          <CardRow4 dateString={dateString} tag1={tag1} tag2={tag2} tag3={tag3} />
```
murilovmachado: This looks good, but since `CardRow1`, `CardRow2`, `CardRow3` and`CardRow4` won't be used elsewhere,  I think we can choose more descriptive names and get rid of the comments, ex:
```
<CardContent>
  <ChildNameAndActions ... />
  <CallTypeAndCounselor ... />
  <Notes ... />
  <DateAndTags ... />
</CardContent>
```

What do you think?

plugin-hrm-form/package.json
```diff
@@ -23,6 +23,7 @@
     "flex-plugin": "^3.0.0",
     "flex-plugin-scripts": "^3.0.4",
     "lodash": "^4.17.15",
+    "moment": "^2.24.0",
```
murilovmachado: Are you familiar with `date-fns` lib? The community is using it instead of `moment` nowadays.
(Don't need to change now, but we can revisit the date lib we use in the future)

plugin-hrm-form/src/components/search/ContactPreview/ContactPreview.jsx
GPaoloni: I wasn't aware of them till yesterday, and reading I found that `date-fns` beats `moment` in performance and size so I'm swapping to it! Thanks

plugin-hrm-form/src/components/search/ContactPreview/ContactPreview.jsx
GPaoloni: Totally agree, wasn't sure about the naming convention to use here. Going with this!

plugin-hrm-form/package.json
nickhurlburt: Haha, Sebastian and I had recommended `moment`.  I guess we're too old school :)

plugin-hrm-form/src/components/search/ContactPreview/ContactPreview.jsx
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Returns first word of the string as uppercase
+ * and maps as specified:
+ * CHILD -> SELF
+ * SOMEONE -> CALLER
+ * anything else is unchanged
+ * @param {string} str
+ * @return {string}
+ */
+export const mapAndToUpper = str => {
```
nickhurlburt: This is still not quite the right behavior.  Don't strip the second word off of `HANG UP` or `WRONG NUMBER`.  We want:
```
Child calling about self -> SELF
Someone calling about a child -> CALLER
Silent -> SILENT
Blank -> BLANK
Wrong number -> WRONG NUMBER
Hang up -> HANG UP
Abusive -> ABUSIVE
```

PR #64, opened 2020-03-25
Search component rework

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-76

This PR:
- Updates `Search` component and inner components to match the new behavior
- Moves all search data into redux

(1. initial search page)
Screen Shot 2020-03-25 at 18 55 50

(2. after clicking search, shows results page)
Screen Shot 2020-03-25 at 18 56 07

(3. after seleting one result, currently clicking on contact's name, shows details page)
Screen Shot 2020-03-25 at 18 56 16

The inner components `SearchForm`, `SearchResults`, `ContactDetail` will be updated in future PRs, so don't worry about how they currently look.


plugin-hrm-form/src/components/Search.js
```diff
@@ -1,39 +1,83 @@
...
+const Search = props => {
+  const handleSearch = searchParams => {
+    const { hrmBaseUrl } = props.context;
+    props.searchContacts(hrmBaseUrl, searchParams);
   };
 
-  state = {
-    results: [],
-  };
+  const goToForm = () => props.changeSearchPage('form');
```
nickhurlburt: Would it be a good idea to define constants for `'form'`, `'results'` and `'details'`?

plugin-hrm-form/src/components/SearchResults.js
```diff
@@ -88,9 +45,11 @@ class SearchResults extends Component {
 
   handleClickCallSummary = selectedCallSumary => this.setState({ selectedCallSumary });
```
nickhurlburt: nit: spelling, `selectedCallSummary`. If this is all getting removed, don't worry about it.

plugin-hrm-form/src/states/SearchContact.js
```diff
@@ -1,11 +1,40 @@
...
 
+export const handleSearchFormChange = (name, value) => ({
+  type: HANDLE_SEARCH_FORM_CHANGE,
+  name,
+  value,
+});
+
+export const searchContacts = dispatch => async (hrmBaseUrl, searchParams) => {
+  try {
+    dispatch({ type: SEARCH_CONTACTS_REQUEST });
```
nickhurlburt: Great!  This was the three-action API fetch we were talking about (or at least I linked to somewhere).

plugin-hrm-form/src/components/Search.js
```diff
@@ -1,39 +1,83 @@
...
+  const goToResults = () => props.changeSearchPage('results');
 
-  handleSearch = async searchParams => {
-    const { hrmBaseUrl } = this.props.context;
-    const results = await searchContacts(hrmBaseUrl, searchParams);
-    this.setState({ results });
-  };
+  const { currentPage, currentContact, searchResult, isRequesting, error, form } = props;
+  console.log({ isRequesting, error });
```
nickhurlburt: Would we change the UI somehow when `isRequesting` is true, in another component somewhere?

murilovmachado: @nickhurlburt Yes, right now we're not using `isRequesting`, but it should be used to show a loading spinner. There's two places where it could be shown:
- On `SearchForm`, after hitting the search button, it could display a loading spinner, and it only goes to `SearchResults` after `isRequesting` is finished; or
- Directly goes `SearchResults` and show a loading spinner in that page while `isRequesting` is true;

This loading scenario was not covered in the mockups. I believe you, Joan or Jana should decide which of these options is best. Or maybe come up with a third/different option.

Does it make sense?

plugin-hrm-form/src/components/SearchResults.js
murilovmachado: It's going to be deleted, but it's 10s fix, so I've fixed it.

murilovmachado:
Fixed code review points. Working on creating unit tests.

plugin-hrm-form/src/components/Search.js
nickhurlburt: Agreed... I'll feed that back.

GPaoloni: I'm just leaving this here a possible solutions to the "isRequesting" behavior.

https://material-ui.com/components/backdrop/
https://material-ui.com/components/progress/

PR #66, opened 2020-03-30
Add contact detail component

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-81

This PR:

- Adds the Contact details component
- Reuses "not implemented yet" dialog on all the Search components (moved one level up in the component tree)
- Splits tests in subfolders


plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,119 @@
...
+        </Button>
+      </NameContainer>
+      <Category category="General details">
+        <CategoryEntry description="Channel" value="Work in progress (hrm)" />
+        {customerNumber && <CategoryEntry description="Phone Number" value={customerNumber} />}
+        <CategoryEntry description="Conversation Duration" value="Work in progress (hrm)" />
+        <CategoryEntry description="Counselor" value={counselor} />
+        <CategoryEntry description="Date/Time" value={formatedDate} />
+      </Category>
+      <Category category="Caller information">
```
nickhurlburt: If the call type is not "Someone calling about a child", there will be no Caller Information.  In that case, this block does not need to appear.

plugin-hrm-form/src/components/search/ContactDetails/Category.jsx
```diff
@@ -0,0 +1,46 @@
...
+import PropTypes from 'prop-types';
+import { Button, Collapse } from '@material-ui/core';
+import { ArrowDropDownTwoTone, ArrowDropUpTwoTone } from '@material-ui/icons';
+
+import { CategoryTitleContainer, CategoryTitleText, ContactDetailsIcon } from '../../../Styles/search';
+
+const ArrowDownIcon = ContactDetailsIcon(ArrowDropDownTwoTone);
+const ArrowUpIcon = ContactDetailsIcon(ArrowDropUpTwoTone);
+
+class Category extends React.Component {
```
nickhurlburt: I would like to avoid calling this a `Category` because that confuses it with the categories that counselors enter.  `Section`?

plugin-hrm-form/src/Styles/HrmStyles.js
```diff
@@ -202,11 +202,15 @@ export const CategoryCheckboxField = styled('div')`
   width: 160px;
 `;
 
-export const StyledTableCell = styled(TableCell)`
-  text-overflow: ellipsis;
-  overflow: hidden;
-  white-space: nowrap;
-`;
+/*
```
nickhurlburt: I'm okay with removing unused things.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,119 @@
...
+    okForCaseWorkerToCall,
+    howDidTheChildHearAboutUs,
+    didYouDiscussRightsWithTheChild,
+    didTheChildFeelWeSolvedTheirProblem,
+    wouldTheChildRecommendUsToAFriend,
+  } = details.caseInformation;
+
+  const nameOrUnknown = name.trim() === '' ? 'Unknown' : name;
+  const nameUpperCase = nameOrUnknown.toUpperCase();
+  const formatedDate = `${format(new Date(dateTime), 'MMM d, yyyy / h:mm aaaaa')}m`;
```
nickhurlburt: spelling: `formattedDate`

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
nickhurlburt: Also, if the call type is standalone (not 'Someone calling about a child' or 'Child calling about self') then we should only show General Details, no other sections.  We only show sections that were present as tabs when the information was first entered.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,119 @@
...
+  const { streetAddress, city, stateOrCounty, postalCode, phone1, phone2 } = location;
+  const fullLocation = streetAddress + withComma(city) + withComma(stateOrCounty) + withComma(postalCode);
```
nickhurlburt: We should use template strings here.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,119 @@
...
+  return (
+    <DetailsContainer>
+      <NameContainer>
+        <DetNameText>{nameUpperCase}</DetNameText>
+        <Button
+          size="small"
+          style={{ padding: 0 }}
+          onClick={() => /* TODO: this must be implemented */ console.log(contact)}
+        >
+          <MoreHorizIcon style={{ color: '#ffffff' }} />
```
nickhurlburt: Let's re-use that "Not implemented yet" dialog.

plugin-hrm-form/src/components/search/ContactDetails/index.jsx
```diff
@@ -1,17 +1,25 @@
...
   <Container>
-    <button type="button" onClick={props.handleBack}>
-      Back
-    </button>
-    <SearchResultDetails details={props.contact} handleClickCallSummary={() => {}} />
+    <Row>
+      <IconButton onClick={handleBack}>
+        <ChevronLeftIcon />
+      </IconButton>
+      <BackText>RETURN TO RESULTS</BackText>
```
nickhurlburt: If the user clicks on either the icon or the text, let's go back to results.

plugin-hrm-form/src/Styles/search/index.js
```diff
@@ -108,3 +108,66 @@ export const TagText = styled(FontOpenSans)`
...
+  margin-bottom: 3px;
+`;
+
+const BoldDetailFont = styled(FontOpenSans)`
+  font-size: 12px;
+  font-weight: 700;
+  line-height: 14px;
+`;
+
+export const BackText = styled(BoldDetailFont)`
```
nickhurlburt: Consider making this and the "< BACK" that renders above it use the same styling to improve DRYness?

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,119 @@
...
+        <CategoryEntry description="Did you discuss right with the child?" value={didYouDiscussRightsWithTheChild} />
+        <CategoryEntry
+          description="Did the child feel we solved their problem?"
+          value={didTheChildFeelWeSolvedTheirProblem}
+        />
+        <CategoryEntry
+          description="Would the child recommend us to a friend?"
+          value={wouldTheChildRecommendUsToAFriend}
+        />
+      </Category>
```
nickhurlburt: Something to think about: We will have many helplines have slightly different fields they capture in each of these sections.  How could we avoid hard-coding every field to make customization easier?  Also good for @murilovmachado to consider.

nickhurlburt:
Also, can we start making the title of the PR a little more clear?  Rather than "Gian contact details", something like "Add contact detail component"?

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
GPaoloni: This was another of the question I had today! Forgot to mention it on slack, but now I know, thanks for clarifying.

plugin-hrm-form/src/components/search/ContactDetails/Category.jsx
GPaoloni: Sure, makes sense

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
GPaoloni: Yes! I was using that instead for easier debug, not going to stay at the end.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
GPaoloni: I've considered, instead of passing each Section Entry as a React children, we could pass an array of touples (e.g. `<Section sectionTitle='Section_1'entries=[ [entry_1, value_1] , [entry_2, value_2] , [entry_3, value_3] ]`) and then with a map generate an `<SectionEntry .../>` for each tuple.
We still will need to manually specify those, but maybe it will be less verbose. Just an idea though, I'm going to further research on this.

nickhurlburt:
Some decisions:
* When the detail card first loads, let's have just the General Details section expanded, and everything else contracted.
* Make the entire row of the section name clickable to expand/contract.  Get rid of the color change when mousing over the chevron.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
nickhurlburt: We don't have to have a really general solution this time around.  But we will eventually, so I just want to raise the idea.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,216 @@
...
+};
+
+/**
+ * @param {string} s
+ */
+const formatCallSummary = s => {
+  const words = s.split(' ');
+  const fixedWords = words.map(w => maxLength30(w));
+  return fixedWords.join(' ');
+};
```
GPaoloni, line 62: Maybe this helper functions can be factored out from here? If so, where should they live?

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,216 @@
...
+          <MoreHorizIcon style={{ color: '#ffffff' }} />
+        </Button>
+      </NameContainer>
+      <Section
+        expanded
+        sectionTitle="General details"
+        entries={[
+          { description: 'Channel', value: channel.charAt(0).toUpperCase() + channel.slice(1) },
+          maybeEntry(Boolean(customerNumber), { description: 'Phone Number', value: customerNumber }),
+          { description: 'Conversation Duration', value: conversationDuration },
```
nickhurlburt, line 139: Let's display the duration as `Hh Mm Ss`, for example, `17s` or `24m 9s` or `1h 12m 0s`.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,216 @@
...
+  return (
+    <DetailsContainer>
+      <NameContainer>
+        <DetNameText>{nameUpperCase}</DetNameText>
+        <Button
+          size="small"
+          style={{ padding: 0 }}
+          onClick={() => /* TODO: this must be implemented */ console.log(contact)}
+        >
+          <MoreHorizIcon style={{ color: '#ffffff' }} />
```
nickhurlburt, line 130: The placement of this doesn't look right: 
image

In the design it's directly over the chevrons:
image

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
nickhurlburt: Still to do ^

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,216 @@
...
+  if (w.length < 30) return w;
+  const init = w.substr(0, 29);
+  const tail = maxLength30(w.substr(29));
+  return `${init}-\n${tail}`;
+};
+
+/**
+ * @param {string} s
+ */
+const formatCallSummary = s => {
```
nickhurlburt, line 58: Notice that this does not support newlines in the call summary.  If I enter into the call summary window:
```
This is a summary of the notes. 

Then a newline. 


Then two newlines.
```

It renders as:
image

We also have the same issue on the ContactPreview:
image


Here is code for how I dealt with this in `plugin-show-recent-contacts`, though there's probably a better JS utility method out there for this: https://github.com/tech-matters/flex-plugins/blob/master/plugin-show-recent-contacts/src/components/RecentContactsView.js#L300

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
nickhurlburt: We could add a `/utility/` or `/tools/` directory in the package.  Maybe @murilovmachado has ideas as well.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,216 @@
...
+          onClick={() => /* TODO: this must be implemented */ console.log(contact)}
+        >
+          <MoreHorizIcon style={{ color: '#ffffff' }} />
+        </Button>
+      </NameContainer>
+      <Section
+        expanded
+        sectionTitle="General details"
+        entries={[
+          { description: 'Channel', value: channel.charAt(0).toUpperCase() + channel.slice(1) },
```
nickhurlburt, line 137: For channel, let's display the following mapping:
```
facebook => Facebook Messenger
web => Chat
voice => Voice
sms => SMS
whatsapp => WhatsApp
```

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,216 @@
...
+          maybeEntry(Boolean(customerNumber), { description: 'Phone Number', value: customerNumber }),
```
nickhurlburt, line 138: If the channel is not voice, sms, or whatsapp, then let's just do '-' for Phone Number.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,216 @@
...
+          { description: 'Conversation Duration', value: conversationDuration },
+          { description: 'Counselor', value: counselor },
+          { description: 'Date/Time', value: formattedDate },
+        ]}
+      />
+      {callType === 'Someone calling about a child' && (
```
nickhurlburt, line 144: Let's use the constants in https://github.com/tech-matters/flex-plugins/blob/master/plugin-hrm-form/src/states/DomainConstants.js#L1 when referencing these.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,216 @@
...
+export const formatAddress = (street, city, state, postalCode) => {
+  const commaSeparated = [street, city, state].filter(s => s.trim()).join(', ');
+  const withPostalCode = [commaSeparated, postalCode].filter(s => s.trim()).join(' ');
+  return withPostalCode;
+};
+
+/**
+ * @param {string} s
+ */
+const checkForDataContact = s => {
```
nickhurlburt, line 33: Consider reusing https://github.com/tech-matters/flex-plugins/blob/master/plugin-hrm-form/src/states/ValidationRules.js#L10 for this.  It can be renamed `isNonDataCallType`.  If reusing that doesn't make sense, then never mind.  It would be nice to stay DRY with this, though.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * If a word length is more than 30 chars, it will break it with a hyphen and line break
+ * @param {string} w
+ */
+const maxLength30 = w => {
+  if (w.length < 30) return w;
+  const init = w.substr(0, 29);
+  const tail = maxLength30(w.substr(29));
+  return `${init}-\n${tail}`;
+};
```
nickhurlburt, line 53: Skeptical about this, as this could still break the word in the middle of a line.  What is the behavior if you don't do this at all?

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,165 @@
...
+        <ButtonBase style={{ padding: 0 }} onClick={handleMockedMessage}>
+          <MoreHorizIcon style={{ color: '#ffffff' }} />
+        </ButtonBase>
+      </NameContainer>
+      <Section
+        expanded
+        sectionTitle="General details"
+        entries={[
+          { description: 'Channel', value: formattedChannel },
+          { description: 'Phone Number', value: channel === 'voice' ? customerNumber : '' },
```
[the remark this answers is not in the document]
GPaoloni: Oh sorry on this, the misread was mine, I actually forgot that we have a phone number when channel is sms/whatsapp.

plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -0,0 +1,165 @@
...
+  const {
+    name: callerNameObj,
+    relationshipToChild: callerRelationship,
+    gender: callerGender,
+    age: callerAge,
+    language: callerLanguage,
+    nationality: callerNationality,
+    ethnicity: callerEthnicity,
+    location: callerLocation,
+  } = details.callerInformation;
```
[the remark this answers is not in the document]
GPaoloni: The renaming for the caller information is just because I can't destructure using the original names the object has, because (when destructuring `callerInformation` object) age is already defined (it was defined when destructuring the `childInformation` object). Do you prefer the children's information to be renamed like `childName`, `childAge`, etc?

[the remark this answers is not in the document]
nickhurlburt: Yes, I think that if we are going to namespace one thing, we should namespace the other.  It gets confusing otherwise.  I think there are two options:
1. Destructure with `callerXXX` and `childXXX`
2. Leave them inside an object, say:
```
const caller = details.callerInformation;
const child = details.childInformation;
function(caller.gender, caller.age, ...);
function(child.gender, child.age, ...);
```
#2 actually makes more sense to me, but I am really happy with either one.  I just think we should do the same for both caller and child.

PR #69, opened 2020-04-02
Remove confirmation and ContactPreview highlighting

GPaoloni:
JIRA: 
https://bugs.benetech.org/browse/CHI-90
https://bugs.benetech.org/browse/CHI-91

- Remove 'Connected' confirmation when copying data into new contact
- Remove highlighting on the ContactPreview card


GPaoloni:
> Seems like there's an off-by-one error in which tab we land on after redirecting. 

Agreed with doing this in [CHI-84](https://bugs.benetech.org/browse/CHI-84)!

PR #71, opened 2020-04-03
Change how data is copied from contact + search state gets cleaned when task completes

GPaoloni:
JIRA:
https://bugs.benetech.org/browse/CHI-84
https://bugs.benetech.org/browse/CHI-86

- Changed how data is copied from a contact to the current form (details can be found [here](https://bugs.benetech.org/browse/CHI-84))
- Landing on form tab where data is copied to
- SearchContact state gets cleaned after task is finished
- Tests related to state have been moved into __tests__/states folder


plugin-hrm-form/src/components/search/SearchResults/index.jsx
```diff
@@ -54,7 +71,7 @@ class SearchResults extends Component {
         }}
       >
         <ConfirmContainer>
-          <ConfirmText>{msg}</ConfirmText>
+          <ConfirmText>{this.state.msg}</ConfirmText>
```
nickhurlburt, line 74: Let's also change L86 to 'yes, copy'

plugin-hrm-form/src/components/TabbedForms.jsx
```diff
@@ -64,9 +64,11 @@ class TabbedForms extends React.PureComponent {
   handleSelectSearchResult = (searchResult, taskId) => {
     this.props.handleSelectSearchResult(searchResult, taskId);
 
-    // For now, redirect to childInformation Tab (will be changed in the future)
-    const isCallerType = this.props.form.callType.value === callTypes.caller;
-    const tab = isCallerType ? 2 : 1;
+    // Redirects to the tab where data is being copied to
+    const currentIsCaller = this.props.form.callType.value === callTypes.caller;
+    const selectedIsChild = searchResult.details.callType === callTypes.child;
+    const tab = currentIsCaller && selectedIsChild ? 2 : 1;
```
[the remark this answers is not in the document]
nickhurlburt: Ah... looking at this again, I think this issue was introduced in https://github.com/tech-matters/flex-plugins/pull/55/files#diff-60517ac9379a4aa6d141d972df716007R50.  So, no need to change that in this PR.  We can handle that elsewhere.  So other than the few other comments here, consider this approved.

plugin-hrm-form/src/states/SearchContact.js
```diff
@@ -100,12 +112,54 @@ function copyNewValues(originalObject, objectWithNewValues) {
...
 export function copySearchResultIntoTask(currentTask, searchResult) {
   const searchResultDetails = searchResult.details;
-  return copyNewValues(currentTask, searchResultDetails);
+  return partiallyCopyValues(currentTask, searchResultDetails);
 }
 
 const initialState = {
+  tasks: {},
+  isRequesting: false,
+  error: null,
```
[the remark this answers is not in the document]
GPaoloni: I've assumed that if `isRequesting`, the user won't be able to perform more actions until completion (succes or failure). We want them to be able to perform a search and in the meanwhile continue working with other task?

plugin-hrm-form/src/___tests__/search/Search.test.js
```diff
@@ -17,21 +17,25 @@ jest.mock('../../services/ServerlessService', () => ({
...
+              phoneNumber: '',
+              dateFrom: '',
+              dateTo: '',
+            },
+            searchResult: searchResult || [],
+          },
         },
-        searchResult: searchResult || [],
         isRequesting: false,
         error: null,
```
murilovmachado, line 40: @GPaoloni @nickhurlburt What I had in mind was to have `searchResults`, `isRequesting` and `error` inside `[taskId]` as well. Disclaimer: I didn't follow the whole discussion, so I'm not sure that would still be a correct solution.

nickhurlburt: I think that sounds right.

PR #73, opened 2020-04-06
Do not save form data when non-data callType

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-87

When saving a non-data callType, no information filled in the form by the user should be sent. In this case, we're recreating a blank form, keeping just `metadata` and `callType`.

### DIFF
--- a/plugin-hrm-form/src/states/DomainConstants.js
+++ b/plugin-hrm-form/src/states/DomainConstants.js
@@ -9,4 +9,6 @@ const callTypes = {
   abusive: 'Abusive',
 };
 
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/states/DomainConstants.js
```diff
@@ -9,4 +9,6 @@ const callTypes = {
   abusive: 'Abusive',
 };
 
+export const isNonDataCallType = callType => ![callTypes.child, callTypes.caller].includes(callType);
```
nickhurlburt, line 12: Can we reuse or replace `ValidationRules.isNonDataCallType` rather than duplicate it?

PR #79, opened 2020-04-14
Implement TabPressWrapper

murilovmachado:
Implements `TabPressWrapper`, a component that handles `tab` and `shift+tab` key presses, and rotate the focus between the contained elements.

Usage:
```
<TabPressWrapper>
  <YourComponent />
</TabPressWrapper>
```

This PR:
- Implements `TabPressWrapper`
- Implement unit tests

### DIFF
--- a/plugin-hrm-form/src/___tests__/TabPressWrapper.test.js
+++ b/plugin-hrm-form/src/___tests__/TabPressWrapper.test.js
@@ -1,0 +1,79 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import './mockStyled';
+import TabPressWrapper from '../components/TabPressWrapper';
+
+const getFirstElement = component => component.getInstance().firstElementRef.current;
+const getLastElement = component => component.getInstance().lastElementRef.current;
+
+test('<TabPressWrapper> children with tabIndexes: 1 and 2', () => {
+  const component = renderer.create(
+    <TabPressWrapper>
+      <button id="firstElement" type="button" tabIndex={1} />
+      <button id="secondElement" type="button" tabIndex={2} />
+    </TabPressWrapper>,
+  );
+
+  const firstElement = getFirstElement(component);
+  const lastElement = getLastElement(component);
+
+  expect(firstElement.props.id).toEqual('firstElement');
+  expect(lastElement.props.id).toEqual('secondElement');
+});
+
+const setActiveElement = element => jest.spyOn(document, 'activeElement', 'get').mockReturnValue(element);
+const event = {
+  preventDefault: jest.fn(),
+  stopPropagation: jest.fn(),
+};
+const pressTab = component => component.getInstance().handleTab('tab', event);
+const pressShiftTab = component => component.getInstance().handleTab('shift+tab', event);
+
+const assertCalledFocusElementWith = (component, element) => {
+  const focusElementMethod = component.getInstance().focusElement;
+  const firstArgument = focusElementMethod.mock.calls[0][0];
+
+  expect(focusElementMethod).toHaveBeenCalled();
+  expect(firstArgument.current).toEqual(element);
+};
+
+test('<TabPressWrapper> lastElement focused and hit "tab"', () => {
+  const component = renderer.create(
+    <TabPressWrapper>
+      <button id="firstElement" type="button" tabIndex={1} />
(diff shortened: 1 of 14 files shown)


plugin-hrm-form/src/components/TabPressWrapper.jsx
```diff
@@ -0,0 +1,135 @@
...
+    }
+  };
+
+  focusElement = (elementRef, event) => {
+    event.preventDefault();
+    event.stopPropagation();
+    elementRef.current.focus();
+  };
+
+  countTabIndexElements = children => {
```
nickhurlburt, line 71: Always love seeing a good recursive tree traversal in the wild :)

plugin-hrm-form/src/components/TabPressWrapper.jsx
```diff
@@ -0,0 +1,135 @@
...
+        childrenCount += 1;
+      } else if (hasChildren) {
+        childrenCount += this.countTabIndexElements(element.props.children);
+      }
+    });
+
+    return childrenCount;
+  };
+
+  addRefToFirstAndLastElement = (children, maxTabIndex) =>
```
nickhurlburt, line 88: And another one!

plugin-hrm-form/src/components/TabPressWrapper.jsx
```diff
@@ -0,0 +1,135 @@
...
+    React.Children.map(children, element => {
+      const hasTabIndex = Boolean(element.props && !isNullOrUndefined(element.props.tabIndex));
+      const hasChildren = Boolean(element.props && !isNullOrUndefined(element.props.children));
+
+      if (hasTabIndex) {
+        const { tabIndex } = element.props;
+        const isFirstElement = tabIndex === 1;
+        const isLastElement = tabIndex === maxTabIndex;
```
nickhurlburt, line 96: Hmm... I think this makes the assumption that the tab indexes start at 1 and are strictly consecutive.  So this works if the indexes are 1, 2, 3, 4, 5.  But it doesn't work if the indexes are 1, 3, 4, 5, 7 or 2, 3, 4, 5, 6.  If the browser works fine in those scenarios, then you could make this more robust by changing `countTabIndexElements` to `findMaxTabIndex` and changing the recursion to return the largest element seen at any level with something like a `reduce` rather than `forEach`.

And it'd be good to include some non-consecutive tab indexes in your tests too :)

murilovmachado: That's a good idea! I'll let you know when I have these changes and unit tests in place.

plugin-hrm-form/src/___tests__/TabPressWrapper.test.js
```diff
@@ -0,0 +1,278 @@
...
+import TabPressWrapper from '../components/TabPressWrapper';
+
+const getFirstElement = component => component.getInstance().firstElementRef.current;
+const getLastElement = component => component.getInstance().lastElementRef.current;
+
+/**
+ * firstElement and lastElement tests
+ */
+
+test('<TabPressWrapper> with no children', () => {
```
nickhurlburt, line 14: Love testing the empty state!

plugin-hrm-form/src/components/TabPressWrapper.jsx
```diff
@@ -68,6 +82,30 @@ class TabPressWrapper extends Component {
...
+      } else if (hasChildren) {
+        return this.reduceChildren(element.props.children, comparator, result);
+      }
+
+      return result;
+    }, initialValue);
+
+  findMinTabIndex = children => this.reduceChildren(children, (tabIndex, minTabIndex) => tabIndex < minTabIndex, null);
+
+  findMaxTabIndex = children => this.reduceChildren(children, (tabIndex, maxTabIndex) => tabIndex > maxTabIndex, null);
```
nickhurlburt, line 107: Nice code reuse here

plugin-hrm-form/src/components/TabPressWrapper.jsx
```diff
@@ -27,35 +27,49 @@ class TabPressWrapper extends Component {
   static displayName = 'TabPressWrapper';
```
nickhurlburt, line 27: May be worth updating the comments above with how it handles duplicate tab indexes.

PR #83, opened 2020-04-21
Small strings updates for Version 0.4

nickhurlburt:
A few small text changes before release:
1. A typo on `ContactDetail`, "right" -> "rights"
Screen Shot 2020-04-21 at 9 21 08 AM
2. Drop the term "Customer" from the search form:
Screen Shot 2020-04-21 at 9 20 39 AM
3. Drop the term "Customer" from the chat window message, "Conversation started with customer":
Screen Shot 2020-04-21 at 9 20 21 AM

Helplines don't generally talk about the people who call in as customers, so we'll drop the term for now.

My biggest question on review is if I should make the `strings` change in another place.  `HrmFormPlugin#init` is getting quite big.

*Deployment*
* I've already deployed this to production so we can get screenshots for instructions
* After review I'll merge to master, then merge down to staging, then deploy staging.

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -11,7 +11,7 @@ import HrmTheme from './styles/HrmTheme';
 import { channelTypes } from './states/DomainConstants';
 
 const PLUGIN_NAME = 'HrmFormPlugin';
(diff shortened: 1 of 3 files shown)


nickhurlburt:
Great, I'll merge for now, but I created an issue to revisit factoring this out.

PR #85, opened 2020-04-23
Queue viz: Contacts Waiting component

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-64

- Adds a component that shows how many tasks are waiting for each channel, for each queue
Screenshot from 2020-04-23 13-31-25


plugin-hrm-form/src/components/queuesStatus/QueueCard.jsx
```diff
@@ -0,0 +1,112 @@
...
+      60 * 1000,
+    );
+
+    this.setState({ intervalId });
+  };
+
+  render() {
+    const { qName, colors, facebook, sms, voice, web, whatsapp, longestWaitingDate } = this.props;
+    const { voiceColor, smsColor, facebookColor, whatsappColor, webColor } = colors;
+    const waitingMinutesMsg = longestWaitingDate === null ? 'none' : formatDistanceToNow(new Date(longestWaitingDate));
```
GPaoloni, line 72: In the docs for (this function)[https://date-fns.org/v2.2.1/docs/formatDistanceToNow] you can find all the possible formats  for the message to be used in the longest wait time. Is this ok?

nickhurlburt:
Something interesting -- when I went into taskrouter in the twilio console and deleted a pending task, the widget did not update from 1 waiting to 0.  When I reloaded the page, it went back to zero.

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -60,6 +62,37 @@ export default class HrmFormPlugin extends FlexPlugin {
       console.error('HRM base URL not defined, you must provide this to save program data');
     }
 
+    flex.AgentDesktopView.Panel1.Content.add(
```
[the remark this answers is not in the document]
GPaoloni: Right, moving it!

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -60,6 +62,37 @@ export default class HrmFormPlugin extends FlexPlugin {
...
+      <QueuesStatusWriter insightsClient={manager.insightsClient} key="queue-status-writer" />,
+      {
+        sortOrder: -1,
+        align: 'start',
+      },
+    );
+
+    // voice color right now is same as web color. Should this change?
+    const voiceColor = flex.DefaultTaskChannels.Chat.colors.main;
```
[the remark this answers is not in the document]
GPaoloni: Yes, it is a function that returns a color depending on the task status. Playing around with this I found a way to retrieve them but seems like they are all the same (#a0a8bd ):
Screenshot from 2020-04-24 10-58-44

Should change to this color?

plugin-hrm-form/src/components/search/ConnectDialog.jsx
```diff
@@ -71,11 +71,15 @@ const ConnectDialog = ({ anchorEl, currentIsCaller, contact, handleConfirm, hand
 
 ConnectDialog.displayName = 'ConnectDialog';
 ConnectDialog.propTypes = {
-  anchorEl: PropTypes.instanceOf(Element).isRequired,
+  anchorEl: PropTypes.instanceOf(Element),
```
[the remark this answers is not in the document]
GPaoloni: Are a consequence of previous merge, and the only "bad thing" was a warning when passing a `null` value (e.g. we are using `anchorEl = null` when popup should be hidden).

plugin-hrm-form/src/styles/queuesStatus/index.js
```diff
@@ -0,0 +1,75 @@
...
+  ${props => props.marginLeft && 'margin-left: 3px;'}
+`;
+
+export const ChannelBox = styled('div')`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 32px;
+  height: 26px;
+  background-color: ${props => transparentize('0.85', props.backgroundColor)};
```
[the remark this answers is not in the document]
GPaoloni: The opacity works for the component and all of their children, and thus the numbers inside the boxes are affected too. Instead of using this transparentize function I can also do this to add a "15% opacity alpha":
Screenshot from 2020-04-24 11-18-24
Should I change to this instead?

GPaoloni:
> Something interesting -- when I went into taskrouter in the twilio console and deleted a pending task, the widget did not update from 1 waiting to 0. When I reloaded the page, it went back to zero.

I've tried this, while keeping an eye to the plugin console. Not event was sent to my plugin instance when canceling nor when deleting a task. In fact, when canceling/deleting from Twilio console, the chats are not properly closed to customer either (no goodbye message sent to whatsapp nor to the chatline).

nickhurlburt:
>> Something interesting -- when I went into taskrouter in the twilio console and deleted a pending task, the widget did not update from 1 waiting to 0. When I reloaded the page, it went back to zero.

> I've tried this, while keeping an eye to the plugin console. No event was sent to my plugin instance when canceling nor when deleting a task. In fact, when canceling/deleting from Twilio console, the chats are not properly closed to customer either (no goodbye message sent to whatsapp nor to the chatline).

Right, I suspect that the insights client doesn't actually handle deleted tasks.  Deleting a task is kind of like `Ctrl+C` -- it's a disruptive action and things don't get cleaned up well.  It's also an edge case that I don't think we have to worry about.  But it's good to remember what the behavior is so it doesn't surprise us.

plugin-hrm-form/src/styles/queuesStatus/index.js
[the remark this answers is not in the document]
nickhurlburt: I'll leave it up to you.  Looks like a tradeoff between adding more stuff into our build that our users have to download, and code clarity/conciseness.  There will never be a clear answer.  Your choice.  Here are a few more possible options to consider:
* Use ``${props => `${props.backgroundColor}{0.15*256}`};`` to make it more self-documenting
* add a comment
* define a util function that does the same

plugin-hrm-form/src/HrmFormPlugin.js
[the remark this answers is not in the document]
nickhurlburt: It's strange that `flex.DefaultTaskChannels.Call.colors.main` is a function but `flex.DefaultTaskChannels.Chat.colors.main` is an object.  Though I did work out that `flex.DefaultTaskChannels.Call.colors.main('Pending')` (or other states) works.  And yes, it all returns `a0a8bd`.  My guess is that there's some inconsistency due to some legacy code.

nickhurlburt:
Further updates:
- Need to figure out why it's not tracking voice tasks (and there's no action sent when the caller hangs up while still in 'pending')
- There are some small design edits coming
- Can you add in a change that will cause an active voice call to appear below the All Tasks bar?  I think we would just set `staticTaskFilter` to always return `false`.  See https://www.twilio.com/docs/flex/ui/configuration#tasklistcontainer.

nickhurlburt:
I think there's something off with the longest wait time calculation.  Create a task with SMS/chat/whatever, let it get assigned to you but don't accept it.  After 2 minutes, you will be set offline and it will go back to 'pending'.  But its "waiting time" will only start from when it was moved back to 'pending', even though it has still been waiting.  So it will be behind by 2 minutes. 

The supervisor view captures the longest wait time correctly.  Are we getting enough information from Twilio to make this correct?

nickhurlburt:
Please also address this item from a previous comment:
* Can you add in a change that will cause an active voice call to appear below the All Tasks bar? I think we would just set `staticTaskFilter` to always return `false`. See https://www.twilio.com/docs/flex/ui/configuration#tasklistcontainer.  I edited the comment to add it, so it may not have come through email to you.

nickhurlburt:
Another issue I'm seeing with respect to accessibility is that there is no way to tab into the Contacts Waiting widget, at least as far as I can tell.  This is something that can be brought up in our meeting tomorrow.

nickhurlburt:
I left a number of comments above.  It's good to see the new zero-color added, and good to see the styling now updated, as well as being able to track pending voice calls.  I think the full list of remaining outstanding issues is:
- Voice tasks don't go down on user hang up
- Move active voice call task item below 'All Tasks'
- Longest wait time isn't correct when a task is reserved but never accepted, and goes back to pending
- Address accessibility.  We should at least add an axe unit test using the original example.  I think we want to be able to tab into it or have some way for a screen reader to access this.

After that I think we're good.  A follow-on, separate issue and PR will have to handle:
- Only showing certain queues based on the counselor's helpline and configuration
- Removing the collapse/expand chevron.

plugin-hrm-form/src/styles/GlobalOverrides.js
```diff
@@ -0,0 +1,10 @@
+import { injectGlobal } from 'react-emotion';
+
+// eslint-disable-next-line no-unused-expressions
+injectGlobal`
+  div.Twilio-ModalPopupWithEntryControl {
```
nickhurlburt, line 5: What does this affect?

plugin-hrm-form/src/styles/HrmTheme.js
```diff
@@ -2,6 +2,7 @@ const HrmTheme = {
   baseName: 'HrmTheme',
   colors: {
     base2: '#f6f6f6',
+    // base4: '#BDC3D2',
```
nickhurlburt, line 5: Do we need to keep this comment?

plugin-hrm-form/src/styles/queuesStatus/index.js
```diff
@@ -0,0 +1,79 @@
...
+`;
+
+export const HeaderContainer = styled(ButtonRow)`
+  width: 100%;
+  justify-items: flex-start;
+  background-color: ${props => props.theme.colors.base2};
+  border-radius: 2px;
+  border-style: solid;
+  border-width: 1px 0px 1px 0px;
+  border-color: ${props => props.theme.colors.base4};
```
nickhurlburt, line 19: Does this need to be in sync with https://github.com/tech-matters/flex-plugins/pull/85/files#diff-56d568040a1945fa74fab27ef71a0e6dR5 ?

plugin-hrm-form/src/components/queuesStatus/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+  render() {
+    const { queuesStatus, error } = this.props.queuesStatusState;
+    const { expanded } = this.state;
+    return (
+      <Container role="complementary">
+        <HeaderContainer
+          onClick={this.handleExpandClick}
+          role="button"
+          aria-label={`Contacts waiting ${this.state.expanded ? 'press to collapse' : 'press to expand'}`}
```
nickhurlburt, line 54: Yay!

plugin-hrm-form/src/components/queuesStatus/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+      <ArrowDropDownTwoTone style={{ padding: 0, fontSize: 18 }} />
+    );
+  };
+
+  render() {
+    const { queuesStatus, error } = this.props.queuesStatusState;
+    const { expanded } = this.state;
+    return (
+      <Container role="complementary">
+        <HeaderContainer
```
nickhurlburt, line 51: We should change this to match the All Tasks style.

plugin-hrm-form/src/___tests__/components/queuesStatus/QueuesStatus.test.js
```diff
@@ -0,0 +1,187 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import renderer from 'react-test-renderer';
+import configureMockStore from 'redux-mock-store';
+import { mount } from 'enzyme';
+import { configureAxe, toHaveNoViolations } from 'jest-axe';
+import { StorelessThemeProvider } from '@twilio/flex-ui';
+
+// import '../../mockStyled';
```
nickhurlburt, line 9: Can we remove the comment?

plugin-hrm-form/src/___tests__/components/queuesStatus/QueuesStatus.test.js
```diff
@@ -0,0 +1,187 @@
...
+      </Provider>
+    </StorelessThemeProvider>,
+  ).root;
+  expect(() => component.findByType(QueueCard)).not.toThrow();
+  expect(() => component.findByType(ErrorText)).not.toThrow();
+  const ErrorProps = component.findByType(ErrorText).props;
+  expect(ErrorProps.children).toBe(queuesStatusState.error);
+});
+
+test('a11y', async () => {
```
nickhurlburt, line 146: yay!

plugin-hrm-form/src/___tests__/mockStyled.js
```diff
@@ -17,6 +17,9 @@ jest.mock('../styles/HrmStyles', () => ({
   Row: 'Row',
   FontOpenSans: 'FontOpenSans',
   Box: 'Box',
+  ErrorText: 'ErrorText',
+  ButtonRow: 'ButtonRow',
+  HiddenText: 'HiddenText',
```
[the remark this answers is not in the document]
GPaoloni: Actually I'm using the the enzyme way in the test, but I forgot to delete this. Thanks for pointing it

plugin-hrm-form/src/styles/GlobalOverrides.js
GPaoloni: This is the class that  overrides the "ALL TASKS" container. By importing this file in HrmFormPluggin we can override the Twilio styles.

PR #86, opened 2020-04-24
Implement accessibility test for NonDataCallTypeDialog

murilovmachado:
This PR:
- Fix Accessibility issue of button without description on `NonDataCallTypeDialog`
- Creates an accessibility unit test for `NonDataCallTypeDialog`

Details:
1. `jest-axe` expects a DOM element as a parameter, so I had to use `enzyme` instead of `react-test-renderer` for this kind of test.
2. Notice that using Material's `StorelessThemeProvider` with our `HrmTheme` I didn't need to mock the `styled components`. I believe we can use the same approach in the other unit tests. Mocking `styled components` is currently a pain point in our development process.

### DIFF
--- a/plugin-hrm-form/package-lock.json
+++ b/plugin-hrm-form/package-lock.json
@@ -2816,6 +2816,12 @@
       "resolved": "https://registry.npmjs.org/aws4/-/aws4-1.9.1.tgz",
       "integrity": "sha512-wMHVg2EOHaMRxbzgFJ9gtjOOCrI80OHLG14rxi28XwOW8ux6IiEbRCGGGqCtdAIg4FQCbW20k9RsT4y3gJlFug=="
     },
+    "axe-core": {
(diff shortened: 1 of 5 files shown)


murilovmachado:
> Looks great, though I can't find a reference to `StorelessThemeProvider` anywhere. Where does the `Storeless` originate from?

@nickhurlburt I couldn't find anything online about `StorelessThemeProvider` as well. I was experimenting with `@twilio/flex-ui` `ThemeProvider`, but I was getting complaints about undefined `store`. So I've noticed that this lib also exported an `StorelessThemeProvider`, and it fitted our needs, it worked like charm.

nickhurlburt:
@murilovmachado oh, interesting, perhaps `ThemeProvider` keeps the theme in redux or some other store.  I've noticed before that some components seem to magically pick up a theme without any context or props, and that might explain it.  Good find!

PR #88, opened 2020-04-27
Localization via Twilio strings

GPaoloni:
This PR adds a way of localizing our app.
The parts that compose this are: 
- Flex UI localization (translating what a counselor see)
  The language can be configured in several ways (descending order of precedence):
  1. `counselorLanguage` property, from worker attributes
  2. `helplineLanguage` property, from worker attributes
  3. `configuredLanguage` property, fom serviceConfiguration attributes

  If none of this is provided, the UI will use the `defaultLanguage` (`en-US`).
  If a language different from `defaultLanguage` is provided, then it will be fetched from our 
  /getTranslation function (Serverless) on plugin initialization. If the translation exists, the UI is translated.
  The strings are stored as Twilio strings and can be accesed:
  1. using the `Template` component from Twilio library (e.g.: `<Template code="Some localized string"/>`)
  2. using` LocalizationContext` (wrapper that have acces to the `manager.strings`)

- System messages localization (translating the messages that are sent to a chat-based contact)
Right now the only localized message is "good bye message" (sent to the contact when the counselor leaves the chat).
  The language can be configured in several ways (descending order of precedence):
  1. `language` property, from task attributes
  2. `helplineLanguage` property, from worker attributes
  3. `configuredLanguage` property, fom serviceConfiguration attributes

  If none of this is provided, the message will use the `defaultLanguage` (`en-US`).
  If a language different from `defaultLanguage` is provided, then it will be fetched from our /getMessages function (Serverless) each time it is required. If the translation exists, send the translated message, and if it doesn't, the message will be the in the `defaultLanguage`


plugin-hrm-form/src/translations/en/translation.json
```diff
@@ -0,0 +1,240 @@
+{
```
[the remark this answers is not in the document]
GPaoloni: Yes, but will only work for the English ones, and as soon as we change e.g. from English to Spanish, when we want go back to English we will need to override all the Spanish strings, so I think we need to have them in the English translation too.

plugin-hrm-form/src/translations/index.js
```diff
@@ -0,0 +1,10 @@
+const translations = {
```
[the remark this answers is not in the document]
GPaoloni: Sorry, old code that shouldn't be there anymore, it was the starting point before I moved to json!

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -7,11 +7,17 @@ import reducers, { namespace } from './states';
 import { Actions } from './states/ContactState';
 import ConfigurationContext from './contexts/ConfigurationContext';
 import LocalizationContext from './contexts/LocalizationContext';
+import Translator from './components/translator';
+import SettingsSideLink from './components/sideLinks/SettingsSideLink';
 import HrmTheme from './styles/HrmTheme';
 import { channelTypes } from './states/DomainConstants';
+import { defaultLanguage, changeLanguage } from './states/ConfigurationState';
+import { configuredLanguage } from './private/secret';
```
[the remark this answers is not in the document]
GPaoloni: `defaultLanguage` is needed for the initial mount, and it will allways be en-US (or the one that we bundle with the code).  `configuredLanguage` is the one that we can use to override the default within the code. I'm not sure if we should have this but it was a way of showing that a helpline can choose a default language via attributes or via code. Should I remove it?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -48,23 +70,15 @@ export default class HrmFormPlugin extends FlexPlugin {
...
-    strings.ChatWelcomeText = 'Conversation started';
-
     // TODO(nick): Eventually remove this log line or set to debug
     console.log(`HRM URL: ${hrmBaseUrl}`);
     if (hrmBaseUrl === undefined) {
       console.error('HRM base URL not defined, you must provide this to save program data');
     }
 
+    // utilities for developers only
+    if (manager.store.getState().flex.worker.attributes.helpline === '') addDeveloperUtils(flex, manager, translateUI);
```
nickhurlburt, line 80: Can it also be undefined?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -102,7 +116,7 @@ export default class HrmFormPlugin extends FlexPlugin {
         value={{ hrmBaseUrl, serverlessBaseUrl, workerSid, helpline, currentWorkspace, getSsoToken }}
         key="custom-crm-container"
       >
-        <LocalizationContext.Provider value={{ strings, isCallTask }}>
+        <LocalizationContext.Provider value={{ strings: manager.strings, isCallTask }}>
```
nickhurlburt, line 119: Is `LocalizationContext` still the right name for this?  Is there a better name to distinguish what this does than what the localization system does?  I'm still a little confused on the difference.

plugin-hrm-form/src/components/callTypeButtons/CallTypeButtons.jsx
```diff
@@ -29,14 +29,14 @@ const CallTypeButtons = props => {
             <Box width="50px" marginRight="5px">
               <FaceIcon />
             </Box>
-            {callTypes.child}
+            <Template code="CallType-child" />
```
nickhurlburt, line 32: cool

plugin-hrm-form/src/utils/pluginHelpers.js
```diff
@@ -0,0 +1,95 @@
...
+    return defaultMessages.GoodbyeMsg;
+  } catch (err) {
+    window.alert(translationErrorMsg);
+    console.error(translationErrorMsg, err);
+    return defaultMessages.GoodbyeMsg;
+  }
+};
+
+/**
+ *
```
nickhurlburt, line 58: Could you add a little more explanation in the comment here as to what this is doing?  The variable names aren't descriptive enough to make it clear from the code.

plugin-hrm-form/src/HrmFormPlugin.js
GPaoloni: You are right, thanks for the catch!

plugin-hrm-form/src/utils/pluginHelpers.js
```diff
@@ -0,0 +1,101 @@
...
+    return defaultMessages.GoodbyeMsg;
+  } catch (err) {
+    window.alert(translationErrorMsg);
+    console.error(translationErrorMsg, err);
+    return defaultMessages.GoodbyeMsg;
+  }
+};
+
+/**
+ * WARNING: the way this is done right now is "hacky", as it changes an object reference (setNewStrings) and then forces a re-render (afterNewStrings). The safe way of doing this would be 1) async init method 2) having acces to a function that updates the ContextProvider state that wraps the entire app. A fallback is to move translations within the code (avoiding the asynchronus operation).
```
nickhurlburt, line 60: spelling: `access`, `asynchronous`

plugin-hrm-form/src/utils/pluginHelpers.js
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * Receives localization config object and initial language. Based on this, translates de UI
```
nickhurlburt, line 62: spelling: `the`

plugin-hrm-form/src/HrmFormPlugin.js
nickhurlburt: Looks like this still needs updating to handle undefined?

PR #90, opened 2020-05-06
Which queues does the counselor see?

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-98

This PR changes the queues that each counselor see in queue viz widget.

If counselor has a helpline, shows the queue that matches that helpline (Desing in this case)
Screenshot from 2020-05-06 16-53-59

If counselor has empty or no helpline, or if counselor has a helpline, but there is no queue with that name,counselor see the Admin queue
Screenshot from 2020-05-06 16-56-30

By changing the [counselorQueues](https://github.com/tech-matters/flex-plugins/blob/a0696c25689068f944b19745ad3ac2f8a3c8eb9c/plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx#L39) to an array of more than one queue, the counselor can subscribe to multiple queues (even when not eligible for task assignment in that queue)
Screenshot from 2020-05-07 11-12-18

### DIFF
--- a/plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx
+++ b/plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx
@@ -7,14 +7,14 @@ import { queuesStatusUpdate, queuesStatusFailure } from '../../states/QueuesStat
 import { initializeQueuesStatus, getNewQueuesStatus } from './helpers';
 import { namespace, queuesStatusBase } from '../../states';
 
-class QueuesStatusWriter extends React.Component {
+export class QueuesStatusWriter extends React.Component {
   static displayName = 'QueuesStatusWriter';
 
   static propTypes = {
     insightsClient: PropTypes.shape({
       liveQuery: PropTypes.func,
     }).isRequired,
-    helpline: PropTypes.string.isRequired,
+    helpline: PropTypes.string,
     queuesStatusState: PropTypes.shape({
       queuesStatus: PropTypes.shape({}),
       error: PropTypes.string,
@@ -23,6 +23,15 @@ class QueuesStatusWriter extends React.Component {
     queuesStatusUpdate: PropTypes.func.isRequired,
     queuesStatusFailure: PropTypes.func.isRequired,
   };
+
+  static defaultProps = {
+    helpline: undefined,
+  };
+
+  constructor(props) {
+    super(props);
+    this.updateQueuesState = this.updateQueuesState.bind(this);
+  }
 
   state = {
     tasksQuery: null,
@@ -42,33 +51,27 @@ class QueuesStatusWriter extends React.Component {
 
       const tasksQuery = await this.props.insightsClient.liveQuery('tr-task', '');
 
-      const updateQueuesState = () => {
-        const tasks = tasksQuery.getItems();
-        const queuesStatus = getNewQueuesStatus(cleanQueuesStatus, tasks);
-        this.props.queuesStatusUpdate(queuesStatus);
-      };
+      this.updateQueuesState(tasksQuery, cleanQueuesStatus);
+      this.setState({ tasksQuery });
 
       const shouldUpdate = status => status === 'pending' || status === 'assigned' || status === 'canceled';
-
-      updateQueuesState();
-      this.setState({ tasksQuery });
 
       tasksQuery.on('itemUpdated', args => {
         console.log('TASK UPDATED', args);
         // eslint-disable-next-line camelcase
         const { status, queue_name } = args.value;
         if (counselorQueues.includes(queue_name) && shouldUpdate(status)) {
-          updateQueuesState();
+          this.updateQueuesState(tasksQuery, cleanQueuesStatus);
         }
       });
 
       tasksQuery.on('itemRemoved', args => {
-        updateQueuesState();
+        this.updateQueuesState(tasksQuery, cleanQueuesStatus);
       });
     } catch (err) {
       const error = "Error, couldn't subscribe to live updates";
       this.props.queuesStatusFailure(error);
-      console.log(error, err);
+      console.error(error, err);
     }
   }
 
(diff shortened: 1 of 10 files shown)


plugin-hrm-form/src/styles/HrmStyles.js
```diff
@@ -250,11 +250,13 @@ export const Row = styled('div')`
...
-  flex-direction: row;
-  align-items: center;
-`;
+/*
+ * export const ButtonRow = styled('button')`
+ *   display: flex;
+ *   flex-direction: row;
+ *   align-items: center;
+ * `;
+ */
```
GPaoloni, line 259: I left this because we will need it for addressing accessibility on search contact (commented for lint reasons)

plugin-hrm-form/src/styles/HrmStyles.js
```diff
@@ -250,11 +250,13 @@ export const Row = styled('div')`
   align-items: center;
 `;
 
-export const ButtonRow = styled('button')`
-  display: flex;
-  flex-direction: row;
-  align-items: center;
-`;
+/*
```
nickhurlburt, line 253: If it's not needed, can we remove it?

plugin-hrm-form/src/components/queuesStatus/index.jsx
```diff
@@ -10,68 +8,52 @@ import { Container, HeaderContainer, QueuesContainer } from '../../styles/queues
...
+    <Container role="complementary" tabIndex={0}>
+      <HeaderContainer>
+        <Box marginTop="12px" marginRight="5px" marginBottom="12px" marginLeft={TLHPaddingLeft}>
+          Contacts waiting
+        </Box>
+      </HeaderContainer>
+      <QueuesContainer>
+        {error && <ErrorText>{error}</ErrorText>}
+        {queuesStatus &&
+          (helpline && queuesStatus[helpline] ? (
```
nickhurlburt, line 24: Interesting.  My initial thought is that we would filter out which queues and tasks we follow [here](https://github.com/tech-matters/flex-plugins/blob/staging/plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx#L33) and [here](https://github.com/tech-matters/flex-plugins/blob/staging/plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx#L40), or at least in those areas, and then leave this component unchanged, as `queuesStatus` would only hold what we want to show.  But my initial thought may very possibly be wrong.  What lead you to prefer this approach?

GPaoloni: The refactor on this component is mainly because if we don't want the expand/collapse anymore (we don't right?) this component can be a stateless one. For filtering the queues, I've implemented your suggestion as it will lead to less state updates, thanks!

plugin-hrm-form/src/styles/HrmStyles.js
GPaoloni: I was thinking that this might be better for using [in here](https://github.com/tech-matters/flex-plugins/blob/a0696c25689068f944b19745ad3ac2f8a3c8eb9c/plugin-hrm-form/src/components/search/ContactDetails/Section.jsx#L53), as this acts as `Row`, can be used as a button and it preserves the accessibility benefits of button. I'll remove it if doesn't make sense tho

nickhurlburt: Sure, sounds fine.  Generally, I don't think we should keep commented-out code around (since it does still exist in the commit history) but if you want to keep it for a limited time, that's fine.

plugin-hrm-form/src/components/queuesStatus/index.jsx
nickhurlburt: Yes, we don't want expand/collapse.  I understand, yes changing this to being stateless is a great idea.  I didn't mean to say we can't make any edits to this component, and changing it for that reason makes a lot of sense.  My question was really on the filtering.

plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx
```diff
@@ -60,7 +67,7 @@ class QueuesStatusWriter extends React.Component {
       });
```
nickhurlburt, line 67: Do we need to add the filtering in L60 here as well?

plugin-hrm-form/src/components/queuesStatus/helpers.js
```diff
@@ -14,21 +14,23 @@ export const newQueueEntry = {
...
 export const initializeQueuesStatus = queues =>
-  Object.values(queues)
-    // eslint-disable-next-line no-nested-ternary
-    .sort((a, b) => (a.queue_name < b.queue_name ? -1 : a.queue_name > b.queue_name ? 1 : 0))
-    .reduce((acc, q) => ({ ...acc, [q.queue_name]: newQueueEntry }), {});
+  // eslint-disable-next-line no-nested-ternary
+  queues.sort((a, b) => (a < b ? -1 : a > b ? 1 : 0)).reduce((acc, qName) => ({ ...acc, [qName]: newQueueEntry }), {});
+
+const isNotWaiting = status => status !== 'pending' && status !== 'reserved';
+const suscribedToQueue = (queue, queues) => Boolean(queues[queue]);
```
nickhurlburt, line 25: spelling: `subscribed`

plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx
nickhurlburt: I see.  There's a tradeoff here.  We can keep as it is, and we may do more updates than we need when a task is removed that isn't in our queue.  Or we add a list of tasks to track to state, and we avoid unnecessary updates, but we have some more complexity in our code and we have to do a lookup every time (and that may be a little fragile since we can never miss an event).

I think we could go with either option.  Which do you think is best?

GPaoloni: I like the idea of avoiding unnecessary calls to `updateQueuesState` as that maps over all of the tasks and dispatches updates to the redux store. So, I'll rather give that extra complexity a chance

nickhurlburt: Sounds good!

PR #91, opened 2020-05-18
[WIP] Implement call to setTaskConversationsAttributes

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-133
Backend: https://github.com/tech-matters/serverless/pull/7

Insights data are being save at `beforeCompleteTask` listener.

Screen Shot 2020-05-25 at 18 28 47

Screen Shot 2020-05-25 at 18 28 41

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -5,7 +5,7 @@ import { FlexPlugin } from 'flex-plugin';
 import CustomCRMContainer from './components/CustomCRMContainer';
 import QueuesStatus from './components/queuesStatus';
 import QueuesStatusWriter from './components/queuesStatus/QueuesStatusWriter';
-import reducers, { namespace } from './states';
+import reducers, { namespace, contactFormsBase } from './states';
 import { Actions } from './states/ContactState';
 import ConfigurationContext from './contexts/ConfigurationContext';
 import LocalizationContext from './contexts/LocalizationContext';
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/services/ServerlessService.js
```diff
@@ -38,3 +40,34 @@ export const getMessages = async (configuration, body) => {
...
+  const conversations = {
+    conversation_attribute_1: callType,
+  };
+
+  let customers = {};
+  if (hasCustomerData) {
+    const { childInformation } = task;
+    customers = {
+      gender: childInformation.gender.value,
+      year_of_birth: childInformation.age.value,
```
nickhurlburt, line 57: There are two issues with this: 1) sounds like year of birth isn't showing up and 2) this isn't the year of birth.  If year of birth isn't showing up, should we use a custom field for this?  Or do you have a different plan?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -135,8 +136,17 @@ export default class HrmFormPlugin extends FlexPlugin {
...
     });
 
-    flex.Actions.addListener('beforeCompleteTask', (payload, abortFunction) => {
+    const saveInsights = async payload => {
+      const { taskSid } = payload.task;
+      const task = manager.store.getState()[namespace][contactFormsBase].tasks[taskSid];
+      const configuration = { serverlessBaseUrl, currentWorkspace, getSsoToken };
+
+      await saveInsightsData(configuration, task, taskSid);
+    };
```
nickhurlburt, line 145: Not specific to this PR -- but this just reminds me that we do need to refactor this overall function.  It's huge.

plugin-hrm-form/src/services/ServerlessService.js
murilovmachado: 1) Ideally, we should contact Twilio and check why it's not showing up on Insights. If we find that it's not gonna be there soon, then we should use a custom field.
2) If we go with the custom option, we don't need to worry. If it's really going to be Year Of Birth, then we need to explicitly document we're actually storing the age, or calculate the year of birth.

plugin-hrm-form/src/HrmFormPlugin.js
murilovmachado: completely agree!

PR #95, opened 2020-05-21
[WIP] Exploring task transfers

GPaoloni:
Added initial approach for warm and cold transfers, for chat and call based tasks


plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -1,22 +1,109 @@
...
+  const workerSid = manager.workerClient.sid;
+  const { helpline } = manager.workerClient.attributes;
+  const currentWorkspace = manager.serviceConfiguration.taskrouter_workspace_sid;
+  const { token } = manager.user;
+
+  return { hrmBaseUrl, serverlessBaseUrl, workerSid, helpline, currentWorkspace, token };
+};
+
+const setUpComponents = () => {
+  Flex.TaskCanvasHeader.Content.add(<TransferButton key="transfer-button" />, {
```
nickhurlburt, line 45: We wouldn't want to show this on a voice task because we have the existing transfer button, I think.

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -1,22 +1,109 @@
...
+  Flex.TaskCanvasHeader.Content.add(<CompleteTransferButton key="complete-transfer-button" />, {
+    sortOrder: 1,
+    if: props => showTransferControls(props.task),
+  });
+
+  Flex.TaskCanvasHeader.Content.add(<RejectTransferButton key="reject-transfer-button" />, {
+    sortOrder: 1,
+    if: props => showTransferControls(props.task),
+  });
+};
```
nickhurlburt, line 63: I'm not sure what we want here if this is a voice task.  Perhaps the button to hang up on Counselor 1 can do the job of `Complete Transfer` and the button to hang up yourself does the job of `Reject Transfer`.

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -1,22 +1,109 @@
...
+    ...payload.task.attributes,
+    transferMeta: {
+      originalTask: payload.task.taskSid,
+      originalReservation: payload.task.sid,
+      originalCounselor: payload.task.workerSid,
+      transferStatus: mode === transferModes.warm ? transferStatuses.transferring : transferStatuses.completed,
+      formDocument: null, // the name of the Sync document where the form is stored (TODO: save it before this)
+    },
+  };
+  await payload.task.setAttributes(updatedAttributes);
```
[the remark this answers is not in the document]
GPaoloni: Yes, both do the same thing. The one I'm using here is a method provided by ITask interface, within Flex. The other way of setting attributes requires hitting a serverless endpoint. I've chosen this one to avoid one extra fetch, but easily can change it to match the serverless approach! Should I?

GPaoloni:
Closing PR as it was already implemented in a more organized and polished way (#99, #101, #106 and #109).

PR #97, opened 2020-05-28
Changes to Gender and Age on Insights

murilovmachado:
This PR:
- Updates Select Gender Options
- Updates Select Age Options
- Saves `age` as `Customer Attribute 1` (instead of `Year of Birth`)


plugin-hrm-form/src/components/tabbedForms/SelectOptions.js
```diff
@@ -0,0 +1,2 @@
+export const genderOptions = ['Boy', 'Girl', 'Non-binary', 'Unknown'];
+export const ageOptions = ['0-3', '4-6', '7-9', '10-12', '13-15', '16-17', '18-25', '>25', 'Unknown'];
```
nickhurlburt, line 2: This gives me the idea that we could use something like @GPaoloni's localization design (with the JSON files for local strings) to customize form fields and options for different helplines.  Default fields and options could be specified by default, then we have modifications with helpline-specific JSON.  No action to take now.  Just a thought.

PR #98, opened 2020-05-28
Use setAttributes instead of Serveless Function

murilovmachado:


plugin-hrm-form/src/services/InsightsService.js
```diff
@@ -0,0 +1,46 @@
...
+      ...conversations,
+    },
+    customers: {
+      ...previousAttributes.customers,
+      ...customers,
+    },
+  };
+}
+
+export async function saveInsightsData(twilioTask, task) {
```
GPaoloni, line 39: Maybe change the name of `task` to `form` / `taskForm` or something like that? Just to emphasize where this is coming from.

PR #99, opened 2020-05-28
HrmFormPlugin refactor

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-140
Refactored HrmFlexPlugin.js file, isolating the code according to concerns:
- added getConfig function -> replaces ConfigurationContext
- setUpLocalization -> localization related code
- setUpComponents -> here we add/replace/remove dynamic components to flex ui
- setUpActions -> here we add/replace/remove actions for the flex actions framework


plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -19,6 +18,183 @@ import { saveInsightsData } from './services/ServerlessService';
...
+
+  const onCompleteTask = async (sid, task) => {
+    if (task.status !== 'wrapping') {
+      if (task.channelType === channelTypes.voice) {
+        await Flex.Actions.invokeAction('HangupCall', { sid, task });
+      } else {
+        await Flex.Actions.invokeAction('WrapupTask', { sid, task });
+      }
+    }
+    // should we await here?
```
GPaoloni, line 109: I'm not sure of this, but based on the recent issue with asynchronus code, should we add and `await` statement to this action?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -19,6 +18,183 @@ import { saveInsightsData } from './services/ServerlessService';
...
+    Flex.Actions.invokeAction('CompleteTask', { sid, task });
```
GPaoloni, line 110: I'm not sure of this, but based on the recent issue with asynchronus code, should we add and await statement to this action?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -19,6 +18,183 @@ import { saveInsightsData } from './services/ServerlessService';
...
+  };
+
+  // TODO(nick): Can we avoid passing down the task prop, maybe using context?
```
GPaoloni, line 113: I think this is old comment. Should I remove it? @nickhurlburt

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -19,6 +18,183 @@ import { saveInsightsData } from './services/ServerlessService';
...
+
+  const saveInsights = async payload => {
+    const { taskSid } = payload.task;
+    const task = manager.store.getState()[namespace][contactFormsBase].tasks[taskSid];
+
+    await saveInsightsData(task, taskSid);
+  };
+
+  Flex.Actions.addListener('beforeCompleteTask', async (payload, abortFunction) => {
+    manager.store.dispatch(Actions.saveContactState(payload.task, abortFunction, hrmBaseUrl, workerSid, helpline));
```
GPaoloni, line 148: Question: wouldn't be better to move this to the "onComplete" function above (to avoid depending on the event)? And this needs to be a redux action? Is not my intention to start changing everything haha, but I was following the way of this and maybe it could be simpler if we do this as a "contact service". Am I missing something here?

GPaoloni: PS: I actually noticed this because of this comment
Screenshot from 2020-05-28 18-55-39

plugin-hrm-form/src/HrmFormPlugin.js
nickhurlburt: Yes, thanks!

plugin-hrm-form/src/HrmFormPlugin.js
GPaoloni: Ha, just found this issue https://github.com/tech-matters/flex-plugins/issues/31, so the answer is yes.

nickhurlburt: That's pretty old code, and it's probably not the best way to do it anymore.  So I'd be totally fine with changing it.  It may be best to save that for a different PR to reduce the complexity of this one, though.

plugin-hrm-form/src/components/callTypeButtons/NonDataCallTypeDialog.jsx
```diff
@@ -24,6 +24,7 @@ const NonDataCallTypeDialog = ({ isOpen, isCallTask, handleConfirm, handleCancel
         <Box marginBottom="32px">
           <Row>
             <ConfirmButton autoFocus tabIndex={1} onClick={handleConfirm}>
+              {/* eslint-disable-next-line react/jsx-max-depth */}
```
nickhurlburt, line 27: Yay!

plugin-hrm-form/src/HrmFormPlugin.js
nickhurlburt: Good question.  I think it's the very last thing we do with the task, so it's fine if things are getting cleaned up afterward.  I don't see a reason to have it await now, but it's something to keep an eye on.

PR #100, opened 2020-05-29
Implement thematic design changes

nickhurlburt:
This is WIP as I plan to wait for #99 before rebasing this and submitting it as a full PR for review.

These are @GPaoloni's original changes with two more things after feedback from product/design:
- Only removing the Twilio logo from the top nav, and not the mute button or user controls
- Adding back in the "Conversation started" string for the chat welcome text, which went missing somehow.

The view with a task now looks like this:
image

The view without a task looks a bit weird now, with 'No selected tasks' floating in the middle with no divider to show that it's part of a panel.  I intend to remove the `NoTaskView` component altogether, but I want to do it in a subsequent PR, so that it's easier to revert if we want to put it back to show a more interesting empty state.

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -235,8 +235,6 @@ export default class HrmFormPlugin extends FlexPlugin {
     }
 
     flex.MainHeader.Content.remove('logo');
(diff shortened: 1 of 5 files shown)


nickhurlburt:
I've rebased and moved lines to match the new refactoring.  Will merge and deploy.

PR #101, opened 2020-06-02
Add transfer task functionality

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-141
This PR is related to https://github.com/tech-matters/serverless/pull/10

Adds the transfer task functionality by adding 3 buttons to TaskHeaderCanvas
- Transfer button: present in all kind of tasks, but only if a) the task is accepted b) it is not being transferred
Screenshot from 2020-06-02 17-13-25
This button will open Worker Directory, where a counselor can choose to transfer to a worker or to a queue, as well as between cold or warm transfer.
- Accept transfer button: transfers the control of the task to the 2nd counselor.
- Reject transfer button: returns the control of the task to the 1st counselor.
Screenshot from 2020-06-02 17-30-24
Note both will appear to the 2nd counselor only, and for warm transfers only. The form is not being blocked (for testing purpose), instead a message is shown above the categories panel.
Screenshot from 2020-06-03 18-00-16

Details on how transfer is working: https://app.lucidchart.com/documents/edit/2234a292-3fda-4c2b-9712-2f1f11236c25/0_0?shared=true


plugin-hrm-form/src/___tests__/helpers.js
```diff
@@ -0,0 +1,136 @@
...
+ *  unhold: () => Promise<ReturnType<createTask>>
+ *  holdParticipant: () => Promise<ReturnType<createTask>>
+ *  unholdParticipant: () => Promise<ReturnType<createTask>>
+ *  cancelTransfer: () => Promise<ReturnType<createTask>>
+ *  kick: () => Promise<ReturnType<createTask>>
+ *  setAttributes: () => Promise<ReturnType<createTask>>
+ *  wrapUp: () => Promise<ReturnType<createTask>>
+ * }}
+ */
+export function createTask(attributes = {}, namesNsids = {}) {
```
GPaoloni, line 59: This function is just a helper used in the following test file. It's factored out to reduce the size of the test file and because it might be useful in the future.

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -0,0 +1,216 @@
...
+  const transformed = [...str].map(char =>
+    shouldReplaceChar(char)
+      ? `_${char
+          .charCodeAt(0)
+          .toString(16)
+          .toUpperCase()}`
+      : char,
+  );
+  return transformed.join('');
+};
```
GPaoloni, line 158: @nickhurlburt here I'm doing some transform to match the way identities are being used in channel members. Should we contact Twilio support make sure this is the transformation being done?

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -0,0 +1,219 @@
+import { Actions, ITask, TaskHelper, StateHelper } from '@twilio/flex-ui';
+
+import { transferChatStart, transferChatResolve } from '../services/ServerlessService';
+import { transferStatuses, transferModes } from '../states/DomainConstants';
+import { getConfig } from '../HrmFormPlugin';
+
+/**
+ * @param {ITask} task
+ */
+export const tranferStarted = task => Boolean(task.attributes.transferMeta);
```
nickhurlburt, line 10: To keep this similar in naming to the others, perhaps `hasTransferStarted(task)`?

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * @param {ITask} task
+ */
+export const isOriginal = task =>
+  Boolean(task.attributes.transferMeta && task.attributes.transferMeta.originalReservation === task.sid);
```
nickhurlburt, line 16: Should we call this `isOriginalReservation`?  It looks like it is comparing reservation SIDs instead of task SIDs, but I could be wrong.

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -0,0 +1,219 @@
...
+  TaskHelper.isTaskAccepted(task) &&
+  task.taskStatus === 'assigned' &&
+  task.status === 'accepted' &&
+  !isTransferring(task);
+
+/**
+ * @param {ITask} task
+ */
+export const showTransferControls = task =>
+  !isOriginal(task) && isTransferring(task) && TaskHelper.isTaskAccepted(task);
```
nickhurlburt, line 56: Would it be better to name these `shouldShowTransferButton` and `shouldShowTransferControls`?  It looks like they don't handle the showing.

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -0,0 +1,219 @@
...
+export const shouldSubmitFormCall = task =>
+  TaskHelper.isCallTask(task) &&
+  (!tranferStarted(task) ||
+    (isOriginal(task) && isTransferRejected(task)) ||
+    (!isOriginal(task) && isTransferAccepted(task)));
+
+/**
+ * @param {ITask} task
+ */
+export const shouldSubmitForm = task => shouldSubmitFormCall(task) || shouldSubmitFormChat(task);
```
nickhurlburt, line 78: I can't figure out what "submit" means here.  I would think that "should submit form" means that it is now time to send a POST request to the HRM to record the contact.

```
if (shouldSubmitForm(task)) {
  saveToHrm(task);
}
```
Otherwise, I don't understand.  What does it mean?

plugin-hrm-form/src/utils/transfer.js
GPaoloni: You are right! I'll make the change

plugin-hrm-form/src/utils/transfer.js
GPaoloni: You've guessed right! I don't know how can I rename this in order to be more explicit. Maybe `shouldSendFormToHrm`?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -2,21 +2,25 @@ import React from 'react';
 import * as Flex from '@twilio/flex-ui';
 import { FlexPlugin } from 'flex-plugin';
 
+import './styles/GlobalOverrides';
```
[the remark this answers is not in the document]
GPaoloni: No we don't have one for the order of the imports (like sort alphabetically). Should we include a plugin like this one? https://github.com/lydell/eslint-plugin-simple-import-sort ?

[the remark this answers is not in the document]
nickhurlburt: @murilovmachado what do you think?

plugin-hrm-form/src/utils/transfer.js
nickhurlburt: But what is confusing is that it is used to decide whether to show the banner about being in the middle of a transfer here: https://github.com/tech-matters/flex-plugins/pull/101/files#diff-d57ccb7c7833ebb892a99d7279002460R54.  It seems like what the function is really doing is telling you if we can edit the form or not.  So should we call it something like `formEditingDisabled` or `formEditingEnabled`?

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -0,0 +1,219 @@
+import { Actions, ITask, TaskHelper, StateHelper } from '@twilio/flex-ui';
```
nickhurlburt, line 1: Note compilation warnings:
```
  Line 1:19:  'ITask' is defined but never used              no-unused-vars
  Line 3:10:  'transferChatStart' is defined but never used  no-unused-vars
```

plugin-hrm-form/src/styles/HrmStyles.js
```diff
@@ -263,3 +263,25 @@ export const HiddenText = styled('p')`
   font-size: 0px;
   line-height: 0px;
 `;
+
+export const StyledButton = styled('button')`
```
nickhurlburt, line 267: Can you re-do the button styling so that it matches https://github.com/tech-matters/flex-plugins/pull/103?

plugin-hrm-form/src/HrmFormPlugin.js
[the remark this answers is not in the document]
murilovmachado: I've took a look at this plugin @GPaoloni suggested and it seems to have a reasonable amount of stars, so I think it's fine to use it. We can install it and see if it plays well with our current set of rules. If everything is ok, I'm in for using it.

PR #103, opened 2020-06-03
Round corners and title case for end conversation button

nickhurlburt:
Changed per design feedback.  We'll want to make most buttons look like this -- for example, the buttons at the bottom right of the forms when we add case management.  If we should put this in a different part of the code in order to make it more generalized, fine by me.

Buttons look like this now:
image (2)
image (3)


nickhurlburt:
In order to get this on staging so we can get screenshots, I'm going to merge and deploy this.  But please do still look at it.

GPaoloni:
Looks great!

PR #106, opened 2020-06-05
Add Sync Client for sharing contact form

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-142
This PR is related to https://github.com/tech-matters/serverless/pull/11

Adds Sync Client to share the state of the form on transfers. The state of the form is saved before initiating the transfer (if possible) and:
- if is a `COLD` transfer the form is retrieved when new counselor accepts the task.
- if is a `WARM` transfer the form is retrieved when new counselor accepts the transfer (`Accept Transfer` button).
The form is restored via a new Redux Action (`RESTORE_ENTIRE_FORM`) to preserve metadata.

This PR also address an existing issue in #101. When a `WARM` transfer is resolved, the unnecesary existing task is moved from `wrapup` to `completed` state via `transferCallResolve` serverless function.


plugin-hrm-form/src/utils/sharedState.js
```diff
@@ -0,0 +1,49 @@
...
+ * Saves the actual form into the Sync Client
+ * @param {*} form form for current contact (or undefined)
+ * @param {import("@twilio/flex-ui").ITask} task
+ */
+export const saveFormSharedState = async (form, task) => {
+  const { sharedStateClient } = getConfig();
+  if (sharedStateClient === undefined || sharedStateClient.connectionState !== 'connected') {
+    window.alert("Can't access Shared State, form won't be saved");
+    return null;
+  }
```
GPaoloni, line 13: I've resolved to alert the counselor that the form is not saved but letting the transfer to continue. Should we prevent the transfer if the form can't be saved in shared state?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -187,6 +221,24 @@ const transferOverride = async (payload, original) => {
   return transferChatStart(body);
 };
 
+const restoreFormIfCold = async payload => {
```
nickhurlburt, line 224: It will make the overall code easier to read if this is `restoreFormIfColdTransfer`

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -187,6 +221,24 @@ const transferOverride = async (payload, original) => {
...
+  if (TransferHelpers.isColdTransfer(payload.task)) {
+    const form = await loadFormSharedState(payload.task);
+    if (form) Flex.Manager.getInstance().store.dispatch(Actions.restoreEntireForm(form, payload.task.taskSid));
+  }
+};
+
+const closeCallIfCold = async payload => {
```
nickhurlburt, line 231: Same comment as above, add `Transfer`

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -225,6 +277,9 @@ const setUpActions = setupObject => {
     manager.store.dispatch(Actions.removeContactState(payload.task.taskSid));
   });
 
+  Flex.Actions.addListener('afterAcceptTask', restoreFormIfCold);
+  Flex.Actions.addListener('afterTransferTask', closeCallIfCold);
```
nickhurlburt, line 281: How do we want to organize the `setUpActions()` function as we add more things?  I'm thinking that maybe we want to order it by the normal order of the task lifecycle -- so accept/reject task first, transfer/hang up/end chat in the middle, complete task at the end.  This will make it easier to read through the function to understand how we are modifying the actions.  So we could put this `afterAcceptTask` line under L258 where there's `beforeAcceptTask`.  I am open to other ways of organizing it.  Most important, I think we do need something natural and consistent.

I also notice that this function has a lot of functions declared inside of it.  Mostly they are right before the `addListener` or `replaceAction` where they are used.  But then it is all mixed up.  I wonder if we should move those inner functions to the top of `setUpActions()`?  Or move them out?  I don't have a very firm idea and we can talk more about it.  There are a few ways we could make this cleaner.

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -243,12 +298,12 @@ const setUpActions = setupObject => {
...
-  const hangupCall = fromActionFunction(saveEndMillis);
+  const hangupCall = fromActionFunction(async payload => {
+    await saveEndMillis(payload);
+  });
 
   const wrapupTask = fromActionFunction(async payload => {
-    if (shouldSayGoodbye(payload.task.channelType)) {
-      await sendGoodbyeMessage(payload);
-    }
+    if (shouldSayGoodbye(payload.task.channelType)) await sendGoodbyeMessage(payload);
```
nickhurlburt, line 306: Did we change a linting rule?  Just curious.

plugin-hrm-form/src/utils/sharedState.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+export const loadFormSharedState = async task => {
+  const { sharedStateClient } = getConfig();
+  if (sharedStateClient === undefined || sharedStateClient.connectionState !== 'connected') {
+    window.alert("Can't access Shared State, form can't be restored");
+    return null;
+  }
+
+  if (!task.attributes.transferMeta) {
+    window.alert('Task not transferred');
```
nickhurlburt, line 38: How could we have gotten to this case, and what should happen?

plugin-hrm-form/src/components/transfer/AcceptTransferButton.jsx
```diff
@@ -1,24 +1,35 @@
...
 };
 
-const AcceptTransferButton = ({ theme, task }) => {
+const AcceptTransferButton = ({ theme, task, restoreEntireForm }) => {
   return (
     <StyledButton
       color={theme.colors.base11}
       background={theme.colors.base2}
-      onClick={() => handleAcceptTransfer(task)}
+      onClick={() => handleAcceptTransfer(task, removeContactState)}
```
nickhurlburt, line 32: This looks like a bug.

plugin-hrm-form/src/HrmFormPlugin.js
GPaoloni: No we didn't, this way was always accepted by our rules. I've dropped an if that were placed just above this one and though that for readability having them in this style was better. I've removed that if statement tho, so I'll go back to the previous version with the brackets.

plugin-hrm-form/src/utils/sharedState.js
GPaoloni: I was testing what would happen if this function ever gets called with a non transferred task. If this function is called only on transferred task, we should never hit this case. Should I remove it?

plugin-hrm-form/src/components/transfer/AcceptTransferButton.jsx
GPaoloni: Related to the work in progress I told you in Slack, sorry.

plugin-hrm-form/src/utils/sharedState.js
nickhurlburt: Whenever we're giving an alert to a user, we need to be clear about what's going on.  Remember that the alert would go to a helpline counselor who doesn't know the code and might not know what 'task' means.  Is an alert the right thing?  Or a console.log?  Or is this something we can handle silently?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -294,6 +355,8 @@ export default class HrmFormPlugin extends FlexPlugin {
     const { translateUI, getGoodbyeMsg } = setUpLocalization(config);
 
     const setupObject = { ...config, translateUI, getGoodbyeMsg };
+
+    if (config.featureFlags.enable_shared_state) setUpSharedStateClient();
```
nickhurlburt, line 359: What is the thinking behind having separate `enable_transfers` and `enable_shared_state` flags?  I am okay with it, just want to understand.

plugin-hrm-form/src/HrmFormPlugin.js
GPaoloni: I agree with the idea of sorting the tasks by order of the task lifecycle! And I suggest moving this helpers out of this file (i.e. to `utils/pluginActions.js` or something like that to improve code redability). Is it ok if I address this in a separate (smaller) PR?

plugin-hrm-form/src/HrmFormPlugin.js
GPaoloni: I did this without thinking too much about it, because it was easier to test each part like this, but having them under the same flag will be ok. Should I merge them?

plugin-hrm-form/src/HrmFormPlugin.js
nickhurlburt: Yes, totally okay to do this in a separate PR

plugin-hrm-form/src/HrmFormPlugin.js
nickhurlburt: Up to you!  I just wanted to understand the logic behind it.

GPaoloni: Ok, let's go with the same for both so there will be less flags to be aware of!

PR #109, opened 2020-06-10
Re-do the way transferred task are being closed

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-152
This PR is related to https://github.com/tech-matters/serverless/pull/12

- Fixed bug where a `COLD` transferred call task was not getting closed.
- After a `WARM` transfer, the new counselor will be able to see the form of the original one as soon as the task is accepted.
- Added `TransferredTaskJanitor`, which is in charge of closing transferred tasks that are not required anymore (via `CompleteTask` action).


plugin-hrm-form/src/utils/devUtils.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/no-unused-modules */
+export const log = arg => {
+  console.log(`%c ${arg}`, 'background: #222; color: #bada55');
+};
```
GPaoloni, line 4: I've been using this a lot in this PR, and might be useful while developing. This makes a very visible console.log (only for strings tho). Should I remove it?

nickhurlburt: If you expect to use it in future work, I'm fine with leaving it.

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -57,29 +57,37 @@ export const shouldShowTransferControls = task =>
   !isOriginalReservation(task) && isTransferring(task) && TaskHelper.isTaskAccepted(task);
 
 /**
+ * Indicates if the current counselor has control over the task. Used to know if counselor should send form to hrm backend and pevent the form from being edited
```
nickhurlburt, line 60: `prevent`

nickhurlburt, line 60: Can we say "sole control"?  That makes it clear that this counselor and only this counselor has the task.

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -57,29 +57,37 @@ export const shouldShowTransferControls = task =>
...
+ * A counselor controls a task if
+ * - transfer was not initiated
+ * - task is original and a transfer was initiated but it was rejected
```
nickhurlburt, line 63: "this is the original reservation and a transfer was initiated and then rejected"

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -57,29 +57,37 @@ export const shouldShowTransferControls = task =>
...
+ * - task non original and a transfer was initiated and it was accepted
```
nickhurlburt, line 64: "this is not the original reservation and a transfer was initiated and then accepted"

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -57,29 +57,37 @@ export const shouldShowTransferControls = task =>
...
- * - task is original and a transfer was initiated but it was rejected
- * - task non original and a transfer was initiated and it was accepted
  * @param {ITask} task
  */
-export const hasTaskControlCall = task =>
-  TaskHelper.isCallTask(task) &&
-  (!hasTransferStarted(task) ||
-    (isOriginalReservation(task) && isTransferRejected(task)) ||
-    (!isOriginalReservation(task) && isTransferAccepted(task)));
+// export const hasTaskControl = task => hasTaskControlCall(task) || hasTaskControlChat(task);
```
nickhurlburt, line 75: remove when ready

PR #110, opened 2020-06-12
Refactor setUpActions

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-155

Refactored the `setUpActions` function on plugin init:
- events are sorted by order of appearance in the task lifecycle 
- moved all functions for `replaceAction` and `addListener` to a separate file (`setUpActions.js`)

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -268,75 +268,87 @@ const fromActionFunction = fun => async (payload, original) => {
   await original(payload);
 };
 
-const setUpActions = setupObject => {
+const saveEndMillis = async payload => {
   const manager = Flex.Manager.getInstance();
+  manager.store.dispatch(Actions.saveEndMillis(payload.task.taskSid));
+};
 
-  const {
-    hrmBaseUrl,
-    workerSid,
-    helpline,
-    helplineLanguage,
-    configuredLanguage,
-    getGoodbyeMsg,
-    featureFlags,
-  } = setupObject;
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -8,21 +8,19 @@ import CustomCRMContainer from './components/CustomCRMContainer';
...
 import QueuesStatusWriter from './components/queuesStatus/QueuesStatusWriter';
 import { TransferButton, AcceptTransferButton, RejectTransferButton } from './components/transfer';
-import reducers, { namespace, contactFormsBase } from './states';
-import { Actions } from './states/ContactState';
+import reducers, { namespace } from './states';
 import LocalizationContext from './contexts/LocalizationContext';
 import HrmTheme from './styles/HrmTheme';
 import { channelTypes, transferModes } from './states/DomainConstants';
 import { addDeveloperUtils, initLocalization } from './utils/pluginHelpers';
+import * as Actions from './utils/setUpActions';
```
nickhurlburt, line 16: We've overloaded the term "Actions" quite a bit.  We use them in three ways:
1. Redux actions that show up in `ActionCreators`, `ActionTypes`, `ContactState` and elsewhere.
2. The Flex Actions framework
3. Functions in `setupActions.js` that we call here with `Actions.<function>`.

Let's think of some other naming so we don't get confused between these.

GPaoloni: As agreed in standup, we are going with `ActionFunctions` to distinguish them from the rest of Actions!

PR #111, opened 2020-06-12
Refactor setUpComponents

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-157

Added TS support (letting it to be compatible with the actual codebase):
- added `tsconfig.json` and installed typescript dependency
- changed `src/HrmFlexPlugin` from JS to TS, and moved types from JSDoc to TS style
- changed src/utils/setUpActions from JS to TS, and moved types from JSDoc to TS style

Refactored the setUpComponents function on plugin init:
- moved all components declarations to a separate file (setUpComponents.tsx). They are now added to dynamic components by calling helpers.


plugin-hrm-form/src/utils/setUpComponents.js
```diff
@@ -0,0 +1,167 @@
...
+  if (task.status !== 'wrapping') {
+    if (task.channelType === channelTypes.voice) {
+      await Flex.Actions.invokeAction('HangupCall', { sid, task });
+    } else {
+      await Flex.Actions.invokeAction('WrapupTask', { sid, task });
+    }
+  }
+
+  Flex.Actions.invokeAction('CompleteTask', { sid, task });
+};
```
nickhurlburt, line 79: It bothers me that we have a function that invokes three actions in `setupComponents` and not `setupActions`.  Someone who reads `setupActions()` and thinks they know all the changes we are making to the actions lifecycle will be surprised.

However, `setupActions()` only has `addListener` and `replaceAction`, nothing that invokes actions.  So, I think this is okay as it is.  We just still need to remember that we have action framework stuff in a few places.

GPaoloni: Right, I though about doing this or not, but couldn't find a better place to put this function. I think we can document this, as well as every appearance of the actions frameworks in our code, and/or maybe have one more helper file where we group all of the function that makes use of the actions framework (and import them from this centralized file).

nickhurlburt: I think those are good ideas.  This is probably enough for now.

PR #113, opened 2020-06-15
Remove transfer/kick/leave buttons

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-158

- Remove directory button from Call Canvas Actions (arrow at the bottom of the call)
Screenshot from 2020-06-15 17-06-01

- Remove hangup button from Task List Actions (both counselors, only on warm transfers)
Screenshot from 2020-06-15 17-07-07

- Remove hangup button from Call Canvas Actions (both counselors, only on warm transfers)
Screenshot from 2020-06-15 17-07-44

- Remove actions (kick/hold) from Participant Canvas (both counselors, only on warm transfers and only on participants that are workers, in order to be able to hold/unhold the caller)
Screenshot from 2020-06-15 17-08-12


plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -136,6 +136,8 @@ const setUpComponents = setupObject => {
   // remove dynamic components
   Components.removeActionsIfWrapping();
   Components.removeLogo();
+  Components.removeDirectoryButton();
+  Components.removeActionsIfTransferring();
```
nickhurlburt, line 140: Remember to wrap these in a feature flag

GPaoloni: Ups, I forgot the flags. Thank you for the reminder!

PR #114, opened 2020-06-16
Murilo open new case

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-145

This PR:
- Decouples saving a Contact from completing a Task
- Moves `saveToHrm` out of redux
- Creates `Save and Add to Case` button
- Implements clicking on `Open New Case`:
  1) Saves the task
  2) Creates a new case
  3) Connects the task to the case
  4) Redirects to Case screen
- Raw Case screen shows:
  - Case id
  - Button `Save` that completes the Twilio Task

This PR doesn't touch any styles so far:
- Button style update not applied
- Case Page without ANY style for now


plugin-hrm-form/src/states/ActionCreators.js
```diff
@@ -20,20 +31,6 @@ export const handleFocus = dispatch => (taskId, parents, name) => {
...
 
-export const handleSubmit = dispatch => (form, handleCompleteTask) => task => {
-  const newForm = validateBeforeSubmit(form);
-  dispatch({
-    type: HANDLE_BLUR, // probably need to rename this
-    form: newForm,
-    taskId: task.taskSid,
-  });
-  if (formIsValid(newForm)) {
-    handleCompleteTask(task.taskSid, task);
```
nickhurlburt, line 31: This creates a problem that is hard to spot.  We record `conversationDuration` on the contact record.  In `setUpActions#wrapupTask` and `setUpActions#hangupCall`, we write the timestamp for when the call/chat ends, then when we send the contact to the HRM we send the difference between start and end as `conversationDuration`.  But now when we save the contact, it may not be the end of the conversation.  `endMillis` has not been set and so we send a number like `-1592355589` (which is `(0 - startMillis)/1000`) as the duration.  Because the conversation doesn't necessarily end when the contact is saved, we need to figure out another way to do this.

Right now, I am not sure how we do this without updating the contact record later.  I'm sure we can think of something better.  Good for @GPaoloni to think about this too.

plugin-hrm-form/src/components/callTypeButtons/CallTypeButtons.jsx
```diff
@@ -20,18 +22,30 @@ const CallTypeButtons = props => {
...
 
+  const handleCallTypeButtonClick = (taskSid, callType) => {
+    props.handleCallTypeButtonClick(taskSid, callType);
+    props.changeRoute('tabbed-forms', taskSid);
+  };
+
+  const handleConfirmNonDataCallType = async () => {
+    const { hrmBaseUrl, workerSid, helpline } = getConfig();
+
+    await saveToHrm(task, form, hrmBaseUrl, workerSid, helpline);
```
nickhurlburt, line 33: Hmm... by putting this here, we will have to handle errors here as well.  This could just fail and then we complete the task anyway.  I'm okay with this bit of debt for now.  Adding API call error handling to the list for future reliability.

plugin-hrm-form/src/components/callTypeButtons/CallTypeButtons.jsx
```diff
@@ -20,18 +22,30 @@ const CallTypeButtons = props => {
   const { form, task, localization } = props;
   const { isCallTask } = localization;
 
+  const handleCallTypeButtonClick = (taskSid, callType) => {
```
nickhurlburt, line 25: Is it good to give this the same name as the prop, but change the behavior?  I am a little conflicted.

plugin-hrm-form/src/components/tabbedForms/BottomBar.jsx
```diff
@@ -0,0 +1,158 @@
...
+
+    try {
+      if (formIsValid(newForm)) {
+        await saveToHrm(task, form, hrmBaseUrl, workerSid, helpline);
+        this.props.handleCompleteTask(task.taskSid, task);
+      } else {
+        this.setState({ isMenuOpen: false });
+        window.alert('There is a problem with your submission.  Please check the form for errors.');
+      }
+    } catch (error) {
```
nickhurlburt, line 94: Note that this catch isn't replicated when we call `saveToHrm` from the non-data call type component.  Though I am unsure what the best error-handling behavior is, really.

plugin-hrm-form/src/services/ContactService.js
```diff
@@ -107,29 +107,33 @@ export function saveToHrm(task, form, abortFunction, hrmBaseUrl, workerSid, help
...
-      // TODO(nick): fix this. this isn't working I don't think the function is working from inside the promise.
-      if (!window.confirm('Unknown error saving form.  Are you sure you want to end the task without recording?')) {
-        abortFunction();
-      }
-    });
+  });
+
+  if (!response.ok) {
+    console.log(`Form error: ${response.statusText}`);
+    if (!window.confirm('Error from backend system.  Are you sure you want to end the task without recording?')) {
```
nickhurlburt, line 118: Hmm.. if we call this from `BottomBar#handleSubmit`, does this error appear twice?

Seeing this may invalidate my comment on the non-data call type component.  Seems our error pop up a little inconsistently, unless I misunderstood.

plugin-hrm-form/src/components/callTypeButtons/CallTypeButtons.jsx
murilovmachado: Actually in this case, it won't complete the task. There's 2 "failing" scenario:
1- If the call returns `response.ok = false`, then the user will get the alert: `Error from backend system.  Are you sure you want to end the task without recording?` (this part is inside `saveToHrm`).
2- If the call throws an unexpected (not handled) error, then it would stop the execution and next line would be never called: `props.handleCompleteTask(task.taskSid, task)`

I agree we need to make huge changes to error handling, but at least in this case the task will not be closed.

plugin-hrm-form/src/components/callTypeButtons/CallTypeButtons.jsx
murilovmachado: I agree it can be confusing, I'll rename it.

plugin-hrm-form/src/services/ContactService.js
murilovmachado: You're right, there's a scenario where the error popup may be shown twice, I'll fix that.

murilovmachado:
@nickhurlburt I've changed `saveToHrm` to not do any error handling within itself. Now, I think the code will be easier to follow. From now on, every time we use `saveToHrm` we'll have to write error handling code around it. Having some error handling inside `saveToHrm` mixed with custom error handling around its call was making the code very confusing.

plugin-hrm-form/src/components/tabbedForms/BottomBar.jsx
murilovmachado: **UPDATE:** I've changed the way we were error handling this function, it should be more clear now.

plugin-hrm-form/src/states/ActionCreators.js
nickhurlburt: Thinking more about this.  To start, we have at least two invariants to work with:
1. We cannot record the duration if the call hasn't finished.
2. We cannot connect a case and a contact if they don't exist yet.

Therefore, if we 1) create a case linked to the contact and 2) allow a conversation to continue while working on the case, then we have to figure out another approach.  Here are some options:
1. Don't save the contact record right away when the user clicks 'Save and add to case'.  Just create the case.  Then when they close the task, save the contact and connect it with the case.
2. Save the contact record before creating the case, but create a `PUT` endpoint on contact and update the duration at the end of the task.  We will likely need to support editing an existing contact record eventually, so this may be a start to that.
3. Store conversation duration in another table, and write that at the end, linking it with the contact record.

Given that right now our goal was to come to a good stopping place and pick this up later, I'll willing to cut some corners to get this checked in, then return after categories.  That corner-cutting could be ensuring that existing functionality (recording duration) works fine so long as we have a feature flags that maintains the existing behavior.  The goal there is that our code is still deployable.

murilovmachado: Good catch, I can't think of another way of doing this without updating the contact later. Also, I've checked and we don't have an endpoint for updating a contact. We need to build that as well.

murilovmachado:
@nickhurlburt I've implemented your first suggestion: 
> Don't save the contact record right away when the user clicks 'Save and add to case'. Just create the case. Then when they close the task, save the contact and connect it with the case.

`conversationDuration` seems to be working fine now. I've also removed `saveEndMillis` from redux (and from Twilio's Hangup action).  `endMillis` is now being set at the time we trigger save.

nickhurlburt:
I'm sorry, there's some context I didn't provide.  `conversationDuration` is the duration of the call or chat, it's not the duration of the overall contact.  We need to handle a situation where the counselor clicks "Hang up"/"End Chat" (and the conversation is over and the timer stops) and then the counselor continues in the 'wrapping' state to add more information to the contact or case.  That's why this is linked to the `HangUp` action and why it's in redux.  Your approach to waiting until the end to store everything makes sense and it looks good, but we have to allow for the `conversationDuration` timer to stop if someone hits "Hang Up" while still working on the task.

murilovmachado:
@nickhurlburt I've put back `saveEndMillis` to be called when both `HangUpAction` and `WrapUpAction` are triggered. It looks like all the scenarios are being captured now. Please take a look to see if it behaves correctly.

murilovmachado:
@nickhurlburt Right, basically all new visual stuff comes from the new `Save And Add to Case...` button (the popup menu and the new case page), and this button is under the `enable_case_management ` feature flag. I'm finishing localization, and I'll ask Gian to take a look at the localization part.

nickhurlburt:
Excellent, sounds good to me.

PR #116, opened 2020-06-18
Fixed voice transfer bugs

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-156
This PR is related to https://github.com/tech-matters/serverless/pull/13

- tasks can't be transferred to offline workers anymore
- if the reservation created because of a transfer to a worker is rejected, the control of the task returns to the original counselor
- several bugs fixed
- Transfer buttons styles changed
Screenshot from 2020-06-19 19-30-41
Screenshot from 2020-06-19 19-54-21


plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -84,11 +84,18 @@ const setUpSharedStateClient = () => {
  */
 const setUpTransferredTaskJanitor = async setupObject => {
   const { workerSid } = setupObject;
-  const query = 'data.attributes.channelSid == "CH00000000000000000000000000000000"';
+  const query =
+    'data.attributes.transferStarted == "true" || data.attributes.channelSid == "CH00000000000000000000000000000000"';
```
GPaoloni, line 88: I've realized that we won't need the fake channelSid anymore when true warm chat transfer is done, so I've introduced a few changes on the way we "track" the transferred tasks with the janitor, as well as how we can determine if a reservation `hasTaskControl`  (in `utils/transfers.js`). I can explain if needed, but I think this is a simpler approach than the first one!

plugin-hrm-form/src/components/transfer/AcceptTransferButton.jsx
```diff
@@ -16,12 +15,6 @@ import { Actions } from '../../states/ContactState';
 const handleAcceptTransfer = async (task, restoreEntireForm) => {
   if (TaskHelper.isChatBasedTask(task)) await closeChatOriginal(task);
   else await closeCallOriginal(task);
-
-  // restore the state of the previous form for warm transfer (if there is any)
-  if (isWarmTransfer(task)) {
-    const form = await loadFormSharedState(task);
-    if (form) restoreEntireForm(form, task.taskSid);
-  }
```
GPaoloni, line 24: This is not necessary as we always load the form in `afterAcceptTask` event.

plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -48,12 +48,45 @@ export const initializeContactForm = payload => {
...
+/**
+ * If the task is transferred, set the reservation sid who controls the task (according to transfer mode)
+ * @param {import('@twilio/flex-ui').ITask} task
+ */
+const setProperControlIfTransfer = async task => {
+  if (TransferHelpers.hasTransferStarted(task) && TransferHelpers.isColdTransfer(task))
+    await TransferHelpers.takeTaskControl(task);
+
+  if (TransferHelpers.hasTransferStarted(task) && TransferHelpers.isWarmTransfer(task))
+    await TransferHelpers.clearTaskControl(task);
```
GPaoloni, line 69: During a warm transfer, until it's resolved, none of both counselors "have task control", so here is set to empty string.

nickhurlburt:
One thing I noticed here is that we are not locking the call type buttons.  I can hit "Back" from the form, click "Silent" and end the call during the warm transfer.  Then the task disappears for me but the other counselor is marooned.

nickhurlburt:
Another problem I'm seeing is that since the form isn't locked until the transfer task is accepted, the original counselor can continue editing data after requesting the transfer, but that data does not get sent over to the second counselor.  So then they are looking at different form data.

It's okay to lock the form once the transfer starts.  The goal for CHI-156 is to make sure that the task doesn't end up stuck forever.  We may have to talk through how we do that.

nickhurlburt:
An update on things I've observed so far today:
- The issues in the bug (CHI-156) look resolved.  Warm and cold voice transfers look like they can't get into a broken state, and the form is locked properly including non-data calls.  Great!
- The counselor can still focus on a form field and can still get a dropdown, though can't choose any options.  And when mousing over or clicking, the CSS still changes.  We may want to disable all of this... however, I think we can wait for product feedback before doing that.
- I'm seeing some problems with chat transfers now.  Cold chat transfers can still be sent to an offline worker, and then it ends up sitting around in the queue.  When the transfer is direct, we should have the same behavior as voice, where we prevent it and show an error.  I think we do need to fix this before asking for product feedback.
- For a cold chat transfer to a queue, the original counselor has no way to give up on the transfer if it takes too long.  I think we can just raise this for product feedback... it may not need to be fixed right away.
- Warm chat transfers can end up in a bad state.  If the second counselor rejects the task, then the first counselor gets form control back as though there's no transfer.  But then the task appears again on the second counselor's UI.  Say the second counselor then accepts the task, and then rejects the transfer.  When that happens, the first counselor's form gets locked again, and there is no way to unlock it.  I had to go into the Twilio console to delete the task.  Since we will probably disable warm chat transfers, I don't think we need to fix this now, but I wanted to capture it.

plugin-hrm-form/src/services/ServerlessService.js
```diff
@@ -38,10 +40,20 @@ export const getMessages = async body => {
...
-  const newTask = await fetchProtectedApi(url, { ...body, Token: token });
-  return newTask;
+  try {
+    const url = `${serverlessBaseUrl}/transferChatStart`;
+
+    const newTask = await fetchProtectedApi(url, { ...body, Token: token });
+    return newTask;
+  } catch (err) {
+    Notifications.showNotification('TransferFailed', {
+      reason: `Worker ${body.targetSid} is not available.`,
```
nickhurlburt, line 51: Joan doesn't like this message (and I agree) but for consistency with current Twilio practice for voice, let's keep as is.

GPaoloni: I believe we can change this (either the message or the entire alert). I'll learn a bit more about the Notifications and look a way of doing it, if desired!

nickhurlburt: Let's get this up on staging as is.  If you figure out how to change the existing Twilio message, then we can change both.

PR #119, opened 2020-06-23
Transfers related code cleanup

GPaoloni:
- Removed various functions that are not longer necessary
- Simplified (and unified) the logic behind who controls a task

### DIFF
--- a/plugin-hrm-form/src/utils/transfer.js
+++ b/plugin-hrm-form/src/utils/transfer.js
@@ -121,10 +121,6 @@ export const updateTransferStatus = newStatus => async task => {
     transferMeta: {
       ...task.attributes.transferMeta,
       transferStatus: newStatus,
-      /*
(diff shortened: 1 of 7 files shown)


plugin-hrm-form/src/utils/transfer.js
```diff
@@ -124,11 +117,15 @@ export const clearTaskControl = async task => {
 export const updateTransferStatus = newStatus => async task => {
   const updatedAttributes = {
     ...task.attributes,
+    transferStarted: true,
     transferMeta: {
       ...task.attributes.transferMeta,
       transferStatus: newStatus,
+      /*
```
nickhurlburt, line 124: Why is this here?

GPaoloni: Sorry it should't be there, removed already.

PR #120, opened 2020-06-24
Style category checkbox

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-162

Screen Shot 2020-06-24 at 14 03 46

### DIFF
--- a/plugin-hrm-form/package.json
+++ b/plugin-hrm-form/package.json
@@ -23,7 +23,6 @@
     "date-fns": "^2.11.1",
     "flex-plugin": "^3.0.0",
     "flex-plugin-scripts": "^3.0.4",
-    "hex-to-rgba": "^2.0.1",
     "lodash": "^4.17.15",
     "prop-types": "^15.7.2",
     "react": "16.5.2",
--- a/plugin-hrm-form/package.json
+++ b/plugin-hrm-form/package.json
@@ -23,6 +23,7 @@
     "date-fns": "^2.11.1",
     "flex-plugin": "^3.0.0",
     "flex-plugin-scripts": "^3.0.4",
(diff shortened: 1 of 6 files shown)


plugin-hrm-form/package.json
```diff
@@ -23,6 +23,7 @@
     "date-fns": "^2.11.1",
     "flex-plugin": "^3.0.0",
     "flex-plugin-scripts": "^3.0.4",
+    "hex-to-rgba": "^2.0.1",
```
nickhurlburt, line 26: Heh... I think I complained to @GPaoloni a month or two ago about adding another package to do something we could do ourselves.  I don't remember what we decided on.  We can discuss quick today.

plugin-hrm-form/src/components/tabbedForms/IssueCategory.jsx
```diff
@@ -1,33 +1,36 @@
...
 const IssueCategory = props => {
-  console.log({ color: props.color });
+  const color = hexToRgba(props.color, '0.6');
   return (
     <div style={{ display: 'flex', flexDirection: 'column' }}>
       <StyledLabel>{props.category}</StyledLabel>
       {props.subcategories.map(subcategoryName => {
         const id = `IssueCategorization_${props.category}_${subcategoryName}`;
+        const { value } = props.form.caseInformation.categories[props.category][subcategoryName];
+        const disabled = false;
```
GPaoloni, line 21: Just a question. This is now a constant, but the idea is that in a future this will be "enabled" or "disabled" from a config provided?

murilovmachado: This will be automatically `disabled` if three other categories had been chosen previously.

PR #123, opened 2020-06-26
Added Case Details component

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-147

Adds Case Details component to Case page.

Screenshot from 2020-06-29 16-44-06


plugin-hrm-form/src/states/ConfigurationState.js
```diff
@@ -1,10 +1,17 @@
...
 export const changeLanguage = language => ({ type: CHANGE_LANGUAGE, language });
 
+export const populateCounselorsState = counselorsList => ({ type: POPULATE_COUNSELORS, counselorsList });
+
 const initialState = {
   language: defaultLanguage,
+  counselors: {
+    list: [],
+    hash: {},
+  },
```
GPaoloni, line 14: I'm not sure where this should live. I've dropped it in here because I'm sure this does not belongs to `activeContacts` nor to `searchState`. Maybe this is the right place but the name (`configuration`) isn't..  What do you think?

GPaoloni:
Deployed

PR #124, opened 2020-06-27
Use Close Contact when conversation ended

nickhurlburt:
Joan mentioned that it's weird that our non-data call type confirmation dialog says "Hang Up" or "End Chat" even when we have already hung up or ended the chat.  This PR changes the messaging to "Close Contact" if we are already in the 'wrapping' state.

When the conversation is still open, it looks like it did before:
Screen Shot 2020-06-26 at 5 33 21 PM
Screen Shot 2020-06-26 at 5 34 19 PM

But when the conversation has been ended, we show 'close contact':
Screen Shot 2020-06-26 at 5 33 37 PM
Screen Shot 2020-06-26 at 5 34 35 PM

### DIFF
--- a/plugin-hrm-form/src/components/callTypeButtons/NonDataCallTypeDialog.jsx
+++ b/plugin-hrm-form/src/components/callTypeButtons/NonDataCallTypeDialog.jsx
@@ -13,7 +13,7 @@ import {
 } from '../../styles/callTypeButtons';
 import TabPressWrapper from '../TabPressWrapper';
 
-const NonDataCallTypeDialog = ({ isOpen, isCallTask, handleConfirm, handleCancel }) => (
+const NonDataCallTypeDialog = ({ isOpen, isCallTask, isInWrapupMode, handleConfirm, handleCancel }) => (
   <CloseTaskDialog onClose={handleCancel} open={isOpen}>
     <TabPressWrapper>
       <NonDataCallTypeDialogContainer>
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/components/callTypeButtons/NonDataCallTypeDialog.jsx
```diff
@@ -24,8 +24,11 @@ const NonDataCallTypeDialog = ({ isOpen, isCallTask, handleConfirm, handleCancel
         <Box marginBottom="32px">
           <Row>
             <ConfirmButton autoFocus tabIndex={1} onClick={handleConfirm}>
-              {/* eslint-disable-next-line react/jsx-max-depth */}
-              <Template code={isCallTask ? 'TaskHeaderEndCall' : 'TaskHeaderEndChat'} />
+              {/* eslint-disable react/jsx-max-depth,no-nested-ternary */}
+              <Template
+                code={isInWrapupMode ? 'CallType-CloseContact' : isCallTask ? 'TaskHeaderEndCall' : 'TaskHeaderEndChat'}
```
nickhurlburt, line 29: I am especially interested in your thoughts on whether we should be disabling `no-nested-ternary` here.  It feels like this nested ternary is actually a very good way to express the logic here.  However, I don't want to go around ignoring linting rules without double-checking.

GPaoloni: I agree, using ternary here is very straightforward!

murilovmachado: For this particular situation, I think it's fine to go with a nested ternary operator. Another option would be to have a `code` variable and fill its value with an `if-elseif-else`, and then use `<Template code={code} ...>`. It would be more verbose, though. I'm fine with either approach.

nickhurlburt:
This has been deployed to staging.

PR #125, opened 2020-06-29
[WIP] Adjustments to IssueCategorizationTab

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-164

This PR put the pieces together in the IssueCategorizationTab.

WIP:
- Fix Tests

Screen Shot 2020-06-29 at 20 25 46
Screen Shot 2020-06-29 at 20 26 06

### DIFF
--- a/plugin-hrm-form/src/components/tabbedForms/IssueCategory.jsx
+++ b/plugin-hrm-form/src/components/tabbedForms/IssueCategory.jsx
@@ -17,15 +17,19 @@ const IssueCategory = props => {
             const { value } = props.form.caseInformation.categories[props.category][subcategoryName];
             const selectedCategories = countSelectedCategories(props.form.caseInformation.categories);
             const disabled = selectedCategories === 3 && !value;
+            const handleClickCheckboxField = e => {
+              e.preventDefault();
+              props.handleCategoryToggle(props.taskId, props.category, subcategoryName, !value);
+            };
             return (
-              <CategoryCheckboxField key={id} color={lighterColor} selected={value} disabled={disabled}>
-                <StyledCategoryCheckbox
-                  disabled={disabled}
-                  color={lighterColor}
-                  checked={value}
-                  id={id}
-                  onClick={() => props.handleCategoryToggle(props.taskId, props.category, subcategoryName, !value)}
-                />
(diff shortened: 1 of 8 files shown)


plugin-hrm-form/src/components/tabbedForms/IssueCategory.jsx
```diff
@@ -14,7 +15,8 @@ const IssueCategory = props => {
           {props.subcategories.map(subcategoryName => {
             const id = `IssueCategorization_${props.category}_${subcategoryName}`;
             const { value } = props.form.caseInformation.categories[props.category][subcategoryName];
-            const disabled = false;
+            const selectedCategories = countSelectedCategories(props.form.caseInformation.categories);
+            const disabled = selectedCategories === 3 && !value;
```
nickhurlburt, line 19: Is it safer to make this `selectedCategories > 2`?  Or are we confident enough that it won't exclude 3?

plugin-hrm-form/src/components/tabbedForms/IssueCategory.jsx
```diff
@@ -14,7 +15,8 @@ const IssueCategory = props => {
...
             return (
               <CategoryCheckboxField key={id} color={lighterColor} selected={value} disabled={disabled}>
```
nickhurlburt, line 21: Currently, the mouseOver and click behavior is that you get a pointer only when mousing over the `StyledCategoryCheckbox`, and you can select/unselect when clicking on entire the checkbox or the label component.

But, when you click elsewhere inside this div, nothing happens.  This feels strange now that there's a border.  Try clicking inside the border but to the right of the text... doesn't do anything.  Can we change this so that we can click on anything inside this div?  (and show a pointer cursor as well)

murilovmachado: Good catch, I've implemented all these fixes.

plugin-hrm-form/src/components/tabbedForms/IssueCategory.jsx
murilovmachado: I'm changing it to `selectedCategories >= 3` to make it safer.

PR #126, opened 2020-06-30
Toggle categories view

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-165

This PR implements:
- Toggle categories view
- Move categories expanded state to redux
- Move search->contact_details expanded state to redux
- Make `Section` a stateless component

WIP:
- Fix tests

Screen Shot 2020-06-30 at 19 07 39


plugin-hrm-form/src/components/search/ContactDetails/Details.jsx
```diff
@@ -76,15 +82,23 @@ const Details = ({ contact, handleOpenConnectDialog, handleMockedMessage }) => {
           <MoreHorizIcon style={{ color: '#ffffff' }} />
         </ButtonBase>
       </NameContainer>
-      <Section expanded sectionTitle="General details">
+      <Section
+        sectionTitle="General details"
```
nickhurlburt, line 86: We use the General details string three times here (and use other strings three times later as well).  Is it worth defining constants for these somewhere?

PR #131, opened 2020-07-07
(WIP) Upgrade to React 16.13.1

nickhurlburt:
[Per Twilio instructions](https://www.twilio.com/docs/flex/plugins/react-versions#updating-a-pre-existing-plugin-to-use-the-latest-react-version), this upgrades React to the latest version.  It also matches our dev dependency on `flex-ui` to 1.20.1.

This seemed like a good time to do a full `npm install`, so there are may `package-lock.json` changes.  Notably, Material UI is still below Version 4.  Our support person was optimistic that V4 would work.  We can attempt on a follow-up PR.

This is WIP because I have not tested it out yet.  You are welcome to do so if you like.  I will eventually.

**Important Note**: after you pull these changes, you will need to run `UNBUNDLED_REACT=true npm install`.  Normal `npm install` will fail.  I am not sure about `npm ci`.

@murilovmachado @GPaoloni @raarielgrace

### DIFF
--- a/plugin-hrm-form/package-lock.json
+++ b/plugin-hrm-form/package-lock.json
@@ -5,39 +5,47 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.8.3",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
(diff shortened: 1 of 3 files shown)


nickhurlburt:
Subsumed by #145.  Closing.

PR #135, opened 2020-07-08
Adding backend pagination

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-179
This PR is related to https://github.com/tech-matters/hrm/pull/39

This PR 
- adds backend pagination for the cases list
- adds error handling if fetch fails

Note: this version has no caching, so each page change implies a new request to the backend.


plugin-hrm-form/src/styles/caseList/index.js
```diff
@@ -138,10 +144,7 @@ export const ButtonText = styled(FontOpenSans)`
   font-weight: ${props => (props.highlight ? 700 : 600)};
 `;
 
-export const LoadingContainer = styled('div')`
-  display: flex;
-  width: 100%;
-  height: 100%;
-  align-items: center;
-  justify-content: center;
+export const SomethingWentWrogText = styled(FontOpenSans)`
```
nickhurlburt, line 147: spelling: `SomethingWentWrongText`

plugin-hrm-form/src/components/caseList/CaseList.jsx
```diff
@@ -3,53 +3,78 @@ import { Template } from '@twilio/flex-ui';
 import { CircularProgress, Dialog, DialogContent } from '@material-ui/core';
 
 import CaseListTable from './CaseListTable';
-import { CaseListContainer, LoadingContainer } from '../../styles/caseList';
+import { CaseListContainer, CenteredContainer, SomethingWentWrogText } from '../../styles/caseList';
 import { getCases } from '../../services/CaseService';
 
+export const CASES_PER_PAGE = 5;
```
nickhurlburt, line 9: In reality I think we'll use 20 for this (to be the same as the search results), but it's good that this is so easily configurable.

plugin-hrm-form/src/components/caseList/CaseList.jsx
```diff
@@ -3,53 +3,78 @@ import { Template } from '@twilio/flex-ui';
...
     page: 0,
     mockedMessage: null,
   };
 
   async componentDidMount() {
-    const caseList = await getCases();
-    console.log('HERE HERE HERE');
-    console.log(caseList);
-    this.setState({ caseList, loading: false });
+    this.fetchCaseList(0);
```
nickhurlburt, line 24: You'll have to remember to display page numbers as `page + 1`

plugin-hrm-form/src/components/caseList/CaseListTable.jsx
```diff
@@ -9,16 +9,17 @@ import { TLHPaddingLeft } from '../../styles/GlobalOverrides';
...
 
 /**
  * This component is splitted to make it easier to read, but is basically a 8 columns Table (7 for data, 1 for the "expand" button)
  */
-const CaseListTable = ({ caseList, page, handleChangePage, handleMockedMessage }) => {
-  const pagesCount = Math.floor(caseList.length / CASES_PER_PAGE) + 1;
-  const from = page * 5;
-  const to = from + 5;
+const CaseListTable = ({ caseList, caseCount, page, handleChangePage, handleMockedMessage }) => {
+  const pagesCount = Math.floor(caseCount / CASES_PER_PAGE) + 1;
```
nickhurlburt, line 18: I'm not sure this formula is right.  `Math.floor(caseCount / CASES_PER_PAGE)` seems right.  If we have 5 total cases and 5 cases per page, we have 1 page, but the formula above will return 2.  I might be missing something.

GPaoloni: Ooh you are right, sorry bout this. I should add 1 to the pageCount only if `caseCount % 5 !== 0` (not multiple of 5), right?

GPaoloni: Actually, what I've said above is same as `Math.ceil` haha

nickhurlburt:
Notice that when clicking somewhere on the case list view, a blue border appears over the whole component.  I don't think we want this:

image

nickhurlburt:
Can you change the search results mouseover look to be the same as for the mouseover in the case list view?  (If it's better in a separate PR, that's fine)
image

nickhurlburt:
The chevron (<) and the numbers aren't quite vertically aligned.  Can we make it better? 

image

plugin-hrm-form/src/components/caseList/CaseList.jsx
nickhurlburt: We will want to change this to 20, but I think it's good to leave it as is for now so that everyone can see how the pagination works with so many pages.

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -106,5 +106,6 @@
   "CaseList-THCategory": "Category",
   "CaseList-ExpandButton": "Open case details",
   "CaseList-PrevPage": "Previous page",
-  "CaseList-NextPage": "Next page"
+  "CaseList-NextPage": "Next page",
+  "CaseList-SomethingWentWrog": "Something went wrong while loading cases list. Please retry in a moment."
```
nickhurlburt, line 110: Spelling: `CaseList-SomethingWentWrong`

plugin-hrm-form/src/components/caseList/CaseListTableRow.jsx
```diff
@@ -27,14 +27,26 @@ import { caseStatuses } from '../../states/DomainConstants';
...
-const renderCategory = category => (
-  <div style={{ width: '100%' }} key={`category-tag-${category}`}>
-    <CategoryTag>
-      <CategoryFont>{category}</CategoryFont>
-    </CategoryTag>
-  </div>
-);
+const renderCategory = category => {
+  const tag =
+    category.length > 17 && !category.toUpperCase().includes('UNSPECIFIED/OTHER')
```
nickhurlburt, line 36: Should we localize this string maybe?  I'm also okay with leaving it for now.

GPaoloni: I though about this but I was unsure. This is coming like this from the backend, maybe we can create a mapper that returns the appropriate string template based on the English category name?

PR #138, opened 2020-07-10
Enable chat capabilities

murilovmachado:
In order for the new `TaskChannels` to work properly, we need to add Chat Capabilities to them. And this is done by creating a `ChatTaskChannel`.

After this PR is merged, we need to go on https://www.twilio.com/console/flex/messaging and link `whatsapp`, `web`, `facebook` and `sms` to the `ChannelSelector Flow`.

~Didn't include `facebook` and `sms` yet, because I'm not sure how to trigger them.~
It worked for `whatsapp`, `web`, `facebook` and `sms`.

Screen Shot 2020-07-13 at 19 51 34
Screen Shot 2020-07-13 at 19 51 56

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -186,6 +186,16 @@ const setUpActions = setupObject => {
   Flex.Actions.addListener('afterCompleteTask', ActionFunctions.removeContactForm);
 };
 
+/**
+ * @param  {...string} taskChannels
+ */
+const enableChatCapabilities = (...taskChannels) => {
+  const customChatChannel = Flex.DefaultTaskChannels.createChatTaskChannel('custom-chat-channel', task =>
+    taskChannels.includes(task.taskChannelUniqueName),
+  );
+  Flex.TaskChannels.register(customChatChannel);
+};
+
 export default class HrmFormPlugin extends FlexPlugin {
   constructor() {
     super(PLUGIN_NAME);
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -186,6 +186,16 @@ const setUpActions = setupObject => {
   Flex.Actions.addListener('afterCompleteTask', ActionFunctions.removeContactForm);
 };
 
+/**
+ * @param  {...string} taskChannels
+ */
+const enableChatCapabilities = (...taskChannels) => {
+  const customChatChannel = Flex.DefaultTaskChannels.createChatTaskChannel('custom-chat-channel', task =>
```
nickhurlburt, line 193: This looks like it's creating a single channel to support both whatsapp and web.  Is that what we want?  I don't know enough to know what this does.  Do the icons under 'all tasks' still look right?  How is it recorded in insights?  Does 'contacts waiting' still get the right colors?

murilovmachado: Yes, it's creating a single `ChatChannel`, I was fixing that right now (creating separate ChatChannels) in order to get the right colors and icons for each one.

About Insights it's still fine, the value comes from the `TaskChannel` not from the `ChatChannel`.

nickhurlburt: Great.  It seems weird that we can't just connect these to the existing `DefaultTaskChannels` for whatsapp or web.  This looks like a good approach, though.

murilovmachado:
@nickhurlburt Everything now works as expected (including visuals)

PR #144, opened 2020-07-16
Created script to copy one studio flow to create and/or update a second

raarielgrace:
This is a standalone script to copy studio flows, likely from staging to production. It copies the flow structure and updates the relevant SIDs.


nickhurlburt:
@raarielgrace please add `node_modules/` to your `.gitignore` file.  You may also need to remove everything under `node_modules` with `git rm`.  This will reduce the size of this PR by thousands of files.  You may also want to look at https://github.com/tech-matters/flex-plugins/blob/master/plugin-hrm-form/.gitignore and copy other relevant lines.

The prefix `plugin-` is used for Twilio Flex plugins specifically, so your directory name probably does not require that prefix.

replicate-twilio/copyFlow.js
```diff
@@ -0,0 +1,158 @@
...
+
+// Requires each value to be unique
+function getKeyByValue(jsonDict, val) {
+  var keys = Object.keys(jsonDict);
+  for (var i = 0; i < keys.length; i++) {
+    var key = keys[i];
+    if (jsonDict[key] === val) {
+      return key;
+    }
+  }
```
nickhurlburt, line 114: This can be implemented using [`Array.find()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find):
`keys.find(key => jsonDict[key] === val)`

It would be worthwhile browsing through the methods on [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array).  I would guess that there's a fairly larger overlap with functions in Python.

replicate-twilio/copyFlow.js
```diff
@@ -0,0 +1,158 @@
...
+}
+
+function recReplaceIds(flow, idDict) {
+  var keys = Object.keys(flow);
+  for (var i = 0; i < keys.length; i++) {
+    var key = keys[i];
+    // Don't copy hidden values
+    if (key.indexOf("_") == 0) {
+      continue;
+    }
```
nickhurlburt, line 75: I would discourage modifying the input parameter `flow`, because it can lead to bugs.  For example, by `continu`ing here, you are still returning this in the flow because you're returning the modified input parameter.  It would be better to build up a new object and return it.  `Array.reduce` may be helpful to you here.  It's also worth remembering `Array.filter` and the spread operator (like the splat operator in Python).

replicate-twilio/copyFlow.js
```diff
@@ -0,0 +1,158 @@
+const fs = require("fs").promises;
+
```
nickhurlburt, line 2: Please add usage documentation so that someone who is new to this tool would be able to use it.  The usage definition on Line 145 is great, though insufficient.  Currently someone would have to read all the code to figure out how to use this.  Please add docs so that the next new person to the team will have an easy time getting this to work.

replicate-twilio/copyFlow.js
```diff
@@ -0,0 +1,158 @@
...
+  return flow;
+}
+
+var argv = require("yargs")
+  .usage(
+    "npm run copy_flow -- --fromIds {file} --toIds {file} [--useFlow {file}] [--update]"
+  )
+  .alias("f", "fromIds")
+  .alias("t", "toIds")
```
nickhurlburt, line 148: Is there an option to just fetch a flow and not upload it to a `toId`?  I can imagine wanting to look it over first.

nickhurlburt:
@raarielgrace any reason not to merge this in?  (Using 'Squash and merge').  With this unmerged, #155 looks a lot bigger than it is.

PR #146, opened 2020-07-16
(WIP) Manual task pull UI

nickhurlburt:
@GPaoloni see email for what this is


GPaoloni:
For the record: the goal of this can be achieved by changing 
`Flex.TaskListContainer.Content.add`
to
`Flex.TaskList.Content.add`

nickhurlburt:
It makes me a little nervous to add a non-task to the task list because of what might break that we don't realize, however, I would be happy to give it a try, and it certainly would be a cleaner way to work with the interface!  I wonder how it behaves when we use the filter (the chevron next to 'All Tasks')

nickhurlburt:
No longer needed.

PR #150, opened 2020-07-21
Gian Add Notes to case

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-151

- Adds "Add Note" button
Screenshot from 2020-07-21 17-30-28

- Adds "Add Note" screen (update case adding a new note to the info column)
Screenshot from 2020-07-21 17-44-32

- Some small modifications to MenuItem proptypes (that were causing warnings)


plugin-hrm-form/src/components/case/AddNote.jsx
```diff
@@ -0,0 +1,84 @@
...
+        <CaseActionTextArea
+          data-testid="Case-AddNoteScreen-TextArea"
+          aria-labelledby="Case-TypeHere-label"
+          rows={25}
+          value={newNote}
+          onChange={e => setNewNote(e.target.value)}
+        />
+      </Box>
+      <div style={{ width: '100%', height: 5, backgroundColor: '#ffffff' }} />
+      <CaseActionButtonBar>
```
murilovmachado, line 53: I think you should reuse `HrmStyles.BottomButtonBar` here. If the styles are a bit different, you can change in `BottomButtonBar` (Its styles keep changing in every version of the mockups =/ ). But this bottom bar should be the same throughout the application.

plugin-hrm-form/src/styles/case/index.js
```diff
@@ -55,3 +60,38 @@ export const DefaultStatusFont = styled(DetailEntryText)`
...
+`;
+
+export const CaseActionDetailFont = styled(FontOpenSans)`
+  font-style: italic;
+  font-size: 12px;
+  line-height: 30px;
+  opacity: 67%;
+`;
+
+export const CaseActionTextArea = styled('textarea')`
```
nickhurlburt, line 85: Can we remove the blue box on focus as we did with the case list view?

plugin-hrm-form/src/styles/case/index.js
```diff
@@ -8,6 +8,11 @@ export const CaseContainer = styled('div')`
   height: 100%;
 `;
 
+export const CenteredContainer = styled(CaseContainer)`
```
nickhurlburt, line 11: Could you annotate these new components with a `displayName` as in #152?

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -90,11 +91,17 @@
...
   "Case-CaseNumber": "Case",
   "Case-CaseDetailsSection": "Case Details",
+  "Case-TimelineSection": "Timeline",
   "Case-CaseDetailsChildName": "Child Name",
   "Case-CaseDetailsOwner": "Counselor",
   "Case-CaseDetailsDateOpened": "Date Opened",
   "Case-CaseDetailsStatusLabel": "Status",
   "Case-CaseDetailsStatusOpen": "Open",
+  "Case-AddNote": "Add Note",
+  "Case-AddNoteTypeHere": "Type here to add  note",
```
nickhurlburt, line 101: There's an extra space between 'add' and 'note'.  And where does this show up on the UI?  Can't see it.

plugin-hrm-form/src/components/case/AddNote.jsx
```diff
@@ -0,0 +1,78 @@
...
+import PropTypes from 'prop-types';
+import { Template } from '@twilio/flex-ui';
+import { ButtonBase } from '@material-ui/core';
+import { Close } from '@material-ui/icons';
+
+import { Box, Row, HiddenText, StyledNextStepButton, BottomButtonBar } from '../../styles/HrmStyles';
+import { CaseContainer, CaseActionTitle, CaseActionDetailFont, CaseActionTextArea } from '../../styles/case';
+
+const AddNote = ({ counselor, handleSaveNote, onClickClose }) => {
+  const [newNote, setNewNote] = useState('');
```
nickhurlburt, line 11: I'm unclear why were aren't using redux here both to save the current view and to save partial information.  If I click over to another tab (say, Case List View), and then return here, I am no longer in the notes view and my partial note text is gone.  We still want the behavior where clicking around different left nav tabs does not mean we lose our location or data.

plugin-hrm-form/src/components/case/AddNote.jsx
```diff
@@ -1,17 +1,18 @@
...
 import { Close } from '@material-ui/icons';
 
+import { getConfig } from '../../HrmFormPlugin';
 import { Box, Row, HiddenText, StyledNextStepButton, BottomButtonBar } from '../../styles/HrmStyles';
-import { CaseContainer, CaseActionTitle, CaseActionDetailFont, CaseActionTextArea } from '../../styles/case';
+import { AddNoteContainer, CaseActionTitle, CaseActionDetailFont, CaseActionTextArea } from '../../styles/case';
 
-const AddNote = ({ counselor, handleSaveNote, onClickClose }) => {
-  const [newNote, setNewNote] = useState('');
+const AddNote = ({ counselor, value, onChage, handleSaveNote, onClickClose }) => {
```
nickhurlburt, line 11: spelling: `onChange`

plugin-hrm-form/src/states/ContactState.js
```diff
@@ -258,6 +266,35 @@ export function reduce(state = initialState, action) {
       };
     }
 
+    case UPDATE_CASE_INFO: {
+      const currentTask = state.tasks[action.taskId];
+      const { metadata } = currentTask;
+      const { connectedCase } = metadata;
+      const updatedCase = { ...connectedCase, info: action.info };
```
nickhurlburt, line 273: It doesn't really feel right to put case info or temporary case info in `metadata`.  Where do you think it belongs?

GPaoloni: To be honest I don't know, because we want each task to have it's own temporary state, and leaving it outside metadata feels even worse (as there is the actual form data). What makes the most sense to me is to factor out the entire case from this reducer, but if we want to do that, should be done in a separate PR. Do you have any suggestion?

nickhurlburt: Making a separate reducer for cases makes sense to me, yes, and I think it'd be fine to do in a separate PR.

plugin-hrm-form/src/components/case/Case.jsx
```diff
@@ -57,73 +63,128 @@ class Case extends Component {
...
   handleSaveAndEnd = async () => {
+    this.setState({ loading: true });
+
     const { task, form } = this.props;
     const { connectedCase } = form.metadata;
     const { hrmBaseUrl, workerSid, helpline, strings } = getConfig();
 
     try {
       const contact = await saveToHrm(task, form, hrmBaseUrl, workerSid, helpline);
+      const caseFromDB = await updateCase(connectedCase.id, { info: connectedCase.info });
```
murilovmachado, line 74: caseFromDB not being used.

plugin-hrm-form/src/components/case/AddNote.jsx
```diff
@@ -0,0 +1,102 @@
...
+import { Close } from '@material-ui/icons';
+
+import { getConfig } from '../../HrmFormPlugin';
+import { Box, Row, HiddenText, StyledNextStepButton, BottomButtonBar } from '../../styles/HrmStyles';
+import { AddNoteContainer, CaseActionTitle, CaseActionDetailFont, CaseActionTextArea } from '../../styles/case';
+import { taskType, formType } from '../../types';
+import { Actions } from '../../states/ContactState';
+import { namespace, contactFormsBase } from '../../states';
+
+const AddNote = ({ task, form, counselor, onClickClose, dispatch }) => {
```
murilovmachado, line 15: I didn't see where dispatch was being passed into this component, does `connect` passes it automatically?. Also, isn't it better to use `mapDispatchToProps`, like the other component does? (just a question)

GPaoloni:
Deployed

PR #152, opened 2020-07-23
Add displayNames to CaseList styling components for developer tab viewing

nickhurlburt:
For review by @GPaoloni but also thanks to @murilovmachado for the idea.

This annotates one of our emotion styling files with `displayName`s, which means they can then be inspected using the Chrome React developer extension.  This allows people to identify which styling components are being used without having to look through the code.  We should start doing this for all new styled components going forward.

image


nickhurlburt:
deployed to staging

PR #153, opened 2020-07-23
Move Insights reported customer attributes to conversation attributes

nickhurlburt:
Now that we've learned that Insights "define customer based on their phone number so.. the same number means the same customer and customer_attribute_x will get overwritten over and over", we'll need to move age and gender to the conversation.  Sadly, this leaves only two free conversation attributes left, though we don't currently see another option.

Another request we have is to put leading zeroes on the select options for age so that they sort alphabetically in the drop-downs in Insights.  While I would prefer to have another layer of abstraction so that we do not couple all this together, this is probably okay for now.

Please think through if there are any other implications to these changes that I don't realize.  When everyone confirms it's good we'll move it to production.

### DIFF
--- a/plugin-hrm-form/src/___tests__/services/InsightsService.test.js
+++ b/plugin-hrm-form/src/___tests__/services/InsightsService.test.js
@@ -28,7 +28,6 @@ test('saveInsightsData for non-data callType', async () => {
       content: 'content',
       conversation_attribute_2: 'Abusive',
     },
-    customers: {},
   };
 
   expect(twilioTask.setAttributes).toHaveBeenCalledWith(expectedNewAttributes);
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/___tests__/services/InsightsService.test.js
```diff
@@ -100,11 +99,11 @@ test('saveInsightsData for data callType', async () => {
       content: 'content',
       conversation_attribute_1: 'Unspecified/Other - Missing children;Bullying;Addictive behaviours',
       conversation_attribute_2: 'Child calling about self',
+      conversation_attribute_3: 'Boy',
+      conversation_attribute_4: '13-15',
     },
     customers: {
       name: 'John Doe',
```
nickhurlburt, line 106: Should we keep this here if we're not making any `customers` changes?

nickhurlburt:
@murilovmachado I've fixed tests so I've pulled this out of WIP.  I'm going to deploy this un-checked-in to staging so Joan can look at it.

plugin-hrm-form/src/___tests__/services/InsightsService.test.js
murilovmachado: I think we should get rid of the `customers` object.

nickhurlburt:
merged and deployed to staging

PR #155, opened 2020-07-28
Ariel new account auto-configuration code

raarielgrace:
Created a script to add workspaces to a given account and configure them to have the same TaskChannels and Activities as the original account.


deploy-scripts/configureNewAccount.js
```diff
@@ -0,0 +1,150 @@
...
+        `and auth token of the account you wish to copy the workspaces of.`
+    );
+    return;
+  }
+
+  // Make the workspaces
+  const oldWSes = await client.taskrouter.workspaces.list();
+  await Promise.all(
+    oldWSes.map(async (oldWS) => {
+      const friendlyN = oldWS["friendlyName"];
```
nickhurlburt, line 31: Why not just call it `friendlyName`?  The abbreviation just removes clarity.

PR #156, opened 2020-07-29
Case timeline

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-150

Backend: https://github.com/tech-matters/hrm/pull/43

WIP:
- Add more tests
- Accessibility check

Screen Shot 2020-07-29 at 14 25 56


plugin-hrm-form/src/components/case/Timeline.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const isCreatingCase = isNullOrUndefined(
+    timeline.find(activity => ['whatsapp', 'facebook', 'web', 'sms'].includes(activity.type)),
+  );
+  if (isCreatingCase) {
+    timeline.push({
+      date: format(Date.now(), 'yyyy-MM-dd HH:mm:ss'),
+      type: task.channelType,
+      text: form.caseInformation.callSummary.value,
+    });
```
GPaoloni, line 37: [We souldn't be mutating the state](https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly) in any way that is not the `setState` function, so this is working but feels like it could be done in a safer way providing an array with this single entry as `initialState`, or inside the `useEffect` call.

plugin-hrm-form/src/styles/case/index.js
```diff
@@ -75,6 +77,17 @@ export const DefaultStatusFont = styled(DetailEntryText)`
 `;
 DefaultStatusFont.displayName = 'DefaultStatusFont';
 
+export const ViewButton = styled(props => <Button roundCorners={false} {...props} />)`
+  color: #192b33;
+  background-color: #ecedf1;
```
GPaoloni, line 82: We are using this colors in several places (I remember adding them in `case` and `caseList`). Maybe we should add them to our theme or something?

plugin-hrm-form/src/components/case/Timeline.jsx
murilovmachado: good catch!

plugin-hrm-form/src/styles/case/index.js
murilovmachado: I've seen that it's actually already at our `HrmTheme` as `categoryTextColor`. I'll reuse it, but we may come up with a better strategy for naming these constants on `HrmTheme`. This is actually a deeper conversation that we can leave for later.

PR #158, opened 2020-07-29
Added Case Summary

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-148

- Added Case Summary component
    Note: this component isn't in it's final placement, as it's going to be "pushed down" in the Case screen as more components are added (perpetratorss, timeline, etc)

Screenshot from 2020-07-29 17-45-06


plugin-hrm-form/src/components/case/CaseSummary.jsx
```diff
@@ -0,0 +1,56 @@
...
+import { connect } from 'react-redux';
+
+import { CaseSectionFont, CaseSummaryTextArea } from '../../styles/case';
+import { namespace, contactFormsBase } from '../../states';
+import { Actions } from '../../states/ContactState';
+import { formType, taskType } from '../../types';
+import { getConfig } from '../../HrmFormPlugin';
+
+const CaseSummary = ({ task, form, updateCaseInfo }) => {
+  const { strings } = getConfig();
```
GPaoloni, line 14: Just realized this today: this is a function with constant complexity, but still a function call that will be performed on every re-render. The same is done in AddNote component.. Do you think that this can affect performance? Should this be passed via Context instead (we can use [useContext](https://reactjs.org/docs/hooks-reference.html#usecontext) now)? @murilovmachado

murilovmachado: I don't think performance will be an issue, but a quick solution to this would be to move `const { strings } = getConfig();` outside of `CaseSummary`, just after the `imports` section. I believe that would avoid calling it on every render.

GPaoloni:
Deployed

PR #160, opened 2020-07-31
Moved case state to a separate reducer

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-189

- moved case state (redux) to a separate reducer
- introduced TS (for the store, some components and styles file)
- added `types/types.ts` to represent some types shared across the project. Note: probably, this file should (eventually) become the `index.d.ts` file, but that can be done in the future to avoid conflicts with our current PropTypes file. Research pending on this
- changed Node version from 10.x to 12.x in CI file to solve some issues with globalThis when running tests (apparently happening since introducing TS)


GPaoloni:
Deployed

PR #161, opened 2020-08-05
Moved routing information into it's own reducer

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-201

- Moved routing information to a new reducer (routing).
- Moved `initializeContactState`, `removeContactState` actions (and it's corresponding types) to the root of `states` folder, as many reducers will react to this actions.


plugin-hrm-form/src/states/routing/reducer.ts
```diff
@@ -0,0 +1,59 @@
...
+    // @ts-ignore TODO: maybe we need a "common" action for this, that triggers this in all of the reducers
+    case RECREATE_SEARCH_CONTACT:
+      return {
+        ...state,
+        tasks: {
+          ...state.tasks,
+          // @ts-ignore
+          [action.taskId]: newTaskEntry,
+        },
+      };
```
GPaoloni, line 41: I haven't included this in the general actions yet, because I wan't to know what do you think about this. [Here](https://github.com/tech-matters/flex-plugins/blob/6aa65c9b49556779a466b33abd8f6a0f69cf4a1a/plugin-hrm-form/src/components/TaskView.jsx#L36) we are checking if the search state exists for the current task, and recreating it if not on mount (possible case after a page reload). Maybe we can have a general action that does this, and every reducer can "recreate" the state for the task (if it does not exists) in the same way (contact form, search and now routing state). Does this makes sense?

murilovmachado: Sure, I like the idea of having a general action that will trigger a recreate in all the reducers.

GPaoloni:
Deployed

PR #162, opened 2020-08-10
[WIP] Perpetrators and Add Perpetrators

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-187

- Added `Perpetrators` section in "new case" screen
- Added `Add Perpetrator` screen

TODO:
- Unit test
- Accessibility check
- Check corner cases (long strings, etc)
- Move shared methods from /state folder to a common one (edit nested fields & createBlankForm). (Pending question: Do we want to move this?)


plugin-hrm-form/src/components/common/forms/CallerForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+};
+
+export type CallerFormValues = FormValues<CallerFormInformation>;
+
+type Props = {
+  callerInformation: CallerFormInformation;
+  defaultEventHandlers: DefaultEventHandlers;
+};
+
+const CallerForm: React.FC<Props> = ({ callerInformation, defaultEventHandlers }) => (
```
GPaoloni, line 39: This is almost a copy paste of the caller information tab form (with tiny changes to make TS compile). Does the naming and the folder structure makes sense to you? I'm not importing this component from the tabbed forms yet, but I'll change to it if you think this is good.

plugin-hrm-form/src/components/common/forms/helpers.ts
```diff
@@ -0,0 +1,34 @@
+import { CallerFormInformation, CallerFormValues } from './CallerForm';
+
+export function getCallerFormValues(callerFormInformation: CallerFormInformation): CallerFormValues {
```
GPaoloni, line 3: The only purpose of this function is to get the values on each leaf of a form with "CallerFormInformation" structure. It's kinda tedious cause I'm manually pulling each value. I've tried to do something more general with `reduce` to get the values from a form with general structure (commented code below this) but couldn't achieve it as some type information is lost. I'll keep trying to do so, but in the meanwhile does this makes sense to you?

GPaoloni: I think this is solved in https://github.com/tech-matters/flex-plugins/pull/162/commits/63a9f8efb660bea914e2689a4b70848e0d03dace

I'm leaving the old code commented out and will remove it before merging if we don't spot any bugs with the current (more general) version.

GPaoloni:
Closed in favor of https://github.com/tech-matters/flex-plugins/pull/165

PR #164, opened 2020-08-12
Revert "Moved case state to a separate reducer"

GPaoloni:
Reverts tech-matters/flex-plugins#160


GPaoloni:
Not needed as merge conflicts were solved.

PR #165, opened 2020-08-13
Perpetrators and Add Perpetrators

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-187

- Added `Perpetrators` section in "new case" screen
- Added `Add Perpetrator` screen

Related future work to do:
- Accessibility
- Move shared methods from /state folder to a common one (edit nested fields & createBlankForm). (Pending question: Do we want to move this?)
- Refactor caller information tab to use the `CallerForm` introduced in this PR


GPaoloni:
Deployed

PR #167, opened 2020-08-14
Ariel pre-survey form code

raarielgrace:


plugin-hrm-form/src/components/TaskView.jsx
```diff
@@ -53,6 +59,65 @@ class TaskView extends Component {
...
+            ageStr = '18-25';
+          } else {
+            ageStr = '>25';
+          }
+        }
+
+        // Change call type and fill form based on survey answers
+        const child = answers.about_self.answer === 'Yes';
+        if (child) {
+          form.callType.value = callTypes.child;
```
GPaoloni, line 108: This (and the equivalent line in the else block) is not the [right way to set values in Redux store](https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns), as you are mutating the state. I think in this case it works because you are updating the store in the legal way just after this, but you can achieve the same in a safer way by assigning `form.callType.value` within the same action you dispatch to pre-populate the form, in the reducer, taking advantage of the fact that you already have 2 actions, one for child and one for caller.

plugin-hrm-form/src/components/TaskView.jsx
```diff
@@ -53,6 +59,65 @@ class TaskView extends Component {
...
+          } else if (ageNum >= 13 && ageNum <= 15) {
+            ageStr = '13-15';
+          } else if (ageNum >= 16 && ageNum <= 17) {
+            ageStr = '16-17';
+          } else if (ageNum >= 18 && ageNum <= 25) {
+            ageStr = '18-25';
+          } else {
+            ageStr = '>25';
+          }
+        }
```
GPaoloni, line 103: Maybe factor out this into a function that returns the proper string for each case, like
```
if (ageNum >= 0 && ageNum <= 3)
  return '0-03';
...
```
and so on? If you decide to do so, `src/utils/mappers` might be a good place to drop it.
Also consider using the predicate [isNaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN) to check for `Unknown` results (if the user can only input numbers that's not needed).

GPaoloni: Please, ignore the `isNaN` part of above comment, I've been testing and it works great. It seems like autopilot already handles those errors, awesome.

plugin-hrm-form/src/components/TaskView.jsx
```diff
@@ -53,6 +59,65 @@ class TaskView extends Component {
       return null;
     }
 
+    // If this task came from the pre-survey
+    if (task.attributes.memory) {
```
GPaoloni, line 63: I've just realized that all of this code is inside the `render` method. It should be in `componentDiMount` instead (after the `recreateContactState` call), [because `render` method is called on each rerender](https://stackoverflow.com/a/39542218/11833888). Sorry that I keep adding comments, haven't realized that on the first pass.

PR #168, opened 2020-08-14
Households and Add Household

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-188

- Added `Households` section in "new case" screen
- Added `Add Household` screen


GPaoloni:
Deployed

PR #169, opened 2020-08-17
Make Case List View show Case Summary instead of Call Summary

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-202
This PR is related to https://github.com/tech-matters/hrm/pull/45


GPaoloni:
Deployed

PR #170, opened 2020-08-18
Use 'Contact' instead of 'Case' or 'Call' where appropriate

nickhurlburt:
In our contact forms, we have been using the terms 'Case' and 'Call' in situations where 'Contact' is more appropriate.  With the introduction of case management, this can confuse the user.  This PR changes the "Call Summary" to "Contact Summary", and changes "Add Case Summary" to "Summary".  The change to "Add Contact Summary" made the tab three lines, so in discussion with our Product team we decided to remove the verbs on the tabs, shortening the names:

image

In the code we still have labels like `caseInformation` which can now be confusing.  I've created [Jira issue CHI-205](https://bugs.benetech.org/browse/CHI-205) for later to consider whether we want to change those.  That change would ripple through the system, though, so it's not a job for now, and we may just want to live with it for awhile.

I think I just need @GPaoloni to review this, but @murilovmachado please do read this over when you have a chance.

### DIFF
--- a/plugin-hrm-form/src/components/tabbedForms/CaseInformationTab.jsx
+++ b/plugin-hrm-form/src/components/tabbedForms/CaseInformationTab.jsx
@@ -13,7 +13,7 @@ const CaseInformationTab = ({ caseInformation, handleCheckboxClick, defaultEvent
       <ColumnarBlock>
         <FieldText
           id="CaseInformation_CallSummary"
-          label="Call summary"
+          label="Contact summary"
           field={caseInformation.callSummary}
           rows={10}
           {...defaultEventHandlers(['caseInformation'], 'callSummary')}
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/components/tabbedForms/CaseInformationTab.jsx
```diff
@@ -13,7 +13,7 @@ const CaseInformationTab = ({ caseInformation, handleCheckboxClick, defaultEvent
       <ColumnarBlock>
         <FieldText
           id="CaseInformation_CallSummary"
-          label="Call summary"
+          label="Contact summary"
```
GPaoloni, line 16: This can be easily localized, but as this entire component will probably be refactored (like caller information tab) we can do that later. Should we have a Jira task to remind that to us?

nickhurlburt:
Deployed!

PR #171, opened 2020-08-18
View timeline item: Contact

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-150

This PR implements clicking `View` on a Connected Contact from the timeline.

WIP:
- ~Add more tests~ (done)

Screen Shot 2020-08-18 at 14 08 42
Screen Shot 2020-08-18 at 14 08 29

### DIFF
--- a/plugin-hrm-form/src/components/case/AddHousehold.tsx
+++ b/plugin-hrm-form/src/components/case/AddHousehold.tsx
@@ -13,7 +13,7 @@ import { namespace, connectedCaseBase } from '../../states';
 import * as CaseActions from '../../states/case/actions';
 import * as RoutingActions from '../../states/routing/actions';
 import { CaseState } from '../../states/case/reducer';
-import { DefaultEventHandlers } from '../common/forms/types';
+import { DefaultEventHandlers, FormValues } from '../common/forms/types';
 import { getFormValues } from '../common/forms/helpers';
 
 // @ts-ignore     TODO: fix this type error (createBlankForm must be typed or maybe create a separate function)
@@ -59,7 +59,7 @@ const AddHousehold: React.FC<Props> = ({
     if (!temporaryCaseInfo || typeof temporaryCaseInfo === 'string') return;
 
     const { info } = connectedCaseState.connectedCase;
-    const household = getFormValues(temporaryCaseInfo);
+    const household = getFormValues(temporaryCaseInfo) as FormValues<CallerFormInformation>;
     const createdAt = new Date().toISOString();
     const newHousehold = { household, createdAt };
     const households = info && info.households ? [...info.households, newHousehold] : [newHousehold];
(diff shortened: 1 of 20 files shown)


plugin-hrm-form/src/components/case/AddHousehold.tsx
```diff
@@ -81,7 +81,10 @@ const AddHousehold: React.FC<Props> = ({
     <CaseActionContainer>
       <CaseActionFormContainer>
         <ActionHeader titleTemplate="Case-AddHousehold" onClickClose={onClickClose} counselor={counselor} />
-        <CallerForm callerInformation={temporaryCaseInfo} defaultEventHandlers={defaultEventHandlers} />
+        <CallerForm
+          callerInformation={temporaryCaseInfo as CallerFormInformation}
```
GPaoloni, line 85: Above suggestion also removes the need for this cast.

plugin-hrm-form/src/components/case/AddPerpetrator.tsx
```diff
@@ -59,7 +59,7 @@ const AddPerpetrator: React.FC<Props> = ({
     if (!temporaryCaseInfo || typeof temporaryCaseInfo === 'string') return;
 
     const { info } = connectedCaseState.connectedCase;
-    const perpetrator = getFormValues(temporaryCaseInfo);
+    const perpetrator = getFormValues(temporaryCaseInfo) as FormValues<CallerFormInformation>;
```
GPaoloni, line 62: The two comments in the previous file apply for this and below casting too.

plugin-hrm-form/src/components/case/ContactDetailsAdapter.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * TODO(murilo): This file replicates some code from the hrm repo. We should implement
+ * a better solution later on.
+ */
```
GPaoloni, line 4: Agreed!

plugin-hrm-form/src/components/case/ViewContact.tsx
```diff
@@ -0,0 +1,85 @@
...
+import ContactDetails from '../ContactDetails';
+import { ViewContact as ViewContactType } from '../../states/case/types';
+import ActionHeader from './ActionHeader';
+import { adaptFormToContactDetails } from './ContactDetailsAdapter';
+
+const mapStateToProps = (state, ownProps: OwnProps) => {
+  const form = state[namespace][contactFormsBase].tasks[ownProps.task.taskSid];
+  const counselorsHash = state[namespace][configurationBase].counselors.hash;
+  const tempInfo = state[namespace][connectedCaseBase].tasks[ownProps.task.taskSid]
+    .temporaryCaseInfo as ViewContactType;
```
GPaoloni, line 20: Maybe change this to type-guards inside the component (return null if `isViewContactType(temporaryCaseInfo)` is false), as there shouldn't be any case where this is called with for example a string value as `temporaryCaseInfo`, but if that ever happens this will error on mount.

murilovmachado:
> @murilovmachado actually my PR (#170) is relevant here because I've changed the titles on the tabbed forms in `flexUI.json`. Looks like `SearchDetails` is using a hardcoded version of those from [`SearchContact.js`](https://github.com/tech-matters/flex-plugins/blob/6aa65c9b49556779a466b33abd8f6a0f69cf4a1a/plugin-hrm-form/src/states/SearchContact.js#L142). Those need to be changed, ideally to use the same strings from `flexUI.json` if that's possible.

@nickhurlburt I've updated the PR to use the updated and localized labels.

murilovmachado:
Deployed to staging.

nickhurlburt:
Hooray!

PR #172, opened 2020-08-20
Case screen fix for call tasks

GPaoloni:
Fixed a bug in timeline component where call tasks were causing infinite rerenders.

Pending: render the proper color icon for calls. For this we might have to send a supoport ticket, as default call channel properties differs from the others.

### DIFF
--- a/plugin-hrm-form/src/components/case/TimelineIcon.jsx
+++ b/plugin-hrm-form/src/components/case/TimelineIcon.jsx
@@ -30,7 +30,7 @@ const TimelineIcon = ({ type }) => <TimelineIconContainer>{getIcon(type)}</Timel
 
 TimelineIcon.displayName = 'TimelineIcon';
 TimelineIcon.propTypes = {
-  type: PropTypes.oneOf(Object.values(channelTypes)).isRequired,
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/components/case/TimelineIcon.jsx
```diff
@@ -5,39 +5,50 @@ import * as Flex from '@twilio/flex-ui';
...
-
-  return <TimelineIconContainer>{icon}</TimelineIconContainer>;
 };
 
+const TimelineIcon = ({ type }) => <TimelineIconContainer>{getIcon(type)}</TimelineIconContainer>;
+
 TimelineIcon.displayName = 'TimelineIcon';
 TimelineIcon.propTypes = {
-  type: PropTypes.oneOf(['whatsapp', 'facebook', 'web', 'sms', 'note']).isRequired,
+  type: PropTypes.oneOf(Object.values(channelTypes)).isRequired,
```
nickhurlburt, line 33: Is this right?  It's missing `'note'`.

GPaoloni: Oh, good catch, thanks.

murilovmachado:
deployed to staging.

nickhurlburt:
Yay!

PR #181, opened 2020-08-25
Case temporary information changes

GPaoloni:
Changed the way we use temporary information property for a case, in order to have type distinction based on the current screen.


nickhurlburt:
Looks like this branch needs some work to show the actual diff against current `staging`.

plugin-hrm-form/src/states/case/types.ts
```diff
@@ -21,6 +22,7 @@ export type ViewContact = {
   counselor: string;
 };
 
+// Should we remove this type-guard
 export function isViewContact(object: any): object is ViewContact {
   return (
```
murilovmachado, line 27: Yes, please remove it if it's not being used anymore.

GPaoloni:
Closed in favor of https://github.com/tech-matters/flex-plugins/pull/183

PR #182, opened 2020-08-25
Fix counselor name at ViewNote screen

murilovmachado:
The issue was that the backend send this info as `twilioWorkerId` and the front end was assuming `counselor`.

Screen Shot 2020-08-25 at 17 00 30


murilovmachado:
Deployed to Prod.

PR #183, opened 2020-08-26
Case temporary information changes

GPaoloni:
Cherry picked relevant commits from https://github.com/tech-matters/flex-plugins/pull/181

Changed the way we use temporary information property for a case, in order to have type distinction based on the current screen.


GPaoloni:
Deployed to staging

PR #185, opened 2020-08-28
Added code to copy autopilot bot across accounts

raarielgrace:
Currently copying tasks works, but copying field values is broken.


raarielgrace:
Bug is fixed and the code works correctly now.

PR #191, opened 2020-09-03
Gian refactor search reducer

GPaoloni:
This PR refactors search reducer in order to match the new structure for our reducers.


plugin-hrm-form/src/states/ActionTypes.js
```diff
@@ -5,21 +5,12 @@ export const HANDLE_CHANGE = 'HANDLE_CHANGE';
 export const SAVE_END_MILLIS = 'SAVE_END_MILLIS';
 export const CHANGE_TAB = 'CHANGE_TAB';
 export const RESTORE_ENTIRE_FORM = 'RESTORE_ENTIRE_FORM';
+export const HANDLE_SELECT_SEARCH_RESULT = 'HANDLE_SELECT_SEARCH_RESULT';
```
GPaoloni, line 8: I let the name of this as it is, but I think we might want to rename it, to make clear that this actually does a "copy contact information".

plugin-hrm-form/src/states/ContactState.js
```diff
@@ -79,6 +79,12 @@ export class Actions {
   static prepopulateFormCaller = (gender, age, taskId) => ({ type: PREPOPULATE_FORM_CALLER, gender, age, taskId });
 }
 
+export const handleSelectSearchResult = (searchResult, taskId) => ({
```
GPaoloni, line 82: Same comment as above.

plugin-hrm-form/src/states/ContactState.js
```diff
@@ -79,6 +79,12 @@ export class Actions {
...
+  type: HANDLE_SELECT_SEARCH_RESULT,
+  searchResult,
+  taskId,
+});
+
```
GPaoloni, line 87: Same comment as above.

plugin-hrm-form/src/types/types.ts
```diff
@@ -24,3 +24,64 @@ export type Case = {
...
+      callSummary: string;
+      referredTo: string;
+      status: string;
+      keepConfidential: boolean;
+      okForCaseWorkerToCall: boolean;
+      howDidTheChildHearAboutUs: string;
+      didYouDiscussRightsWithTheChild: boolean;
+      didTheChildFeelWeSolvedTheirProblem: boolean;
+      wouldTheChildRecommendUsToAFriend: boolean;
+    };
```
GPaoloni, line 66: If we ever factor out the children and case forms as we did for caller information form, the type here will be much simple (like above callerInformation).

plugin-hrm-form/src/states/search/types.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+
+// this was inside search reducer, but it's used in ContactDetails and Timeline components. Should this be moved to a more common/shared place?
+export const ContactDetailsSections = {
+  GENERAL_DETAILS: 'General details',
+  CALLER_INFORMATION: 'Caller information',
+  CHILD_INFORMATION: 'Child information',
+  ISSUE_CATEGORIZATION: 'Issue categorization',
+  CONTACT_SUMMARY: 'Contact summary',
+} as const;
```
murilovmachado, line 33: I agree we can find a more general place to put this type.

murilovmachado (approved):
Looks good to me!

PR #193, opened 2020-09-04
Added queues status (alternative)

GPaoloni:
Alternative approach to https://github.com/tech-matters/flex-plugins/pull/192


nickhurlburt (changes requested):
Let's do the other one instead.

GPaoloni:
Closed in favor of https://github.com/tech-matters/flex-plugins/pull/192

PR #194, opened 2020-09-08
Category Color Redesign

murilovmachado:
This PR removes all color information from `ContactFormStateFactory`, so it can be easily reusable just by calling `getCategoryColor(category)`.

Since we won't need to fetch different colors for different helplines in the near future, we didn't need to implement any function on the Serverless repo nor any additional setup code for it.


GPaoloni (approved):
This is great! Just one question, if we add a category to ContactFormStateFactory but forgot to add in colors, it will default to undefined, maybe is worth adding a default color for that case? Just a though, that may never happen.

murilovmachado:
> This is great! Just one question, if we add a category to ContactFormStateFactory but forgot to add in colors, it will default to undefined, maybe is worth adding a default color for that case? Just a though, that may never happen.

I'll merge this one, since my next PR adds the default color behaviour.

PR #200, opened 2020-09-16
Fix linting on deploy-scripts

GPaoloni:
Jira : https://bugs.benetech.org/secure/RapidBoard.jspa?rapidView=19&projectKey=CHI&view=detail&selectedIssue=CHI-242

Added `.prettierrc` file with rules matching the rest of the repos.


nickhurlburt (approved):
Awesome, thanks!

PR #201, opened 2020-09-16
Change Sync Client error log

GPaoloni:
Changed logs when there's a Synnc Client initialization to be an error instead, in order to be reported to Rollbar.


nickhurlburt (approved):
Great!

PR #204, opened 2020-09-21
Search pagination

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-227

This PR adds pagination to the Search Results.
PS: This PR will only work when the backend gets deployed.

WIP: Fix tests

image


plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { Template } from '@twilio/flex-ui';
+
+import ContactPreview from '../ContactPreview';
+import { SearchContactResult } from '../../../types/types';
+import { Row } from '../../../styles/HrmStyles';
+import { BackIcon, BackText, ResultsHeader, ListContainer, ScrollableList } from '../../../styles/search';
+import ConnectDialog from '../ConnectDialog';
+import Pagination from '../../Pagination';
+
+export const CONTACTS_PER_PAGE = 10;
```
nickhurlburt, line 13: We have currently been assuming 20.  Is there a reason to do 10 instead?

plugin-hrm-form/src/services/ContactService.js
```diff
@@ -4,12 +4,26 @@ import { isNonDataCallType } from '../states/ValidationRules';
 import { channelTypes } from '../states/DomainConstants';
 import { getConversationDuration, fillEndMillis } from '../utils/conversationDuration';
 import { getConfig } from '../HrmFormPlugin';
+import { isNullOrUndefined } from '../utils';
 
-export async function searchContacts(searchParams) {
+function getLimitAndOffsetParams(limit, offset) {
```
nickhurlburt, line 9: Is this functionality best kept separate between `CaseService` and `ContactService`, or shared?

nickhurlburt (commented):
After a quick pass, looks good!  @GPaoloni can give it a closer look.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { BackIcon, BackText, ResultsHeader, ListContainer, ScrollableList } from '../../../styles/search';
+import ConnectDialog from '../ConnectDialog';
+import Pagination from '../../Pagination';
+
+export const CONTACTS_PER_PAGE = 20;
+
+type SearchResultsProps = {
+  currentIsCaller: boolean;
+  results: SearchContactResult;
+  handleSelectSearchResult: (contact) => void;
```
GPaoloni, line 18: I might be wrong but assuming `contact` is of type `SearchContact`, you can type this as `  handleSelectSearchResult: (contact: SearchContact) => void;` to make the caller of this component be more strict about the function that's being passed. Same for `handleViewDetails` and `handleSearch: (offset: number) => void`. Just pointing to this functions notation, nothing wrong with the code.

plugin-hrm-form/src/states/search/actions.ts
```diff
@@ -24,12 +24,16 @@ export const handleSearchFormChange = (taskId: string) => <K extends keyof t.Sea
 export const searchContacts = (dispatch: Dispatch<any>) => (taskId: string) => async (
   searchParams: any,
   counselorsHash: ConfigurationState['counselors']['hash'],
+  limit: number,
+  offset: number,
 ) => {
   try {
+    console.log({ limit, offset });
```
GPaoloni, line 31: Do we wanna keep this one for while?

GPaoloni (approved):
This is great!

plugin-hrm-form/src/states/search/actions.ts
murilovmachado: Good catch, I'll remove it

PR #205, opened 2020-09-22
Revert "Added queues status to AgentDesktopPanel when there are no tasks"

GPaoloni:
Reverts tech-matters/flex-plugins#192


nickhurlburt (approved):
Thanks!

PR #208, opened 2020-09-24
Fixed bug: stuck in agent-desktop

GPaoloni:
With the current way of showing contacts waiting (`QueuesStatus`) when there are no active tasks, there's a bug. If you have exactly one task, when clicking a `SideNav` button to change the current page, you see a quick flicker and get redirected back to agent desktop. Also you can see an error like this happening too often in the console

Screenshot from 2020-09-24 17-12-03

Both arises from the way we re-render the UI when `QueuesStatus` disapears from `NoTaskView` and appears in `AgentDesktop` instead.

This PR fixes both bugs.

### DIFF
--- a/plugin-hrm-form/src/utils/setUpComponents.js
+++ b/plugin-hrm-form/src/utils/setUpComponents.js
@@ -34,10 +34,22 @@ export const setUpQueuesStatusWriter = setupObject => {
   );
 };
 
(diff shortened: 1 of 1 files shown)


nickhurlburt (approved):
Looks good, thanks for spotting and fixing this!

GPaoloni:
Deployed

PR #209, opened 2020-09-25
Support "Unknown" for Pre-survey age question

nickhurlburt:
Before now, our pre-survey chatbot could only produce a number.  Now, in order to support 'Prefer not to answer', it can have a value of 'Unknown' as well.  `mapAge()` assumes that it gets a numeric value, and `mapAge('Unknown')` was returning '>25'.  This PR fixes that.


nickhurlburt:
@GPaoloni please take a look

GPaoloni (approved):
Looks great!

GPaoloni:
Deployed

PR #212, opened 2020-09-30
Adds implementation of manual pull

nickhurlburt:
Code from VPS that calls serverless function to implement manual pull.


nickhurlburt:
Replaced with #218

PR #215, opened 2020-10-01
Upgrade Flex to 1.22.0

nickhurlburt:
I've upgraded Flex on staging to 1.22.0, and this upgrades our dev dependency to the same.  To do this I've run `npm install` which has made minor updates to many packages.

I attempted to upgrade to Material UI 4, but it caused a lot of issues.  This may be something we attempt the next time we see a big reason to.

This PR is mostly just informational, but let me know if you have any concerns.

### DIFF
--- a/plugin-hrm-form/package-lock.json
+++ b/plugin-hrm-form/package-lock.json
@@ -13,9 +13,9 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.10.4",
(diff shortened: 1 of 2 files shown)


nickhurlburt:
Looks like there are some issues.  I'm going to roll back our Flex configuration in the console for now.

nickhurlburt:
Closing this without merging.  After our next release, I'll upgrade to 1.23.

PR #220, opened 2020-10-09
Disabling the manual pull button when appropriate

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-255

This PR adds the following to manual pull button:
- Disabled if no pending chat.
- Disabled when max capacity reached.
- Notification shown if worker is not elegible for task assignment.
- Styles to hover and disabled states.

Note: Is easier to read this PR commit by commit, as some of them involves small refactors and conversion to TS.

WIP:
- Fix CI.
- Add tests for changes in configuration reducer (action and reducer).


plugin-hrm-form/src/components/ManualPullButton/index.tsx
```diff
@@ -1,29 +1,75 @@
...
+      Notifications.showNotification('NoTaskAssignableNotification');
+      await adjustChatCapacity('decrease');
+    }, 5000);
+
+    workerClient.once('reservationCreated', eventHandler);
+
+    await adjustChatCapacity('increase');
+  };
+
+  const maxCapacityReached = chatChannelCapacity === workerAttr.maxMessageCapacity;
```
nickhurlburt, line 44: I changed my `maxMessageCapacity` in the Twilio console after I had loaded up the page, from 3 to 2.  I have 2 tasks but the button is not disabled:

Screen Shot 2020-10-09 at 5 06 30 PM

Do we expect disabling to work if the `maxMessageCapacity` is changed after the page is loaded?  (The `workerClient`'s attributes do get updated correctly).

nickhurlburt: Actually, I'm not sure if this is getting disabled on `maxCapacityReached` at all.  I've reloaded the page so that everything should be up to date, and with a `maxMessageCapacity` of 2, I still have an active button when I have two active tasks.

As we discussed yesterday, it's okay if we have a notification saying that you're at capacity, rather than disabling the button.  But this code makes it appear that the button should be disabled if the current capacity is equal to the max capacity.

nickhurlburt (commented):
Looks good!  Though there seems to be a bug with disabling when at max capacity.

plugin-hrm-form/src/components/ManualPullButton/index.tsx
GPaoloni: Yes the intention on this is to disable the button, is this ok or the notification is preferred? Fun fact on this one, after struggling a while why this was not working: 
We are setting `"maxMessageCapacity":"3"`, which means "3 as string", and thus the comparison   const `maxCapacityReached = chatChannelCapacity === workerAttr.maxMessageCapacity;` is never true. I've configured my worker to be `"maxMessageCapacity": 3` and it's working as expected now haha. Thanks for the catch! Should we add a `parseInt` in case `maxMessageCapacity` is accidentally a string?

nickhurlburt: Yes, unfortunately it may need to be a string.  We intend to update the attributes in Okta, which then propagates to Twilio, but it propagates as a string.  So let's get `parseInt` in.

Let's go ahead and merge and put this on staging (it's okay if it overwrites Mu's branch).  We may need to edit to make it a notification later, but people can play with it on staging.

nickhurlburt (approved):
After adding `parseInt`, please merge and deploy to staging

GPaoloni:
Deployed to staging.

PR #223, opened 2020-10-13
Records channelType in Flex Insights

nickhurlburt:
Now that we have unfortunately needed to abandon using separate TaskChannels for different channel types, we can no longer rely on the Communication Channel in Flex Insights to record the channel of a contact.  This PR uses Custom Conversation Attribute 6 to record `channelType` using the same words as we had in Communication Channel previously.

One note to remember:  we will need to localize some of these words ('Call', 'Web') for non-English helplines.

### DIFF
--- a/plugin-hrm-form/src/services/InsightsService.js
+++ b/plugin-hrm-form/src/services/InsightsService.js
@@ -1,6 +1,7 @@
 /* eslint-disable camelcase */
 import { isNonDataCallType } from '../states/ValidationRules';
 import { isNotCategory, isNotSubcategory } from '../states/ContactFormStateFactory';
+import { mapChannelForInsights } from '../utils/mappers';
 
 function getSubcategories(task) {
   if (!task || !task.caseInformation || !task.caseInformation.categories) return [];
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/services/InsightsService.js
```diff
@@ -50,7 +57,7 @@ function mergeAttributes(previousAttributes, { conversations }) {
 }
 
 export async function saveInsightsData(twilioTask, task) {
-  const conversations = buildConversationsObject(task);
+  const conversations = buildConversationsObject(twilioTask.attributes, task);
```
GPaoloni, line 60: Just out of curiosity, what's the difference between `task` and `twilioTask`?

GPaoloni (approved):
Looks good!

murilovmachado (approved):

plugin-hrm-form/src/services/InsightsService.js
murilovmachado: `task` is our `contact`, `twilioTask` is really the task from twilio.

nickhurlburt:
Deployed to staging

PR #226, opened 2020-10-14
Fix package-lock.json. Lodash was not updated

murilovmachado:
Our code had inconsistencies between `package.json` and `package-lock.json` introduced by this [PR](https://github.com/tech-matters/flex-plugins/pull/203). `Lodash` was updated in `package.json`, but not in `package-lock.json`. This was causing `npm ci` to fail.


GPaoloni (approved):
Great, thanks!

PR #227, opened 2020-10-15
v0.9 bug fixes

GPaoloni:
This PR fixes two bugs:
- Chat capacity is decreased after a task is transferred to another counselor (1st commit).
- After accepting a task, if it's a transferred one, the state of the form saved by previous counselor is loaded into current form, OR chatbot memory is loaded otherwise (2nd commit). Previous behavior loaded chatbot memory in every scenario, sometimes overriding original counselor's input (e.g. if chatbot memory have unknown age but 1st counselor manage to capture that data, we want it to be preserved after a transfer).

### DIFF
--- a/plugin-hrm-form/src/utils/setUpActions.js
+++ b/plugin-hrm-form/src/utils/setUpActions.js
@@ -173,16 +173,6 @@ export const afterCancelTransfer = payload => {
 export const hangupCall = fromActionFunction(saveEndMillis);
 
 /**
- * @param {ReturnType<typeof getConfig> & { translateUI: (language: string) => Promise<void>; getMessage: (messageKey: string) => (language: string) => Promise<string>; }} setupObject
- * @returns {ActionFunction}
- */
-export const beforeWrapupTask = setupObject => async payload => {
-  const { featureFlags } = setupObject;
-  const { task } = payload;
-  if (featureFlags.enable_manual_pulling && task.taskChannelUniqueName === 'chat') await adjustChatCapacity('decrease');
-};
-
-/**
  * Helper to determine if the counselor should send a message before leaving the chat
  * @param {string} channel
  */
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -226,6 +218,25 @@ export const sendInsightsData = setupObject => async payload => {
...
+  if (featureFlags.enable_manual_pulling && task.taskChannelUniqueName === 'chat') await adjustChatCapacity('decrease');
+};
+
+/**
+ * @param {ReturnType<typeof getConfig> & { translateUI: (language: string) => Promise<void>; getMessage: (messageKey: string) => (language: string) => Promise<string>; }} setupObject
+ * @returns {ActionFunction}
+ */
+export const beforeCompleteTask = setupObject => async payload => {
+  await sendInsightsData(setupObject)(payload);
+  await decreaseChatCapacity(setupObject)(payload);
```
nickhurlburt, line 237: VPS put this in `beforeWrapupTask` rather than `beforeCompleteTask` because they were worried about race conditions.  If the capacity decrease did not finish before the task was completed, then a new task could be assigned before the capacity was decreased.  I am fine with you changing this as long as you are confident that this race condition will not occur.

plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -97,12 +102,8 @@ export const afterAcceptTask = setupObject => async payload => {
   const { featureFlags } = setupObject;
   const { task } = payload;
 
-  if (featureFlags.enable_transfers) {
-    await takeControlIfTransfer(task);
-    await restoreFormIfTransfer(task);
-  }
-
-  prepopulateForm(task);
+  if (featureFlags.enable_transfers && TransferHelpers.hasTransferStarted(task)) handleTransferredTask(task);
```
nickhurlburt, line 105: As we create more feature flags, they may make our code harder to work with.  We can talk about removing some of the old ones if they are getting in the way.  The decision to remove would be based on whether we think there are some helplines that will not want the feature.

nickhurlburt (approved):
Looks great.  See my note about race conditions and do what you think makes sense.

Another thing I notice is that no tests needed to be changed.  This is a sign that our test coverage is not good here.  I realize that testing this kind of thing is hard.  We don't need to fix that right now.  But something to keep in mind.

When this is on staging, please raise with Joan whether she wants to ship it to production or wait.

plugin-hrm-form/src/utils/setUpActions.js
GPaoloni: I've added a long timeout (10 seconds) and tried several times and they always execute in the same order, so I think we shouldn't face any race condition here. Nice to be aware and keep an eye just in case tho.


Screenshot from 2020-10-15 15-42-08
Screenshot from 2020-10-15 15-39-59

nickhurlburt: Great!

PR #232, opened 2020-10-16
Gian console bash (part 1)

GPaoloni:
This PR addresses some of the warnings displayed in console, listed [here](https://benetech.app.box.com/notes/730821213051), under "Gian - part 1" section.


nickhurlburt (approved):
Looks good to me!  Out of curiosity, why does Github show `CustomCRMContainer.jsx` and `CustomCRMContainer.tsx` as completely different files, but the other files are marked as being moved?

GPaoloni:
@nickhurlburt To be honest I don't know.. I've rewritten several parts of it, as it was previously a `React.Component` and I turned into a `React.FunctionalComponent`, so it's likely related to that.

plugin-hrm-form/src/components/CustomCRMContainer.tsx
```diff
@@ -0,0 +1,50 @@
...
+        const counselorsList = await populateCounselors();
+        dispatch(populateCounselorsState(counselorsList));
+      } catch (err) {
+        // TODO (Gian): probably we need to handle this in a nicer way
+        console.error(err.message);
+      }
+    };
+
+    fetchPopulateCounselors();
+  }, [dispatch]);
```
murilovmachado, line 30: Do we need dispatch here in the dependency array?

murilovmachado (approved):

plugin-hrm-form/src/components/CustomCRMContainer.tsx
GPaoloni: Actually it was added by a react-hooks eslint rule..I've been reading about its recomended to use it, but if you think this can cause unnecessary re-calls to this effect, we can remove it.

murilovmachado: no, it's fine. I was just curious about it. I already had the impression that it was suggested by eslint.

PR #234, opened 2020-10-16
(WIP) Convert FieldSelect to TS

nickhurlburt:
@GPaoloni I'm getting stuck on this because I'm getting lost in the types especially for the form fields.  It seems we are creating parallel types for prop types and TypeScript for all the form types.  Help!


nickhurlburt:
The initial warning I was trying to fix is:

> Failed prop type: Invalid prop `label` of type `object` supplied to `FieldSelect`, expected `string`.

nickhurlburt:
With the customization changes I think this is obsolete.  Closing.

PR #235, opened 2020-10-16
Move FieldText to TS and clear up DOM element warnings in TabbedForms

nickhurlburt:
Eliminates the following console warnings:
* Failed prop type: Invalid prop `label` of type `object` supplied to `FieldText`, expected `string`.
* React does not recognize the `searchTab` prop on a DOM element
* React does not recognize the `isPlaceholder` prop on a DOM element
* Failed prop type: Invalid prop color of value #506BA599 supplied to Checkbox, expected one of [“primary”,“secondary”,“default”].

Moves `FieldText` to TS for the first one.  For the others, deconstructing the named property and passing `...rest` to the nested component.

### DIFF
--- a/plugin-hrm-form/src/components/FieldText.tsx
+++ b/plugin-hrm-form/src/components/FieldText.tsx
@@ -11,9 +11,10 @@ type OwnProps = {
   placeholder?: string;
   field: FormFieldType;
   rows?: number;
-  handleBlur: any;
-  handleChange: any;
-  handleFocus: any;
+  // Someday we should make this consistent with DefaultEventHandlers in /components/common/forms/types.ts
+  handleBlur: React.FocusEventHandler<HTMLDivElement | HTMLTextAreaElement | HTMLInputElement | HTMLSelectElement>;
+  handleChange: React.ChangeEventHandler<HTMLTextAreaElement | HTMLInputElement | HTMLSelectElement>;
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/components/FieldText.tsx
```diff
@@ -1,14 +1,32 @@
...
-const FieldText = ({ id, label, placeholder, field, rows, handleBlur, handleChange, handleFocus, ...rest }) => (
+type OwnProps = {
+  id: string;
+  label?: string | JSX.Element;
+  placeholder?: string;
+  field: FormFieldType;
+  rows?: number;
+  handleBlur: any;
+  handleChange: any;
+  handleFocus: any;
```
murilovmachado, line 16: We can define these handlers like this:
```
handleBlur: React.FocusEventHandler<HTMLTextAreaElement | HTMLInputElement | HTMLSelectElement>;
handleFocus: React.FocusEventHandler<HTMLTextAreaElement | HTMLInputElement | HTMLSelectElement>;
handleChange: React.ChangeEventHandler<HTMLTextAreaElement | HTMLInputElement | HTMLSelectElement>; 
```

See `/components/common/forms/types.ts` to see an example

murilovmachado (approved):

GPaoloni (approved):
Great!

plugin-hrm-form/src/components/FieldText.tsx
nickhurlburt: Great, I've added these, and left a note that we should try to standardize with the other file.

PR #236, opened 2020-10-16
Murilo warnings hunt

murilovmachado:
This PR:
- Removes unnecessary console.logs
- Fixes `invalid propType searchResult` passed into `Search` component
- Makes `Search` a functional component
- Transform `Search` into a typescript file


plugin-hrm-form/src/components/search/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+  currentIsCaller?: boolean;
+  handleSelectSearchResult: (contact: SearchContact) => void;
+};
+
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ReturnType<typeof mapStateToProps> & ReturnType<typeof mapDispatchToProps>;
+
+const Search: React.FC<Props> = props => {
+  const [mockedMessage, setMockedMessage] = useState('');
+  const [searchParams, setSearchParams] = useState<any>({});
```
murilovmachado, line 34: We should use a "real" type in here instead of `any`, but I wanna leave it for a future PR.

nickhurlburt (approved):
This looks fine to me.

PR #238, opened 2020-10-20
[WIP] Prototyping form customization (with rhf)

GPaoloni:
Notes on this task: https://benetech.app.box.com/notes/729346531651

This PR is a prototype for generating forms from a config file (json) introducing [React Hook Forms](https://github.com/react-hook-form/react-hook-form) library. The form state is saved internally in the form component, and copied into Redux onBlur event, validation is handled by rhf components, updates to global state are dispatched on each onBlur event (can be easily changed to be other event tho) and structure for the global state needs to change for this approach. Initially, the child tab info is the one being generated this way.

### DIFF
--- a/plugin-hrm-form/package-lock.json
+++ b/plugin-hrm-form/package-lock.json
@@ -1190,9 +1190,9 @@
       }
     },
     "@craco/craco": {
-      "version": "5.6.4",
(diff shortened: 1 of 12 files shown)


GPaoloni:
Closed as this PR served the purpose of protoyping.

PR #240, opened 2020-10-30
[WIP] Customization prototyping (using ZA specs)

GPaoloni:
This PR is related to https://github.com/tech-matters/hrm/pull/55
Note: In order to run contact searches for this PR (test `ContactDetails`), you must use the above linked version of the backend. Contact search will break otherwise.

This PR is a proof of concept for customizing tabbed forms using rhf, following the ZA specifications.

Links to specs: 
Form: https://benetech.app.box.com/file/729083763749?comment_id=353433215&sb=/activity/comments/353433215
Locations: https://benetech.app.box.com/file/727816126730?comment_id=353597671&sb=/activity/comments/353597671


plugin-hrm-form/src/services/ContactService.js
```diff
@@ -43,19 +43,31 @@ const createCategory = (obj, [category, { subcategories }]) => ({
...
+  const { callType, metadata, caseInformation } = form;
+
+  const callerInformation = groupName(form.callerInformation);
+  const childInformation = groupName(form.childInformation);
 
   const categoriesObject = createCategoriesObject();
   const { categories } = form.categories.reduce((acc, path) => set(path, true, acc), { categories: categoriesObject });
 
   const transformed = {
+    definitionVersion: 'v1', // TODO: put this in config (like feature flags). Question: Should this be inside rawForm or be a separate column for each contact?
```
nickhurlburt, line 70: First off... everything I'm going to say below can be ignored in the short term while we only have one new version.  You can take shortcuts here that we can improve later.

It makes more sense to me to put the definition version inside the form definition files (eg, `callerForm.json`) themselves.  Then our core code doesn't change with different versions.  The problem with that is that it would need to be added to all the separate files.  It may be better to have a single form definition file for all the contact and case tabs makes more sense than separating them out.  Another option would be to have a subdirectory of `formDefinitions` named "v1" with all the files inside it, and/or have the config files named for the version, such as `callerForm_v1.json` or `v1_callerForm.json`.

You probably do need to have a Twilio configuration entry for the "live" form.  Then that is what you load for new input, but when you load old contacts out of the HRM, you look at their versions and pull the needed configuration for them.

In the short term, I think what you're doing is fine, since we only have one version.  But you'll need to change it pretty soon.

Lastly, we'll probably want to name these by helpline, for example, "zm_v1" or "zm_1_0_0".  I think that will be easier to follow with multiple helplines.  Again, in the very first version, it doesn't matter too much.

nickhurlburt: And I think it makes sense to live in the `rawForm`.  Then when you read back the form data from the HRM, the form is self-describing, and you can then look up the relevant version and display it.

plugin-hrm-form/src/components/ContactDetails.tsx
```diff
@@ -62,6 +54,9 @@ const Details = ({
     CONTACT_SUMMARY,
   } = ContactDetailsSections;
 
+  const deTransform = (e: FormItemDefinition, obj: { [k: string]: any }) =>
+    ['firstName', 'lastName'].includes(e.name) ? obj.name[e.name] : obj[e.name];
+
```
nickhurlburt, line 59: I don't really like detransforming this here, since I think it belongs at the layer where we first get it back from the API.  Then inside the rest of the plugin, everything is clean.  (However, if that's a ton of work, we can keep this for now.)

nickhurlburt (commented):
@GPaoloni I left some comments on the latest commit, hopefully they are not too confusing.

plugin-hrm-form/src/services/ContactService.js
```diff
@@ -43,19 +43,31 @@ const createCategory = (obj, [category, { subcategories }]) => ({
...
  * VisibleForTesting
  * @param {TaskEntry} form
  */
 // eslint-disable-next-line import/no-unused-modules
 export function transformForm(form) {
-  const { callType, callerInformation, childInformation, metadata, caseInformation } = form;
+  const { callType, metadata, caseInformation } = form;
+
+  const callerInformation = groupName(form.callerInformation);
+  const childInformation = groupName(form.childInformation);
```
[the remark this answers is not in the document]
GPaoloni: Totally agree, will do!

plugin-hrm-form/src/services/ContactService.js
GPaoloni: > Another option would be to have a subdirectory of formDefinitions named "v1" with all the files inside it

I was thinking about this option, as is the easiest to work with (to avoid having a 1000+ lines json).

> You probably do need to have a Twilio configuration entry for the "live" form

Load the latest one when the pluggin first mounts as you suggest (current version can live alongside feature flags, in config)

> Then that is what you load for new input, but when you load old contacts out of the HRM, you look at their versions and pull the needed configuration for them

Yes, and maybe cache them (Redux seems as a good place for this).

> Lastly, we'll probably want to name these by helpline, for example, "zm_v1" or "zm_1_0_0".

I though that this will live in serverless repo, and that each helpline will have it's own account and hence avoiding any collision, but let me know if I'm wrong.

plugin-hrm-form/src/components/ContactDetails.tsx
GPaoloni: Agreed!

plugin-hrm-form/src/services/ContactService.js
nickhurlburt: >> Another option would be to have a subdirectory of formDefinitions named "v1" with all the files inside it

> I was thinking about this option, as is the easiest to work with (to avoid having a 1000+ lines json).

Sure.

>> You probably do need to have a Twilio configuration entry for the "live" form

> Load the latest one when the pluggin first mounts as you suggest (current version can live alongside feature flags, in config)

Yes

>> Then that is what you load for new input, but when you load old contacts out of the HRM, you look at their versions and pull the needed configuration for them

> Yes, and maybe cache them (Redux seems as a good place for this).

Yes, caching is a great idea.  Redux makes sense.

>> Lastly, we'll probably want to name these by helpline, for example, "zm_v1" or "zm_1_0_0".

> I though that this will live in serverless repo, and that each helpline will have it's own account and hence avoiding any collision, but let me know if I'm wrong.

That sounds right, though namespacing them by helpline buys us flexibility.  We could, for example, in the short term just push all the configuration files with every account, especially if there's nothing secret about them.  Or, we may end up having one big S3 bucket where we store them and pull from that.  I've just had enough times where later on I've said "oh shoot, I wish I would have kept this info" that it feels like a good idea :)

GPaoloni: Makes total sense, I'll follow your advice!

GPaoloni:
Closed as this PR served the purpose of protoyping.

PR #241, opened 2020-11-05
[285] - add search results tabs

andvirga:
- Added new tabs to Search result components
- Applied styling (tried to mimic code styling)

### DIFF
--- a/plugin-hrm-form/src/components/search/SearchResults/index.tsx
+++ b/plugin-hrm-form/src/components/search/SearchResults/index.tsx
@@ -1,7 +1,9 @@
 /* eslint-disable react/prop-types */
 import React, { useState } from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
 import { ButtonBase } from '@material-ui/core';
-import { Template } from '@twilio/flex-ui';
+import { Template, Tab as TwilioTab, ITask } from '@twilio/flex-ui';
 
 import ContactPreview from '../ContactPreview';
 import { SearchContactResult, SearchContact } from '../../../types/types';
@@ -15,13 +17,24 @@ import {
   StyledFormControlLabel,
   StyledSwitch,
   SwitchLabel,
+  StyledLink,
+  StyledTabs,
+  StyledSearchTabs,
+  StyledSearchText,
+  StyledTabLabel,
+  StyledFolderIcon,
+  BoldText,
 } from '../../../styles/search';
 import ConnectDialog from '../ConnectDialog';
 import Pagination from '../../Pagination';
+import * as SearchActions from '../../../states/search/actions';
+import { SearchPages, SearchPagesType } from '../../../states/search/types';
+import { namespace, searchContactsBase } from '../../../states';
 
 export const CONTACTS_PER_PAGE = 20;
 
-type SearchResultsProps = {
+type OwnProps = {
+  task: ITask;
   currentIsCaller: boolean;
   results: SearchContactResult;
   onlyDataContacts: boolean;
@@ -30,9 +43,14 @@ type SearchResultsProps = {
   toggleNonDataContacts: () => void;
   handleBack: () => void;
   handleViewDetails: (contact: SearchContact) => void;
+  changeSearchPage: (SearchPagesType) => void;
+  currentPage: SearchPagesType;
 };
 
-const SearchResults: React.FC<SearchResultsProps> = ({
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ReturnType<typeof mapStateToProps> & typeof mapDispatchToProps;
+
+const SearchResults: React.FC<Props> = ({
   currentIsCaller,
   results,
   onlyDataContacts,
@@ -41,6 +59,8 @@ const SearchResults: React.FC<SearchResultsProps> = ({
   toggleNonDataContacts,
   handleBack,
   handleViewDetails,
+  changeSearchPage,
+  currentPage,
 }) => {
   const [currentContact, setCurrentContact] = useState(null);
   const [anchorEl, setAnchorEl] = useState(null);
@@ -74,6 +94,17 @@ const SearchResults: React.FC<SearchResultsProps> = ({
   const { contacts, count } = results;
   const pagesCount = Math.ceil(count / CONTACTS_PER_PAGE);
 
+  const toggleTabs = () => {
+    // eslint-disable-next-line no-unused-expressions
+    currentPage === SearchPages.resultsContacts
+      ? changeSearchPage(SearchPages.resultsCases)
+      : changeSearchPage(SearchPages.resultsContacts);
+  };
+
+  const tabSelected = tabName => {
+    changeSearchPage(tabName);
+  };
+
   return (
     <>
       <ResultsHeader>
@@ -85,15 +116,25 @@ const SearchResults: React.FC<SearchResultsProps> = ({
             </BackText>
           </ButtonBase>
         </Row>
-        <Row style={{ paddingLeft: '24px' }}>
-          <BackText>
-            {count}
-            {count === 1 ? (
-              <Template code="SearchResultsIndex-Result" />
-            ) : (
-              <Template code="SearchResultsIndex-Results" />
-            )}
-          </BackText>
+        <Row style={{ justifyContent: 'center' }}>
+          <div style={{ width: '300px' }}>
+            <StyledTabs selectedTabName={currentPage} onTabSelected={tabSelected}>
+              <TwilioTab label="Contacts" uniqueName={SearchPages.resultsContacts}>
+                {[]}
+              </TwilioTab>
+              <TwilioTab
+                label={
+                  <StyledTabLabel>
+                    <StyledFolderIcon />
+                    Cases
+                  </StyledTabLabel>
+                }
+                uniqueName={SearchPages.resultsCases}
+              >
+                {[]}
+              </TwilioTab>
+            </StyledTabs>
+          </div>
         </Row>
       </ResultsHeader>
       <ConnectDialog
@@ -105,6 +146,26 @@ const SearchResults: React.FC<SearchResultsProps> = ({
       />
       <ListContainer>
         <ScrollableList>
+          <StyledSearchTabs>
+            <StyledSearchText>
+              There are&nbsp;
+              <BoldText>
+                {/* Intentionally we must show the option different at the one currently selected */}
+                {`${count} ${currentPage === SearchPages.resultsContacts ? 'cases' : 'contacts'}`}
+              </BoldText>
+              &nbsp;returned in this search.&nbsp;
+            </StyledSearchText>
+            <StyledLink onClick={toggleTabs}>
+              <Template
+                code={
+                  // Intentionally we must show the option different at the one currently selected
+                  currentPage === SearchPages.resultsContacts
+                    ? 'SearchResultsIndex-ViewCases'
+                    : 'SearchResultsIndex-ViewContacts'
+                }
+              />
+            </StyledLink>
+          </StyledSearchTabs>
           <StyledFormControlLabel
             control={<StyledSwitch checked={!onlyDataContacts} onChange={handleToggleNonDataContact} />}
             label={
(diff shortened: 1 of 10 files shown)


plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -15,12 +15,22 @@ import {
...
+  StyledSearchText,
+  StyledTabLabel,
+  BoldText,
+} from "../../../styles/search";
+import ConnectDialog from "../ConnectDialog";
+import Pagination from "../../Pagination";
 
 export const CONTACTS_PER_PAGE = 20;
 
+const SEARCH_RESULTS_TABS = Object.freeze({
```
nickhurlburt, line 29: Interesting... `Object.freeze` might be a method we could use more of.  Immutability is often a good practice.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -105,8 +143,35 @@ const SearchResults: React.FC<SearchResultsProps> = ({
       />
       <ListContainer>
         <ScrollableList>
+          <StyledSearchResults>
+            {activeTab === SEARCH_RESULTS_TABS.CONTACTS && (
+              <StyledSearchText>
+                There are &nbsp;{<BoldText>{count} cases</BoldText>}&nbsp; returned in this search.&nbsp;
```
nickhurlburt, line 149: Why the extra `&nbsp;`s?  This is creating a double space around the case count, though there's no reason to do that.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -105,8 +143,35 @@ const SearchResults: React.FC<SearchResultsProps> = ({
...
+              </StyledSearchText>
+            )}
+            {activeTab === SEARCH_RESULTS_TABS.CASES && (
+              <StyledSearchText>
+                There are more than &nbsp;<BoldText>{count} contacts</BoldText>&nbsp; returned in this search.&nbsp;
+              </StyledSearchText>
+            )}
+            <StyledLink onClick={toggleTabs}>
+              <Template
```
nickhurlburt, line 158: Good to see you've picked up on the localization standards!

plugin-hrm-form/src/styles/search/index.tsx
```diff
@@ -112,6 +116,27 @@ export const PrevNameText = styled(FontOpenSans)`
   color: #182b33;
 `;
 
+export const StyledSearchResults = styled('div')`
```
nickhurlburt, line 119: `StyledSearchResults` seems like a strange name to me, since this is just the tabs.  "`StyledSearchTabs`" perhaps?

nickhurlburt, line 119: In general, the styling of these tabs seems to depart from the mockups.  There's a slightly different icon for cases, the underline is a different thickness and length,  and the boldness of the letters doesn't seem right.  The mockups are an attempt to duplicate the style here:

image

Except that the underline is black instead of blue.

Ideally we want to match the mockups, but if there is a technical consideration that makes that difficult, we can discuss it.  @murilovmachado and @GPaoloni may have thoughts on this too.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -105,8 +143,35 @@ const SearchResults: React.FC<SearchResultsProps> = ({
...
         <ScrollableList>
+          <StyledSearchResults>
+            {activeTab === SEARCH_RESULTS_TABS.CONTACTS && (
+              <StyledSearchText>
+                There are &nbsp;{<BoldText>{count} cases</BoldText>}&nbsp; returned in this search.&nbsp;
+              </StyledSearchText>
+            )}
+            {activeTab === SEARCH_RESULTS_TABS.CASES && (
+              <StyledSearchText>
+                There are more than &nbsp;<BoldText>{count} contacts</BoldText>&nbsp; returned in this search.&nbsp;
```
nickhurlburt, line 154: The 'more than' is actually a mistake in the mockups (though you have no way to know that!).  As long as we are getting an accurate count of the cases, we can just say `There are {count} contacts`.

plugin-hrm-form/src/styles/search/index.tsx
nickhurlburt, line 119: For newly added styles, we should also add a `displayName` property.  For example:
```
StyledSearchResults.displayName = 'StyledSearchResults';
```
We have some legacy code that does not have this, but all new code should add this.  It makes navigating the developer console much easier.

nickhurlburt (commented):
Great start.  A few comments here, and Murilo and Gian may have some more comments as well.

A few notes on our conventions/standards (that we haven't actually told you about yet!):
* There are several linting errors here -- `npm run lint:fix` should take care of most of them.
* Generally, branches have the prefix of the person working on them.  For example, `andres_add_search_tabs`.  This just helps to know whose branch it is.  But no need to change this one.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Since `<BoldText>` add an extra `<p>` in the text, the spaces around it are  ignored. However I added manual spaces inside bold text and that solves this issue.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Thanks!

plugin-hrm-form/src/styles/search/index.tsx
andvirga: Good catch!

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Done!

plugin-hrm-form/src/styles/search/index.tsx
andvirga: Done, I've added that prop for all the new style that I've created.

murilovmachado:
Nice work! You got the behavior right. The next step to improve this would be:

- Moving the tab state to redux. Right now you have `const [activeTab, setActiveTab] = useState("contacts");` to handle the state, which is a good first step. But suppose the following scenario:
  1. User is on search results
  2. User selects **cases** tab
  3. User clicks on **categories** tab, up above
  4. User clicks back on the **search** tab
  5. Right now, the user will be on the **contacts** tab instead of **cases**

By moving the state to redux, we'll avoid this issue. More specifically, we thought of putting this into the **routing** state. Take a look at `/states/routing/types`. We probably want to add a new route to `AppRoutes`. You may want to coordinate this with @GPaoloni , since he was doing some rework on this area too.

There're other points to take a look (most of them @nickhurlburt has already made comments):
- Getting the styles closer to the mockups
- Fixing lint errors (You could also set your editor to alert in real type about lint errors)
- Creating tests (and possibly you'll need to fix some existing tests)

plugin-hrm-form/src/styles/search/index.tsx
andvirga: Done! Now i'm using `import { Tabs, Tab } from '@twilio/flex-ui'` instead of Material UI Tabs. They already have the format expected, I did only a minor tweak to have the indicator color as black.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Thanks!

murilovmachado:
Some comments on styling:

image
- You can drop the **7 results** from the header part
- There should be no vertical scrolling on the header part (probably **overflow-y** is doing this)
- Remove **brackets []** from below the tabs

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -15,12 +17,27 @@ import {
...
+import * as SearchActions from '../../../states/search/actions';
+import { SearchPages } from '../../../states/search/types';
+import { namespace, searchContactsBase } from '../../../states';
 
 export const CONTACTS_PER_PAGE = 20;
 
+const SEARCH_RESULTS_TABS = Object.freeze({
+  CONTACTS: 'results.contacts',
+  CASES: 'results.cases',
+});
```
murilovmachado, line 39: Now, you won't need this const anymore, since you have these values defined on redux:
`export const SearchPages = {
  form: 'form',
  resultsContacts: 'results.contacts',
  resultsCases: 'results.cases',
  details: 'details',
} as const;`

You can use `SearchPages` `resultsContacts` and `resultsCases` directly.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -30,6 +47,8 @@ type SearchResultsProps = {
   toggleNonDataContacts: () => void;
   handleBack: () => void;
   handleViewDetails: (contact: SearchContact) => void;
+  changeSearchPage: (string) => void;
```
murilovmachado, line 50: Here you can be more restrict and change it to be `changeSearchPage: (SearchPagesType) => void;`

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -30,6 +47,8 @@ type SearchResultsProps = {
...
+  currentPage: string;
```
murilovmachado, line 51: and this could be `currentPage: SearchPagesType;`

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -95,6 +127,26 @@ const SearchResults: React.FC<SearchResultsProps> = ({
             )}
           </BackText>
         </Row>
+        <Row style={{ justifyContent: 'center' }}>
+          <div style={{ width: '300px', padding: '20px' }}>
+            <StyledTabs selectedTabName={currentPage} onTabSelected={tabSelected}>
+              <TwilioTab label="Contacts" uniqueName={SEARCH_RESULTS_TABS.CONTACTS}>
```
murilovmachado, line 133: `SearchPages.resultsContacts`

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -95,6 +127,26 @@ const SearchResults: React.FC<SearchResultsProps> = ({
...
+                []
+              </TwilioTab>
+              <TwilioTab
+                label={
+                  <StyledTabLabel>
+                    <StyledFolderIcon />
+                    Cases
+                  </StyledTabLabel>
+                }
+                uniqueName={SEARCH_RESULTS_TABS.CASES}
```
murilovmachado, line 143: `SearchPages.resultsCases`

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -105,6 +157,26 @@ const SearchResults: React.FC<SearchResultsProps> = ({
       />
       <ListContainer>
         <ScrollableList>
+          <StyledSearchTabs>
+            <StyledSearchText>
+              There are&nbsp;
+              <BoldText>
+                {/* Intentionally we must show the option different at the one currently selected */}
+                {`${count} ${currentPage === SEARCH_RESULTS_TABS.CONTACTS ? 'cases' : 'contacts'}`}
```
murilovmachado, line 165: `SearchPages.resultsContacts`

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -105,6 +157,26 @@ const SearchResults: React.FC<SearchResultsProps> = ({
...
+              </BoldText>
+              &nbsp;returned in this search.&nbsp;
+            </StyledSearchText>
+            <StyledLink onClick={toggleTabs}>
+              <Template
+                code={
+                  // Intentionally we must show the option different at the one currently selected
+                  currentPage === SEARCH_RESULTS_TABS.CONTACTS
```
murilovmachado, line 173: `SearchPages.resultsContacts`

murilovmachado (approved):
Looks good!

For the code, see my suggestion on removing the const `SEARCH_RESULTS_TABS` and use `SearchPages` instead. Also, using `SearchPagesType` on some definitions to make them more restrictive.

I've also made some notes on stylings, please take a look at that.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: You're right! I've removed that const

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Good catch!

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Done!

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Done!

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Done!

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Done!

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: Done!

PR #242, opened 2020-11-06
Offline contact (data entry without a call/chat)

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-305 
     + https://bugs.benetech.org/browse/CHI-311
     + https://bugs.benetech.org/browse/CHI-337

This PR is related to:
- https://github.com/tech-matters/serverless/pull/27
- https://github.com/tech-matters/hrm/pull/60

This task is about ~~prototyping~~ implementing the "add offline task" button to add a new task without a contact.

Notes: 
- This PR includes `forms/formGenerators.ts` among a few other pieces that are part of the bigger customization task. Unit testing for this is pending and will be addressed with customization.
- To test this PR, change serverless url to be dev environment one.


plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -149,6 +149,7 @@ const setUpComponents = setupObject => {
   if (featureFlags.enable_case_management) Components.setUpCaseList();
 
   if (featureFlags.enable_manual_pulling) Components.setUpManualPulling();
+  Components.setUpOfflineContact();
```
GPaoloni, line 152: Should this be under a feature flag?

plugin-hrm-form/src/components/tabbedForms/TabbedForms.jsx
```diff
@@ -99,30 +113,62 @@ const TabbedForms = props => {
...
-      <BottomBar
-        tabs={tabs.length}
-        form={form}
-        changeTab={props.changeTab}
-        handleCompleteTask={props.handleCompleteTask}
-        handleValidateForm={props.handleValidateForm}
-      />
-    </TabbedFormsContainer>
+    <FormProvider {...methods}>
+      <form style={{ height: '100%' }}>
```
GPaoloni, line 150: Here I wrap everything in a form for a11y purposes, but this is not really necessary.

plugin-hrm-form/src/components/tabbedForms/TabbedForms.jsx
```diff
@@ -99,30 +113,62 @@ const TabbedForms = props => {
...
+        <TabbedFormsContainer>
+          <TopNav>
+            <TransparentButton onClick={handleBackButton}>&lt; BACK</TransparentButton>
+          </TopNav>
+          <StyledTabs name="tab" variant="scrollable" scrollButtons="auto" value={tab} onChange={handleTabsChange}>
+            {tabs}
+          </StyledTabs>
+          {task.attributes.isContactlessTask && <ContactlessTaskTab display={tab === 1} />}
```
GPaoloni, line 158: Instead of handling this in the same way we've been doing it, rhf is able to validate the form if the inputs that define it are mounted. This is adding the `ContactlessTaskTab` to be allways mounted, but hiding the entire UI unless the tab is 1. Same approach can be seen in https://github.com/tech-matters/flex-plugins/pull/240, but for all tabs.

plugin-hrm-form/src/services/InsightsService.js
```diff
@@ -56,10 +56,38 @@ function mergeAttributes(previousAttributes, { conversations }) {
...
 
-export async function saveInsightsData(twilioTask, task) {
-  const conversations = buildConversationsObject(twilioTask.attributes, task);
+const overrideAttributes = (attributes, form) => {
+  const { date, time } = form.contactlessTask;
+
+  const [y, m, d] = date.split('-').map(s => parseInt(s, 10));
+  const [hh, mm] = time.split(':').map(s => parseInt(s, 10));
+
+  const dateUTC = Date.UTC(y, m - 1, d, hh, mm);
```
GPaoloni, line 65: @nickhurlburt this fixes the off-by-one error when creating the date, but I'm still not sure if this is the timezone Twilio uses. Is this something worth asking?

plugin-hrm-form/src/states/ContactFormStateFactory.js
```diff
@@ -581,7 +584,7 @@ const recursivelyCreateBlankForm = formDefinition => {
       }
     });
 
-  return initialState;
+  return { ...initialState, contactlessTask: contactlessTaskTabDefinition.reduce(createFormItem, {}) };
```
GPaoloni, line 587: Adding this after the entire form is created, because if handled by rhf, we don't want this properties to be objects like { value, touched, error }, but plain values instead.

plugin-hrm-form/src/services/InsightsService.js
nickhurlburt: We should be sending Insights a `date` value of milliseconds past the epoch.  This does not have a time zone.  It does look like printing the date as you're doing below does yield epoch millis.

I would expect that counselors would be inputting the date and time in their local time zone, so we should create a date object based on the date/time in their local time zone.

plugin-hrm-form/src/services/InsightsService.js
```diff
@@ -56,10 +56,38 @@ function mergeAttributes(previousAttributes, { conversations }) {
...
+    dateUTC,
+    new Date(dateUTC),
+    ' and channel ',
+    form.contactlessTask.channel,
+  );
+  return {
+    ...attributes,
+    conversations: {
+      ...attributes.conversations,
+      communication_channel: form.contactlessTask.channel,
```
nickhurlburt, line 78: So a couple issues:
* The values that the counselor can choose in the dropdown should be properly capitalized.  They should be the options that would be available from the `mapChannelForInsights` function. 
* I think we want to use `communication_channel`, but elsewhere in this file we're still using Custom attribute 6.  I think we just need to ask Joan what she'd like, and then we use that consistently everywhere.

plugin-hrm-form/src/components/tabbedForms/ContactlessTaskTabDefinition.ts
```diff
@@ -0,0 +1,26 @@
...
+    type: 'select',
+    label: 'Channel',
+    options: channelOptions,
+    required: { value: true, message: 'RequiredFieldError' },
+  },
+  {
+    name: 'date',
+    type: 'date-input',
+    label: 'Date of Contact',
+    required: { value: true, message: 'RequiredFieldError' },
```
nickhurlburt, line 18: How hard would it be to prevent them from picking a date/time in the future?

nickhurlburt (commented):
Left several comments.  I think we should try to demo this to Joan tomorrow, as there are several questions.  A short working list:
- What will we do able the scrolled tabs?
- What do we want the call type options to be, and which field do we put them in?
- What happens when Offline Contact Record is clicked a second time?
- How do we want the Offline Contact Record button to look when there are no active tasks?

Some more questions/comments:

The fonts here for all three inputs and the required field error look different than elsewhere.  They should look the same.  And the date field uses a different format than the date field in search.  We should be consistent.

image

Having the scroll bar on the tabs doesn't look great:
image

Though I'm not all that sure what to do.  Is there any way we can recover some of the space from the conversation pane, maybe by decreasing its width?


What is the intended behavior when Offline Contact Record is clicked a second time?  It doesn't seem to do anything at the moment.

GPaoloni:
> What is the intended behavior when Offline Contact Record is clicked a second time? It doesn't seem to do anything at the moment.

@nickhurlburt I've tested clicking twice and I got 2 different tasks. I saw that there were 4 "pending" tasks (in Twilio console), so is there chance that you were offline? We can prevent getting more than one if desired tho.
Screenshot from 2020-11-18 10-34-30

plugin-hrm-form/src/components/tabbedForms/ContactlessTaskTabDefinition.ts
GPaoloni: If the definition is in JSON format, this might be a bit harder, but I believe we can achieve the same with a JS file), combined with a `validate` function like:

```
rules={{
  validate: (data) => {
    const date = new Date(data);
    return date.getDate() !== 13;
  },
}}
```
As far as I know, we can also provide [async validations](https://github.com/react-hook-form/react-hook-form/blob/9422033ad469901cd5d34ba2868d61de79b23e35/examples/asyncFieldValidation.tsx#L23), so I can imagine hitting a validation endpoint with more advanced requirements, o use an [schema validation like Yup](https://react-hook-form.com/get-started#schemavalidation), tho  I haven't tried those features..

plugin-hrm-form/src/components/tabbedForms/TabbedForms.jsx
GPaoloni: Yes, and I think everything works fine, with the only difference that errors are cleared when you change tasks, so the UI does not show them anymore, until next submission attempt. We can change this behavior by triggering a validation on every mount but the first one.

Another difference that is intentionally left on purpose right now is the fact that when you attempt a submission, the Contact Tab is validated first, if it fails the others are not validated, if it succeeds the other ones are validated. This is caused by the way `handleSubmit`  (provided y rhf) works. Every tab handled by rhf will be validated at the same time so this will go away if we start using it for all tabs, but in the meanwhile we can achieve the same behavior with just a few changes, if desired.

nickhurlburt: Okay.  I'm not too concerned about this, especially since it's temporary.

nickhurlburt:
A couple more issues to work on/think about:
- Can we add the offline contact button to the No Tasks View, even if the design is not currently clear?
- Looks like the counselor will need to be available to create an offline contact.  In our first, data-entry-only version, this is fine, since there will be no other contacts coming in.  But later, that would be an issue.  
- We have been delaying making Jana's style changes to the tabs, which make the tab titles Title Case (eg, "Child Info") and make them a little smaller (see https://marvelapp.com/prototype/f7f0434/screen/73656631).  It may be time to do that to try to fit them in so we don't get a scroll bar on the tabs.

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -0,0 +1,444 @@
...
+            );
+          }}
+        </ConnectForm>
+      );
+    default:
+      return null;
+  }
+};
+
+export const createFormFromDefinition = (definition: FormDefinition) => (parents: string[]) => (
```
andvirga, line 388: Since this is a core function I would strongly suggest adding a JSDoc description like this one:
```javascript
/**
 * Creates a RHF Form based in JSON definition.
 * @param definition form JSON definition (schema).
 * @param parents Elements to extract from structure.
 * @param updateCallback callback called to update form state.
export const createFormFromDefinition = (definition: FormDefinition) => (parents: string[]) => (
  updateCallback: () => void,
): JSX.Element[] => definition.map(getInputType(parents, updateCallback));
 */
```

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -0,0 +1,444 @@
...
+  updateCallback: () => void,
+): JSX.Element[] => definition.map(getInputType(parents, updateCallback));
+
+// eslint-disable-next-line react/no-multi-comp
+export const makeFormColumns = (formItems: JSX.Element[]) => {
```
andvirga, line 393: Same here, please add a JSDocdescription to this function too

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -0,0 +1,444 @@
...
+    case 'textarea':
+      return '';
+    case 'date-input':
+      return '';
+    case 'time-input':
+      return '';
+    default:
+      return null;
+  }
+};
```
andvirga, line 439: You can group conditions with the same result: (Don't trust me, verify it ;) )
```javascript
  switch (def.type) {
    case 'select':
      return def.options[0].value;
    case 'dependent-select':
      return def.defaultOption.value;
    case 'checkbox':
      return false;
    case 'mixed-checkbox':
      return def.initialChecked === undefined ? 'mixed' : def.initialChecked;
    case 'input':
    case 'numeric-input':
    case 'textarea':
    case 'date-input':
    case 'time-input':
    default:
      return null;
  }
```

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -0,0 +1,444 @@
...
+
+  const [l, r] = [items.slice(0, m), items.slice(m)];
+
+  return (
+    <TwoColumnLayout>
+      <ColumnarBlock>{l}</ColumnarBlock>
+      <ColumnarBlock>{r}</ColumnarBlock>
+    </TwoColumnLayout>
+  );
+};
```
andvirga, line 410: Also, this function needs to be renamed, I would suggest something like `buildTwoColumnFormLayout` and not a generic name since this function will always return a two-column layout form.

plugin-hrm-form/src/components/tabbedForms/TabbedForms.jsx
```diff
@@ -99,30 +113,62 @@ const TabbedForms = props => {
...
+              const { isContactlessTask, ...rest } = task.attributes;
+              await task.setAttributes(rest); // skip insights override
+              Actions.invokeAction('CompleteTask', { task });
+            },
+          },
+        ]
+      : undefined;
+
+  console.log('>>> errors: ', methods.errors);
+  console.log('>>> getValues: ', methods.getValues());
```
andvirga, line 146: These are great for debugging, but we can leave it commented to avoid showing sensitive data.

plugin-hrm-form/src/states/ContactState.js
```diff
@@ -353,6 +365,18 @@ export function reduce(state = initialState, action) {
         },
       };
     }
+    case UPDATE_CONTACTLESS_TAKS: {
```
andvirga, line 368: small typo, not big deal

plugin-hrm-form/src/states/ActionTypes.js
```diff
@@ -6,6 +6,7 @@ export const SAVE_END_MILLIS = 'SAVE_END_MILLIS';
 export const CHANGE_TAB = 'CHANGE_TAB';
 export const RESTORE_ENTIRE_FORM = 'RESTORE_ENTIRE_FORM';
 export const HANDLE_SELECT_SEARCH_RESULT = 'HANDLE_SELECT_SEARCH_RESULT';
+export const UPDATE_CONTACTLESS_TAKS = 'UPDATE_CONTACTLESS_TAKS';
```
andvirga, line 9: same typo here ;)

andvirga:
The PR looks great! I've only left some minor comments, I will review it again once you finish.

andvirga (approved):

plugin-hrm-form/src/components/tabbedForms/ContactlessTaskTabDefinition.ts
nickhurlburt: I think we will need to check that the date/time is not in the future

plugin-hrm-form/src/components/tabbedForms/TabbedForms.jsx
```diff
@@ -73,6 +85,8 @@ const TabbedForms = props => {
     />,
   );
 
+  if (task.attributes.isContactlessTask) body.push(<div style={{ display: 'none' }} />); // fake it so lenght is accurate
```
nickhurlburt, line 88: `s/lenght/length/`

nickhurlburt: Are we doing this because contactless task is rendered in a different way?

plugin-hrm-form/src/utils/setUpComponents.js
```diff
@@ -13,10 +16,58 @@ import CaseList from '../components/caseList';
...
+      whatsappColor,
+    }}
+  />
+);
+
+/**
+ * Returns de UI for the "Add..." section
+ * @param {ReturnType<typeof getConfig> & { translateUI: (language: string) => Promise<void>; getMessage: (messageKey: string) => (language: string) => Promise<string>; }} setupObject
+ */
+const addButonsUI = setupObject => {
```
nickhurlburt, line 54: `s/Butons/Buttons/`

plugin-hrm-form/src/utils/setUpComponents.js
```diff
@@ -13,10 +16,58 @@ import CaseList from '../components/caseList';
...
+      webColor,
+      facebookColor,
+      smsColor,
+      whatsappColor,
+    }}
+  />
+);
+
+/**
+ * Returns de UI for the "Add..." section
```
nickhurlburt, line 51: `s/de/the/`

plugin-hrm-form/src/styles/HrmStyles.tsx
```diff
@@ -600,4 +610,205 @@ export const ManualPullButtonBase = withStyles({
...
+  color: ${props => props.theme.colors.errorColor};
+  font-size: 10px;
+  line-height: 1.5;
+  letter-spacing: normal;
+`;
+FormError.displayName = 'FormError';
+
+type FormInputProps = { error?: boolean };
+export const FormInput = styled('input')<FormInputProps>`
+  /* ---------- Input ---------- */
```
nickhurlburt, line 643: I am no expert on CSS, but this large section feels like it may be hard to work with later.

plugin-hrm-form/src/components/OfflineContactButton/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+type OwnProps = {};
+
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ConnectedProps<typeof connector>;
+
+const OfflineContactButton: React.FC<Props> = ({ tasks }) => {
+  const disabled = Array.from(tasks.values()).some(t => t.channelType === 'default');
+
+  const onClick = async () => {
+    await assignMeContactlessTask();
```
nickhurlburt, line 18: I clicked on this button while I was set to 'Offline' and, from a user perspective, nothing happened.  I think we want to communicate with users better about what's going on.  One suggestion: if the button is clicked but the worker is not Available, send an error notification saying you must be available to start a new offline contact.

plugin-hrm-form/src/states/DomainConstants.ts
```diff
@@ -21,6 +21,18 @@ export const channelTypes = {
 
 export type ChannelTypes = typeof channelTypes[keyof typeof channelTypes];
 
+export const otherContactChannels = {
```
nickhurlburt, line 24: Helplines will need to customize which of these options they show in the dropdown.  I think we can keep this as is for now, but keep that in mind.

plugin-hrm-form/src/___tests__/services/ContactService.test.js
```diff
@@ -86,6 +91,11 @@ describe('transformForm', () => {
         status: '',
         callSummary: 'My summary',
       },
+      contactlessTask: {
```
nickhurlburt, line 94: We need to have this come back from the HRM as if the contact was created at the time specified.  And that needs to show up in search and other places where the contact details are visible.

For example, I just submitted a task (at about 11:40am my time) with these values:
Screen Shot 2020-11-20 at 11 41 45 AM

But when I search for it I see:
image
Search and Insights should match.  Helplines shouldn't need to know that they are two systems.

Unfortunately this adds a bunch more work on the HRM... and may include adding another field for the date (rather than just the auto `dateCreated` field) and doing searches on that.  And I think we do still want to record in the HRM the fact that this was an offline contact.

nickhurlburt (commented):
@GPaoloni Here's a partial review so far.  Might have more, still looking.

nickhurlburt:
Additionally, I've tested creating one contactless task for whether the channel, date and time get added to Insights as expected.  Looks good.

nickhurlburt:
One more thing I noticed... the [mockups](https://marvelapp.com/prototype/f7f0434/screen/73656631) have the tabs more narrow.  Can we make the tabs more narrow to try to keep from needing to scroll?

plugin-hrm-form/src/components/tabbedForms/TabbedForms.jsx
GPaoloni: For React Hook Form to run the validations, the form must be mounted, so we can't render the proper tab with `body[tab]` (only 1 tab will be mounted and hence we'll miss validations), but instead mount them all and only hide the UI of all the non-selected ones with `display: none`. This will be the way all of them will be rendered soon (like in https://github.com/tech-matters/flex-plugins/pull/240), but in the meanwhile, this does the trick with this mixed approach to calculate the accurate length of tabs.

plugin-hrm-form/src/styles/HrmStyles.tsx
GPaoloni: Yes, I agree, it was just a copy paste of the CSS file I had, but making it work in styled components instead. I'll now split them so it will be easier to work with them!

plugin-hrm-form/src/___tests__/services/ContactService.test.js
GPaoloni: Should I address this in this task or are we delaying it for a next task?

plugin-hrm-form/src/components/tabbedForms/ContactlessTaskTab.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }, [dispatch, getValues, task.taskSid]);
+
+  const date = watch('contactlessTask.date');
+  const time = watch('contactlessTask.time');
+
+  React.useEffect(() => {
+    if (date && time) {
+      const [y, m, d] = splitDate(date);
+      const [mm, hh] = splitTime(time);
+      if (new Date(y, m - 1, d, mm, hh).getTime() > Date.now())
```
andvirga, line 44: Since we're using date-fns we can use `isFuture()` to check this condition, but we can change this in a separate PR (I don't want to delay this one)
[Doc: datefns - isFuture](https://date-fns.org/v2.16.1/docs/isFuture)

andvirga (approved):
Approved again! Just a small comment that can be addressed in another PR

plugin-hrm-form/src/components/tabbedForms/ContactlessTaskTab.tsx
GPaoloni: Didn't knew about this, but works great. Thanks!!
Screenshot from 2020-11-24 18-15-56

nickhurlburt:
One more request on this: can we put AM on top of PM?  Since it's earlier in the day, having it first makes more sense:
image

GPaoloni:
@nickhurlburt [from MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time)
> In Chrome/Opera the time control is simple, with slots to enter hours and minutes in 12 or 24-hour format depending on operating system local.

In my computer the input is looking like your request, and is AM while trying it, so maybe is related to that? I don't think this can be changed, but I can spend some time investigating if you want.
Screenshot from 2020-11-25 11-14-20

nickhurlburt:
> In my computer the input is looking like your request, and is AM while trying it, so maybe is related to that? I don't think this can be changed, but I can spend some time investigating if you want.

I see, it's the widget itself doing that.  And it's probably putting whatever's current first.  I don't think it's worth investigating unless we get complaints.  Thanks!

GPaoloni:
Deployed to staging.

PR #244, opened 2020-11-19
Connect to search cases backend

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-287

This PR connects to the search cases backend.

In order for this branch to work against staging, this [PR](https://github.com/tech-matters/hrm/pull/59) needs to be merged first.

WIP:
- Tests

### DIFF
--- a/plugin-hrm-form/src/components/search/SearchResults/index.tsx
+++ b/plugin-hrm-form/src/components/search/SearchResults/index.tsx
@@ -33,7 +33,7 @@ import { SearchPages, SearchPagesType } from '../../../states/search/types';
 import { namespace, searchContactsBase } from '../../../states';
 
 export const CONTACTS_PER_PAGE = 20;
-export const CASES_PER_PAGE = 5;
+export const CASES_PER_PAGE = 20;
 
 type OwnProps = {
   task: ITask;
--- a/plugin-hrm-form/src/components/search/SearchResults/index.tsx
+++ b/plugin-hrm-form/src/components/search/SearchResults/index.tsx
@@ -7,7 +7,7 @@ import { Template, Tab as TwilioTab, ITask } from '@twilio/flex-ui';
 
 import ContactPreview from '../ContactPreview';
 import CasePreview from '../CasePreview';
-import { SearchContactResult, SearchContact } from '../../../types/types';
+import { SearchContactResult, SearchCaseResult, SearchContact } from '../../../types/types';
 import { Row } from '../../../styles/HrmStyles';
 import {
   BackIcon,
@@ -33,16 +33,18 @@ import { SearchPages, SearchPagesType } from '../../../states/search/types';
 import { namespace, searchContactsBase } from '../../../states';
 
 export const CONTACTS_PER_PAGE = 20;
-const CASES_PER_PAGE = 5;
+export const CASES_PER_PAGE = 5;
 
 type OwnProps = {
   task: ITask;
   currentIsCaller: boolean;
-  results: SearchContactResult;
+  searchContactsResults: SearchContactResult;
+  searchCasesResults: SearchCaseResult;
   onlyDataContacts: boolean;
   closedCases: boolean;
   handleSelectSearchResult: (contact: SearchContact) => void;
-  handleSearch: (offset: number) => void;
+  handleSearchContacts: (offset: number) => void;
+  handleSearchCases: (offset: number) => void;
   toggleNonDataContacts: () => void;
   toggleClosedCases: () => void;
   handleBack: () => void;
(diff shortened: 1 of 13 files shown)


plugin-hrm-form/src/styles/search/index.tsx
```diff
@@ -5,8 +5,6 @@ import { Tabs, TabsProps } from '@twilio/flex-ui';
 import Folder from '@material-ui/icons/Folder';
 import ChevronLeft from '@material-ui/icons/ChevronLeft';
 import Link from '@material-ui/icons/Link';
-import CropFreeIcon from '@material-ui/icons/CropFree';
-import CallMergeIcon from '@material-ui/icons/CallMerge';
```
andvirga, line 9: Good catch! I forgot to delete these icons

plugin-hrm-form/src/states/search/types.ts
```diff
@@ -1,4 +1,4 @@
-import { SearchContact } from '../../types/types';
+import { SearchContact, Case } from '../../types/types';
```
andvirga, line 1: If the change isn't too big or complex, I would suggest to rename `SearchContact`to `Contact`

andvirga (approved):
The PR looks great, I've only left one small comment.

plugin-hrm-form/src/states/search/types.ts
murilovmachado: It's because search contacts does not return a plain `Contact`. But for search cases, it returns a plain `Case`.
I agree we can make search contacts return plain `Contact` in the future, but we can leave it for future improvements.

plugin-hrm-form/src/states/search/types.ts
```diff
@@ -33,7 +36,15 @@ export const SearchPages = {
 
 export type SearchPagesType = typeof SearchPages[keyof typeof SearchPages];
 
-export type DetailedSearchResult = ReturnType<typeof addDetails>;
+export type DetailedSearchContactsResult = {
+  count: number;
+  contacts: ReturnType<typeof addDetails>;
+};
+
+export type DetailedSearchCasesResult = {
```
GPaoloni, line 44: Related with above @andvirga's comment, won't this be a plain `CaseResult`? Just pointing to this to be consistent with the notation.

GPaoloni (approved):
Works and looks great! 

Just one comment, not really related to this PR, but raising the question:
Should we hide the "show more" button if the summary is not long enough (cases result preview)?
Screenshot from 2020-11-23 11-37-52

murilovmachado:
> Works and looks great!
> 
> Just one comment, not really related to this PR, but raising the question:
> Should we hide the "show more" button if the summary is not long enough (cases result preview)?
> Screenshot from 2020-11-23 11-37-52

@GPaoloni yeah, **show more** should not be shown if the whole text is already visible

plugin-hrm-form/src/states/search/types.ts
murilovmachado: Yeah, agreed.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -33,16 +33,18 @@ import { SearchPages, SearchPagesType } from '../../../states/search/types';
 import { namespace, searchContactsBase } from '../../../states';
 
 export const CONTACTS_PER_PAGE = 20;
-const CASES_PER_PAGE = 5;
+export const CASES_PER_PAGE = 5;
```
nickhurlburt, line 36: I would expect we would show 20 cases, just like we're showing 20 contacts.

nickhurlburt (commented):
I see a few appearance issues here:
- The 'Opened' and 'Updated' dates have a couple issues.  Once is that there should be a colon between the word and the date.  Another is that the font seems bigger than the mockups.  See https://marvelapp.com/prototype/6dc75bc/screen/74539628
- We've backed away from the extra functionality in the original mockups.  The '...' in the upper right should just be an icon that goes to the search detail page.  See again: https://marvelapp.com/prototype/6dc75bc/screen/74539628

I don't have a strong preference on whether this is addressed in this PR or a follow-on PR.

murilovmachado:
> I see a few appearance issues here:
> 
> * The 'Opened' and 'Updated' dates have a couple issues.  Once is that there should be a colon between the word and the date.  Another is that the font seems bigger than the mockups.  See https://marvelapp.com/prototype/6dc75bc/screen/74539628
> * We've backed away from the extra functionality in the original mockups.  The '...' in the upper right should just be an icon that goes to the search detail page.  See again: https://marvelapp.com/prototype/6dc75bc/screen/74539628
> 
> I don't have a strong preference on whether this is addressed in this PR or a follow-on PR.

@nickhurlburt I've created a Jira card to work on this and on another fix that Gian found on that component:
https://bugs.benetech.org/browse/CHI-332

plugin-hrm-form/src/components/search/SearchResults/index.tsx
andvirga: That was my fault, I had left that number for testing purposes and I didn't remember to revert that change.

PR #245, opened 2020-11-20
[313] -  Fixed timeline activities bug

andvirga:
The issue was located in one of the `useEffect()` hooks that were causing infinite re-renders.
I've merged the logic Into a single `useEffect()` hook. And calling `setTimeline()` in a single spot to avoid having multiple re-renders.

### DIFF
--- a/plugin-hrm-form/src/components/case/Timeline.jsx
+++ b/plugin-hrm-form/src/components/case/Timeline.jsx
@@ -30,29 +30,33 @@ const Timeline = ({ task, form, caseId, changeRoute, updateTempInfo }) => {
   const [timeline, setTimeline] = useState([]);
 
   useEffect(() => {
-    async function updateTimeline() {
+    /**
+     * Gets the activities timeline from current caseId
+     * If the case is just being created, adds the case's description as a new activity.
+     */
+    const getTimeline = async () => {
       const activities = await getActivities(caseId);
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/components/case/Timeline.jsx
```diff
@@ -100,19 +104,21 @@ const Timeline = ({ task, form, caseId, changeRoute, updateTempInfo }) => {
...
-            <TimelineIcon type={activity.type} />
-            <TimelineText>{activity.text}</TimelineText>
-            <Box marginLeft="auto" marginRight="10px">
-              <ViewButton onClick={() => handleOnClickView(activity)}>View</ViewButton>
-            </Box>
-          </TimelineRow>
-        );
-      })}
+      {timeline &&
+        timeline.length > 0 &&
```
GPaoloni, line 108: Given that the initial state is for `timeline` is `[]`(empty array) and it will always be updated with arrays, the first check seems unnecessary , but I might be wrong, and the `timeline.length > 0` makes no difference as if `timeline` is empty, `timeline.map` will be an empty array (renders nothing). Not saying that is wrong, but they might be extra lines of code that can be removed.

GPaoloni (approved):
Great! Is really weird that this bug is not happening to me (tried while in a call with @andvirga), but he shown to me that it happens to him, so thanks for fixing it!

PR #247, opened 2020-11-26
Changes spelling of counselor to counsellor

nickhurlburt:
Per https://bugs.benetech.org/browse/CHI-302: "It's currently using the US spelling and more of the helplines will use British English so we may as well change it now for the first few implementations."

This does not change any of our variables, just the strings that the user will see.  I have tried to catch everything but please point out anything I've missed.

I have also introduced localization in two places: the General Details section of the Contact Details, and the Search Form.  I think these seem like reasonable things to localize, as the strings will change per helpline but the fields are less likely to change.  I don't recommend localizing the other sections of the Contact Details, since we will be customizing many of those values.

A companion PR for webchat is here: https://github.com/tech-matters/webchat/pull/12

And we will also need to change two messages in the Autopilot configuration.

### DIFF
--- a/plugin-hrm-form/src/components/ContactDetails.tsx
+++ b/plugin-hrm-form/src/components/ContactDetails.tsx
@@ -96,11 +96,17 @@ const Details = ({
         expanded={detailsExpanded[GENERAL_DETAILS]}
         handleExpandClick={() => handleExpandDetailsSection(GENERAL_DETAILS)}
       >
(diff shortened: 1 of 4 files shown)


GPaoloni (approved):
Looks great! :rocket:

andvirga (approved):
Great job Nick!

murilovmachado (approved):

PR #248, opened 2020-11-26
[295] - Saving Households and Perpetrators right after clicking 'save'.

andvirga:
Description:
- Added API calls to save hh and perp right after clicking save.
- Added logic to save data and redirect the user to a blank form if "save and stay" option is selected.
- Modified unit tests.


murilovmachado (approved):
Looks good!

nickhurlburt (approved):
looks good!

PR #249, opened 2020-11-26
Display standalone search

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-343

Screen Shot 2020-11-26 at 19 15 17
Screen Shot 2020-11-26 at 19 15 30

### DIFF
--- a/plugin-hrm-form/src/translations/en-US/flexUI.json
+++ b/plugin-hrm-form/src/translations/en-US/flexUI.json
@@ -41,6 +41,8 @@
 
   "NonDataCallTypeDialog-CloseConfirm": "Are you sure?",
 
+  "SearchContactsAndCases-Title": "Search for Contacts and Cases",
+
   "SearchResultsIndex-Back": "Return to search criteria",
   "SearchResultsIndex-Result": " result",
   "SearchResultsIndex-Results": " results",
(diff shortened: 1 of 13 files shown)


plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -49,6 +51,8 @@
   "SearchResultsIndex-ViewContacts": "View Contacts",
   "SearchResultsIndex-ViewCases": "View Cases",
 
+  "StandaloneSearch-SideNav": "Standalone Search",
```
nickhurlburt, line 54: For anything that's user-facing, let's just say "Search".

plugin-hrm-form/src/utils/setUpComponents.js
```diff
@@ -275,6 +277,21 @@ export const setUpCaseList = () => {
   );
 };
 
+export const setUpStandaloneSearch = () => {
+  Flex.ViewCollection.Content.add(
+    <Flex.View name="standalone-search" key="standalone-search-view">
```
nickhurlburt, line 282: I have a slight preference to make the name just 'search', so the URL is just 'search'.

nickhurlburt (approved):
Looks good!  A couple comments on user-facing naming.  Also, a screen reader just says 'button' when you focus on the side nav button.  However... I think I'm happy to wait for Twilio to provide better hooks for this to work properly, since they need to do it for the core `SideLink` components anyway.

PR #250, opened 2020-11-27
Fix bug when sending Welcome Message (webchat)

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-340

This PR:
- fixes a bug when sending the welcome message: if the channel is not ready, wait 200ms and retry, repeat until channel is ready and the message succeeds or 2000ms (2 seconds) elapsed, reporting the following error to Rollbar:
Screenshot from 2020-11-30 10-32-08
- fixes bug on contactless task form validation: currently, if `date` + `time` are future, UI reflects an error, but you can still submit the form because `onSubmit`, validation triggers but there is no validation that fails at this point (the UI error is set from an `useEffect`, and cleared when validation triggers). This PR fixes this, introducing a new (invisible and unused) field `isFutureAux` in the form with a validation that fails if `date` + `time` are future, and if `isFutureAux` have an error, it is then replicated to `time` input (again, from an `useEffect`, but that runs **after** the validation was triggered, making the form invalid when trying to submit it).

### DIFF
--- a/plugin-hrm-form/src/utils/setUpActions.js
+++ b/plugin-hrm-form/src/utils/setUpActions.js
@@ -158,7 +158,6 @@ export const customTransferTask = setupObject => async (payload, original) => {
   if (mode === transferModes.warm) {
     await TransferHelpers.clearTransferMeta(payload.task);
     window.alert(Manager.getInstance().strings['Transfer-ChatWarmNotAllowed']);
-    return Promise.resolve();
   }
 
   const memberToKick = mode === transferModes.cold ? TransferHelpers.getMemberToKick(payload.task, identity) : '';
--- a/plugin-hrm-form/src/utils/setUpActions.js
+++ b/plugin-hrm-form/src/utils/setUpActions.js
@@ -1,5 +1,5 @@
 // eslint-disable-next-line no-unused-vars
-import { Manager, TaskHelper, Actions as FlexActions, ActionFunction, ReplacedActionFunction } from '@twilio/flex-ui';
+import { Manager, TaskHelper, Actions as FlexActions, StateHelper } from '@twilio/flex-ui';
 
 // eslint-disable-next-line no-unused-vars
 import { DEFAULT_TRANSFER_MODE, getConfig } from '../HrmFormPlugin';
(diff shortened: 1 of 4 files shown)


plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -107,8 +107,17 @@ export const afterAcceptTask = setupObject => async payload => {
 
   // To enable for all chat based task, change condition to "if (TaskHelper.isChatBasedTask(task))"
   if (task.attributes.channelType === channelTypes.web) {
-    // Ignore event payload as we already have everything we want in afterAcceptTask arguments
-    manager.chatClient.once('channelJoined', () => setTimeout(() => sendWelcomeMessage(setupObject)(payload), 0));
+    const trySendWelcomeMessage = ms => {
+      setTimeout(() => {
+        const channelState = StateHelper.getChatChannelStateForTask(task);
+        // if channel is not ready, wait 200ms and retry
+        if (channelState.isLoadingChannel) trySendWelcomeMessage(200);
```
nickhurlburt, line 114: I'd recommend having a finite number of retries to prevent against infinite loops.  You can do this with an additional parameter, for example:
```
const trySendWelcomeMessage = (ms, retries) => {
      setTimeout(() => {
        const channelState = StateHelper.getChatChannelStateForTask(task);
        // if channel is not ready, wait 200ms and retry
        if (channelState.isLoadingChannel) {
            if (retries < 3) trySendWelcomeMessage(200, retries++);  // 3-5 retries is probably right
            else console.error('failed to send message'); // and add more context and make sure this is recorded in rollbar
        }
        else sendWelcomeMessage(setupObject)(payload);
      }, ms);
    };

manager.chatClient.once('channelJoined', () => trySendWelcomeMessage(0, 0));

```

Though certainly you can re-format that however you want.  The important thing is to block the risk of infinite loops.

nickhurlburt (approved):
Looks good, with one suggestion on the retry behavior.  And I assume this has been tested to work on users that are consistently failing?

GPaoloni:
Deployed to staging.

PR #253, opened 2020-12-01
Add support for custom logo

nickhurlburt:
Allows a logo URL to be set using Twilio configuration.  I've added that config to staging, and we now have the Aselo logo as shown below.

I am likely to cherry-pick this commit onto master and then deploy prod before Thursday so we can use it during a demo.

thumbnail_image001


murilovmachado (approved):
Looks good to me!

A next step would be to show this logo up above on the browser's tab and change tab's title to **Aselo** instead of **Twilio Flex**.

image

andvirga (approved):
Approved!

GPaoloni (approved):
:rocket: 

>I am likely to cherry-pick this commit onto master and then deploy prod before Thursday so we can use it during a demo.

I wonder, is there way to cherry pick and then avoid future merge conflicts when we merge staging into master? Maybe doing merge instead of squash for this one?

PR #255, opened 2020-12-02
[352] - Showing Orphaned Cases in UI

andvirga:
Important note:
- A case is considered "orphaned" when `connectedCases = []`
- Since we're sending to the API `closedCases` as parameter I think that the API should use that parameter to also return the list of orphaned cases.


nickhurlburt (approved):
Looks good!  Though we should look again once the backend changes are in.

PR #256, opened 2020-12-02
Forms Generators

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-317

This PR:
- Improves styles for the forms, making them look more like what we have now.
- Reorganizes styles and some UI generation helpers

NOTE: in this PR, a second "settings button" is included in the left nav. That button opens a screen full with a bunch of inputs in order to make testing easier. The code related to this has been removed, in order to try it out, checkout to the commit `9ba7363`.

______________

The idea of this PR is to finish the required work on the form generators. The main code is contained in `src/components/forms/formGenerators.tsx` (introduced in a previous PR but reworked here).
The key entry point for using this is the `createFormFromDefinition` function:
```
export const createFormFromDefinition = (definition: FormDefinition) => (parents: string[]) => (
  updateCallback: () => void,
): JSX.Element[] => definition.map(getInputType(parents, updateCallback));

```
The basic idea of this generators is that if you provide a form definition (`definition: FormDefinition` following the schema described at the bottom of `src/components/forms/types.ts`), the `createFormFromDefinition` function will return an array of html inputs, each one connected to the React-Hook-Form wrapper (for this to work, the generated array must be child of a `FormProvider`, [examples here](https://react-hook-form.com/api)). The inputs are connected to the provider via the `ConnectForm` high order component, which will provide all of the form methods (refer to [rhf docs](https://react-hook-form.com/api) for details). 
The second parameter is an array of string called `parents`, that specifies the levels of nesting that this will have inside of the broader form (many "arrays" of inputs can be connected to the same form, each of them representing a different property of the entire form, like `childInformation`, `callerInformation`, etc). Each component will receive the `parents` array, and a `name` describing itself. The resulting string from concatenating the `parents` and `name` will be used to refer to this particular input, following the "dot joined conventions" that [rhf follows](https://react-hook-form.com/api#register). The definition of each input contains, besides its name, a lot more of important information, like the label (localized template string), the type of the input, different options depending on the type and validation rules.
The `createFormFromDefinition` receives a 3rd parameter `updateCallback: () => void`, that is the function that will be called `onChange` / `onBlur` (depending on the input type), and can be used to trigger local or global state updates, api calls, etc. This function is now being called in specific events, but can be better abstracted to be an object of event handlers, in order to provide better granularity and more reusability (future work).
The responsibility on how to display the inputs is delegated to the component rendering the form. The styles for each input is coupled in this version, but can be abstracted as an extra parameter (future work).

For the rest of the functions in the `formGenerators.tsx` file:
- `getInputType` creates an individual input from a given `FormItemDefinition`, using the `parents` and `updateCallback` mentioned above. The definition for each input contains all the required information like name, type, validation, etc.
- `getInitialValue` is a helper that returns the "initial state" for each input type, used for example, when first creating the "empty contact state" for a freshly accepted task.
- `createStateItem` uses above helper and combined with `Array.reduce` make easy to create the "empty state" for all the components described in a `FormDefinition`.
- rest of the functions are just UI helpers for the tabbed forms particular use case, barely abstracted.

The rest of the files changing in this PR are for styling.

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -1,5 +1,5 @@
 import * as Flex from '@twilio/flex-ui';
-import { FlexPlugin } from 'flex-plugin';
+import { FlexPlugin, loadCSS } from 'flex-plugin';
 import SyncClient from 'twilio-sync';
 
 import './styles/GlobalOverrides';
@@ -205,8 +205,10 @@ export default class HrmFormPlugin extends FlexPlugin {
    * @param manager { import('@twilio/flex-ui').Manager }
    */
   init(flex, manager) {
+    loadCSS('https://use.fontawesome.com/releases/v5.15.1/css/solid.css');
+
     const monitoringEnv = manager.serviceConfiguration.attributes.monitoringEnv || 'staging';
-    if (!process.env.NO_MONITORING) setUpMonitoring(this, manager.workerClient, monitoringEnv);
+    if (process.env.NODE_ENV !== 'development') setUpMonitoring(this, manager.workerClient, monitoringEnv);
 
     console.log(`Welcome to ${PLUGIN_NAME} Version ${PLUGIN_VERSION}`);
     this.registerReducers(manager);
--- a/plugin-hrm-form/src/components/common/forms/formGenerators.tsx
+++ b/plugin-hrm-form/src/components/common/forms/formGenerators.tsx
@@ -14,9 +14,15 @@ import {
   FormError,
   Row,
   FormInput,
+  FormDateInput,
+  FormTimeInput,
+  FormCheckBoxWrapper,
+  FormCheckbox,
+  FormMixedCheckbox,
   FormSelect,
   FormOption,
   FormSelectWrapper,
+  FormTextArea,
 } from '../../../styles/HrmStyles';
 import type { FormItemDefinition, FormDefinition, SelectOption, MixedOrBool } from './types';
 
@@ -38,7 +44,7 @@ const getRules = (field: FormItemDefinition): ValidationRules =>
   pick(field, ['max', 'maxLength', 'min', 'minLength', 'pattern', 'required', 'validate']);
 
 /**
- * Creates a Form with each input conntected to RHF's wrapping Context, based on the definition.
+ * Creates a Form with each input connected to RHF's wrapping Context, based on the definition.
  * @param {string[]} parents Array of parents. Allows you to easily create nested form fields. https://react-hook-form.com/api#register.
  * @param {() => void} updateCallback Callback called to update form state. When is the callback called is specified in the input type.
  * @param {FormItemDefinition} def Definition for a single input.
@@ -61,13 +67,14 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
                     {rules.required && <RequiredAsterisk />}
                   </Box>
                 </Row>
-                <input
+                <FormInput
                   id={path}
                   name={path}
+                  error={Boolean(error)}
                   aria-invalid={Boolean(error)}
                   aria-describedby={`${path}-error`}
                   onBlur={updateCallback}
-                  ref={register(rules)}
+                  innerRef={register(rules)}
                 />
                 {error && (
                   <FormError>
@@ -92,13 +99,14 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
                     {rules.required && <RequiredAsterisk />}
                   </Box>
                 </Row>
-                <input
+                <FormInput
                   id={path}
                   name={path}
+                  error={Boolean(error)}
                   aria-invalid={Boolean(error)}
                   aria-describedby={`${path}-error`}
                   onBlur={updateCallback}
-                  ref={register({
+                  innerRef={register({
                     ...rules,
                     pattern: { value: /^[0-9]+$/g, message: 'This field only accepts numeric input.' },
                   })}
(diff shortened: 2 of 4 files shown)


plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -393,6 +397,16 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
   }
 };
 
+/**
+ * Creates a Form with each input conntected to RHF's wrapping Context, based on the definition.
```
nickhurlburt, line 401: s/conntected/connected

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -418,26 +432,30 @@ export const getInitialValue = (def: FormItemDefinition) => {
...
+export const splitInHalf = (formItems: JSX.Element[]) => {
   const m = Math.ceil(formItems.length / 2);
 
-  const [l, r] = [items.slice(0, m), items.slice(m)];
+  const [l, r] = [formItems.slice(0, m), formItems.slice(m)];
+
+  return [l, r];
+};
+
+export const buildTwoColumnFormLayout = (formItems: JSX.Element[]) => {
```
nickhurlburt, line 455: Without having looked through all the code, this looks like it's automatically splitting the first and second halves of the `formDefinition` into the two columns.  There may be times when we need a manual split, for example, if we have several small inputs and then a big `textarea`.  If that's still possible with the code as-is, then no problem.

nickhurlburt (approved):
I haven't looked in great detail, but I think I understand the high level, and looks good to me!

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -205,8 +205,10 @@ export default class HrmFormPlugin extends FlexPlugin {
    * @param manager { import('@twilio/flex-ui').Manager }
    */
   init(flex, manager) {
+    loadCSS('https://use.fontawesome.com/releases/v5.15.1/css/solid.css');
```
andvirga, line 208: Can we move this to an environment variable? To easily update or change the CDN if needed.

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -393,6 +397,16 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
...
+ * @param {string[]} parents Array of parents. Allows you to easily create nested form fields. https://react-hook-form.com/api#register.
+ * @param {() => void} updateCallback Callback called to update form state. When is the callback called is specified in the input type (getInputType).
+ */
+export const createFormFromDefinition = (definition: FormDefinition) => (parents: string[]) => (
+  updateCallback: () => void,
+): JSX.Element[] => definition.map(getInputType(parents, updateCallback));
+
 /**
  * Utility functions to create initial state from definition
  * @param {FormItemDefinition} def Definition for a single input of a Form
```
andvirga, line 412: The area that I wanted to comment is out of the PR (lines 415 - 432)
``` javascript
switch (def.type) {
    case 'input':
    case 'numeric-input':
    case 'textarea':
    case 'date-input':
    case 'time-input':
      return '';
    case 'select':
      return def.options[0].value;
    case 'dependent-select':
      return def.defaultOption.value;
    case 'checkbox':
      return false;
    case 'mixed-checkbox':
      return def.initialChecked === undefined ? 'mixed' : def.initialChecked;
    default:
      return null;
  }
```

Can we create an object that holds all the different types of Inputs? It should look like this:

```javascript
const InputTypes = Object.freeze({
  INPUT: 'input',
  NUMERIC_INPUT: 'numeric-input',
  TEXTAREA: 'textarea',
  DATE_INPUT: 'date-input',
  TIME_INPUT: 'time-input',
  SELECT: 'select',
})
```

plugin-hrm-form/src/HrmFormPlugin.js
GPaoloni: Makes sense. Where do think we should place it? W don't have a .env in this repo, but maybe Twilio account setting?

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
GPaoloni: Actually that information is already in the type. We can do that if you consider it to be useful of course, but I don't see the extra benefit over TS.
Screenshot from 2020-12-14 17-55-12

andvirga (changes requested):
This PR looks great! I've tested your branch and everything seems to work fine! 
I've only left two small comments, it would be good if you can address them.

plugin-hrm-form/src/HrmFormPlugin.js
andvirga: I would say to choose the easiest / simpler solution. I'm not very familiarized with the Twilio account setting but if that option is simple let's go with that.

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
andvirga: I believe that having this object will be very useful, for example, if you want to define a new input field in the JSON you can do something like this, and you can easily know which are the available types for the field you want to create.
Of course, let's do this only if this change doesn't add too much complexity.

```javascript  
{
    name: 'test',
    type: InputTypes.DATE,
    label: 'Test',
  },
```

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -215,21 +223,19 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
...
+                    id={path}
+                    name={path}
+                    type="checkbox"
+                    aria-invalid={Boolean(error)}
+                    aria-describedby={`${path}-error`}
+                    onChange={updateCallback}
+                    innerRef={register(rules)}
+                  />
+                  <Template code={`${def.label}`} />
+                  {rules.required && <RequiredAsterisk />}
```
murilovmachado, line 237: I'm curious how these inputs work fine when I look at this line:
`{rules.required && <RequiredAsterisk />}`.

Currently, `RequiredAsterisk` needs a `field` param:
```
const RequiredAsterisk = ({ field }) => {
  const isRequired = field.validation && field.validation.includes(ValidationType.REQUIRED);
  return isRequired && <span style={{ color: 'red' }}>*</span>;
};
```
Calling `<RequiredAsterisk />` (without `field` param) should give an error like: 
`Tried calling validation of undefined`.

But it looks like the contactless tab is using them without throwing an error.

murilovmachado: **Update:** Oh, this is not using our old `RequiredAsterisk`, but a custom one defined in this same file. I got it.

murilovmachado (approved):
Looks good!

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
andvirga: **Update:** as we discussed, TS already helps you to identify input types, we don't need an additional frozen object.

plugin-hrm-form/src/HrmFormPlugin.js
GPaoloni: What do you think @murilovmachado ? Should we put this in config or leave it hardcoded? I've been thinking and I don't see this changing very often.

nickhurlburt: This is really up to you all, but if you decide to keep it hard coded I don't see that as a problem.

PR #257, opened 2020-12-02
Added categoriesTabGenerator

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-318

Added categoriesTabGenerator file, which includes helpers to generate the categories tab (tabbed forms).

WIP:
- styles (will be addressed with tabbed form customization).
- tests.


andvirga (approved):
Amazing work!

murilovmachado (approved):
Looks good, but correct me if I'm wrong: this `categoriesTabGenerator` is not being used yet, so we cannot really see it working when running the branch, is that right?

PR #258, opened 2020-12-03
Define Redux structure to persist customized form data

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-322

This PR adds the Redux code for holding (and persisting between tasks) the form information.


plugin-hrm-form/src/states/contacts/types.ts
```diff
@@ -0,0 +1,74 @@
...
+
+// Action types
+export const UPDATE_FORM = 'UPDATE_FORM';
+export const SAVE_END_MILLIS = 'SAVE_END_MILLIS';
+export const SET_CATEGORIES_GRID_VIEW = 'SET_CATEGORIES_GRID_VIEW';
+export const HANDLE_EXPAND_CATEGORY = 'HANDLE_EXPAND_CATEGORY';
+export const HANDLE_SELECT_SEARCH_RESULT = 'HANDLE_SELECT_SEARCH_RESULT';
+export const PREPOPULATE_FORM_CHILD = 'PREPOPULATE_FORM_CHILD';
+export const PREPOPULATE_FORM_CALLER = 'PREPOPULATE_FORM_CALLER';
+export const RESTORE_ENTIRE_FORM = 'RESTORE_ENTIRE_FORM';
```
andvirga, line 12: Would be great to have all these values in a frozen object:

```javascript
export const ActionTypes = Object.freeze({
  UPDATE_FORM = 'UPDATE_FORM',
  SAVE_END_MILLIS = 'SAVE_END_MILLIS',
  etc...
})
```

plugin-hrm-form/src/states/contacts/types.ts
```diff
@@ -0,0 +1,74 @@
...
+
+// INCOMPLETE
+
+type UpdateFormAction = {
+  type: typeof UPDATE_FORM;
```
andvirga, line 17: Just a question (since I'm not super expert with TS), why are you using `typeof` here? Since the result is always `string` right?.

plugin-hrm-form/src/types/types.ts
```diff
@@ -27,45 +27,17 @@ export type Case = {
...
+  definitionVersion?: number;
+  callType: CallTypes;
+  childInformation: { [key: string]: string | boolean };
+  callerInformation: { [key: string]: string | boolean };
+  caseInformation: { [key: string]: string | boolean } & { categories: {} };
+  contactlessTask: { [key: string]: string | boolean };
+  metadata: {
+    startMillis: number;
+    endMillis: number;
+    recreated: boolean;
```
andvirga, line 40: These line reductions are great!

andvirga (commented):
Looks great! I've only left one comment, however, I'm not able to run this on my local environment, it's giving me the following error (see screenshot), maybe I'm missing some additional step to run this branch alongside other Customization changes.

Captura de pantalla de 2020-12-15 12-29-21

murilovmachado:
@GPaoloni, I'm getting the same error as @andvirga when running this branch.

GPaoloni:
@murilovmachado @andvirga sorry guys, forgot tom mention that this PR is for review, but the code added here is not yet being used in the rest of the code, resulting on this errors. Next PR is the one where you can test this!

murilovmachado (approved):
Good to see our contact state following the same pattern as the other states (actions/reducer/helper/types).

plugin-hrm-form/src/states/contacts/types.ts
GPaoloni: TL;DR you can't dispatch invalid action strings (action creators won't type).

Image that will answer better than I could: 
Screenshot from 2020-12-15 15-19-11
The `b2` constant errors with the following message:
`Type '"another"' is not assignable to type '"UPDATE_FORM"'.ts(2322)`

Basically, this is restricting the type, you can't assign any string, just that one.
In this context, the `ContactsActionType` is the union of all individual actions
Screenshot from 2020-12-15 15-27-19

and an action creator will type only if the `action.type` is one of the available ones, so if an action creator typed like
`export const updateCallType = (taskId: string, callType: CallTypes | ''): t.ContactsActionType => ({ ...})`
compiles, then you creating a valid action object, with the corresponding valid payload.
Idea from the [Redux docs with TS](https://redux.js.org/recipes/usage-with-typescript#type-checking-actions--action-creators).

PR #259, opened 2020-12-04
Fix Unknown Child Name

murilovmachado:
Fixes the bug the Joan found where Child Name was displayed **Unknown**.

Screen Shot 2020-12-04 at 11 09 08

### DIFF
--- a/plugin-hrm-form/src/components/case/Case.tsx
+++ b/plugin-hrm-form/src/components/case/Case.tsx
@@ -142,7 +142,7 @@ const Case: React.FC<Props> = props => {
 
   const isMockedMessageOpen = Boolean(mockedMessage);
 
-  const firstConnectedContact = connectedCase.connectedContacts && connectedCase.connectedContacts[0];
+  const firstConnectedContact = connectedCase && connectedCase.connectedContacts && connectedCase.connectedContacts[0];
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -131,8 +141,9 @@ const Case: React.FC<Props> = props => {
     );
 
   const isMockedMessageOpen = Boolean(mockedMessage);
-  const { firstName, lastName } = form.childInformation.name;
-  const name = formatName(`${firstName.value} ${lastName.value}`);
+
+  const firstConnectedContact = connectedCase.connectedContacts[0];
```
andvirga, line 145: I'm thinking in the orphaned scenario, I believe in that scenario connectedContacts should be an empty array but just in case it would be null we can do this:
`connectedCase.connectedContacts &&  connectedCase.connectedContacts[0]`

andvirga: Or even better:
`connectedCase && connectedCase.connectedContacts && connectedCase.connectedContacts[0]`

andvirga (approved):

andvirga:
Looks good! Just make sure to test all scenarios

PR #262, opened 2020-12-07
[Case View revamp] - Header - Change title

andvirga:
- Added new structure for Case View Header
- Added childName, CaseId, and space for ZA Office.
- Styling

Note: This PR won't be merged to staging until 12/9 PROD release is completed.


plugin-hrm-form/src/styles/case/index.tsx
```diff
@@ -218,3 +211,39 @@ export const NoteContainer = styled('div')`
...
+
+DetailsHeaderCaseId.displayName = 'DetailsHeaderCaseId';
+
+export const DetailsHeaderOfficeName = styled(Typography)`
+  padding-left: 10px;
+  font-size: 0.7rem !important;
+  font-weight: 300 !important;
+`;
+
+DetailsHeaderOfficeName.displayName = 'DetailsHeaderOfficeName';
```
murilovmachado, line 249: Can you add all these styled-components into `mockStyled.js` as well? (including the ones that previously exist)
I ask this because sometimes we have rendering issues while running the tests, that's hard to spot the problem, and it's usually just missing a mock version of these styled-components on `mockStyled.js`.

It would be something like this:
```
jest.mock('../styles/case', () => ({
  DetailsHeaderContainer: 'DetailsHeaderContainer',
  DetailsHeaderChildName: 'DetailsHeaderChildName',
  DetailsHeaderCaseContainer: 'DetailsHeaderCaseContainer',
  DetailsHeaderCaseId: 'DetailsHeaderCaseId',
  DetailsHeaderOfficeName: 'DetailsHeaderOfficeName',
 // and all other styled-compents from this file
}));
```

murilovmachado (approved):
Looks good.

Also, according to the mockups I believe the Child Name on the header should have a higher **font-weight** (maybe 600):
Screen Shot 2020-12-10 at 12 29 37

plugin-hrm-form/src/styles/case/index.tsx
andvirga: Good catch! I'll do that!

andvirga: ChildName already has `font-weight: 600px`, I'll leave that value for now, but I can easily change it if required.

PR #263, opened 2020-12-07
Case view bug fixes

GPaoloni:
This PR is related to https://github.com/tech-matters/hrm/pull/64

Fixed the following bugs for contactless/offline tasks: 
1. The icon isn’t right in the timeline for standalone-entered tasks
2. The contact summary doesn’t appear in timeline
3. When clicking on View it says “Not yet implemented”

### DIFF
--- a/plugin-hrm-form/src/components/case/Timeline.jsx
+++ b/plugin-hrm-form/src/components/case/Timeline.jsx
@@ -17,12 +17,10 @@ import * as CaseActions from '../../states/case/actions';
 import * as RoutingActions from '../../states/routing/actions';
 import { ContactDetailsSections } from '../common/ContactDetails';
 import { getConfig } from '../../HrmFormPlugin';
-import { channelTypes } from '../../states/DomainConstants';
+import { channelsAndDefault } from '../../states/DomainConstants';
 import { namespace, routingBase } from '../../states';
 
-const channelsArray = Object.values(channelTypes);
-
-const isConnectedCaseActivity = activity => channelsArray.includes(activity.type);
+const isConnectedCaseActivity = activity => Boolean(channelsAndDefault[activity.type]);
 
 const sortActivities = activities => activities.sort((a, b) => b.date.localeCompare(a.date));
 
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/components/case/TimelineIcon.jsx
```diff
@@ -21,16 +22,18 @@ const getIcon = type => {
     case channelTypes.voice:
       return <DefaultIcon defaultTaskChannel={Flex.DefaultTaskChannels.Call} />;
     case 'note':
-    default:
       return <NoteIcon style={{ opacity: 0.62, fontSize: '20px' }} />;
+    // defaulting to otherContactChannels.includes(icon). Maybe at some point we need to address this in a different way.
+    default:
```
nickhurlburt, line 27: Having a different default is a good idea here

plugin-hrm-form/src/components/case/Timeline.jsx
```diff
@@ -112,7 +113,7 @@ const Timeline = ({ task, form, caseObj, changeRoute, updateTempInfo, route }) =
           return (
             <TimelineRow key={index}>
               <TimelineDate>{date}</TimelineDate>
-              <TimelineIcon type={activity.type} />
+              <TimelineIcon icon={activity.icon} />
```
nickhurlburt, line 116: I see, here we do use the `icon` from the backend.  But why not reuse the same idea further up in this file, `const icon = task.channelType === 'default' ? form.contactlessTask.channel : task.channelType;` but `activity.type` instead of `task.channelType`?  It makes so much more sense to me to decide what icon to use in the frontend.

nickhurlburt (changes requested):
I added a chat offline contact and created a new case from it and that worked fine, but when I searched for it I got this:
image

(I had also added a note.)  This is just using the HRM on staging and not the HRM with the recent changes, so that may be the reason why it's failing.

plugin-hrm-form/src/components/case/Timeline.jsx
```diff
@@ -42,12 +40,15 @@ const Timeline = ({ task, form, caseObj, changeRoute, updateTempInfo, route }) =
       const isCreatingCase = !timelineActivities.some(isConnectedCaseActivity);
 
       if (isCreatingCase) {
+        const icon = task.channelType === 'default' ? form.contactlessTask.channel : task.channelType;
```
andvirga, line 43: Are we 100% sure that when `task.channelType = 'default'` --> `form.contactlessTask` has a value and it's not `undefined`?

andvirga (approved):
Looks good! I've only one concern that I've left in a comment.

plugin-hrm-form/src/components/case/Timeline.jsx
GPaoloni: Yes, it's required in order to submit the form!

andvirga (approved):
Approved!

nickhurlburt (approved):
Looks good to me

murilovmachado (approved):

GPaoloni:
Deployed to staging.

PR #264, opened 2020-12-08
Do not let warm text transfers go through

nickhurlburt:
@GPaoloni please read [CHI-380](https://bugs.benetech.org/browse/CHI-380) for background.

This PR fixes CHI-380.  After warning that we can't do a warm text transfer, we don't `return`, so we continue to try to do the transfer.  This PR adds a `return` and also moves that check up earlier in the function so that we don't have to start and then roll back all of the transfer metadata preparation.

The grand mystery is that this bug is **not** happening on production, and I don't understand how that could be happening.

### DIFF
--- a/plugin-hrm-form/src/utils/setUpActions.js
+++ b/plugin-hrm-form/src/utils/setUpActions.js
@@ -144,24 +144,26 @@ const safeTransfer = async (transferFunction, task) => {
  * @returns {import('@twilio/flex-ui').ReplacedActionFunction}
  */
 export const customTransferTask = setupObject => async (payload, original) => {
+  const mode = payload.options.mode || DEFAULT_TRANSFER_MODE;
+
+  // Currently (as of 2 Dec 2020) warm text transfers are not supported.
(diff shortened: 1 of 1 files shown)


nickhurlburt:
@GPaoloni it looks like the bug was introduced [here](https://github.com/tech-matters/flex-plugins/pull/250/files#diff-5ad47b0fe01e356ae7f76caf27619a8f4355ac81796f2f0f3cf89c1b25d0be72L152) where a `Promise.resolve()` was removed.  I'm not sure if the right approach to fixing this bug is to re-introduce the `Promise.resolve()` or with the changes in this PR.  What do you think?

GPaoloni (approved):
Yes it looks like I've introduced this bug when deleting that `return Promise.resolve()`.. I assume any return value will be enough, the idea was to break the execution flow. Don't know why tho, so sorry about that, and glad you spotted it.
I'd keep your changes, as you are stopping the execution flow before, so it's less time wasted updating and clearing the task attributes. Thanks!!

PR #270, opened 2020-12-10
[Case View revamp] - Header - Opened and Last Updated/Closed fields

andvirga:
- Added Opened and LastUpdated / Closed date fields
- Added a simple logic to not display LastUpdated Date when matches with Opened Date


murilovmachado:
Functionality is working fine.

About styling, there're 2 minor adjustments:
1. Both `Opened` and `Last Updated` need a `padding-left` (the values are stick to the left of the input)
2. `Date Opened` label should be only `Opened` according to this [mockup](https://marvelapp.com/prototype/6dc75bc/screen/74539629) (not sure if other mockups are different)

Branch:
Screen Shot 2020-12-11 at 10 54 06

Mockup:
Screen Shot 2020-12-11 at 10 54 53

plugin-hrm-form/src/components/case/CaseDetails.jsx
```diff
@@ -26,29 +34,52 @@ const renderCaseStatus = status => {
   }
 };
 
-const CaseDetails = ({ caseId, name, counselor, date, status }) => {
+const CaseDetails = ({ caseId, name, counselor, openedDate, lastUpdatedDate, followUpDate, status }) => {
+  const lastUpdatedClosedDate = openedDate === lastUpdatedDate ? ' ̶ ' : lastUpdatedDate;
```
murilovmachado, line 38: there's a trailing space on `' ̶ '`

plugin-hrm-form/src/styles/case/index.tsx
```diff
@@ -247,3 +240,15 @@ export const DetailsHeaderOfficeName = styled(Typography)`
...
+  width: 150px;
+  height: 36px;
+  color: #000000;
+  background-color: ${props => props.theme.colors.inputBackgroundColor};
+  font-weight: 600;
+  border: none;
+  margin-top: 7px;
+`;
+
+StyledInputField.displayName = 'StyledDateField';
```
murilovmachado, line 254: Here the `displayName` is different than the component name.
You can change the component's name to be `StyledDateField` or change the `displayName`  to be `StyledInputField `.

murilovmachado (approved):

plugin-hrm-form/src/components/case/CaseDetails.jsx
andvirga: Good catch, It was a weird issue with VSCode but finally I could remove the spaces around the hyphen

PR #273, opened 2020-12-14
Generate TabbedForms from a .JSON definition

GPaoloni:
Jira:
+ https://bugs.benetech.org/browse/CHI-321
+ https://bugs.benetech.org/browse/CHI-331
+ https://bugs.benetech.org/browse/CHI-334

This PR:
- bundles JSON definitions for tabbed forms (just to generate forms, will update later on).
- generates tabbes forms from above definitions (caller, child, categories and summary tabs).
- includes functionality for populate/copy search result/transfer.
- small updates in the reducers and actions.
- add styles for categories tab.
- adapts Calltypes & Noncalltypes buttons, Bottom Bar, TaskView components.
- adapts insights service.

Note: big PR as introduces JSONs, and more TS. Easier to review commit by commit.


plugin-hrm-form/src/components/TaskView.jsx
```diff
@@ -43,12 +44,12 @@ class TaskView extends Component {
     const { contactFormStateExists, routingStateExists, searchStateExists } = this.props;
     if (!contactFormStateExists || !routingStateExists || !searchStateExists) {
       // (Gian) maybe this can be used to recreate the form too?
```
andvirga, line 46: I think this question is no longer necessary right?

plugin-hrm-form/src/components/callTypeButtons/CallTypeButtons.jsx
```diff
@@ -102,12 +113,19 @@ const CallTypeButtons = props => {
...
   localization: localizationType.isRequired,
-  changeRoute: PropTypes.func.isRequired,
   handleCompleteTask: PropTypes.func.isRequired,
+  dispatch: PropTypes.func.isRequired,
+};
+
+const mapStateToProps = (state, ownProps) => {
+  const contactForm = state[namespace][contactFormsBase].tasks[ownProps.task.taskSid];
+  return { contactForm };
 };
```
andvirga, line 127: It's not big deal, but why you haven't used `mapDispatchToProps?`

plugin-hrm-form/src/components/tabbedForms/CallerInformationTab.tsx
```diff
@@ -1,16 +1,55 @@
...
+      const { callerInformation } = getValues();
+      dispatch(updateForm(task.taskSid, 'callerInformation', callerInformation));
+    };
+
+    // TODO: fix this typecasting
+    const callerFormDefinition = createFormFromDefinition(CallerTabDefinition as FormDefinition)(['callerInformation'])(
+      updateCallback,
+    );
+
+    return splitInHalf(disperseInputs(7)(callerFormDefinition));
```
andvirga, line 33: Can we move that value to a constant?
You would need to change that in CaseInformationTab and ChildInformationTab.
It's a small change but my OCD didn't let me pass this.
```javascript
 const marginValue = 7;
 return splitInHalf(disperseInputs(marginValue)(callerFormDefinition));
```

andvirga (approved):
You did a great job! I've left only some small comments. Thanks for converting many .jsx files to .tsx. The code looks very solid!

murilovmachado:
@GPaoloni There're some issues on the categories checkboxes. Not sure if they will be addressed in this PR, but it's worth pointing them:

In the branch they look like this:
Screen Shot 2020-12-16 at 14 18 28

But it should be like this (staging):
Screen Shot 2020-12-16 at 14 18 47

Also, when toggling to view side by side, in the branch they don't change:
Screen Shot 2020-12-16 at 14 19 07
It's probably happening because they are taking the whole line, you can even click on them when clicking outside of the check box.

But it should be like this (staging):
Screen Shot 2020-12-16 at 14 18 59

murilovmachado:
@GPaoloni Another issue that I'll put here (but also, I'm not sure if we plan to fix in this PR) is about the bottom bar not being stick to the bottom.

In this branch:
Screen Shot 2020-12-16 at 14 32 49

The way it should be (staging):
Screen Shot 2020-12-16 at 14 33 10

plugin-hrm-form/src/styles/HrmStyles.tsx
```diff
@@ -864,6 +793,69 @@ export const FormOption = styled('option')`
...
+  }
+
+  svg {
+    font-size: 16px;
+  }
+`;
+CategoryCheckbox.displayName = 'CategoryCheckbox';
+
+type CategoryCheckboxLabelProps = { disabled?: boolean };
+export const CategoryCheckboxLabel = styled('label')<CategoryCheckboxLabelProps>`
```
murilovmachado, line 819: Please add this styled component to `/__tests__/mockStyled.js` to prevent future issues in unit tests.

plugin-hrm-form/src/components/tabbedForms/ChildInformationTab.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+ChildInformationTab.displayName = 'ChildInformationTab';
+
+const mapStateToProps = (state: RootState, ownProps: OwnProps) => ({});
+
+const connector = connect(mapStateToProps);
+const connected = connector(ChildInformationTab);
+
+export default withTaskContext<Props, typeof connected>(connected);
```
murilovmachado, line 55: Since these three tabs are 99% the same code, can we have something like a `tabGenerator` in the future? Where we could just pass as a param `section: 'childInformation' | 'callerInformation' | 'caseInformation'`?

murilovmachado (approved):
Looks really good! Left a couple of comments, but nothing major.

PR #274, opened 2020-12-14
Add Referral Page

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-373

This PR implements the `+ Referral` feature, although it doesn't include showing it in the Timelime.

WIP:
- Tests

Screen Shot 2020-12-14 at 18 32 57
Screen Shot 2020-12-14 at 18 33 32


nickhurlburt:
@murilovmachado as discussed in our last sprint retro, who do you want to designate as the primary reviewer?

murilovmachado:
> @murilovmachado as discussed in our last sprint retro, who do you want to designate as the primary reviewer?

I'll mark @andvirga as the primary reviewer for this one.

nickhurlburt:
One thing about the mockups that's easy to miss is that the time of day has been added.  Per [mockup](https://marvelapp.com/prototype/fca99ce/screen/74694033):
image

These are being added to all of these case items (note, referral, HH member...)

andvirga (approved):
Good job! I've tested it and it's working great!

nickhurlburt (commented):
I realize this is already merged, but a couple comments:
1. It's using the white-background textarea rather than the gray-background in the mockup.
2. This looks like the two-column layout has different behavior than other forms.  There is much more space between the two columns than with others:
image

vs

image

I am okay with cleaning up the first one when we convert it all to the new customization elements, but I'm curious why the layout is different?

PR #275, opened 2020-12-15
Send the customized form to the backend

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-333

This PR modifies the form in local state before sending it to the backend. The only field that is nested at this point is `name` (caller & child info), in order to preserve the current code for search in the backend.


andvirga (approved):
Looks good!

murilovmachado:
> Jira: https://bugs.benetech.org/browse/CHI-333
> 
> This PR modifies the form in local state before sending it to the backend. The only field that is nested at this point is `name` (caller & child info), in order to preserve the current code for search in the backend.

Actually `phone1` and `phone2` (this latest I think doesn't exist anymore) is also a nested field on `location` (caller & child Info). Can you check if searching for phones still works @GPaoloni ?

Besides that, the code looks fine.

nickhurlburt:
@GPaoloni one thing we've discussed is adding the form version into the JSON sent to the backend.  In which task/PR will that be included?

GPaoloni:
@nickhurlburt The form version is included in this PR, in ContactService file. I'm using a string as the version identifier for flexibility, is that ok?
Screenshot from 2020-12-17 11-04-56

nickhurlburt:
Great!  I just didn't see it.

PR #276, opened 2020-12-16
Make contact details to show custom information

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-335

This PR updates ContactDetails component to consume all of the defined input fields.

Important note: this PR is not taking into consideration form versioning. While The form is saved with the version number (so it can be used in the future), we are now only using the v1 definition to consume the data. This will change soon, when we move to the final structure for fetching definition versions if needed.

### DIFF
--- a/plugin-hrm-form/src/___tests__/services/ContactService.test.js
+++ b/plugin-hrm-form/src/___tests__/services/ContactService.test.js
@@ -1,93 +1,30 @@
-import { transformForm, saveToHrm } from '../../services/ContactService';
-import { FieldType, ValidationType, recreateBlankForm } from '../../states/ContactFormStateFactory';
+import { set } from 'lodash/fp';
+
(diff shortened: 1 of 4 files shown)


andvirga (approved):
Looks good! I've tested the "Save contact" functionality and it's working great

murilovmachado (approved):

nickhurlburt (approved):
Looks good.  Does this fail for old contacts already stored in the database?

GPaoloni:
@nickhurlburt Yes, it fails, as the "legacy compatibility" has not been included (that's why I understood we wanted from last meeting, but correct me if I'm wrong).

nickhurlburt:
@GPaoloni that's fine, I just wanted to be clear on it.  We just want to be clear to everyone what to expect when this gets merged into staging.

PR #277, opened 2020-12-16
[Case View] - Header - Add Follow Up Date field

andvirga:
Related to: 
https://bugs.benetech.org/browse/CHI-365

Changes description:
- Added Date Selector and logic to store 'followUpDate' into Case info.
- Replaced Material UI components with native components (this would make it easier to update this view to use the new Customization framework).

### DIFF
--- a/plugin-hrm-form/src/components/case/Case.tsx
+++ b/plugin-hrm-form/src/components/case/Case.tsx
@@ -128,7 +128,7 @@ const Case: React.FC<Props> = props => {
     props.changeRoute({ route, subroute: 'view-perpetrator' }, props.task.taskSid);
   };
 
-  const handleFieldChange = (fieldName, value) => {
+  const onFollowupDateChange = (fieldName, value) => {
     const { connectedCase } = props.connectedCaseState;
     const { info } = connectedCase;
     const newInfo = info ? { ...info, [fieldName]: value } : { [fieldName]: value };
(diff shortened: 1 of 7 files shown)


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -127,6 +128,13 @@ const Case: React.FC<Props> = props => {
     props.changeRoute({ route, subroute: 'view-perpetrator' }, props.task.taskSid);
   };
 
+  const onFollowupDateChange = (fieldName, value) => {
```
murilovmachado, line 131: I think here you don't need to have `fieldName` as a parameter since it's value will always be `followUpDate`.

murilovmachado (approved):
Looks good to me. But after we implement `Update Case` we have to check if follow-up date is being saved just fine.

nickhurlburt (approved):
Looks great!  Does this input field support localized date formatting?  If you put in today's date, do you see 12/17/2020 or 17/12/2020?  (Ideally we want to support localized formatting, and if we can't do that now, we should use day/month/year.)

plugin-hrm-form/src/components/case/Case.tsx
andvirga: Yes, you're right! Actually, I've renamed this method in a posterior PR, so I'll do that change in that PR

andvirga:
> Looks great! Does this input field support localized date formatting? If you put in today's date, do you see 12/17/2020 or 17/12/2020? (Ideally we want to support localized formatting, and if we can't do that now, we should use day/month/year.)

In my case, I'm getting 18/12/2020 when I select today's date. Since we're using a native control I think that we're not supporting localization yet.

nickhurlburt:
> I've tested it in Windows (where I have en-US as language) and I'm seeing the MM/DD/YYYY format. However, in Linux, I have es-AR as language and I'm seeing DD/MM/YYYY. So I think that localization is working great. The date will be always stored in DB as YYYY-MM-DD format

Perfect.

PR #279, opened 2020-12-17
[Case View]- Header - Add Case Status dropdown

andvirga:
Story related:
https://bugs.benetech.org/browse/CHI-366

Note:
This PR needs to be merged first: https://github.com/tech-matters/flex-plugins/pull/277

Changes description:
- Added a native select control for case status.
- Added a new reducer and action to change status in the state.
- Added only two states for now (as per ZM requirements): open and closed.


murilovmachado:
@andvirga The status dropdown is dropping into a new line for me:
image

plugin-hrm-form/src/components/case/CaseDetails.jsx
```diff
@@ -74,17 +81,36 @@ const CaseDetails = ({ caseId, name, counselor, openedDate, lastUpdatedDate, fol
                 <Template code="Case-CaseDetailsFollowUpDate" />
               </label>
             </DetailDescription>
-            {/* Replace this with proper component in next story */}
             <StyledInputField
+              type="date"
```
nickhurlburt, line 85: This component is producing the console warning:
```
Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
    in input (created by StyledInputField)
    in StyledInputField (at CaseDetails.jsx:84)
```

plugin-hrm-form/src/components/case/CaseDetails.jsx
```diff
@@ -74,17 +81,36 @@ const CaseDetails = ({ caseId, name, counselor, openedDate, lastUpdatedDate, fol
...
           <Grid item xs role="gridcell" tabIndex={-1}>
-            <div style={{ height: '100%', display: 'flex', flexDirection: 'column', justifyContent: 'center' }}>
-              {renderCaseStatus(status)}
-            </div>
+            <DetailDescription>
+              <label id="CaseDetailsStatusLabel">
+                <Template code="Case-CaseDetailsStatusLabel" />
+              </label>
+            </DetailDescription>
+            <StyledSelectWrapper>
```
nickhurlburt, line 99: Further downstream this is causing the console warning:
```
plugin-hrm-form.js:295211 Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
    in option (created by FormOption)
    in FormOption (at CaseDetails.jsx:108)
    in select (created by FormSelect)
    in FormSelect (at case/index.tsx:255)
    in Unknown (created by StyledSelectField)
    in StyledSelectField (at CaseDetails.jsx:100)
    in div (created by StyledSelectWrapper)
    in StyledSelectWrapper (at CaseDetails.jsx:99)
```

plugin-hrm-form/src/components/case/CaseDetails.jsx
nickhurlburt: Also creating this warning when used:
```
plugin-hrm-form.js:295211 Warning: A component is changing an uncontrolled input of type date to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
    in input (created by StyledInputField)
    in StyledInputField (at CaseDetails.jsx:84)
```
(perhaps better addressed in #277)

plugin-hrm-form/src/components/case/CaseDetails.jsx
```diff
@@ -10,33 +11,39 @@ import CaseDetailsHeader from './caseDetails/CaseDetailsHeader';
...
-          </HiddenText>
-          <Template code="Case-CaseDetailsStatusOpen" />
-        </OpenStatusFont>
-      );
-    default:
-      return <DefaultStatusFont>{status}</DefaultStatusFont>;
-  }
-};
+const statusOptions = [
+  { label: 'Open', value: 'open', color: 'green' },
```
nickhurlburt, line 21: It's fine to do this now but we will need to make these options customizable for South Africa.

nickhurlburt (commented):
It doesn't look like the value of the dropdown is being persisted.  When I create a new case, set it to 'Closed' and then save it... and then I search for it later, the search says it's open.  Should that be working now or is there a future PR planned?

murilovmachado:
@andvirga About the issue the @nickhurlburt found that the status is not being saved when creating and saving the case, this is happening because right now the `Save and End` button only currently only save `info` changes, but status is not within info:
```
const handleSaveAndEnd = async () => {
  [...]
    try {
      const contact = await saveToHrm(task, form, hrmBaseUrl, workerSid, helpline);
      await updateCase(connectedCase.id, { info: connectedCase.info });
      await connectToCase(hrmBaseUrl, contact.id, connectedCase.id);
      props.handleCompleteTask(task.taskSid, task);
    } catch (error) {
  [...]
  };
```

Notice the line `await updateCase(connectedCase.id, { info: connectedCase.info });`, it passes only the info as a change.
One option would be to add `status` like this:
`await updateCase(connectedCase.id, { info: connectedCase.info, status: connectedCase.status });`
Another easier option to maintain would be to pass the whole `connectedCase`:
`await updateCase(connectedCase.id, connectedCase);`

I believe this should fix the issue. Thanks for spotting it @nickhurlburt.

plugin-hrm-form/src/components/case/CaseDetails.jsx
andvirga: Good catch! It's solved now

plugin-hrm-form/src/components/case/CaseDetails.jsx
andvirga: Sure, we will change that for ZA.

murilovmachado (approved):

plugin-hrm-form/src/components/case/CaseDetails.jsx
andvirga: Thanks! I've replaced `selected` with `defaultValue` prop

PR #280, opened 2020-12-17
Customization followup

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-427

This PR:
- Merges staging into this branch.
- Fixes tabs ugliness.
- Moved contact service to TS
- Transforms "midex" values into null before submitting (3-way checkbox)
- De-transforms null values into "mixed" before copying contact info (3-way checkbox)
- Adds support for ContactDetails to automatically display "Unknown" when value is null (3-way checkbox)
- Moves helpers and function to where they are used
- Code cleanup
- Addresses Jana's style changes for several inputs.
- Addressed the issues in the comments of #273

Notes: 
- This PR is better reviewd commit by commit.
- Everything should be working just fine in this one (contacts, cases, search - except for phone search -).


plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -11,6 +11,7 @@ import {
   ColumnarBlock,
   TwoColumnLayout,
   FormLabel,
+  DependantSelectLabel,
```
nickhurlburt, line 14: s/Dependant/Dependent/.

plugin-hrm-form/src/components/tabbedForms/IssueCategorizationTab.tsx
```diff
@@ -27,23 +28,23 @@ const IssueCategorizationTab: React.FC<Props> = ({ dispatch, task, display, cate
...
+    dispatch(setCategoriesGridView(!gridView, task.taskSid));
   };
 
   const toggleExpandCategory = (category: string) => {
     dispatch(handleExpandCategory(category, task.taskSid));
   };
 
   return (
-    <div style={{ display: display ? 'block' : 'none' }}>
+    <TabbedFormTabContainer display={display}>
```
nickhurlburt, line 39: When selecting and unselecting the subcategories, they seem to shift a little left and right.  I don't think it's worth a lot of effort to fix, but it's kind of weird.

plugin-hrm-form/src/components/tabbedForms/TabbedForms.tsx
```diff
@@ -144,15 +153,31 @@ const TabbedForms: React.FC<Props> = ({ dispatch, routing, contactForm, ...props
...
+              />
+            )}
+            <TabbedFormTab
+              tabPath="childInformation"
+              definition={ChildTabDefinition as FormDefinition}
+              display={subroute === 'childInformation'}
+            />
+            <IssueCategorizationTab display={subroute === 'categories'} />
+            <TabbedFormTab
+              tabPath="caseInformation"
```
nickhurlburt, line 176: This is the main tab referenced in https://bugs.benetech.org/browse/CHI-205, where it would be good to rename things as "contact information", etc, since it can be confusing that we now have cases.  Does not need to be handled in this PR.

plugin-hrm-form/src/formDefinitions/tabbedForms/CaseInformationTab.json
```diff
@@ -21,7 +21,7 @@
       { "value": "Direct interventions by Childline", "label": "Direct interventions by Childline" }, 
       { "value": "Other", "label": "Other" }
     ],
-    "required": true
+    "required": { "value": true, "message": "RequiredFieldError" }
   },
   {
     "name": "howDidTheChildHearAboutUs",
```
nickhurlburt, line 27: Probably all of the checkboxes here will be three-way.  Though no need to change in this PR.

plugin-hrm-form/src/formDefinitions/tabbedForms/CaseInformationTab.json
```diff
@@ -21,7 +21,7 @@
       { "value": "Direct interventions by Childline", "label": "Direct interventions by Childline" }, 
       { "value": "Other", "label": "Other" }
     ],
-    "required": true
+    "required": { "value": true, "message": "RequiredFieldError" }
```
nickhurlburt, line 24: This suggests the field is required, but when I click in and don't select a value, there's no error.

nickhurlburt (commented):
I have some questions, but generally this is okay.

plugin-hrm-form/src/components/tabbedForms/TabbedForms.tsx
GPaoloni: I'll rather do this when we are closer to cleaning the DB, as this will also change the name being stored in the rawJson column, and that will trigger changes in other parts of the system (probably case related stuf and backend, other than tabbed forms). Does that sounds right?

plugin-hrm-form/src/formDefinitions/tabbedForms/CaseInformationTab.json
GPaoloni: That is because of the way validations are triggered. If you click on empty value, the field is still not "dirty" so it won't revalidate. When you try to submit, it will, and error will appear. Another way of triggering validation other than `onSubmit` is picking a different (non empty) value, then going back to empty value, there will appear an error as the field is "dirty". Same goes for required text inputs: you focus but don't type anything, then `onBlur` the field is not dirty and thus not validated, but if you type something and then delete it all, the field will error. All of the validations are triggered `onSubmit`, whether the field is "dirty" or not, but validations `onChange` runs if the value has changed.

GPaoloni: PS: above is true because the validation `mode` that I've set in the code is `onChange`, but that can be a different one. Should I change to `onBlur` instead? The [docs](https://react-hook-form.com/api) warns that onChange mode can affect performance.

plugin-hrm-form/src/components/tabbedForms/TabbedForms.tsx
nickhurlburt: Yes

plugin-hrm-form/src/formDefinitions/tabbedForms/CaseInformationTab.json
nickhurlburt: I think the way you have it is fine, and we don't need to do anything different unless we get complaints.

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -42,38 +44,47 @@ export async function searchContacts(searchParams, limit, offset) {
...
-export const createCategoriesObject = () => Object.entries(categoriesFormDefinition).reduce(createCategory, {});
+const createCategoriesObject = () => Object.entries(categoriesFormDefinition).reduce(createCategory, {});
 
 /**
  * Transforms the form to be saved as the backend expects it
  * VisibleForTesting
- * @param {import('../states/contacts/reducer').TaskEntry} form
  */
-export function transformForm(form) {
+export function transformForm(form: TaskEntry): ContactRawJson {
```
andvirga, line 97: This is just a question, I have used this function in Case.tsx file because I need to convert the "raw form" to a mapped version of that version that allows me to display the categories in Case View.
```javascript
  const getCategories = firstConnectedContact => {
    if (firstConnectedContact?.rawJson?.caseInformation) {
      return firstConnectedContact.rawJson.caseInformation.categories;
    }
    const mappedForm = transformForm(form);
    return mappedForm?.caseInformation?.categories;
  };
```
My question is: Since you're modifying this function, do I need to change something in my code?

plugin-hrm-form/src/components/tabbedForms/TabbedFormTab.tsx
```diff
@@ -0,0 +1,64 @@
...
+  );
+};
+
+TabbedFormTab.displayName = 'TabbedFormTab';
+
+const mapDispatchToProps = {
+  updateForm: actions.updateForm,
+};
+
+const connector = connect(null, mapDispatchToProps);
```
andvirga, line 61: This null is here because we're not using `mapStateToProps`, right?

andvirga (approved):
I've tested it on my local environment and everything seems to be working as expected. You've done an incredible job, Gian!

plugin-hrm-form/src/services/ContactService.ts
GPaoloni: I'm not sure, as I don't know what is your code doing, but the form will be transformed as we store it in the backend. We can check that later before merging this into staging, by merging your changes into this. Also consider that when merging this series of PRs into staging, the form in Redux state won't be the same as it is now, for example the categories will be an array, like you can see in the `TaskEntry` type (`src/states/contacts/reducer.tsx`), so maybe you won't even need to transform anything.

plugin-hrm-form/src/components/tabbedForms/TabbedFormTab.tsx
GPaoloni: That's right!

plugin-hrm-form/src/styles/HrmStyles.tsx
```diff
@@ -553,12 +553,18 @@ FormItem.displayName = 'FormItem';
...
-  font-size: 13px;
-  letter-spacing: 1px;
+  font-size: 14px;
+  letter-spacing: 0;
   min-height: 18px;
+  color: #000000;
 `;
 FormLabel.displayName = 'FormLabel';
 
+export const DependantSelectLabel = styled(FormLabel)<{ disabled: boolean }>`
```
murilovmachado: I think it would be good to put this on `mockStyled.js` to avoid breaking tests in the future.

plugin-hrm-form/src/styles/HrmStyles.tsx
```diff
@@ -69,6 +69,12 @@ export const TabbedFormsContainer = styled('div')`
 `;
 TabbedFormsContainer.displayName = 'TabbedFormsContainer';
 
+export const TabbedInputsContainer = styled('div')`
```
murilovmachado: Put on `mockStyled.js`

murilovmachado:
One thing that I've noticed is that the categories checkbox styles still doesn't match our current ones.

**Unchecked**
Screen Shot 2020-12-23 at 11 17 17
vs
Screen Shot 2020-12-23 at 11 16 56


**Checked**
Screen Shot 2020-12-23 at 11 17 24
vs
Screen Shot 2020-12-23 at 11 17 07

If it's a limitation on styling the native html checkbox, that's fine. Just needed to point these differences.

murilovmachado:
Another issue, when canceling a case, the app goes back to tabbed forms with no tab rendered.

Screen Shot 2020-12-23 at 11 35 12
Screen Shot 2020-12-23 at 11 34 45

murilovmachado (approved):
Code looks good. There're just a few comments I've left about styling and an issue about not rendering tabbed forms properly. When you address that, you're good to go. I don't feel I need no take another look. Great job!

PR #281, opened 2020-12-17
[Case View] - Header - Show categories

andvirga:
Related to: 
https://bugs.benetech.org/browse/CHI-367

Primary reviewer: 
@murilovmachado 

Changes description:
- Added `<CaseTags>` component to display categories in Case View.


nickhurlburt (commented):
This isn't showing the categories when the case is first being created.  Open a new contact form, add required fields, then choose 'Open New Case'.  The categories from the contact don't appear:
image

It's possible that one problem is the positioning of the `<CaseTags>` element:
image

In search, it looks as expected, so maybe there's also an issue with accessing the information of a contact that has not yet been saved.

murilovmachado:
Not sure if it's because `Status` component is in another line, but the categories are showing up in the right of the sreen:
image

But in the mockups they're on the left

nickhurlburt:
Yes, the Opened, Last Updated, Follow Up and Status should all be on a line, and shouldn't wrap if the window is too small.  Categories should be left-aligned on a separate line below.

nickhurlburt:
@andvirga I have a request to add some additional styling to this PR.  Notice that in the [original mockup](https://marvelapp.com/prototype/6dc75bc/screen/74539629), the left edge of the "Opened" date and the categories is almost flush with the left edge of the Case Title and the Timeline.  However, what we've built has more left margin than that, for example:
image

Additionally, the width of the Date fields in the header looks unnecessarily wide, and wider than the mockups.  There is a lot of unnecessary space to the right of the dates:
image

Can we make these match the mockups more closely by reducing the left margin and reducing the width of the fields?  The reason is that for smaller screens, the header field line is very wide and causing displays like this:

thumbnail_image002

We may not be able to solve this for all small screens.  But we can at least match the margin and width in the mockups to save at least a few dozen pixels of width.

andvirga:
@nickhurlburt I've removed the left padding and replaced it with a right one, that should solve the first issue.
About date picker width, I couldn't make it smaller than 130 px, If I put a smaller value the date starts to be hidden by the calendar icon. Here is an example:
image

I've tried to solve that issue with no luck so far, I've tried to customize the styling of the calendar's pseudo-element provided by Chrome `(::-webkit-calendar-picker-indicator)`

However, I think that using a width of 130 px is a good alternative and we don't have to deal with pseudo-element (that will affect all the date-time pickers in the app). Here is how the view looks with these values:
image

With shrunk screen:
image

Let me know if this looks good or if you prefer to do more changes.

nickhurlburt:
@andvirga thank you, that looks good!  If we need to make it more narrow then I think we have to go back to the UI design to see how that can be changed.  We don't need to worry about that now.

murilovmachado (approved):
Looks good to me!

PR #284, opened 2020-12-22
[Case View] - Timeline - UI Changes

andvirga:
Related to: https://bugs.benetech.org/browse/CHI-371

Primary reviewer:  @murilovmachado

Changes description:
- Added `CallTypeIcon` component.
- Obtained call type prop from rawJson or form, depending on the scenario.
- Styling

### DIFF
--- a/plugin-hrm-form/src/components/case/Timeline.jsx
+++ b/plugin-hrm-form/src/components/case/Timeline.jsx
@@ -7,8 +7,16 @@ import PropTypes from 'prop-types';
 import Dialog from '@material-ui/core/Dialog';
 import DialogContent from '@material-ui/core/DialogContent';
 
+import CallTypeIcon from '../common/icons/CallTypeIcon';
 import TimelineIcon from './TimelineIcon';
-import { CaseSectionFont, ViewButton, TimelineRow, TimelineDate, TimelineText } from '../../styles/case';
+import {
+  CaseSectionFont,
+  ViewButton,
+  TimelineRow,
+  TimelineDate,
+  TimelineText,
+  TimelineCallTypeIcon,
+} from '../../styles/case';
 import { Box, Row } from '../../styles/HrmStyles';
 import { taskType, formType } from '../../types';
 import CaseAddButton from './CaseAddButton';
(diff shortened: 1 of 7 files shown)


murilovmachado:
@andvirga There's 2 things that I've noticed:
1. Not all timeline items should display the callType icon, but only the connected contacts activities
2. There should be more spacing right after the callType icon

Screen Shot 2020-12-22 at 17 19 25

plugin-hrm-form/src/components/case/Timeline.jsx
```diff
@@ -97,6 +105,14 @@ const Timeline = ({ task, form, caseObj, changeRoute, updateTempInfo, route }) =
     changeRoute({ route, subroute: 'add-referral' }, task.taskSid);
   };
 
+  /*
+   * If case has not been created yet, we should return value from the form.
+   * Else If case was already created we should return rawJson value.
+   */
+  const getCallType = () => {
+    return form.callType.value || caseObj.connectedContacts[0]?.rawJson?.callType;
+  };
```
GPaoloni, line 114: Just asking: if case exists then we should return that one at first, right? Can this fall in a case where we return the form but the case should be returned instead? Also, as a note: `form.callType.value` is one of the things that will break after customization gets merged, just to be aware!

GPaoloni (approved):
Great!!

nickhurlburt:
I looked quickly, and have the same comments as Murilo.  I also tested that it truncates long summaries correctly -- this looks good.

plugin-hrm-form/src/components/case/Timeline.jsx
andvirga: I think is the other way round, I have noticed that if the case already exists and we're retrieving it from the DB then `form.callType.value` is blank, therefore `callType` value is taken from `rawJson`.
Anyway, we will change this once the Customization framework is deployed to staging. Thanks for the reminder!

plugin-hrm-form/src/components/case/helpers.ts
```diff
@@ -0,0 +1,7 @@
+import { channelsAndDefault } from '../../states/DomainConstants';
+
+/**
+ * Returns true if the activity provided is a connected case activity (included in channelsAndDefault const object)
+ * @param activity Activity (note, referral, web, facebook, sms, etc.)
+ */
+export const isConnectedCaseActivity = (activity: any): Boolean => Boolean(channelsAndDefault[activity.type]);
```
GPaoloni, line 7: Given that this is a TS file maybe you can create a type (don't know where it belongs, perhaps `types/types.ts`) and use it in here? That will help us track the activities we have so far. I believe that right now it should like
```
export type ActivityTpe = typeof channelsAndDefault[keyof typeof channelsAndDefault] | 'note' | 'referral';
```

Note that for this to work as a union type you'll need to add 
```
export const channelsAndDefault = { ...channelTypes, default: 'default' } as const; // add "as const" to the end of channelsAndDefault declaration
```

murilovmachado (approved):
Looks great, I've seen everything has been fixed!

plugin-hrm-form/src/components/case/helpers.ts
GPaoloni: My bad, this already exists in https://github.com/tech-matters/hrm/blob/ae852ed638bb319338968a25a761c525d0503344/controllers/activities.js#L1

PR #289, opened 2020-12-24
Bug fix: dependent select & 3-way checkbox losing state

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-438

This PR fixes the 2nd bug described in https://github.com/tech-matters/flex-plugins/pull/280#issuecomment-750327284, where dependent select & 3-way were being incorrectly initialized with redux/searchResult/transfer data.
This has been fixed by including 1 extra parameter to the `form generator` function (1st commit), which is an `initialValues` object (may be null).
If the provided `initialValues` does not contains a key for certain input, it will default to the "empty state" value associated to the input type.
Second and third commits only makes use of this new approach in tabbed forms.

TODO:
- strictness types for the `form generator` function.

### DIFF
--- a/plugin-hrm-form/src/components/common/forms/formGenerators.tsx
+++ b/plugin-hrm-form/src/components/common/forms/formGenerators.tsx
@@ -472,12 +472,12 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
 export const createFormFromDefinition = (definition: FormDefinition) => (parents: string[]) => (initialValues: any) => (
   updateCallback: () => void,
 ): JSX.Element[] => {
-  const f = getInputType(parents, updateCallback);
+  const bindGetInputType = getInputType(parents, updateCallback);
 
   return definition.map((e: FormItemDefinition) => {
-    const v = get(initialValues, e.name);
-    const initialValue = v === undefined ? getInitialValue(e) : v;
-    return f(e)(initialValue);
+    const maybeValue = get(initialValues, e.name);
+    const initialValue = maybeValue === undefined ? getInitialValue(e) : maybeValue;
+    return bindGetInputType(e)(initialValue);
   });
 };
 
--- a/plugin-hrm-form/src/components/common/forms/formGenerators.tsx
+++ b/plugin-hrm-form/src/components/common/forms/formGenerators.tsx
@@ -27,6 +27,41 @@ import {
 } from '../../../styles/HrmStyles';
 import type { FormItemDefinition, FormDefinition, SelectOption, MixedOrBool } from './types';
 
+/**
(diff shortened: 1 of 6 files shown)


plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -166,25 +206,35 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
...
+
             const dependeePath = [...parents, def.dependsOn].join('.');
             const dependeeValue = watch(dependeePath);
 
-            React.useEffect(() => setValue(path, def.defaultOption.value, { shouldValidate: true }), [
-              setValue,
-              dependeeValue,
-            ]);
+            React.useEffect(() => {
+              if (isMounted.current && prevDependeeValue.current && dependeeValue !== prevDependeeValue.current) {
```
GPaoloni, line 216: Explanation: if 
- `isMounted.current` -> this input is already mounted (after 1st render where entire form is undefined)
- `prevDependeeValue.current` -> there were a value on the input that this one is dependent on
- `dependeeValue !== prevDependeeValue.current` -> and it was different than the current one
and this function is executed (will only happen when `setValue` or `dependeeValue`, per dependencies array), only then clear the state of this dependent select. Any other case is either a false change (dependeeValue === prevDependeeValue.current) or is the first mount and we want to preserve the state that is being copied from Redux state.

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -40,6 +42,12 @@ const nestName = (information: { firstName: string; lastName: string }) => {
   return { ...rest, name: { firstName, lastName } };
 };
 
+export const unNestInforationObject = (
```
nickhurlburt: spelling - Information

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -413,45 +469,18 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
...
-      return false;
-    case 'mixed-checkbox':
-      return def.initialChecked === undefined ? 'mixed' : def.initialChecked;
-    default:
-      return null;
-  }
+  return definition.map((e: FormItemDefinition) => {
+    const v = get(initialValues, e.name);
+    const initialValue = v === undefined ? getInitialValue(e) : v;
+    return f(e)(initialValue);
```
andvirga, line 480: Can we use a more descriptive name for `f` and `v`? (`e` is ok since it's an element inside `map` iteration)

andvirga (approved):
I've tested it and it looks good!

PR #291, opened 2020-12-29
[Case View] - Add “Update Case” button

andvirga:
Jira ticket:
https://bugs.benetech.org/browse/CHI-360

Primary reviewer: @GPaoloni 

Changes description:
- Added a new **Update** button to the bottom bar that updates case information.
- Added a logic to avoid updates of closed cases. (If the user changes the Status dropdown to **closed**, all edition controls are turned disabled).
- Fixed [[Case View] - Bottom bar is inaccessible when window is resized](https://bugs.benetech.org/browse/CHI-396).
- Fixed styles and unit tests.


plugin-hrm-form/src/components/case/CaseAddButton.jsx
```diff
@@ -6,9 +6,10 @@ import { Template } from '@twilio/flex-ui';
 
 import { CaseAddButtonFont } from '../../styles/case';
 
-const CaseAddButton = ({ templateCode, onClick, withDivider }) => {
+const CaseAddButton = ({ status, templateCode, onClick, withDivider }) => {
   return (
     <ButtonBase
+      disabled={status === 'closed'}
```
GPaoloni, line 12: Maybe you can pass down this prop to `CaseAddButtonFont` too, and make the buttons "grey out" when they are disabled? That will give some visual hint about this case not being editable anymore.

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -146,6 +164,14 @@ const Case: React.FC<Props> = props => {
...
   };
 
+  /**
+   * Setting this flag in the first render.
+   */
+  useEffect(() => {
+    setIsEditing(props.connectedCaseState?.connectedCase?.status === 'open');
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
```
GPaoloni, line 174: I feel like using `useState` + `useEffect` for tracking if the case is editable or not is an overkill. Maybe something like
```
const [isEditable, setEditable] = useState(props.connectedCase && props.connectedCase.status === 'open');
```
can achieve the same with less logic involved. I'm not sure tho, as there might be corner cases I'm missing. What do you think?

plugin-hrm-form/src/components/case/CaseDetails.jsx
```diff
@@ -103,6 +105,7 @@ const CaseDetails = ({
                 id="Details_CaseStatus"
                 name="Details_CaseStatus"
                 aria-labelledby="CaseDetailsStatusLabel"
+                disabled={!isEditing}
```
GPaoloni, line 108: Question here: why is this using `disabled={!isEditing}` but above (line 91) is using `disabled={status === 'closed'}`. Aren't they logically equivalent? If so, I think we should use the same in both cases, if not, what am I missing?

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -218,44 +245,46 @@ const Case: React.FC<Props> = props => {
...
               <Perpetrators
                 perpetrators={perpetrators}
+                status={status}
                 onClickAddPerpetrator={onClickAddPerpetrator}
                 onClickView={onClickViewPerpetrator}
               />
             </Box>
             <Box marginLeft="25px" marginTop="25px">
-              <CaseSummary task={props.task} readonly={props.readonly} />
+              <CaseSummary task={props.task} readonly={props.isCreating} />
```
GPaoloni, line 267: This is not being disabled after the case is closed, hence I can write in stuff but the "update" button does not show. I have the feeling that this should also be disabled.

Screenshot from 2020-12-30 12-20-15

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -218,44 +245,46 @@ const Case: React.FC<Props> = props => {
...
+              <MenuItem
+                Icon={AddIcon}
+                text={<Template code="BottomBar-AddThisContactToExistingCase" />}
+                onClick={handleMockedMessage}
+              />
+              <MenuItem
+                red
+                Icon={CancelIcon}
+                text={<Template code="BottomBar-CancelNewCaseAndClose" />}
+                onClick={handleCancelNewCaseAndClose}
```
GPaoloni, line 282: I think this does not belong to the general case, but below `isCreating` one. This causes an standalone search case to show the following.

Screenshot from 2020-12-30 12-47-05

GPaoloni (commented):
In general the code looks great, just a couple comments on the behavior.

I've also found an ugly bug, but I think you are not introducing it, we can realize of this now because of having the ability to edit cases.
When a case is successfully updated, the list of case search results is not, so the sync is lost.
You can
- close a case
- go back to search results
- open the same case using the same search results
and you are able to edit stuff. This will lead to inconsistencies.
I don't think you have to address this bug in this PR, just raising the issue.
Possible fix: the case is returned from the backend after it's updated. We can `find` the corresponding case in the case search results list and replace the out-of-date value with new-returned-from-DB one.

plugin-hrm-form/src/components/case/Case.tsx
andvirga: Cool! It seems to be equivalent to use `useEffect(..., []);` I'll do some testing but so far looks promising!

plugin-hrm-form/src/components/case/CaseDetails.jsx
andvirga: Since this is the Case Status dropdown, we need to use `isEditing` instead of `disabled={status === 'closed'}`. If not the dropdown itself will also get disabled when the user changes the value to closed.

plugin-hrm-form/src/components/case/Case.tsx
andvirga: Good catch! I've changed that button to the "Close" one

plugin-hrm-form/src/components/case/CaseAddButton.jsx
andvirga: > In general the code looks great, just a couple comments on the behavior.
> 
> I've also found an ugly bug, but I think you are not introducing it, we can realize of this now because of having the ability to edit cases.
> When a case is successfully updated, the list of case search results is not, so the sync is lost.
> You can
> 
> * close a case
> * go back to search results
> * open the same case using the same search results
>   and you are able to edit stuff. This will lead to inconsistencies.
>   I don't think you have to address this bug in this PR, just raising the issue.
>   Possible fix: the case is returned from the backend after it's updated. We can `find` the corresponding case in the case search results list and replace the out-of-date value with new-returned-from-DB one.
>   At the same time, backend should not accept update requests for closed cases.

Based on your suggestions I've added a new action and reducer that updates the case right after that case gets updated in the DB.

plugin-hrm-form/src/states/search/reducer.ts
```diff
@@ -217,6 +217,25 @@ export function reduce(state = initialState, action: t.SearchActionType | Genera
...
+
+      return {
+        ...state,
+        tasks: {
+          ...state.tasks,
+          [action.taskId]: {
+            ...task,
+            searchCasesResult: {
+              ...searchCasesResult,
+              cases: [...filteredResults, action.updatedCase].sort((a, b) => b.id - a.id),
```
GPaoloni, line 233: This works fine but it's actually more expensive than just replacing the proper element. Not saying you have to, but you can change to something like
```
updatedCases = map(c  => c.id === action.updatedCase.id ? action.updatedCase : c)
```

plugin-hrm-form/src/styles/case/index.tsx
```diff
@@ -83,11 +83,17 @@ export const ViewButton = styled<ViewButtonProps>(props => <Button roundCorners=
 `;
 ViewButton.displayName = 'ViewButton';
 
-export const CaseAddButtonFont = styled(FontOpenSans)`
+type CaseAddButtonFontProps = {
+  disable: boolean;
```
GPaoloni, line 87: Rename to `disabled` to keep consistency with the rest of our code base please.

plugin-hrm-form/src/states/search/types.ts
```diff
@@ -81,6 +82,12 @@ type SearchExpandSectionAction = {
   taskId: string;
 };
 
+type SearchCasesUpdateAction = {
+  type: typeof SEARCH_CASES_UPDATE;
+  taskId: string;
+  updatedCase: any;
```
GPaoloni, line 88: Use `Case` type here? We have that defined in `src/types/types`

Screenshot from 2020-12-31 17-16-10

Screenshot from 2020-12-31 17-16-03

GPaoloni (approved):
Looking great!

plugin-hrm-form/src/states/search/reducer.ts
andvirga: You're right! This is much better! I'll change it

plugin-hrm-form/src/components/case/CaseAddButton.jsx
nickhurlburt: > Maybe you can pass down this prop to CaseAddButtonFont too, and make the buttons "grey out" when they are disabled? That will give some visual hint about this case not being editable anymore.

There's a current UI Design task to mock up how this "greying out" should look.  We can wait for that.

PR #293, opened 2020-12-29
CHI-430 - Add back the number of returned search results

andvirga:
Jira ticket: https://bugs.benetech.org/browse/CHI-430

Primary reviewer: @GPaoloni 

Changes description:
- Added contacts and cases search results count back to the layout.
- Fixed styling and unit tests.


GPaoloni (approved):
Looks great!

PR #297, opened 2020-12-31
(WIP) Add Insights Support for Zambia (Contact Forms Only)

nickhurlburt:
Note: some mixed checkboxes are on string attributes... is this an issue?
We should test what happens if some of the optional fields don't get data -- everything still work?


nickhurlburt:
Closing in favor of #299

PR #299, opened 2021-01-03
Add Insights Support for Zambia

nickhurlburt:
This adds a framework for handling insights support for helplines, and includes configuration for Zambia.

Primary reviewer: @GPaoloni 
Fulfills: [CHI-383](https://bugs.benetech.org/browse/CHI-383)

Here are the key overall design concepts:
- The `InsightsService` applies a series of cascading functions that create a final version of the insights attributes.  This allows for a base set of updates for all helplines (such as storing categories) as well as configurable updates per helpline.
- Updates for a helpline can be expressed by a configuration language very similar to the JSON for a `FormDefinition`.  This is available in `insightsConfig/zambia.ts`.  This is currently a JS object, but the idea is that it can be merged into the `FormDefinition` JSON as a future revision.
- The configuration language doesn't handle situations where two fields should be combined into one attribute, so there's a small amount of custom code written in `zambiaUpdates()` to apply those changes.

File-by-file, here are the major changes:
- `InsightsService.test.js`: Adds tests for supporting three-way checkboxes (converting to `int(1)`, `int(0)` and `null`; Adds tests for processing the configuration language; Adds test for custom Zambia additions.
- `insightsConfig/types.ts`: Adds types used in insights configuration.  Types at the bottom of the file are composed of types at the top.
- `insightsConfig/zambia.ts`: The configuration for Zambia.  Comments are added where there's a hard-coded conversion not captured here, or where there are outstanding questions about Zambia configuration (to be resolved in a future PR), or for incidents which is not yet running.  This configuration could be merged with existing `FormDefinition` files by adding the `insights` property to relevant fields.  The case definition strays from the definition of `Case` elsewhere so that it matches the pattern of the contact forms.  There is a transform function in `InsightsService.ts` for it.  
- `InsightsService.ts`: Changes the existing functions to only add base configuration, removing insights fields for gender and age that could be custom per helpline.  Adds `processHelplineConfig` and supporting functions for running through configuration like `zambia.ts`.  Adds a transform for `Case` to make the data structure better fit the model of the contact forms which have a form -> subForm pattern.
- `setupActions.js`: Adds access to the case form from Redux.  Sends the case form to `saveInsightsData()`.  Adds a config param that activates storing Zambia config in Insights: this is hard-coded for now but can made configurable using `getConfig()` or whichever method we use for specifying which `FormDefinition`s to use in different helplines.

Some notes:
- Support for incidents is written but commented out, since we don't have incidents working yet.
- I'm still a little concerned about using ints vs strings in some of the fields, and this will need to be tested.  I'm also concerned about nulls vs "".  We use "" or "Unknown" when certain values aren't filled in by the user.
- Are there any pass-by-reference dangers here?  Could we be modifying data that will be stored elsewhere, for example, saved to the HRM?
- This also fixes [CHI-450](https://bugs.benetech.org/browse/CHI-450), which I broke in #295 by making the `date` attribute a string.  It needs to be a number.
- This will only grab the first element of Perpetrators or Referrals for Insights.  Possibly in the future we may need to add one of these to Insights if and only if it was added on this contact.  This will matter when we have cases with multiple connected contacts.

### DIFF
--- a/plugin-hrm-form/src/services/InsightsService.ts
+++ b/plugin-hrm-form/src/services/InsightsService.ts
@@ -104,9 +104,13 @@ enum InsightsObject {
   Customers = 'customers',
   Conversations = 'conversations',
 }
+enum FieldType {
+  MixedCheckbox = 'mixed-checkbox',
+}
 type InsightsFieldSpec = {
   name: string;
   insights: [InsightsObject, string];
+  type?: FieldType;
 };
 type InsightsSubFormSpec = InsightsFieldSpec[];
 type InsightsFormSpec = { [key: string]: InsightsSubFormSpec };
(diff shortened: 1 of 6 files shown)


andvirga (approved):
Looks good to me! However, I don't have too much experience on this topic.

nickhurlburt:
I went ahead and deployed this so that Joan could get going on it, but a quick review would still be good.  I'll definitely have to make some future edits anyway.

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -106,39 +251,28 @@ const mergeAttributes = (previousAttributes: TaskAttributes, newAttributes: Insi
   };
 };
 
-const contactlessAttributes = (attributes: TaskAttributes, contactForm: TaskEntry): InsightsAttributes => {
-  if (!attributes.isContactlessTask) {
-    return {};
+// In TS, how do we say that we are returning a function?
```
GPaoloni, line 254: Functions in the type system are declared with the same arrow notation, [like](https://www.typescripttutorial.net/typescript-tutorial/typescript-function-types/)
```
let add: (x: number, y: number) => number; // we are only specifying the type here
add = (x, y) => x + y; // a valid implementation, as it matches the type
add = (x, y) => 'this errors'; // type error
```
Assuming that each of the functions of `functionArray` have the type, 
```
(attributes: TaskAttributes, contactForm: TaskEntry, caseForm: Case) => InsightsAttributes;
```
and `getInsightsUpdateFunctionsForConfig` returns an array of such functions, you can type it like
```
const getInsightsUpdateFunctionsForConfig = (config: any): ((attributes: TaskAttributes, contactForm: TaskEntry, caseForm: Case) => InsightsAttributes
)[ ] => {...}
```
I suggest factoring out the type of the inner functions so it ends up being something like
```
type InsightsUpdateFunction = (attributes: TaskAttributes, contactForm: TaskEntry, caseForm: Case) => InsightsAttributes;

const getInsightsUpdateFunctionsForConfig = (config: any): InsightsUpdateFunction[ ];
```

GPaoloni (commented):
Looks great to me!
> Are there any pass-by-reference dangers here? Could we be modifying data that will be stored elsewhere, for example, saved to the HRM?

I couldn't notice anything dangerous, as the only place where you are "modifying" Redux information is when "Flatten out the Perpetrator object", and that is done like
```
const thePerp = caseForm.info.perpetrators[0];
const untypedPerp: any = { ...thePerp }; // this is a new object reference. You should not update the inner properties (as those are the same references as the original one), but deleting or adding stuff to untypedPerp is safe
```

PR #300, opened 2021-01-04
Fix Case Screen Background

murilovmachado:
Primary reviewer: @andvirga 

This PR:
- Fixes `Case` Screen to have a white background
- Makes `ViewReferral` use `CaseLayout` (as the other screens are now using)


andvirga (approved):
Looks good!

PR #302, opened 2021-01-04
CHI-449 [Case View] - Blank screen when clicking any "+" Button

andvirga:
Jira related:
https://bugs.benetech.org/browse/CHI-449

Changes description:
- Added `<Case>` component inside TabbedForms to be rendered properly.
- Added a logic inside `handleClose` to redirect to the corresponding page (In Contact or Standalone search).
 - Unified properties passed to View components (ViewContact, ViewNote, etc.)

### DIFF
--- a/plugin-hrm-form/src/components/case/Case.tsx
+++ b/plugin-hrm-form/src/components/case/Case.tsx
@@ -112,9 +112,14 @@ const Case: React.FC<Props> = props => {
     }
   };
 
+  // Redirects to the proper view when the user clicks 'Close' button.
   const handleClose = () => {
     props.updateTempInfo(null, props.task.taskSid);
-    props.changeRoute({ route }, props.task.taskSid);
+    if (route === 'select-call-type') {
+      props.changeRoute({ route: 'select-call-type' }, props.task.taskSid);
+    } else {
+      props.changeRoute({ route: 'tabbed-forms', subroute: 'search' }, props.task.taskSid);
+    }
   };
 
   const onClickAddHousehold = () => {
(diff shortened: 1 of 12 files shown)


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -217,15 +222,15 @@ const Case: React.FC<Props> = props => {
...
+      return <ViewNote {...addScreenProps} />;
     case 'view-household':
-      return <ViewHousehold task={props.task} onClickClose={handleClose} />;
+      return <ViewHousehold {...addScreenProps} />;
     case 'view-perpetrator':
-      return <ViewPerpetrator task={props.task} onClickClose={handleClose} />;
+      return <ViewPerpetrator {...addScreenProps} />;
     case 'view-referral':
-      return <ViewReferral taskSid={props.task.taskSid} />;
+      return <ViewReferral {...addScreenProps} />;
```
murilovmachado, line 233: Very nice!

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -112,9 +112,14 @@ const Case: React.FC<Props> = props => {
...
 
+  // Redirects to the proper view when the user clicks 'Close' button.
   const handleClose = () => {
     props.updateTempInfo(null, props.task.taskSid);
-    props.changeRoute({ route }, props.task.taskSid);
+    if (route === 'select-call-type') {
+      props.changeRoute({ route: 'select-call-type' }, props.task.taskSid);
+    } else {
+      props.changeRoute({ route: 'tabbed-forms', subroute: 'search' }, props.task.taskSid);
+    }
```
murilovmachado, line 122: The idea is good, but there's one scenario that's not being covered and causes the app to bug:
1) Fill the contact form and click **Open New Case**
2) Click **+ Note**
3) Close **Add Note** screen

The app redirects to **Tabbed Form** inside **Search tab**.
Expected: It should just go back to the **Open New Case** screen.

This is happening because there's one route not being handled: `route: 'new-case'`.

murilovmachado (changes requested):
Code looks much better and standardized. It fixes the blank screen issue, but it creates a new bug as described in one comment below. Let me know if you need more information.

plugin-hrm-form/src/components/case/Case.tsx
andvirga: Thanks for spotting this issue! I've added a new condition that handles `new-case` route

murilovmachado (approved):
looks good!

PR #303, opened 2021-01-05
Bug fix: Only one category showing on Case View Header

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-443
Primary reviewer: @andvirga 

This PR fixes a bug in the Case Header: "Only one category showing on Case View Header". This happened when picking more than one subcategory from the same category.


andvirga (approved):
Tested it and it looks great!

murilovmachado (approved):

GPaoloni:
Deployed to staging.

PR #304, opened 2021-01-06
CHI-455 - Other Case View Changes

andvirga:
Jira related:
https://bugs.benetech.org/browse/CHI-455

Primary reviewer: @GPaoloni 

Changes description:
- Added Counsellor back to Case Header,
- Disable Update button if a case has not been edited.
- Fixed a blank screen bug when open any VOIDED case (voided cases don't have child name or categories info).
- Fixed a bug that prevented the user to enter a Case summary when the case has just been created.
- Fixed styling of Close button (secondary).


GPaoloni (approved):
Looks good to me!

PR #305, opened 2021-01-06
Show Contact Info Tab for Offline Non-Data Contacts

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-425

Primary reviewer: @GPaoloni 

When creating an offline contact, and marking the contact as a non-data call, instead of showing up the close dialog it should display the tabbed form with one tab only (Contact Info)

Screen Shot 2021-01-06 at 13 30 35

### DIFF
--- a/plugin-hrm-form/src/components/tabbedForms/BottomBar.jsx
+++ b/plugin-hrm-form/src/components/tabbedForms/BottomBar.jsx
@@ -19,6 +19,8 @@ import { createCase } from '../../services/CaseService';
 import { saveToHrm } from '../../services/ContactService';
 import { hasTaskControl } from '../../utils/transfer';
 import { namespace, contactFormsBase } from '../../states';
+import { isNonDataCallType } from '../../states/ValidationRules';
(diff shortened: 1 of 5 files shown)


plugin-hrm-form/src/components/tabbedForms/BottomBar.jsx
```diff
@@ -19,6 +19,8 @@ import { createCase } from '../../services/CaseService';
 import { saveToHrm } from '../../services/ContactService';
 import { hasTaskControl } from '../../utils/transfer';
 import { namespace, contactFormsBase } from '../../states';
+import { isNonDataCallType } from '../../states/ValidationRules';
```
GPaoloni, line 22: This file is about to be deprecated I believe, as it's related to the previous form state schema. Should we move this function to a helper file in the `state/contacts` folder? Not suggesting to do that in this PR tho.

GPaoloni (approved):
Excellent! :rocket:

PR #306, opened 2021-01-06
Fixed type error

GPaoloni:
This PR fixes the following error:

Screenshot from 2021-01-06 13-28-30

introduced in https://github.com/tech-matters/flex-plugins/pull/301.


murilovmachado:
Approved! Looks good to me.

PR #309, opened 2021-01-06
Fix date errors when creating offline contact

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-433

Primary reviewer: @GPaoloni 

This fixes points 2 and 3 of the Jira card:
> When creating an offline contact and creating a new case from it, multiple issues related to dates:
> 2. The date of the contact in the Timeline is the current date. But it should be the date entered into the contact in the "Contact Info" screen.
> 3. After clicking on "View" next to the contact in the Timeline, the contact detail shows the current date and time. But it should be the date entered into the contact in the "Contact Info" screen.

PS: The other fixes will come later in a backend PR, but it doesn't block this one from being merged.

### DIFF
--- a/plugin-hrm-form/src/components/case/Timeline.jsx
+++ b/plugin-hrm-form/src/components/case/Timeline.jsx
@@ -31,6 +31,15 @@ import { isConnectedCaseActivity } from './caseHelpers';
 
 const sortActivities = activities => activities.sort((a, b) => b.date.localeCompare(a.date));
 
+const getDateFromNotSavedContact = (task, form) => {
+  if (!task.attributes.isContactlessTask) {
+    return Date.now();
+  }
+
+  const { date: dateString, time } = form.contactlessTask;
+  return new Date(`${dateString}T${time}:00`);
+};
+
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/components/case/Timeline.jsx
```diff
@@ -31,6 +31,15 @@ import { isConnectedCaseActivity } from './caseHelpers';
...
 const sortActivities = activities => activities.sort((a, b) => b.date.localeCompare(a.date));
 
+const getDateFromNotSavedContact = (task, form) => {
+  if (!task.attributes.isContactlessTask) {
+    return Date.now();
+  }
+
+  const { date: dateString, time } = form.contactlessTask;
+  return new Date(`${dateString}T${time}:00`);
+};
```
GPaoloni, line 41: I suggest putting the special case inside the if block, and the general case (`!task.attributes.isContactlessTask`) in the "regular return", just in case we add some more special cases in the future.

GPaoloni (approved):
Great!

PR #312, opened 2021-01-07
More timeline date fixes

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-433 and https://bugs.benetech.org/browse/CHI-461
Backed PR: https://github.com/tech-matters/hrm/pull/84 (merged)

Primary reviewer: @GPaoloni 

This PR fixes point **5** from **CHI-433**:
> When creating an offline contact and creating a new case from it, multiple issues related to dates:
> 2. The date of the contact in the Timeline is the current date. But it should be the date entered into the contact in the "Contact Info" screen.
> 3. After clicking on "View" next to the contact in the Timeline, the contact detail shows the current date and time. But it should be the date entered into the contact in the "Contact Info" screen.
> 5. After going into the detail of a search result, same issues as #2 and #3 above. The detail view for a Contact search result works correctly, but viewing the contact detail when coming through Case search shows the wrong date.

**CHI-433** has also some notes on the search feature, that is not addresed in this PR.
**CHI-461** (referral's date) is fully addressed in this PR.

PS: you may notice the use of **parseISO** instead of using the **Date constructor** in order to avoid timezone issues when converting strings to date.

### DIFF
--- a/plugin-hrm-form/src/components/case/ContactDetailsAdapter.js
+++ b/plugin-hrm-form/src/components/case/ContactDetailsAdapter.js
@@ -35,7 +35,7 @@ export const retrieveCategories = categories => {
 };
 
 export const adaptContactToDetailsScreen = (contact, counselorName) => {
-  const dateTime = contact.createdAt;
+  const dateTime = contact.timeOfContact;
   const name = `${contact.rawJson.childInformation.name.firstName} ${contact.rawJson.childInformation.name.lastName}`;
   const customerNumber = contact.number;
   const { callType, caseInformation } = contact.rawJson;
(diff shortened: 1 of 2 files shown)


GPaoloni (approved):
Great! Just a question: contacts are showing the timeOfContact date in the "View contact" header, but referrals are showing the added date. Is this intentional? Should one of both be a different one?

Screenshot from 2021-01-08 10-09-04

Screenshot from 2021-01-08 10-06-42
Screenshot from 2021-01-08 10-07-10

murilovmachado:
> Great! Just a question: contacts are showing the timeOfContact date in the "View contact" header, but referrals are showing the added date. Is this intentional? Should one of both be a different one?
> 
> Screenshot from 2021-01-08 10-09-04
> 
> Screenshot from 2021-01-08 10-06-42
> Screenshot from 2021-01-08 10-07-10

That's a good question. I agree with you, I think it makes more sense to have the **Added** on contacts header the same as the others. I'll make that change, but I need to put up a very simple PR on the backend so I can get the added date for contact activity. Thanks for spotting this.

andvirga (approved):
Looks great!

PR #314, opened 2021-01-08
Fix 12am

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-466

It was always 12am because at some point we used `date.toLocaleDateString()`, and this function ignores the **time** part of the **dateTime**. Replacing it with `date.toISOString()` solves this problem.

Screen Shot 2021-01-08 at 18 54 56


nickhurlburt (approved):
Looks great!

PR #315, opened 2021-01-08
Fix blank screen when viewing contact

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-465

WIP: Fix tests

The error was happening due to a recent change where I've added a `createdAt` field for a contact activity. So when the activities where coming from backend everything was fine. But in the scenario where we're creating the contact, we fake that activity. This fake activity was missing `createdAt`. 

There were more some adjustments to distinguish between `createdAt` and `timeOfContact`, but the main issue and solution is the one explained above.

### DIFF
--- a/plugin-hrm-form/src/components/case/Timeline.jsx
+++ b/plugin-hrm-form/src/components/case/Timeline.jsx
@@ -60,6 +60,7 @@ const Timeline = ({ status, task, form, caseObj, changeRoute, updateTempInfo, ro
         const { workerSid } = getConfig();
         const connectCaseActivity = {
           date: format(date, 'yyyy-MM-dd HH:mm:ss'),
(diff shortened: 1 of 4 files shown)


plugin-hrm-form/src/components/case/Timeline.jsx
```diff
@@ -82,15 +83,15 @@ const Timeline = ({ status, task, form, caseObj, changeRoute, updateTempInfo, ro
       const info = {
         note: activity.text,
         counselor: twilioWorkerId,
-        date: parseISO(activity.date).toISOString(navigator.language),
```
murilovmachado, line 85: toISOString actually receives no params

nickhurlburt (approved):
Looks great!

PR #318, opened 2021-01-13
Child is at risk checkbox

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-376

Primary reviewer: @andvirga 

Screen Shot 2021-01-13 at 17 05 44


GPaoloni (approved):
Looks good!

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -174,6 +174,14 @@ const Case: React.FC<Props> = props => {
     props.updateCaseStatus(value, props.task.taskSid);
   };
 
+  const onClickChildIsAtRisk = () => {
+    const { connectedCase } = props.connectedCaseState;
+    const { info } = connectedCase;
+    const childIsAtRisk = !Boolean(info && info.childIsAtRisk);
```
andvirga, line 180: I have checked that the logic is working great, but I'm just curious about this line of the code, why did you need to add the logical NOT (!) here?

andvirga (approved):
Looks good!

plugin-hrm-form/src/components/case/Case.tsx
GPaoloni: This is negating the state of the `props.connectedCaseState.info.childIsAtRisk`, that would be the "previous state". I almost ask the same here haha.

nickhurlburt:
We should hold off on merging this for now.  There's a question on whether we want it available for ZM Production or not.  See #aselo.

PR #320, opened 2021-01-14
CHI-474 - Counsellor not saving correctly on Household, Perpetrators …

andvirga:
…and Incidents

Jira related: https://bugs.benetech.org/browse/CHI-474

Primary reviewer: @GPaoloni 

Changes description:
- Used the current counselor logged in to display in Case Details, Add Household, Add Perpetrator, and Add Incidents views.
- Used the counselor that performed the save action to display in View Household, View Perpetrator, and View Incidents views.

### DIFF
--- a/plugin-hrm-form/src/components/case/Case.tsx
+++ b/plugin-hrm-form/src/components/case/Case.tsx
@@ -227,7 +227,9 @@ const Case: React.FC<Props> = props => {
   const name = firstConnectedContact ? getNameFromContact(firstConnectedContact) : getNameFromForm(form);
   const categories = getCategories(firstConnectedContact);
   const { createdAt, updatedAt, twilioWorkerId, status, info } = connectedCase || {};
-  const counselor = counselorsHash[twilioWorkerId];
+  const { workerSid } = getConfig(); // -- Gets the current counselor that is using the application.
+  const caseCounselor = counselorsHash[twilioWorkerId];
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -227,15 +227,17 @@ const Case: React.FC<Props> = props => {
   const name = firstConnectedContact ? getNameFromContact(firstConnectedContact) : getNameFromForm(form);
   const categories = getCategories(firstConnectedContact);
   const { createdAt, updatedAt, twilioWorkerId, status, info } = connectedCase || {};
-  const counselor = counselorsHash[twilioWorkerId];
+  const { workerSid } = getConfig(); // -- Gets the current counselor that is using the application.
+  const caseCounselor = counselorsHash[twilioWorkerId];
```
GPaoloni, line 231: Just out of curiosity, if `twilioWorkerId` does not exists (because falling in the second term of the `||`) in
`const { createdAt, updatedAt, twilioWorkerId, status, info } = connectedCase || {};`
won't that result in an `undefined` value for `caseCounselor`? Or is ensured that there will always be a `connectedCase` when we are here?

GPaoloni (approved):
Looks and works good!

nickhurlburt:
Deployed to Staging and ZM Staging.

PR #323, opened 2021-01-14
Version 0.11: Zambia staging release

nickhurlburt:
This version (which will be released as Version 0.10 to helplines) includes:
- A redesigned Case View including date updated, an editable follow-up date and ability to edit case status
- Ability to add incidents and referrals to a case
- Ability to update a case found from Search
- The new customization framework, allowing contact and case forms to be specified with JSON
- Customization forms specific to Childline Zambia
- Addition of new input types, such as conditional select dropdowns and three-way checkboxes
- Additional Insights logging, and a configurable way to set Insights recording
- API calls now specify the Account ID in the path, supporting a multi-tenant backend

The customization framework introduces data model changes that are not backward-compatible, so searching and displaying past records will not fully work.

No need to review this line-by-line, reviewers are tagged mainly for notification.


nickhurlburt:
With the Zambia staging release now complete, I'll merge this.  I'm not intending to deploy for now.

PR #325, opened 2021-01-14
CHI-387 - Implement the “View” [ ] button to display a case.

andvirga:
Jira related: 
https://bugs.benetech.org/browse/CHI-387

Primary reviewer: @GPaoloni 

Changes description:
- Added logic in the reducer to make sure that when a case is updated, it is updated across all tasks.
- Added logic to update the case in All Cases View.
- Added some types.

### DIFF
--- a/plugin-hrm-form/src/components/caseList/CaseList.tsx
+++ b/plugin-hrm-form/src/components/caseList/CaseList.tsx
@@ -1,10 +1,18 @@
-import React, { useEffect, useReducer } from 'react';
+import React, { useState, useEffect, useReducer } from 'react';
+import PropTypes from 'prop-types';
+import { connect, ConnectedProps } from 'react-redux';
 import { Template } from '@twilio/flex-ui';
 import { CircularProgress, Dialog, DialogContent } from '@material-ui/core';
 
+import Case from '../case';
+import { Case as CaseType } from '../../types/types';
 import CaseListTable from './CaseListTable';
 import { CaseListContainer, CenteredContainer, SomethingWentWrongText } from '../../styles/caseList';
 import { getCases } from '../../services/CaseService';
+import { CaseLayout } from '../../styles/case';
+import * as CaseActions from '../../states/case/actions';
+import { StandaloneSearchContainer } from '../../styles/search';
+import { StandaloneITask } from '../StandaloneSearch';
 
 export const CASES_PER_PAGE = 5;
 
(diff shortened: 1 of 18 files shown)


plugin-hrm-form/src/components/caseList/CaseList.tsx
```diff
@@ -0,0 +1,180 @@
...
+}
+
+const standaloneTask: StandaloneITask = {
+  taskSid: 'standalone-task-sid',
+  attributes: { isContactlessTask: false },
+};
+
+type OwnProps = {
+  setConnectedCase: (currentCase: CaseType, taskSid: string, caseHasBeenEdited: Boolean) => void;
+};
```
GPaoloni, line 76: If you are typing the component like
`type Props = OwnProps & ConnectedProps<typeof connector>;`
you don't need to specify this on `OwnProps`, as `setConnectedCase` will be pulled out from `mapDispatchToProps`'s return (that's what `ConnectedProps<typeof connector>` does). So I think here you can leave `OwnProps = {}`

GPaoloni (approved):
Works great!
One thing I've spotted: If you have a case open in task search, then go to standalone search (or all cases) and update that same case, the old version of it is still open in task search (vice-versa is also true). I don't know if we should update it, or if "closing all open cases that matches with the updated one" is a better approach.

plugin-hrm-form/src/components/caseList/CaseList.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ConnectedProps<typeof connector>;
+
+const CaseList: React.FC<Props> = ({ setConnectedCase }) => {
+  // const [showCaseDetails, setDisplayCase] = useState(false);
```
murilovmachado, line 82: Looks like you can just delete this comment.

murilovmachado (approved):
Looks very good. I only have two comments:
1. This CaseList (All Cases) component doesn't keep its state when navigating to StandaloneSearch or to the main App. It's not related to this PR, so don't worry about it, but it's one little inconsistency in our App.
2. I wouldn't worry about trying to keep everything about Cases in sync, but since you've already implemented it, this is good the way you did it. But there are other harder scenarios that will continue to exist, such as two different counselours editing the same case at the same time. (No need to worry about it)

PR #328, opened 2021-01-15
Remove Show Recent Contacts plugin

nickhurlburt:
Removes this plugin, which is no longer in use.


andvirga (approved):
I've tested it and everything seems to be working without issues.

GPaoloni (approved):

murilovmachado (approved):

PR #329, opened 2021-01-15
[WIP] Security upgrades

nickhurlburt:
Upgrades a number of packages to resolve security vulnerabilities.  As part of this, linting has begun reporting unnecessary exports of `type`s, so types that do not need to be exported no longer are.

### DIFF
--- a/plugin-hrm-form/package-lock.json
+++ b/plugin-hrm-form/package-lock.json
@@ -13,51 +13,133 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.10.4",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.10.4.tgz",
-      "integrity": "sha512-t+rjExOrSVvjQQXNp5zAIYDp00KjdvGl/TpDX5REPr0S9IAIPQMTilcfG6q8c0QFmj9lSTVySV2VTsyggvtNIw==",
-      "requires": {
-        "browserslist": "^4.12.0",
-        "invariant": "^2.2.4",
-        "semver": "^5.5.0"
-      },
-      "dependencies": {
-        "semver": {
-          "version": "5.7.1",
-          "resolved": "https://registry.npmjs.org/semver/-/semver-5.7.1.tgz",
-          "integrity": "sha512-sauaDf/PZdVgrLTNYHRtpXa1iRiKcaebiKQ1BJdpQlWH2lCvexQdX55snPFyK7QzpudqbCI0qXFfOasHdyNDGQ=="
-        }
-      }
+      "version": "7.12.7",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.12.7.tgz",
+      "integrity": "sha512-YaxPMGs/XIWtYqrdEOZOCPsVWfEoriXopnsz3/i7apYPXQ3698UFhS6dVT1KN5qOsWmVgw/FOrmQgpRaZayGsw=="
     },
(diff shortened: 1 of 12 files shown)


nickhurlburt:
@GPaoloni can you please take a quick (no more than 20 minutes) look at the current build error.  It looks like the latest version of `craco-config-flex-plugin` depends on `webpack 4.42.0` exactly, but the recent version of `react-scripts` (which I updated to solve a security issue in one of its dependencies) requires `webpack ^4.44.2`.  This is causing our build to complain.

We have several options for solving this, including just ignoring it for now until `craco-config-flex-plugin` updates.  But I'm curious if you have any quick ideas, and I'm also curious if `react-scripts` or some of the other dependencies listed in `package.json` are even still relevant.

If after a few minutes you're not sure, we can punt on this for now.

GPaoloni:
@nickhurlburt The build error you are receiving is related to React Hook Form types? If so, I've already fixed it (code stashed), but if not, it's not happening in my local.

nickhurlburt:
@GPaoloni you can see the errors in the [test run here](https://github.com/tech-matters/flex-plugins/pull/329/checks?check_run_id=1724501317), also copied below:
```
Run npm test

> plugin-hrm-form@0.0.0 test /home/runner/work/flex-plugins/flex-plugins/plugin-hrm-form
> flex-plugin test --env=jest-environment-jsdom-sixteen


There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "webpack": "4.44.2"

Don't try to install it manually: your package manager does it automatically.
However, a different version of webpack was detected higher up in the tree:

  /home/runner/work/flex-plugins/flex-plugins/plugin-hrm-form/node_modules/webpack (version: 4.42.0) 

Manually installing incompatible versions is known to cause hard-to-debug issues.

If you would prefer to ignore this check, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
That will permanently disable this message but you might encounter other issues.

To fix the dependency tree, try following the steps below in the exact order:

  1. Delete package-lock.json (not package.json!) and/or yarn.lock in your project folder.
  2. Delete node_modules in your project folder.
  3. Remove "webpack" from dependencies and/or devDependencies in the package.json file in your project folder.
  4. Run npm install or yarn, depending on the package manager you use.

In most cases, this should be enough to fix the problem.
If this has not helped, there are a few other things you can try:

  5. If you used npm, install yarn (http://yarnpkg.com/) and repeat the above steps with it instead.
     This may help because npm has known issues with package hoisting which may get resolved in future versions.

  6. Check if /home/runner/work/flex-plugins/flex-plugins/plugin-hrm-form/node_modules/webpack is outside your project directory.
     For example, you might have accidentally installed something in your home folder.

  7. Try running npm ls webpack in your project folder.
     This will tell you which other package (apart from the expected react-scripts) installed webpack.

If nothing else helps, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
That would permanently disable this preflight check in case you want to proceed anyway.

P.S. We know this message is long but please read the steps above :-) We hope you find them helpful!

npm ERR! Test failed.  See above for more details.
Error: Process completed with exit code 1.
```

GPaoloni:
Right, sorry. I couldn't find a way around, except by adding the `SKIP_PREFLIGHT_CHECK=true` flag. which we are already using when running the local environment.

nickhurlburt:
This has been solved via a flex plugins version upgrade.  Closing.

PR #331, opened 2021-01-19
Fix Update Case button enabled/disabled

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-463

Primary reviewer: @andvirga 

After updating the case, the Update button was still enabled. This PR fixes this behavior.


andvirga (approved):
Good job! It's working great

PR #332, opened 2021-01-19
Bug fix: Duplicate contacts created with slow internet

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-38
This PR is related to https://github.com/tech-matters/hrm/pull/91

To test this:
- Make hrm url to be localhost (linked version of hrm, server should be running).
- Fill a contact form and create a new case, but do not submit yet.
- Go to Chrome console, Network tab and add "throttling" (force the internet connection to be very slow).
- Submit the form, once the request has been made, disconnect internet , in order to prevent Twilio from finishing the task.
- Check that the contact and case have been created.
- Reconnect internet and resubmit the form.
- Only one contact should exist, and only one time connected to the created case.


andvirga (approved):
Looks great!

PR #333, opened 2021-01-20
Styling when Case is closed

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-470

Primary reviewer: @andvirga 

For closed cases:
- Remove the dropdown arrow from the case status field
- gray out the '+' signs on each section (e.g. + Referral).

Screen Shot 2021-01-20 at 15 17 52


andvirga (approved):
This looks great! However, I've noticed that the **Child at risk** checkbox is still editable after a case is closed. Not sure if we want to handle that in this task, or if we want to create a new Jira ticket.

GPaoloni (approved):

murilovmachado:
> This looks great! However, I've noticed that the **Child at risk** checkbox is still editable after a case is closed. Not sure if we want to handle that in this task, or if we want to create a new Jira ticket.

Good catch, I'll do that in a separate PR.

PR #336, opened 2021-01-21
Child is at risk should be disabled when case is closed

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-482#

Primary reviewer: @andvirga 

This PR fixes the issue where child is at risk checkbox was still editable after case status was closed.


GPaoloni (approved):
Looks good!

PR #340, opened 2021-01-22
Bug fix: Bug(s) with transferring

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-309

This PR:
> The automated message “Hi, this is the counselor. How can I help you?” was sent again which we definitely do not want. Many of the helplines will transfer a text to a new counselor without letting the child know.

Fixed by only sending the welcome message if the accepted task was NOT transferred.

> It came through to the new counselor back at the Call Type screen rather than on the Child Info screen. (The data is still in the form though.)

Fixed by dispatching the proper route-subroute depending on the selected call type.

### DIFF
--- a/plugin-hrm-form/src/states/DomainConstants.ts
+++ b/plugin-hrm-form/src/states/DomainConstants.ts
@@ -1,4 +1,4 @@
-const callTypes = {
+export const callTypes = {
   child: 'Child calling about self',
   caller: 'Someone calling about a child',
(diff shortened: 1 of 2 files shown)


nickhurlburt (approved):
I have not tested this but the code looks good to me.  If you've tested it, I think it's good to go.

andvirga (approved):
Great job!

PR #343, opened 2021-01-25
CHI-469 - [Case View] - Loading indicator

andvirga:
Jira Related: 
https://bugs.benetech.org/browse/CHI-469

Comments:
- This is the reopening of this previously reverted PR (https://github.com/tech-matters/flex-plugins/pull/338)
- This PR now fixes the issue reported by Dee: when you type into the case summary text box, the page refreshes after each key type.

Changes description:
- Added case id, notes, and referrals as dependencies of `useEffect` (the intention here is to trigger a re-fetch of activities each time that a new note or referral is added, or a new connected case is loaded).
- Moved `useEffect` out from Timeline.tsx to Case.tsx (that component has now the responsibility of showing/hiding the circular layout spinner).
Passing `setLoaded` as callback into `getActivities` API.
Converted Timeline.jsx to Typescript and added some useful types (Activity, NoteActivity, etc.).

### DIFF
--- a/plugin-hrm-form/src/components/case/Case.tsx
+++ b/plugin-hrm-form/src/components/case/Case.tsx
@@ -109,9 +109,9 @@ const Case: React.FC<Props> = props => {
           date: format(date, 'yyyy-MM-dd HH:mm:ss'),
           createdAt: new Date().toISOString(),
           type: task.channelType,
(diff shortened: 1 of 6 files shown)


murilovmachado (approved):
Looks very good!

PR #346, opened 2021-02-01
Bug fix: Gender values not captured in pre-survey

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-504

### DIFF
--- a/plugin-hrm-form/src/utils/prepopulateForm.js
+++ b/plugin-hrm-form/src/utils/prepopulateForm.js
@@ -3,6 +3,10 @@ import { Manager } from '@twilio/flex-ui';
 import { mapAge, mapGender } from './mappers';
 import * as RoutingActions from '../states/routing/actions';
 import { prepopulateFormCaller, prepopulateFormChild } from '../states/contacts/actions';
+import ChildInformationTab from '../formDefinitions/tabbedForms/ChildInformationTab.json';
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/utils/prepopulateForm.js
```diff
@@ -3,6 +3,10 @@ import { Manager } from '@twilio/flex-ui';
 import { mapAge, mapGender } from './mappers';
 import * as RoutingActions from '../states/routing/actions';
 import { prepopulateFormCaller, prepopulateFormChild } from '../states/contacts/actions';
+import ChildInformationTab from '../formDefinitions/tabbedForms/ChildInformationTab.json';
+import CallerInformationTab from '../formDefinitions/tabbedForms/CallerInformationTab.json';
+
+const getGenderOptions = definition => definition.find(e => e.name === 'gender').options.map(e => e.value) || [];
```
GPaoloni, line 9: This assumes that `gender` is a `select` type field. Should we care about it being a different one?

nickhurlburt (approved):
Looks good, I left some comments.  If you want to try to convert `mappers.js` to TS, it is worth a try, though I'd argue for doing that first as a separate PR, then following with this one.  However converting everything in here to TS may end up being harder than it appears.

murilovmachado (approved):

PR #348, opened 2021-02-02
[Don't merge] - React PDF Print Document Proof of Concept

andvirga:
Jira Related: https://bugs.benetech.org/browse/CHI-501

Description:
This PR contains a POC of a printed document using react-pdf library. (https://react-pdf.org/)
Please feel free to add comments/suggestions/concerns.

**ReactPDF Pros:**
- Provides a nice API with a well-defined set of components to use (`Page`, `View`, `Text`, `Image`, etc).
- Automatically handles page and text wrapping.
- Generates a **vectorized** pdf (the text can be highlighted and copied).
- Provides a PDF Viewer with download, zoom, and print functionalities.

**ReactPDF Cons:**
- Native components (such as checkboxes, dropdowns, etc.) cannot be used.
- Using custom fonts (such as Open Sans) is not trivial and requires an extra amount of work.
- Some flex instructions (such as FlexBasis doesn't work in the same way that in HTML) (need to investigate more about this).


plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
```diff
@@ -0,0 +1,165 @@
...
+  ],
+};
+
+const notesSection = {
+  sectionName: 'Notes',
+  fieldValues: [
+    { label: 'Field One', value: 'Yes' },
+    { label: 'Field Two', value: 'No' },
+  ],
+};
```
nickhurlburt, line 98: In the mockup, there are multiple notes.  We will need to handle that.

And one thing that is missing from the mockups is that there is a possibility that we could have multiple household members, perpetrators, referrals or incidents.  The mockup needs to be updated to allow this.  Though it is also good to think about how to handle an arbitrary number of each of these items.

plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
```diff
@@ -0,0 +1,165 @@
...
+    <CasePrintViewContainer>
+      <ButtonBase onClick={onClickClose} style={{ marginLeft: 'auto' }} data-testid="CasePrint-CloseCross">
+        <HiddenText>
+          <Template code="Case-CloseButton" />
+        </HiddenText>
+        <Close />
+      </ButtonBase>
+      <PDFViewer style={{ height: '100%' }}>
+        <Document>
+          <Page size="A4" style={styles.page}>
```
nickhurlburt, line 114: In the mockup there is a header on each page... is that possible to implement?

plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
```diff
@@ -0,0 +1,165 @@
...
+});
+
+const callerInfoSection = {
+  sectionName: 'Caller Information',
+  fieldValues: [
+    { label: 'Field One', value: 'Yes' },
+    { label: 'Field Two', value: 'No' },
+    { label: 'Field Three', value: 'Yes' },
+    { label: 'Field Four', value: 'Yes' },
+    { label: 'Field Five', value: 'No' },
```
nickhurlburt, line 34: Ideally we would re-use methods that create the `ContactDetails` to grab the values.

plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
```diff
@@ -0,0 +1,165 @@
...
+            <View>
+              <Text style={styles.caseDetailsLabel}>Case Details</Text>
+              <View style={styles.caseDetailsSection}>
+                <View style={styles.caseDetailsItem}>
+                  <Text>Case Status</Text>
+                  <Text style={{ marginTop: 5 }}>OPEN</Text>
+                </View>
+                <View style={styles.caseDetailsItem}>
+                  <Text>Opened</Text>
+                  <Text style={{ marginTop: 5 }}>11/10/2020</Text>
```
nickhurlburt, line 131: Dates should be formatted based on the browser locale.  If that's not possible, default to day/month/year.

nickhurlburt (commented):
This is a great start.  I'm interested to see what this looks like when we use real values rather than hard-coded ones, and when we need to use the FormDefinitions to populate this.  When we need to present a different view on our model, it may surface any design flaws of the model.

andvirga:
I've fixed some styles, added a Header, and finally, I was able to include Open Sans font. At this point I'm pretty confident with this tool, I believe that it will fit our PDF requirements (Even with multiple notes, households, etc.)

I'll attach an example in the engineering channel (GitHub doesn't allow me to upload PDF)

andvirga:
This was a PoC, closing this without merging it

PR #352, opened 2021-02-09
[ZA Forms Customization] - Add South Africa Form definitions

andvirga:
Jira related: 
https://bugs.benetech.org/browse/CHI-517

Primary Reviewer: @GPaoloni 

Changes description: 
- Added ZA form field definitions created by Dee.
- Changed the `tabbedForms` structure to allow ZA and ZM definitons.
- Changed the references to use ZA form definitions by default (instead of ZM).
- Fixed unit tests.


nickhurlburt (commented):
This is okay as a first step, though I would like to understand your plan for how we:
- Set the live definition version in configuration
- Know where to fetch the right JSON files for that definition version (instead of hard-coding them here)
- Support reading old versions, including multiple versions for ZA (eg, 'za_001', 'za_002', etc

Is the directory structure you're adding here part of a bigger plan, or will we need to change the directory structure on the next PR?

andvirga:
> This is okay as a first step, though I would like to understand your plan for how we:
> 
> * Set the live definition version in configuration
> * Know where to fetch the right JSON files for that definition version (instead of hard-coding them here)
> * Support reading old versions, including multiple versions for ZA (eg, 'za_001', 'za_002', etc
> 
> Is the directory structure you're adding here part of a bigger plan, or will we need to change the directory structure on the next PR?

Yes, this is a first draft, and we shouldn't merge this until the changes that @GPaoloni  is currently working on are finished. I've only pushed this to share the work with him.

Regarding your questions:
1) Set the live definition version in the configuration.
This configuration will be stored in `HrmFormPlugin.js` and it will be different for each environment (ZA, ZM; etc.).
Now it is like:
`const definitionVersion = 'v1'; // will be moved to service configuration later on`
And it will be something like this:
`const { definitionVersion } = manager.serviceConfiguration.attributes;`

2) Know where to fetch the right JSON files for that definition version (instead of hard-coding them here)
For now, JSON files are going to be bundled within the code, and we'll expose a service that given a version name as a parameter it will mimic the fetching of such version. To be clear: JSON files will be in a folder structure in the code. That saves us from creating a new API to provision these form definitions.

3) Support reading old versions, including multiple versions for ZA (eg, 'za_001', 'za_002', etc)
That's part of the work that Gian is currently doing in the reducer. As a comment, there is a big chance that I would need to make some changes to this PR as well, once Gian's work is done.

Please let me know if you still have any questions!

nickhurlburt:
Thanks, Andres, sounds good.  It makes sense that this is connected to the other work that Gian is doing.

> For now, JSON files are going to be bundled within the code, and we'll expose a service that given a version name as a parameter it will mimic the fetching of such version. To be clear: JSON files will be in a folder structure in the code. That saves us from creating a new API to provision these form definitions.

Sure.  For the purpose of my question, whether the files are in a folder structure in the code vs grabbed from Twilio Assets or S3 isn't the important thing.  The current path has no way to grab a specific version, it just divides by country.  I'm looking forward to a way to select based on version.

andvirga:
Closing this PR as it was included in #354

PR #353, opened 2021-02-09
Create reducer to get form defs for a helpline.

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-518

Notes:
- This PR is easier to review commit by commit.
- Only one version for the forms is being used in this PR.

This PR creates a Redux structure to hold the form definitions (dropped inside configuration reducer, but can be moved to it's own if preferred). 
Two new properties are added to the reducer:
```
formsVersions: { [version: string]: FormsVersion | undefined };
currentDefinitionVersion: FormsVersion | undefined;
```
where `formsVersions` is mapping version names to the corresponding definition, and `currentDefinitionVersion` is just for easy access to the currently active definition (the one that will be used to store new forms). The later one can be removed but in that case, every component that needs the information of a form based on the current version will also need to know which is such a version (i.e. call `getConfig()` helper to get `definitionVersion` property).

A `getFormsVersions` method is introduced to be used in those places where we need a form definition but is not a React component (i.e. function calls), like bot memory population helper and contact services.

The JSON files for the `v1` version are grouped in a folder and exposed from a JS file, in order to save us the tedious task of importing one by one all of the different form definitions. This version is exposed to the application from `getFormsVersion` service, that is mimicking the fetch of a specified version (once this are living in a remote location instead than bundled within the code).

Some useful comments related to the naming:
- A `FormDefinition` is the definition for a single form (e.g. child tab form, add referral form, etc).
- A `DefinitionVersion` is the set of all the `FormDefinition`s required for a single version.
- `definitionVersion` is used to denote one single `DefinitionVersion`.
- `definitionVersions` is used to denote the set of `DefinitionVersion` coexisting in the app.
~Please feel free to suggest changes in the names while you review the PR, as this was developed progressively and some things may not make sense anymore.~
Renamed FormsVersion to DefinitionVersion.

The proposed structure here does not contemplates the fact that different helplines might share version names. I have two proposal to solve this issue and would like to hear your opinions:
- Each helpline should add it's unique identifier (ZM/SA/etc) to the version name, in order to "scope" properly. Then a contact that has `definitionVersion = 'v1'` and `helpline = 'ZM'`, will be named as the `'ZM-v1'` definition (just an example).
- Deeper nesting in the global state holding the definitions: instead than 
`formsVersions: { [version: string]: FormsVersion | undefined };`
we could have something like
`formsVersions: { [helpline: string]: { [version: string]: FormsVersion | undefined }};`
and same as above item, use `helpline` alongside `definitionVersion` to access the proper one.

WIP:
- ~CI~
- ~Add tests~
- Tech debt (you will see some TODOs in the code)
- Better error handling in the case that the app is not able to fetch a forms version (suggestions welcomed here)


GPaoloni:
@nickhurlburt 
> It's this sentence that caused me to write the above comment. What is "helpline" here? That helpline attribute, or an account, or the idea of an organization? I would be in favor of having the definitionVersion stored in rawJson be all we need to know how to read and display the data inside rawJson, rather than having to combine other data (such as from a helpline attribute). But I am open to argument on this.

About thihs, I refer to `helpline` as "that helpline attribute" that we are storing in our DB and comming from service configuration. This is not a 1-1 relation to real world helplines, Lifeline vs Helpline can have different values here, and that attribute could be used to scope. Just an alternative tho.

> I would be in favor of having the definitionVersion stored in rawJson be all we need to know how to read and display the data inside rawJson.

I agree with this

plugin-hrm-form/src/states/configuration/actions.ts
```diff
@@ -1,4 +1,5 @@
 import * as t from './types';
+import type { FormsVersion } from '../../components/common/forms/types';
```
andvirga, line 2: I agree with NIck on this item. I would prefer another name, such as: 
`Configuration` or `FormsCollection`

andvirga: In addition, I'd recommend avoiding using the "version" word on this variable. Since that word only refers to the version itself (v1, v2, zav1, zav2, etc. ).
These could be some examples:

- ConfigCollection
- FormsCollection
- FormsSet
- ConfigSet
- FieldDefinitions

plugin-hrm-form/src/services/ServerlessService.ts
```diff
@@ -100,3 +101,9 @@ export const sendSystemMessage = async (body: { taskSid: ITask['taskSid']; messa
 
   return response;
 };
+
+/**
+ * Function that mimics the fetching of a version definition for all the forms used within the app.
+ * Later on this will be fetched in async way.
+ */
+export const getFormsVersion = async (version: string) => formsVersions[version];
```
andvirga, line 109: Since this function is not performing an async operation yet, (don't have an await) I'm not sure if we should leave that declaration. Does `eslint` throw any warning?

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -173,6 +185,9 @@ const setUpComponents = setupObject => {
 const setUpActions = setupObject => {
   const { featureFlags } = setupObject;
 
+  // Is this the correct place for this call?
```
andvirga, line 188: To answer your comment, I think it is the right place. as long as the current `FormsVersion` gets loaded here it makes sense to me, leaving this method right here.

plugin-hrm-form/src/services/ServerlessService.ts
GPaoloni: I'm intentionally wrapping the result int a Promise (that's what `async` actually does). When we start fetching this from a remote location, any call to `getFormsVersion` won't even notice, as they are already using it that way (`await`ing it).

nickhurlburt:
> About thihs, I refer to helpline as "that helpline attribute" that we are storing in our DB and comming from service configuration. This is not a 1-1 relation to real world helplines, Lifeline vs Helpline can have different values here, and that attribute could be used to scope. Just an alternative tho.

Sure, makes sense.  So the helpline attribute is something that _might_ affect which forms we use later, but it won't at first.  And we don't know _how_ it will affect it... Dee is doing some work on that now.  So it would be reasonable for now to either 1) ignore it or 2) leave ourselves a path toward later building to work with it.  Designing to do much with it now feels premature.

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -135,6 +136,23 @@ const Case: React.FC<Props> = props => {
...
+    const fetchFormsVersions = async (v: string) => {
+      const formsVersion = await getFormsVersion(version);
+      updateFormsVersion(version, formsVersion);
+    };
+
+    if (version && !formsVersions[version]) {
+      fetchFormsVersions(version);
+    }
+  }, [formsVersions, updateFormsVersion, version]);
+
```
andvirga, line 155: It's not your fault, but this component is growing too much, having more than useEffect is dangerous, we need to refactor this ASAP. I would create a ticket for this.

andvirga: Ticket created: https://bugs.benetech.org/browse/CHI-523

GPaoloni: I disagree with the multiple `useEffect`s statement ([my reasons](https://reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns)), but agreed that this component is huge. A refactor seems like a good idea!

andvirga: Yes, maybe I didn't express my idea correctly, as you stated, using multiple effects to separate concerns is a good practice. However, we should verify if all these effects are really necessary (which seems to be in this case), or if they can be solved with a different strategy.

andvirga (approved):
Tested the functionality and everything is working as expected (With ZM forms).
The next step will be to integrate ZA Form Definitions.

murilovmachado (approved):
Very good PR, Gian! Learned a lot by reading it.

plugin-hrm-form/src/formDefinitions/index.js
```diff
@@ -0,0 +1,5 @@
+import v1 from './v1';
+
+export default {
```
nickhurlburt, line 3: So whenever we add a new DefinitionVersion (for example, for South Africa), we'll add a directory here, and then import it and add to this object, right?

nickhurlburt (approved):
I can't say I completely understand everything that's going on, but I do understand the overall architecture and it looks sound.  Let's do it!

plugin-hrm-form/src/formDefinitions/index.js
GPaoloni: Right! You can see how I'm adding it in [the follow up PR](https://github.com/tech-matters/flex-plugins/pull/354). That's until we want the `getDefinitionVersion` service to actually fetch something from an remote location.

PR #354, opened 2021-02-12
Add South Africa Form definitions

GPaoloni:
This PR 
- Uses the structure built in https://github.com/tech-matters/flex-plugins/pull/353
- Adds ZA definitions from https://github.com/tech-matters/flex-plugins/pull/352
- Adds a handy helper for fetching multiple definitions.
- Fixes a bug: after a search, if a case or a contact has a definition that is not loaded in the global state, we want to fetch it and populate the state before actually showing any search results. This is handled inside the "thunks" that takes care of fetching contacts and cases.

~WIP:~
- ~Do the same for cases list as for search.~
- ~Check errors that are appearing in the console.~
- Error handling.

### DIFF
--- a/plugin-hrm-form/src/formDefinitions/index.js
+++ b/plugin-hrm-form/src/formDefinitions/index.js
@@ -1,5 +1,7 @@
 import v1 from './v1';
+import zaV1 from './za-v1';
 
 export default {
   v1,
+  'za-v1': zaV1,
 };
--- a/plugin-hrm-form/src/formDefinitions/index.js
+++ b/plugin-hrm-form/src/formDefinitions/index.js
@@ -2,6 +2,6 @@ import v1 from './v1';
 import zaV1 from './za-v1';
 
 export default {
-  v1,
-  'za-v1': zaV1,
+  v1, // Zambia V1
+  'za-v1': zaV1, // South Africa v1
 };
--- a/plugin-hrm-form/src/states/search/actions.ts
+++ b/plugin-hrm-form/src/states/search/actions.ts
@@ -8,9 +8,8 @@ import { searchContacts as searchContactsApiCall } from '../../services/ContactS
 import { searchCases as searchCasesApiCall } from '../../services/CaseService';
 import { ContactDetailsSectionsType } from '../../components/common/ContactDetails';
 import { addDetails } from './helpers';
-import { getDefinitionVersions } from '../../HrmFormPlugin';
-import { getMissingDefinitionVersions } from '../../services/ServerlessService';
 import { updateDefinitionVersion } from '../configuration/actions';
+import { getContactsMissingVersions, getCasesMissingVersions } from '../../utils/definitionVersions';
 
 // Action creators
 export const handleSearchFormChange = (taskId: string) => <K extends keyof t.SearchFormValues>(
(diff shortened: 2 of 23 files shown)


plugin-hrm-form/src/states/search/actions.ts
```diff
@@ -50,6 +67,21 @@ export const searchCases = (dispatch: Dispatch<any>) => (taskId: string) => asyn
...
+      searchResult.cases.reduce(
+        (accum, c) =>
+          definitionVersions[c.info.definitionVersion] ? accum : { ...accum, [c.info.definitionVersion]: true },
+        {},
+      ),
+    );
+
+    // Batch all the missing ones to global state (if any)
+    const definitions = await getMissingDefinitionVersions(missingDefinitionVersions);
+    definitions.forEach(d => dispatch(updateDefinitionVersion(d.version, d.definition)));
```
murilovmachado, line 83: I believe I've seen this exact same code three times in this PR. Could we extract this code and make it reusable?

murilovmachado (approved):
Good to see more things being typed. Looks good to me, left just one note.

plugin-hrm-form/src/formDefinitions/index.js
```diff
@@ -1,5 +1,7 @@
 import v1 from './v1';
+import zaV1 from './za-v1';
 
 export default {
   v1,
+  'za-v1': zaV1,
```
andvirga, line 6: Can we add a comment here just to clarify that v1 is in fact 'Zambia V1'?

plugin-hrm-form/src/states/search/actions.ts
```diff
@@ -35,6 +38,20 @@ export const searchContacts = (dispatch: Dispatch<any>) => (taskId: string) => a
...
 
+    // Look for not loaded definitionVersions
+    const { definitionVersions } = getDefinitionVersions();
+    const missingDefinitionVersions = Object.keys(
+      searchResultRaw.contacts.reduce(
+        (accum, c) =>
+          definitionVersions[c.details.definitionVersion] ? accum : { ...accum, [c.details.definitionVersion]: true },
+        {},
+      ),
+    );
```
andvirga, line 49: This logic is repeated, maybe it would be better to define it in a helper and call it across different places.

plugin-hrm-form/src/states/search/actions.ts
andvirga: Same comment as Murilo! If it isn't too hard it would be great to extract that code

andvirga (approved):
Overall code looks great! Only left one suggestion!

plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -268,7 +268,7 @@ const saveInsights = async payload => {
   const contactForm = getStateContactForms(taskSid);
   const caseForm = getStateCaseForms(taskSid);
 
-  await saveInsightsData(payload.task, contactForm, caseForm, { useZambiaInsights: true });
+  await saveInsightsData(payload.task, contactForm, caseForm, {}); // TODO: build new insight configs for ZA
```
GPaoloni, line 271: @nickhurlburt I've just commented this out, but will break Zambia if we deploy it. Tagging you so we can decide how to handle the behavior here.

nickhurlburt: The error that appears is `Bad mixed checkbox value [undefined], defaulting to null for Insights value`.  I added this as a fail-safe to make sure we were correctly marking mixed checkbox fields as the correct type, since they get [special handling](https://github.com/tech-matters/flex-plugins/blob/master/plugin-hrm-form/src/services/InsightsService.ts#L116).  Looking over the South Africa fields, they don't have the mixed checkboxes `livingInConflictZone`, `livingInPoverty`, `childWithDisability`, and `repeatCaller`.  I'm checking with Dee to make sure that is desired.  Commenting out those fields removes errors.

I would actually be fine with keeping `useZambiaInsights: true` in for now and living with the errors.  The default behavior is sensible (it defaults to `null`, meaning "Unknown").  And getting this onto Development is blocking other things.

But we should fix this before we release.  Part of that depends on what Dee says (whether these fields should be present after all) and part of it depends on bringing Insights configuration into the same model as the new customization version support you've added.

PR #360, opened 2021-02-19
[Ignore] Counselor dropdown for contactless task

GPaoloni:
This PR is issued to show the diff as some of this code will be used in a different one.


GPaoloni:
Closed as this is included in https://github.com/tech-matters/flex-plugins/pull/361.

PR #362, opened 2021-02-23
Replaced deploy steps in Actions

fgonzalez-techmatters:
Replacing the deploy steps in GitHub actions, due new version of the Flex-plugins.


andvirga (approved):
Looks great!

PR #363, opened 2021-02-23
Added Changelog input and corrected env variables

fgonzalez-techmatters:
New input to add a change log description to the deploy process step. Corrected the naming convention for South Africa and Zambia account variables.


andvirga (approved):
Great work! The changelog description will be useful to identify the version that we have released in Twilio.

PR #364, opened 2021-02-24
Changes to support ZA insights

nickhurlburt:
@GPaoloni this WIP PR is a companion to CHI-534.  It shows what fields will need to be changed to support insights for ZA-v1.  In order to make the necessary changes to support ZA insights in the short term, we could either:

1. Use this PR and add feature flag support for `useZambiaInsights` and `useSouthAfricaInsights`
2. Implement a more robust approach per CHI-534

We should certainly do #2 in the longer term, but could do #1 in the short term.

I've added some extra comments inside to show what the actual diff between the two insights configurations are.


plugin-hrm-form/src/insightsConfig/za-v1.ts
```diff
@@ -0,0 +1,163 @@
...
+        type: FieldType.MixedCheckbox,
+      },
+      {
+        name: 'childOnTheMove',
+        insights: [InsightsObject.Customers, 'email'],
+      },
+      {
+        name: 'childWithDisability',
+        insights: [InsightsObject.Customers, 'customer_attribute_3'],
+      },
```
nickhurlburt, line 87: The only thing different in this file vs `zambia.ts` is that the `mixed-checkbox` type has been removed.  `childWithDisability` is a dropdown for ZA.

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -259,6 +260,29 @@ export const zambiaUpdates = (
...
+
+  const attsToReturn: InsightsAttributes = processHelplineConfig(contactForm, caseForm, southAfricaV1InsightsConfig);
+
+  /*
+   * Custom additions:
+   *  Add province, municipality district into area
+   */
+  attsToReturn[InsightsObject.Customers].area = [
+    contactForm.childInformation.province,
+    contactForm.childInformation.municipality,
```
nickhurlburt, line 279: The only difference between this function and `zambiaUpdates` is the addition of the `municipality` field.

plugin-hrm-form/src/insightsConfig/za-v1.ts
```diff
@@ -0,0 +1,163 @@
...
+        type: FieldType.MixedCheckbox,
+      },
+      {
+        name: 'memberOfAnEthnic',
+        insights: [InsightsObject.Conversations, 'conversation_measure_4'],
+        type: FieldType.MixedCheckbox,
+      },
+      {
+        name: 'childOnTheMove',
+        insights: [InsightsObject.Customers, 'email'],
```
GPaoloni, line 82: This is resulting on a bad "mixed" value as `childOnTheMove` is 3-way checkbox in the form. Should we revisit this?

plugin-hrm-form/src/insightsConfig/za-v1.ts
GPaoloni: It seems like there are a few fields that are not part of the ZA form definitions, and trying to pull them results in `"Bad mixed checkbox value [${v}], defaulting to null for Insights value"`. Should we remove those?

GPaoloni:
@nickhurlburt I've pushed to this branch as I've done some changes but trying to reuse as much of your work as possible. You can checkout  to [this commit](https://github.com/tech-matters/flex-plugins/pull/364/commits/429ad87107cbfde189000f8aeb82f1d4cb9df827) to see that the code I've added results in the same as `southAfricaUpdates` function (after a task is completed, you can search for '>>>>>>>' in the console, and both objects should be printed there). I'll write tests to ensure this tho.
In the future I think we can merge the approaches and do all of insights customizations using a single file. What the approach I'm adding allows us is that we can grab more information that may not come from the forms, and we can combine case and contact forms values into the same attributes, (something like "in attribute X, add contact.childInfo.name and case.info.perpetrator.name" if we ever want to). The last commit should have everything in place but I may be mistaken, so please do a deep review on this one. Cleanup pending, and naming corrections/conventions are as usual, welcome!

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -236,30 +248,47 @@ export const processHelplineConfig = (
...
+  attributes: InsightsAttributes,
+  customUpdate: InsightsCustomUpdate,
+) => {
+  // concatenate the values, taken from dataSource using paths (e.g. 'contactForm.childInformation.province')
+  const value = customUpdate.paths
+    .map(path => get(dataSource, path))
+    .filter(Boolean)
+    .join(delimiter);
+
+  // if (!value) return attributes; // Do we want this case to leave insights unaffected?
```
GPaoloni, line 261: What's the desired behavior if all the paths results in "falsy" values (and thus `value === ''`)?

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -236,30 +248,47 @@ export const processHelplineConfig = (
...
 
-// Visible for testing
-export const zambiaUpdates = (
-  attributes: TaskAttributes,
-  contactForm: TaskEntry,
-  caseForm: Case,
-): InsightsAttributes => {
+const applyCustomUpdate = (dataSource: { contactForm: TaskEntry; caseForm: Case }) => (
+  attributes: InsightsAttributes,
+  customUpdate: InsightsCustomUpdate,
```
GPaoloni, line 253: The idea of this function is to bind a `dataSource` where we drop all the possible objects to grab values from, so we can get paths from `contactForm` and `caseForm` in the same attribute (or some other objects that we may add in the future). Once this `dataSource` is binded the returned function is of the form `(attributes: InsightsAttributes, customUpdate: InsightsCustomUpdate) => InsightsAttributes` so we can use it in a `reduce`.

plugin-hrm-form/src/insightsConfig/za-v1.ts
nickhurlburt: Oh -- it looks like `childOnTheMove` has been changed to a [select type](https://github.com/tech-matters/flex-plugins/blob/staging/plugin-hrm-form/src/formDefinitions/za-v1/tabbedForms/ChildInformationTab.json#L1313).  So we should just remove the `type: mixed-checkbox` here but keep reporting to insights.  (When the type is a select or input, no `type` field is needed here.)

plugin-hrm-form/src/insightsConfig/types.ts
```diff
@@ -16,3 +16,10 @@ export type InsightsConfigSpec = {
   contactForm?: InsightsFormSpec;
   caseForm?: InsightsFormSpec;
 };
+
+export type InsightsCustomUpdate = {
+  insightObject: InsightsObject; // In which attributes object this goes
```
nickhurlburt, line 21: I have a preference for calling this `insightsObject` with an "s".  If you have a reason to go with `insightObject`, that is fine with me, though.

plugin-hrm-form/src/insightsConfig/za-v1.ts
```diff
@@ -0,0 +1,164 @@
+import { FieldType, InsightsConfigSpec, InsightsObject } from './types';
+
+// Should we remove this file? This information is now part of definition version
```
nickhurlburt, line 3: Yes, we can remove.

plugin-hrm-form/src/insightsConfig/za-v1.ts
nickhurlburt: The newest updates of za-v1 should have all those fields present.  This branch may have been branched off of an older version of staging that doesn't have them.  Are there any fields that are still missing here after the newer updates of the definition version?

plugin-hrm-form/src/services/InsightsService.ts
nickhurlburt: That shouldn't happen with current helplines because the values are all required.  But if it _did_ happen, I think we would want to leave it as is and send, for example, ";;".  There are insights reports that use ";" to pull them back apart, so having them get pulled apart and having '' for each would make sense to me.

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -236,30 +248,47 @@ export const processHelplineConfig = (
...
+  // concatenate the values, taken from dataSource using paths (e.g. 'contactForm.childInformation.province')
+  const value = customUpdate.paths
+    .map(path => get(dataSource, path))
+    .filter(Boolean)
+    .join(delimiter);
+
+  // if (!value) return attributes; // Do we want this case to leave insights unaffected?
+
+  return {
+    ...attributes,
```
nickhurlburt, line 264: We are re-doing the work of `mergeAttributes` here... could we just return the new attributes from the `customUpdate` and let the machinery above us do the rest of the work?

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -236,30 +248,47 @@ export const processHelplineConfig = (
...
-   *  Add province and district into area
-   */
-  attsToReturn[InsightsObject.Customers].area = [
-    contactForm.childInformation.province,
-    contactForm.childInformation.district,
-  ].join(delimiter);
+  const customUpdatesFun = applyCustomUpdate(dataSource);
 
-  return attsToReturn;
+  return customConfigObject.customUpdates.reduce(customUpdatesFun, processedAtts);
```
nickhurlburt, line 285: This structure seems unnecessarily confusing.  In `saveInsightsData` we have a process by which we handle all the separate little bits to add to Insights and then reduce them all together.  That means that each separate piece of that can be kept simple.  But here we are adding an extra `reduce` for the two different things coming from the `definitionVersion`.  This adds another level of complexity, and it's hard to understand. 

Perhaps in `getInsightsUpdateFunctionsForConfig` we could return functions for processing the `configSpec` and the `customAttributes` back separately as part of the bigger array?  That seems simpler.  Maybe I have overlooked something -- please tell me if so.

nickhurlburt (commented):
Since I started this PR I can't click 'Approve', but consider this approved.  It's a good idea to just have two different specs for insights config that map opposite directions!  I added an idea for simplying `InsightsService`, but otherwise I'd be fine with getting this onto ZA Staging.  Thanks!

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -236,30 +248,40 @@ export const processHelplineConfig = (
...
 
-// Visible for testing
-export const zambiaUpdates = (
-  attributes: TaskAttributes,
-  contactForm: TaskEntry,
-  caseForm: Case,
-): InsightsAttributes => {
-  const { callType } = contactForm;
-  if (isNonDataCallType(callType)) return {};
+const applyCustomUpdate = (customUpdate: InsightsCustomUpdate): InsightsUpdateFunction => {
```
nickhurlburt, line 251: Yes, good to see we are no longer passing in the other attributes here

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -236,30 +248,40 @@ export const processHelplineConfig = (
...
-   *  Add province and district into area
-   */
-  attsToReturn[InsightsObject.Customers].area = [
-    contactForm.childInformation.province,
-    contactForm.childInformation.district,
-  ].join(delimiter);
+  const customUpdatesFuns = customConfigObject.customUpdates.map(applyCustomUpdate);
 
-  return attsToReturn;
+  return [getProcessedAtts, ...customUpdatesFuns];
```
nickhurlburt, line 278: Good to see we are no longer doing a `reduce` here, and letting the main `reduce` do the work.

plugin-hrm-form/src/components/common/forms/types.ts
```diff
@@ -163,4 +164,8 @@ export type DefinitionVersion = {
   };
   callTypeButtons: CallTypeButtonsDefinitions;
   layoutVersion: LayoutVersion;
+  insights: {
+    configSpec: InsightsConfigSpec;
+    customUpdates: InsightsCustomUpdates;
```
nickhurlburt, line 169: I am not sure about these names, but unfortunately I don't have better ideas.  It feels like the `configSpec` is a function mapping form fields to insights fields, and `customUpdates` is a function mapping insights fields to form fields.  It's just two different ways to define how the form and insights fields relate.  In a sense they are both configuration specifications, and they are both custom updates.

If you have a good idea on renaming, go with it.  If you don't, we can leave it as is, but please add some more comments so the next person who looks at this (maybe you or me in a few months) can tell what's going on.

nickhurlburt (commented):
Looks good.  I haven't done anything to test this... please take some time to do that.  Otherwise good to go.

plugin-hrm-form/src/components/common/forms/types.ts
GPaoloni: I though the same while coding this. What about `FieldToInsights` and `FieldsToInsights` (or `PathsToInsights`) as a way to describe the intentions? This is because configSpec maps one field to one attribute, while customUpdates can map several fields to a single attribute. Does that makes more sense?

nickhurlburt: `OneToOneConfigSpec` and `OneToManyConfigSpec`?  Just `FieldToInsights` and `FieldsToInsights` is a little too subtle, I think.

PR #366, opened 2021-02-24
Fixed Non-data contacts dialog size is too small

andvirga:
Jira: https://bugs.benetech.org/browse/CHI-544

Primary reviewer: @GPaoloni 

Description:
Added `classes = { paper: 'paper' }` to CloseTaskDialog, with proper ts typing.

CloseTaskDialog now looks like the previous version:

image


GPaoloni (approved):
Great, thanks!

PR #371, opened 2021-03-02
[Prototype] Addressed frontend changes

GPaoloni:
This PR is a continuation of https://github.com/tech-matters/flex-plugins/pull/361.

You can point to dev env for serverless in order to test this.

This PR:
- Adds `addOfflineContact` boolean property to routing reducer to tell the app that the user is in the process of completing a contacttless task form (even if switch between tasks, state is preserved). The offline contact forms are shown if a) there is no selected Twilio task, and b) `addOfflineContact` is set to `true`.
- "Add Offline Task" button now dispatches a `recreateTaskState` action, to reuse all of our existing reducers logic. "Cancel Offline Task" simply dispatches a `removeTaskState`
- Rework on our TaskView components tree. Now only one single place is using Twilio's `withTaskContext` helper, and from there and based on above routing information, renders the proper forms. A fake Task (`CustomITask`) is passed down the component tree if the user is filling an offline task.
- `assignOfflineContact` serverless function is used to create-assing-complete a new Twilio task in behalf of someone else, but now we use the taskSid of such a task as the unique identifier for the form being submited by the user. Still have some gotchas on this, as if a step fails here, we can end up with inconsistent states between Insights and HRM.
- The form being submitted to HRM is still stored as if the issuer was the creator. We need to decide which approach to take in order to resolve the search issues mentioned in previous PR.
> The twilioWorkerId column of contact stored in the backend is the one of the issuer, not the target counselor. This causes the contact to appear in behalf of the issuer and not the target counselor. For this to be fixed we can either use the target counselor's sid in twilioWorkerId (but then the information that this task was submitted by issuer is lost, unless saved inside the raw json or a new column) or change the search query when contacts are offline contacts (this may slow down the query and increase it's complexity). Same happens for cases.


plugin-hrm-form/src/components/ContactDetails.tsx
```diff
@@ -20,7 +20,6 @@ import { namespace, configurationBase, RootState } from '../states';
 import * as ConfigActions from '../states/configuration/actions';
 import { getDefinitionVersion } from '../services/ServerlessService';
 
-// TODO: complete this type
 type OwnProps = {
   contact: any;
   detailsExpanded: any;
```
andvirga, line 25: Can we define these types? If it's too complex ignore my comment.

plugin-hrm-form/src/components/CustomCRMContainer.tsx
```diff
@@ -28,22 +38,28 @@ const CustomCRMContainer: React.FC<Props> = ({ tasks, dispatch }) => {
...
+      {renderOfflineContactTask && <TaskView task={offlineContactTask} key={`controller-${selectedTaskSid}`} />}
     </Absolute>
   );
 };
 
 CustomCRMContainer.displayName = 'CustomCRMContainer';
 
 const mapStateToProps = (state: RootState) => {
+  const { selectedTaskSid } = state.flex.view;
+  const { addOfflineContact } = state[namespace][routingBase];
```
andvirga, line 56: Please rename this to: `isAddingOfflineContact` 😉

plugin-hrm-form/src/components/OfflineContactButton/index.tsx
```diff
@@ -1,36 +1,53 @@
...
 
 const mapStateToProps = (state: RootState) => {
+  const { currentDefinitionVersion } = state[namespace][configurationBase];
+  const { addOfflineContact } = state[namespace][routingBase];
+  const { selectedTaskSid } = state.flex.view;
+
   return {
-    worker: state.flex.worker,
+    selectedTaskSid,
+    addOfflineContact,
```
andvirga, line 42: Also, you can rename it here 🙏🏻

plugin-hrm-form/src/components/TaskView.tsx
```diff
@@ -0,0 +1,69 @@
...
+TaskView.displayName = 'TaskView';
+
+const mapStateToProps = (state: RootState, ownProps: OwnProps) => {
+  const { task } = ownProps;
+  // Check if the entry for this task exists in each reducer
+  const contactFormStateExists = Boolean(task && state[namespace][contactFormsBase].tasks[task.taskSid]);
+  const routingStateExists = Boolean(task && state[namespace][routingBase].tasks[task.taskSid]);
+  const searchStateExists = Boolean(task && state[namespace][searchContactsBase].tasks[task.taskSid]);
+
+  const shouldRecreateState = !contactFormStateExists || !routingStateExists || !searchStateExists;
```
andvirga, line 53: Just double-checking the logic: the state should be recreated if **any** of these conditions apply:
- contact form state doesn't exist
- routing state doesn't exist
- search state doesn't exist

Do I understand this right?

plugin-hrm-form/src/components/callTypeButtons/CallTypeButtons.jsx
```diff
@@ -48,7 +52,7 @@ const CallTypeButtons = props => {
   };
 
   const handleNonDataClick = (taskSid, callType) => {
-    if (task.attributes.isContactlessTask) {
+    if (isOfflineContactTask(task)) {
```
andvirga, line 55: I really appreciate these improvements

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -156,9 +156,10 @@ export function transformForm(form: TaskEntry): ContactRawJson {
  * @param hrmBaseUrl
  * @param workerSid
  * @param helpline
+ * @param uniqueIdentifier
  * @param shouldFillEndMillis
  */
-export async function saveToHrm(task, form, workerSid, helpline, shouldFillEndMillis = true) {
+export async function saveToHrm(task, form, workerSid, helpline, uniqueIdentifier, shouldFillEndMillis = true) {
```
andvirga, line 162: Is this `uniqueIdentifier` the twilioWorkerId of the target counselor?

plugin-hrm-form/src/services/formSumbissionHelpers.ts
```diff
@@ -26,37 +27,31 @@ export const completeContactTask = async (task: ITask) => {
...
   const { workerSid, helpline } = getConfig();
 
-  if (task.attributes.isContactlessTask) {
+  if (isOfflineContactTask(task)) {
     const targetSid = contactForm.contactlessTask.createdOnBehalfOf as string;
     const initialAttributes = { helpline, channelType: 'default', isContactlessTask: true, isInMyBehalf: true };
     const finalAttributes = buildInsightsData(initialAttributes, contactForm, caseForm, {});
-    await assignOfflineContact(targetSid, finalAttributes);
+    const inBehalfTask = await assignOfflineContact(targetSid, finalAttributes);
+    const contact = await saveToHrm(task, contactForm, workerSid, helpline, inBehalfTask.sid);
```
andvirga, line 49: It seems that I've found the answer for this my previous question 🤣 
In this case, I would suggest to rename `uniqueIdentifier` In `saveToHrm` function to a more descriptive name (i.e: `InBehalfTwilioWorkerId` or `TargetCounselorTwilioWorkerId`)

plugin-hrm-form/src/services/formSumbissionHelpers.ts
```diff
@@ -26,37 +27,31 @@ export const completeContactTask = async (task: ITask) => {
...
+
+  console.log('>>>>>>', inBehalfTask, contact);
```
andvirga, line 51: Remove this `console.log` line

plugin-hrm-form/src/types/types.ts
```diff
@@ -108,3 +109,21 @@ export type SearchCaseResult = {
...
+    isContactlessTask: true;
+  };
+  channelType: 'default';
+};
+
+export type CustomITask = ITask | OfflineContactTask;
+
+export function isOfflineContactTask(task: CustomITask): task is OfflineContactTask {
+  return task.taskSid === offlineContactTaskSid;
+}
```
andvirga, line 129: Bravo!

andvirga (changes requested):
I believe that I mostly get the idea of your changes, however, I'm not quite sure which is the best approach for storing the right Counselor in the DB. 
In my humble opinion, I would prefer to store the TargetCounselor as an extra parameter in the DB, instead of changing the search queries (that could lead to confusions),

andvirga (approved):
Tested it in my local environment and it's working great:
I've created an offline task with you as Conselor, this is the result:

```json
   "contactlessTask":{
      "date":"2021-03-05",
      "time":"10:35",
      "channel":"whatsapp",
      "createdOnBehalfOf":"WKef74b75b8b09f4aac2482f3ee419a302"
   },
```

nickhurlburt:
This is showing the "No active tasks" Panel.  Is that intentional, or an oversight?

image

nickhurlburt:
Minor thing:  notice where I circle here that the corners are offset.  This may not be a problem if we remove the "No Tasks" component.

Screen Shot 2021-03-08 at 6 28 42 PM

nickhurlburt:
Can we alphabetize the options in the counsellor dropdown?  I think this will make it better to find people in the list and also better to use the keyboard to select counsellors.

nickhurlburt:
If you get an incoming contact and accept it, then you are unable to switch between that contact and the offline contact.  You have to finish the incoming contact before you can return to the offline contact.  

Ideally we could switch between them.  However, it may be fine to not support this now.  Please check with Dee in how important that is, if you haven't already.

nickhurlburt:
It doesn't look like this is displaying correctly in Search.  When I save a contact as being from another counsellor, in search it appears as me, both in the search results and the search detail.  Perhaps that's not part of this PR.

image

nickhurlburt:
Interesting... when I select this Location/Municipality combination, the District stays disabled:
image

That is probably not related to this PR, but is this something we need to investigate, or is it already fixed elsewhere?

nickhurlburt:
When I create a case for an offline contact, that is also still recording me as the counsellor.  But again, perhaps this doesn't address that yet:

image

nickhurlburt:
The form layout looks correct:
image

Otherwise see several other comments above.

GPaoloni:
Closed in favor of https://github.com/tech-matters/flex-plugins/pull/374.

PR #373, opened 2021-03-05
[Case Print PDF] - Create Categories Layout

andvirga:
Jira: https://bugs.benetech.org/browse/CHI-516

Changes description: 
- Reused the logic inside `CaseTags.tsx` to generate "printable" categories.
- Changed the layout to move the print icon to the right upper corner.
- Fixed styling.

### DIFF
--- a/plugin-hrm-form/src/components/case/caseDetails/CaseDetailsHeader.tsx
+++ b/plugin-hrm-form/src/components/case/caseDetails/CaseDetailsHeader.tsx
@@ -17,7 +17,7 @@ import {
   ChildIsAtRiskWrapper,
   StyledPrintButton,
 } from '../../../styles/case';
-import { Flex, Box, FormCheckbox, FormLabel, FormCheckBoxWrapper } from '../../../styles/HrmStyles';
+import { Box, FormCheckbox, FormLabel } from '../../../styles/HrmStyles';
 import { CaseStatus } from '../../../types/types';
 
 type OwnProps = {
--- a/plugin-hrm-form/src/components/case/caseDetails/CaseDetailsHeader.tsx
+++ b/plugin-hrm-form/src/components/case/caseDetails/CaseDetailsHeader.tsx
@@ -12,7 +12,9 @@ import {
   DetailsHeaderOfficeName,
   DetailsHeaderCounselor,
   DetailsHeaderContainer,
-  DetailsHeaderPrintContainer,
+  DetailsHeaderTextContainer,
+  DetailsHeaderChildAtRiskContainer,
+  ChildIsAtRiskWrapper,
   StyledPrintButton,
 } from '../../../styles/case';
 import { Flex, Box, FormCheckbox, FormLabel, FormCheckBoxWrapper } from '../../../styles/HrmStyles';
@@ -41,7 +43,7 @@ const CaseDetailsHeader: React.FC<OwnProps> = ({
 }) => {
   return (
     <DetailsHeaderContainer>
-      <Flex flexDirection="column">
+      <DetailsHeaderTextContainer>
         <DetailsHeaderChildName variant="h6">{childName}</DetailsHeaderChildName>
         <DetailsHeaderCaseContainer>
           <DetailsHeaderCaseId id="Case-CaseId-label">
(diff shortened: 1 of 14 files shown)


plugin-hrm-form/src/components/case/caseDetails/CaseDetailsHeader.tsx
```diff
@@ -73,9 +74,12 @@ const CaseDetailsHeader: React.FC<OwnProps> = ({
               />
             </Box>
             <Template code="Case-ChildIsAtRisk" />
-          </FormCheckBoxWrapper>
+          </ChildIsAtRiskWrapper>
         </FormLabel>
-      </DetailsHeaderPrintContainer>
+      </DetailsHeaderChildAtRiskContainer>
+      <div>
```
GPaoloni, line 80: Is this `div` necessary? Asking because it seems to me that is not changing the UI.

plugin-hrm-form/src/components/case/casePrint/CasePrintCategories.tsx
```diff
@@ -0,0 +1,34 @@
...
+  };
+  definitionVersion: string;
+};
+
+type Props = OwnProps;
+
+const CasePrintCategories: React.FC<Props> = ({ categories, definitionVersion }) => {
+  const { strings } = getConfig();
+
+  console.log({ categories });
```
GPaoloni, line 23: Remove this maybe?

plugin-hrm-form/src/components/search/CasePreview/CaseTags.tsx
```diff
@@ -1,11 +1,13 @@
 /* eslint-disable react/prop-types */
```
GPaoloni, line 1: I don't like this file living under `/components/search/` folder if it's also being used in other places. Can we move it to a more "shared" space? Maybe just `/components/` or `components/case/` makes sense.

GPaoloni (approved):
Great!!

plugin-hrm-form/src/components/search/CasePreview/CaseTags.tsx
```diff
@@ -14,14 +16,37 @@ type OwnProps = {
...
-const CaseTags: React.FC<Props> = ({ categories, definitionVersion }) => {
+const CaseTags: React.FC<Props> = ({ categories, definitionVersion, printPDF }) => {
   const [category1, category2, category3] = getContactTags(definitionVersion, retrieveCategories(categories));
 
+  if (printPDF) {
+    return (
+      <View style={styles.categoryContainer}>
+        {category1 && (
+          <View style={styles.categoryView}>
+            <Text style={{ ...styles.categoryText, backgroundColor: category1.color }}>• {category1.label}</Text>
```
murilovmachado, line 33: I've noticed the categories background have very vivid colors. In our app, we're adding a 0.6 alpha to these colors, so we achieve a lighter version:
`const lighterColor = `${color}99`; // Hex with alpha 0.6`

Maybe using this lighter version here is better. In the mockups they use the lighter color?

murilovmachado (approved):
Check the code and behavior and it looks good to me! Left just one note about the background color of a category.

plugin-hrm-form/src/components/case/caseDetails/CaseDetailsHeader.tsx
andvirga: Actually, it's necessary. That `div` is wrapping the Print Icon (which is a `<button>`) and positioning that icon at the right upper corner, the deletion of that `div` generates that the icon is located in a wrong place:

image

plugin-hrm-form/src/components/search/CasePreview/CaseTags.tsx
andvirga: As you wish! 🤝

plugin-hrm-form/src/components/search/CasePreview/CaseTags.tsx
andvirga: Thanks for this suggestion! I had to implement an opacity element since the library doesn't support the alpha addition. But the result is pretty much similar:

Captura de pantalla de 2021-03-08 16-06-24

PR #374, opened 2021-03-09
Offline Contacts created in behalf of another counselor

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-500
This PR is related to tech-matters/serverless#32

Add Counselor Name to Contact Info tab for Offline Contact
You can point to dev env for serverless in order to test this.

This PR is a reopen of https://github.com/tech-matters/flex-plugins/pull/361 and https://github.com/tech-matters/flex-plugins/pull/371 into the current state of staging.

This PR:
- Adds counselor dropdown to offline contact tab, in order to select in behalf of who is being submitted the contact.
- Adds `isAddingOfflineContact` boolean property to routing reducer to tell the app that the user is in the process of completing a contacttless task form (even if switch between tasks, state is preserved). The offline contact forms are shown if a) there is no selected Twilio task, and b) addOfflineContact is set to true.
- Rework on our TaskView components tree. Now only one single place is using Twilio's withTaskContext helper, and from there and based on above routing information, renders the proper forms. A fake Task (CustomITask) is passed down the component tree if the user is filling an offline task.
- When an offline contact is submitted, a new serverless function (assignOfflineContact) is called. The purpose of this function is to create-assign-accept-complete a new Twilio task to the target counselor in order to create the insights for this record. The form is stored in the backend in behalf of the target counselor (`twilioWorkerId` column in the DB).

You can toogle between regular tasks and the offline contact by clicking the "+ Offline Contact" button in the left side of the agent panel. Follow up PR will change this UI.


plugin-hrm-form/src/components/case/AddNote.tsx
```diff
@@ -59,6 +60,7 @@ const AddNote: React.FC<Props> = ({
     const note = Object.values(transformValues(NoteForm)(temporaryCaseInfo.info));
     const notes = info && info.notes ? [...info.notes, ...note] : [...note];
     const newInfo = info ? { ...info, notes } : { notes };
+    // @ts-ignore TODO: fix this (bug in backend involved, see createAddNoteActivity in hrm)
```
andvirga, line 63: Can you explain which is the bug involved? It's something related to the types?

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -186,11 +194,7 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
                     innerRef={register(rules)}
                     defaultValue={initialValue}
                   >
-                    {def.options.map(o => (
-                      <FormOption key={`${path}-${o.label}`} value={o.value} isEmptyValue={o.value === ''}>
-                        {o.label}
-                      </FormOption>
-                    ))}
+                    {def.options.map(createSelectOptions)}
```
andvirga, line 197: Do we have the certainty that `def.options` is never `null` or `undefined`?
If yes, move ahead, if not you can use `{def && def.options && def.options.map.... }`

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -257,11 +263,7 @@ const getInputType = (parents: string[], updateCallback: () => void) => (def: Fo
                     disabled={disabled}
                     defaultValue={initialValue}
                   >
-                    {options.map(o => (
-                      <FormOption key={`${path}-${o.value}`} value={o.value} isEmptyValue={o.value === ''}>
-                        {o.label}
-                      </FormOption>
-                    ))}
+                    {options.map(createSelectOptions)}
```
andvirga, line 266: Same comment as above 😉

plugin-hrm-form/src/services/formSubmissionHelpers.ts
```diff
@@ -0,0 +1,53 @@
...
+  await Actions.invokeAction('CompleteTask', { sid, task });
+};
+
+export const removeOfflineContact = () => {
+  Manager.getInstance().store.dispatch(removeContactState(offlineContactTaskSid));
+};
+
+export const completeContactlessTask = async (task: CustomITask) => {
+  removeOfflineContact();
+};
```
andvirga, line 36: This is redundant and can be removed I guess.

plugin-hrm-form/src/states/configuration/reducer.ts
```diff
@@ -34,10 +34,14 @@ export function reduce(state = initialState, action: t.ConfigurationActionType):
...
-      const counselorsHash = createCounselorsHash(action.counselorsList);
+      const sortedList = action.counselorsList.sort((c1, c2) => c1.fullName.localeCompare(c2.fullName));
+      const counselorsHash = createCounselorsHash(sortedList);
       return {
         ...state,
-        counselors: { list: action.counselorsList, hash: counselorsHash },
+        counselors: {
+          list: sortedList,
+          hash: counselorsHash,
+        },
```
andvirga, line 44: Great Job! 🚀

plugin-hrm-form/src/components/tabbedForms/ContactlessTaskTabDefinition.ts
```diff
@@ -12,28 +13,42 @@ const channelOptions = [{ value: '', label: '' }].concat(
...
+    },
+    {
+      name: 'date',
+      type: 'date-input',
+      label: 'Date of Contact',
+      required: { value: true, message: 'RequiredFieldError' },
+      validate: date => {
+        const [y, m, d] = splitDate(date);
+        return isFuture(new Date(y, m - 1, d)) ? 'DateCantBeGreaterThanToday' : null;
+      },
```
andvirga, line 45: Just wondering, can we move this validation to a helper? Or is it very specific for this component?

plugin-hrm-form/src/utils/setUpComponents.js
```diff
@@ -133,43 +132,12 @@ const setUpManualPulling = () => {
...
-    id: 'YouMustBeAvailableToPerformThisOp',
-    content: <Flex.Template code="YouMustBeAvailableToPerformThisOp" />,
-    timeout: 5000,
-    type: Flex.NotificationType.warning,
+  // This is causing some bad scenarios, cause AgentDesktopView.Panel1 not re-rendering. Possible solutions: a) force a "change view" b) allways remove the no tasks view c) replace it with our own view that is connected to the store and conditionally appears when appropiate
+  Flex.AgentDesktopView.Panel1.Content.remove('no-tasks', {
+    if: props =>
+      props.route.location.pathname === '/agent-desktop/' &&
+      !props.selectedTaskSid &&
+      manager.store.getState()[namespace][routingBase].isAddingOfflineContact, // while this is inefficient because of calling getState several times in a short period of time (re-renders), the impact is minimized by the short-circuit evaluation of the AND operator
```
andvirga, line 140: I wouldn't concern so much about re-renderings as long as the performance is not being highly affected. For the three options, I'd say that the last one is the most convenient.

andvirga (approved):
Excellent work! I've tested it and didn't find any issues. I left some small comments.

plugin-hrm-form/src/components/case/AddNote.tsx
GPaoloni: The types are wrong, but even if the types were correct, when I add a note in a case created by someone else, it's added in behalf of that person and not me.

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
GPaoloni: The requirement is the following type for the definition of selects. Given that this is coming from a JSON file, it may not be true, but in that case the problem is with the validation of the created file and no the generators (that expects a certain shape in the provided input files). I think that we'll make sure all of this types are satisfied when generating the JSONs once we have the UI for that, but in the meantime, we can just ensure this kind of things by ourselves, there will be too many cases to validate otherwise (missing `name`, missing `label`). We could write some kind of parser to validate the definitios tho.

Screenshot from 2021-03-12 11-51-28

plugin-hrm-form/src/components/tabbedForms/ContactlessTaskTabDefinition.ts
GPaoloni: Actually moving this to a helper is a good idea if we plan to add a validation rule `dateIsNotFuture` for date type inputs! But I'll do that in another PR as this one is already too big haha.

PR #378, opened 2021-03-12
Update ZA forms

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-559
Primary reviewer: @GPaoloni 

Small fixes on the ZA forms:
- In Child Information tab, "Child on the move" values "Voluntary" and "Involuntary" are missing labels. (The label can match the field name)
- In Child Information tab, "Region" possible values should be Rural Areas, Towns & semi-dense areas, and cities (match Zambia)
- 
Layout:
- Caller Info tab - move District to first column to stay grouped with Province and Municipality

Screenshots of the 3 fixes:
image
image
image

### DIFF
--- a/plugin-hrm-form/src/formDefinitions/za-v1/tabbedForms/ChildInformationTab.json
+++ b/plugin-hrm-form/src/formDefinitions/za-v1/tabbedForms/ChildInformationTab.json
@@ -1312,8 +1312,8 @@
     "label": "Child on the move or in migration",
     "type": "select",
     "options": [
-      { "value": "Voluntary" },
-      { "value": "Involuntary" }
+      { "value": "Voluntary", "label": "Voluntary" },
+      { "value": "Involuntary", "label": "Involuntary" }
     ]
   },
   {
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/formDefinitions/za-v1/tabbedForms/ChildInformationTab.json
```diff
@@ -1312,8 +1312,8 @@
     "label": "Child on the move or in migration",
     "type": "select",
     "options": [
-      { "value": "Voluntary" },
-      { "value": "Involuntary" }
+      { "value": "Voluntary", "label": "Voluntary" },
+      { "value": "Involuntary", "label": "Involuntary" }
```
andvirga, line 1316: Your changes are correct. However, if not quite sure if it's right that this field is always obligatory (you must choose one of the two options).
Based on this [document](https://benetech.app.box.com/file/758862659326), it seems that this field should include the blank option. We can ask Dee or Joan about it.

andvirga (approved):
The PR and the branching strategy look great to me! 
However, I left one comment that we need to ask, just to make sure that the field is working as expected.

plugin-hrm-form/src/formDefinitions/za-v1/tabbedForms/ChildInformationTab.json
GPaoloni: Agreed with @andvirga here. You can add the following option at the beginning of the options list.
`{ "value": "Unknown", "label": "" },`

GPaoloni (approved):
Looks good!

PR #379, opened 2021-03-12
[Case Print PDF] - Update service configuration with image source URLs

andvirga:
JIra: https://bugs.benetech.org/browse/CHI-557

Primary reviewer: @fgonzalez-techmatters 

Changes Description:
- Getting `pdfImagesSource` from Service Configuration. (this URL is going to be different for each environment).


fgonzalez-techmatters (approved):
It looks good to me.

PR #380, opened 2021-03-12
Follow up - Offline Contact

GPaoloni:
This PR is related to https://github.com/tech-matters/hrm/pull/99.

This PR:
- Adds fake task component for in progress offline contacts in the task list.
- Address a small change related with the backend (we want to store `twilioWorkerId` as the counselor who took the contact, and add a `createdBy` column to store the value of the person submitting the record to the DB).


plugin-hrm-form/src/components/OfflineContact/OfflineContactTask.tsx
```diff
@@ -0,0 +1,52 @@
...
+  OfflineContactTaskButton,
+} from '../../styles/HrmStyles';
+
+type OwnProps = { selectedTaskSid: string };
+
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ConnectedProps<typeof connector>;
+
+const OfflineContactTask: React.FC<Props> = ({ isAddingOfflineContact, selectedTaskSid, ...props }) => {
+  console.log('>>>>>>>>', props);
```
murilovmachado, line 22: we can remove this logging

plugin-hrm-form/src/services/CaseService.ts
```diff
@@ -13,12 +13,14 @@ export async function createCase(task: CustomITask, contactForm: ContactForm) {
         helpline,
         status: 'open',
         twilioWorkerId: contactForm.contactlessTask.createdOnBehalfOf,
+        createdBy: workerSid,
```
murilovmachado, line 16: Nice!

murilovmachado (approved):
Looks good to me. There's only one thing that looks different: the offline task icon and text (not sure if it's intentional).

before:
Screen Shot 2021-03-12 at 18 42 16

after:
Screen Shot 2021-03-12 at 18 42 23

andvirga (commented):
I've found a weird bug, here are the reproduction steps:
1. Click on "+ Offline Contact Record"
2. Select yourself as the Counsellor (the same account that you're currently using to create the case) and fill in the rest of the data.
3. Click on "Save Contact"

As soon as you click this button you'll notice a fast re-render of some screens (for example Call type screen), if you inspect the records in the DB you'll notice that the same record was created twice (with some seconds of difference and two different taskIds).

Please take a look at it, it's a weird issue and I'm not 100% percent sure if I double-clicked the save button the first time that I saw this.

GPaoloni:
> I've found a weird bug, here are the reproduction steps:
Click on "+ Offline Contact Record"
Select yourself as the Counsellor (the same account that you're currently using to create the case) and fill in the rest of the data.
Click on "Save Contact"

@andvirga  After several attempts and different scenarios trying to replicate the bug, I couldn't. Can this be a double click on submit issue? Can you guys confirm if this bug happens to you? @nickhurlburt @murilovmachado

PR #381, opened 2021-03-16
Added new Datadog App ID for monitoring

fgonzalez-techmatters:
Adding new Datadog application ID variable, to handle RUM monitoring with different application IDs and tokens per Flex environments.

### DIFF
--- a/.github/workflows/aselo_development.yml
+++ b/.github/workflows/aselo_development.yml
@@ -25,6 +25,29 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: '12.x'
+      # Setup credentials to access AWS for parameters
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/aselo_development.yml
```diff
@@ -25,6 +25,29 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+      # Get AWS parameters
+      - name: Set Datadog Application ID
+        uses: "marvinpinto/action-inject-ssm-secrets@latest"
+        with:
+          ssm_parameter: "DEV_DATADOG_AS_APP_ID"
```
GPaoloni, line 39: The prefixes will differ here for the actions, according to what environment we want to deploy to, right?

GPaoloni (approved):
Looks great to me! We will still have Rollbar reporting to a single project, right?

.github/workflows/aselo_development.yml
fgonzalez-techmatters: Both comments are correct. I didn't get the chance to look into Rollbar yet, so for now we are going to keep the same key.

Regarding the parameter prefixes, the value will change for each GitHub action according to the environment.

PR #383, opened 2021-03-17
Use Office value in Flex-plugins

andvirga:
Jira: https://bugs.benetech.org/browse/CHI-522

Changes:
1. Getting Office value from worker attributes.
2. Sending that value in each request to HRM (this was already done by sending the helpline attribute).
3. Displaying Office value in Case View Header and in Case Print PDF.
4. Mockup Case Manager values per each Office.

### DIFF
--- a/plugin-hrm-form/src/components/case/casePrint/mockedData.ts
+++ b/plugin-hrm-form/src/components/case/casePrint/mockedData.ts
@@ -1,8 +1,20 @@
 export const caseManagers = [
   {
+    office: 'National Office',
+    name: 'Bhavna Lutchman',
+    phone: '031 201 2059',
+    email: 'research@childlinesa.org.za',
+  },
+  {
     office: 'Gauteng',
     name: 'Bhavna Lutchman',
     phone: '031 201 2059',
     email: 'research@childlinesa.org.za',
+  },
+  {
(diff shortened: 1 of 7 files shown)


murilovmachado (approved):
Looks good!

plugin-hrm-form/src/components/case/casePrint/mockedData.ts
```diff
@@ -1,3 +1,20 @@
...
+    phone: '031 201 2059',
+    email: 'research@childlinesa.org.za',
+  },
+  {
+    office: 'Western Cape',
+    name: 'Nicole Paulsen',
+    phone: '21 762 8198',
+    email: 'info@childlinewc.org',
+  },
+];
```
GPaoloni, line 20: Maybe not in this PR, but it will nice to have this somewhere in the form definitions, so we can easily customize this per helpline. Just an idea.

GPaoloni (approved):
:rocket:

plugin-hrm-form/src/components/case/casePrint/mockedData.ts
andvirga: Sure, I'll address that in the next PR! Thanks!

PR #384, opened 2021-03-19
Support workers in multiple queues

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-548
This PR is related to https://github.com/tech-matters/serverless/pull/34
Primary reviewer: @andvirga 

This PR:
- Fetches all the matching queues for the worker, and subscribes to all of them. This is shown in the "Contacts Waiting" widget, and also allows the worker to "Add another task" from any of those queues.
- If worker attributes are updated, call the above again just in case the worker matches different queues this time.

To test this PR:
- Point to serverless dev env.
- Add `"routing":{"skills":["chat"]} `to your worker attributes in Twilio console.
- Check that both queues works for you, that you can pull tasks from both.
- Check that if you change your worker's attributes (to unsubscribe from a queue then re-subscribe), the queues widget should update without a page reload.

### DIFF
--- a/plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx
+++ b/plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx
@@ -40,6 +40,7 @@ export class InnerQueuesStatusWriter extends React.Component {
   async componentDidMount() {
     try {
       await this.subscribeToQueuesUpdates();
+
       const workerQuery = await this.props.insightsClient.liveQuery(
         'tr-worker',
         `data.worker_sid == "${this.props.workerSid}"`,
@@ -50,7 +51,7 @@ export class InnerQueuesStatusWriter extends React.Component {
         await this.subscribeToQueuesUpdates();
       });
     } catch (err) {
-      handleSubscribeError(err);
+      this.handleSubscribeError(err);
     }
   }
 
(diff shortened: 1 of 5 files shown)


andvirga (approved):
Great Job! 
I've tested what happens when the user is subscribing / unsubscribing the "chat" queue. Everything is working as expected, if you subscribe to the Chat queue you start receiving incoming tasks from that queue. Then If you unsubscribe the system will automatically hide those tasks in the UI.

plugin-hrm-form/src/components/queuesStatus/QueuesStatusWriter.jsx
```diff
@@ -25,29 +26,48 @@ export class InnerQueuesStatusWriter extends React.Component {
...
     try {
-      const q = await this.props.insightsClient.liveQuery('tr-queue', '');
-      const queues = this.getQueuesNames(q.getItems());
-      q.close();
+      await this.subscribeToQueuesUpdates();
+
+      const workerQuery = await this.props.insightsClient.liveQuery(
+        'tr-worker',
+        `data.worker_sid == "${this.props.workerSid}"`,
+      );
```
nickhurlburt, line 47: This is great!  It's also worth keeping in mind that this won't change if the matching working expression on the _queues_ change, so we would miss those real-time updates.  Fine for now, though.

GPaoloni: You are right. We can subscribe to the matching queues and listen for changes there too!

PR #386, opened 2021-03-22
Added optional default value for select input type

GPaoloni:
This PR is addressing a comment @dee-luo left in the task https://bugs.benetech.org/browse/CHI-500.
> 2. Is there anyway we're able to default the counselor drop-down in the form to the counselor that is logged in? Let me know if that's not easy to do.

In order to do this:
- `defaultOption` value is added as optional property of select input definitions.
- `defaultOption` with value of self counselor added to offline form tab definition.


nickhurlburt (approved):
Looks great

PR #390, opened 2021-03-24
PDF error fixing and improvements

andvirga:
This PR contains changes requested for [CHI-515](https://bugs.benetech.org/browse/CHI-515).

Primary reviewer: @GPaoloni 

Changes summary:

1. Added a tiny space between the two text columns that makes long texts occupy another line instead of being too long (for example: Is the perpetrator still in contact with the child? now occupies two lines).
2. Renamed "OK for case worker to call?" to "May social worker call/SMS?" and fixed the incorrect value display (this field is now a multi checkbox and its values can be: Yes, No, Unknown).

### DIFF
--- a/plugin-hrm-form/src/components/case/casePrint/CasePrintSection.tsx
+++ b/plugin-hrm-form/src/components/case/casePrint/CasePrintSection.tsx
@@ -27,7 +27,7 @@ const CasePrintSection: React.FC<Props> = ({ sectionName, values, definitions, u
           return (
             <View key={i} style={i % 2 === 0 ? styles.sectionItemRowOdd : styles.sectionItemRowEven}>
               <View style={styles.sectionItemFirstColumn}>
-                <Text style={{ width: '95%' }}>{def.label}</Text>
+                <Text style={{ marginRight: '10px' }}>{def.label}</Text>
               </View>
               <View style={styles.sectionItemSecondColumn}>
                 <Text>{presentValue(unNestInfo ? unNestInformation(def, values) : values[def.name])(def)}</Text>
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/components/case/casePrint/CasePrintSection.tsx
```diff
@@ -26,10 +26,12 @@ const CasePrintSection: React.FC<Props> = ({ sectionName, values, definitions, u
         {definitions.map((def, i) => {
           return (
             <View key={i} style={i % 2 === 0 ? styles.sectionItemRowOdd : styles.sectionItemRowEven}>
-              <Text style={styles.sectionItemFirstColumn}>{def.label}</Text>
-              <Text style={styles.sectionItemSecondColumn}>
-                {presentValue(unNestInfo ? unNestInformation(def, values) : values[def.name])(def)}
-              </Text>
+              <View style={styles.sectionItemFirstColumn}>
+                <Text style={{ width: '95%' }}>{def.label}</Text>
```
GPaoloni, line 30: This does not seem to have the desired effect (at least in my monitor). Maybe try to do some padding/margin right instead?
Screenshot from 2021-03-24 18-28-25

GPaoloni (approved):
Looks good! Just one small comment.

plugin-hrm-form/src/components/case/casePrint/CasePrintSection.tsx
andvirga: Can you please test again this PR with the latest changes that I've pushed. It's weird but reducing the screen size (using Chrome responsive function) is producing that the PDF is not being shown at all (no error traces, and no extra info).
I wonder if this issue is also happening to you.

GPaoloni: Seems fixed for me! (resolution at 1366x768)
Screenshot from 2021-03-25 21-39-50
Screenshot from 2021-03-25 21-39-42

PR #392, opened 2021-03-24
Bug fix: Completing Offline Contact does not re-render screen

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-582

This PR:
- Fixes a misspelling.
- Avoid selecting a task if is one created in behalf (offline contact process).
- Re renders agent desktop each time Task View unmounts an offline contact form.

### DIFF
--- a/plugin-hrm-form/src/components/TaskView.tsx
+++ b/plugin-hrm-form/src/components/TaskView.tsx
@@ -9,7 +9,8 @@ import { RootState, namespace, contactFormsBase, searchContactsBase, routingBase
 import * as GeneralActions from '../states/actions';
 import { hasTaskControl } from '../utils/transfer';
 import type { ContactFormDefinition } from '../states/types';
-import { CustomITask, isOfflineContactTask } from '../types/types';
+import { CustomITask, isOfflineContactTask, isInMyBehalfITask } from '../types/types';
+import { reRenderAgentDesktop } from '../HrmFormPlugin';
 
 type OwnProps = {
   task: CustomITask;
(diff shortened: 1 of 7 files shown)


nickhurlburt (approved):
Looks good.  When ending an offline contact with no other tasks, it re-renders the NoTasksView correctly.

When ending an offline contact when there are other tasks, it is not auto-selecting another task, but rather keeps the page clear.  This is the normal functionality when closing an online contact as well, though, so I don't think that's a problem.
image

plugin-hrm-form/src/components/TaskView.tsx
```diff
@@ -20,18 +21,28 @@ type Props = OwnProps & ConnectedProps<typeof connector>;
...
     if (shouldRecreateState) {
       recreateContactState(currentDefinitionVersion.tabbedForms)(task.taskSid);
     }
   }, [currentDefinitionVersion.tabbedForms, recreateContactState, shouldRecreateState, task.taskSid]);
 
+  // Force a re-render on unmount (temporary fix NoTaskView issue with Offline Contacts)
+  React.useEffect(() => {
+    return () => {
+      if (isOfflineContactTask(task)) reRenderAgentDesktop();
+    };
```
andvirga, line 35: Is this equivalent to do? (I'm not getting why are you returning a function instead, but it's just curiosity):  
```javascript 
if (isOfflineContactTask(task)) { 
  reRenderAgentDesktop();
}
```

andvirga (approved):
Looks good, I've only one question ;)

plugin-hrm-form/src/components/TaskView.tsx
GPaoloni: The returned function is the [cleanup of the effect](https://reactjs.org/docs/hooks-effect.html#effects-with-cleanup). In this case serves the purpose of `componentWillUnmount` (or when the `task` dependency changes, but that's also triggering an unmount of `TaskView`).

PR #393, opened 2021-03-24
Map helpline value to Insights

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-584

This PR:
- Adds task attributes (as `taskAttributes`) to the possible paths to retrieve data from in `applyCustomUpdate`.
- `taskAttributes.helpline` added to `conversation_attribute_8` for ZA.

### DIFF
--- a/plugin-hrm-form/src/formDefinitions/za-v1/insights/oneToManyConfigSpecs.json
+++ b/plugin-hrm-form/src/formDefinitions/za-v1/insights/oneToManyConfigSpecs.json
@@ -7,5 +7,10 @@
       "contactForm.childInformation.municipality",
       "contactForm.childInformation.district"
     ]
+  },
+  {
+    "attributeName": "conversation_attribute_8",
+    "insightsObject": "conversations",
+    "paths": ["taskAttributes.helpline"]
   }
 ]
--- a/plugin-hrm-form/src/services/InsightsService.ts
+++ b/plugin-hrm-form/src/services/InsightsService.ts
@@ -249,10 +249,10 @@ export const processHelplineConfig = (
 };
 
 const applyCustomUpdate = (customUpdate: OneToManyConfigSpec): InsightsUpdateFunction => {
-  return (attributes, contactForm, caseForm) => {
+  return (taskAttributes, contactForm, caseForm) => {
     if (isNonDataCallType(contactForm.callType)) return {};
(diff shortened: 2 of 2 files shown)


plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -249,10 +249,10 @@ export const processHelplineConfig = (
 };
 
 const applyCustomUpdate = (customUpdate: OneToManyConfigSpec): InsightsUpdateFunction => {
-  return (attributes, contactForm, caseForm) => {
+  return (taskAttributes, contactForm, caseForm) => {
     if (isNonDataCallType(contactForm.callType)) return {};
 
-    const dataSource = { contactForm, caseForm };
+    const dataSource = { contactForm, caseForm, taskAttributes };
```
nickhurlburt, line 255: I have a slight preference for ordering this `{ taskAttributes, contactForm, caseForm }` to keep the same order as the parameters of the returned function.

plugin-hrm-form/src/formDefinitions/za-v1/insights/oneToManyConfigSpecs.json
```diff
@@ -7,5 +7,10 @@
       "contactForm.childInformation.municipality",
       "contactForm.childInformation.district"
     ]
+  },
+  {
```
nickhurlburt, line 11: Per Slack conversation, we may someday want to implement this for all helplines, in which case we'd add it to `baseUpdates()`.  It could still be here and have no effect.  But for now we can leave it here.

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -249,10 +249,10 @@ export const processHelplineConfig = (
...
     // concatenate the values, taken from dataSource using paths (e.g. 'contactForm.childInformation.province')
     const value = customUpdate.paths.map(path => get(dataSource, path, '')).join(delimiter);
```
nickhurlburt, line 258: While you're in here, would you mind knocking out L298: `// In TS, how do we say that we are returning a function?`.  This is an old, unnecessary comment.

nickhurlburt (approved):
Looks great, and nice to see the `oneToManyConfigSpecs` have a new use!

PR #394, opened 2021-03-24
If first name and last name are not provided we don't show 'No data' …

andvirga:
…verbiage

Jira: [CHI-537](https://bugs.benetech.org/browse/CHI-537)

Change description:
- Removed 'No Data' verbiage when firstName or lastName are not defined. We're matching the behavior with the contact search results

### DIFF
--- a/plugin-hrm-form/src/components/search/CasePreview/CaseHeader.tsx
+++ b/plugin-hrm-form/src/components/search/CasePreview/CaseHeader.tsx
@@ -6,6 +6,7 @@ import CropFreeIcon from '@material-ui/icons/CropFree';
 import Dialog from '@material-ui/core/Dialog';
 import DialogContent from '@material-ui/core/DialogContent';
 
+import { getConfig } from '../../../HrmFormPlugin';
 import {
   CaseHeaderContainer,
   CaseHeaderCaseId,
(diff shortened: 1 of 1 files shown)


nickhurlburt (approved):
Looks good!  This is good to go, we can save @murilovmachado a review

plugin-hrm-form/src/components/search/CasePreview/CaseHeader.tsx
```diff
@@ -54,15 +56,7 @@ const CaseHeader: React.FC<Props> = ({
...
-          {isOrphanedCase ? (
-            <Template code="CaseHeader-Voided" />
-          ) : (
-            <>
-              {firstName ? firstName : <Template code="CaseHeader-NoData" />}
-              &nbsp;
-              {lastName ? lastName : <Template code="CaseHeader-NoData" />}
-            </>
-          )}
+          {isOrphanedCase ? strings['CaseHeader-Voided'] : `${firstName} ${lastName}`}
```
GPaoloni, line 59: Just a reminder that we should use `Template` whenever possible (out of react PDF scope): 
```
{isOrphanedCase ? <Template code="'CaseHeader-Voided" /> : `${firstName} ${lastName}`}
```

PR #396, opened 2021-03-25
Upgrade to Flex 1.25.2 (second attempt)

nickhurlburt:
In an earlier PR, I attempted to do this upgrade, but our Github Actions would not deploy.  They kept getting stuck on the deployment step using the Twilio CLI.  Through investigating the build logs, I think there are two potential issues:
1. We don't need to be using `SKIP_PREFLIGHT_CHECK=true` but we do need to be using `UNBUNDLED_REACT=true` per the docs on [selecting react versions](https://www.twilio.com/docs/flex/developer/plugins/react-versions).  I think the docs have been changed since we originally set up our actions.  Replacing `SKIP...` with `UNBUNDLE...` seems to reduce complaints from the build logs.
2. The Flex Admin console provides the option for choosing React `~16.13.1`, not `^16.13.1`.  The instructions in the linked doc above do say you can use `"react": "^16.13.1"`, but I suspect this is an error.  We have started using `16.14` and I think that may cause trouble.  I downgraded it :(

With those two changes made, this now builds and deploys successfully.

One thing that @GPaoloni will need to look at:  without the change I made to the `types.ts` file below, I get the error `Property 'isInMyBehalf' does not exist on type '{ isContactlessTask: true; } | Record<string, any> | (Record<string, any> & { isContactlessTask: true; isInMyBehalf: true; })'.  Property 'isInMyBehalf' does not exist on type '{ isContactlessTask: true; }'.ts(2339)`  I suspect this is due to a dependency of the React downgrade.  I'm guessing that the change I made is fine, but please take a look.

image


plugin-hrm-form/src/types/types.ts
```diff
@@ -120,6 +120,7 @@ export type OfflineContactTask = {
   taskSid: typeof offlineContactTaskSid;
   attributes: {
     isContactlessTask: true;
+    isInMyBehalf?: boolean;
```
GPaoloni, line 123: To avoid confusion between `OfflineContactTask` and `InMyBehalfITask` I suggest doing this instead
```
export function isInMyBehalfITask(task: CustomITask): task is InMyBehalfITask {
  return task.attributes && task.attributes.isContactlessTask && (task.attributes as any).isInMyBehalf;
}
```
I think we should avoid typecasting, but is useful on type guards, and I've seen this being done in [the docs](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards).

GPaoloni (approved):
Looks and works good!

andvirga (approved):
Looks good! Haven't spotted any issue running local environment

fgonzalez-techmatters (approved):
Looks good to me too.

plugin-hrm-form/src/types/types.ts
nickhurlburt: Thanks, that works!  And I think I understand why... I'm not aware of a type filter construct like this (`<param> is <type>`) existing in any other language.  So much TypeScript magic!

PR #398, opened 2021-03-25
Front-end Permissions 

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-571
Primary Reviewer: @nickhurlburt 

~This initial commit already applies permission rules to the front-end. Still working on making it less hardcoded than it is right now.~

Comments:
- For editing **Child is at Risk** and **Follow Up Date**, I've let the same logic as before because we don't have them defined. But I've let the code in a state where it'll be super easy to change it;
- Logic for adding notes, households, perpetrators they are all the same, so for now I have a generic rule for all of them;

Future TODOs:
- Get set of permission rules from **accountSid** instead of case's **definitionVersion**.

### DIFF
--- a/plugin-hrm-form/src/permissions/index.ts
+++ b/plugin-hrm-form/src/permissions/index.ts
@@ -22,9 +22,7 @@ type Version = 'v1' | 'za-v1';
 type Rule = (isSupervisor: boolean, isCreator: boolean, isCaseOpen: boolean) => boolean;
 type Rules = {
   canEditCaseSummary: Rule;
-  canEditChildIsAtRisk: Rule;
   canEditGenericField: Rule;
-  canEditFollowUpDate: Rule;
 };
 
 const rulesMap: { [version in Version]: Rules } = {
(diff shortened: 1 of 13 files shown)


plugin-hrm-form/src/permissions/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  ADD_INCIDENT: 'addIncident',
+  EDIT_INCIDENT: 'editIncident',
+  EDIT_CASE_SUMMARY: 'editCaseSummary',
+};
+
+type PermissionActionsKeys = keyof typeof PermissionActions;
+export type PermissionActionType = typeof PermissionActions[PermissionActionsKeys];
+type Version = 'v1' | 'za-v1';
+
+type canEditCaseSummaryType = (isSpervisor: boolean, isCreator: boolean, isCaseOpen: boolean) => boolean;
```
nickhurlburt: Spelling: `isSupervisor`

nickhurlburt (approved):
This is great!  And adding the abstraction of `can()` is good for allowing us to hard-code at first and then increasingly improve quality and extensibility.

andvirga (approved):
Looks good!

nickhurlburt (changes requested):
I dug a little deeper, and did find some issues which need to be resolved.  All in all I think this is a great approach for this tight timeline!

PR #399, opened 2021-03-25
Counselor Name missing from Case Search results

andvirga:
Jira: [CHI-487](https://bugs.benetech.org/browse/CHI-487)

Primary reviewer: anyone ;)

Change description:
- Added Counselor information to Case Cards (taken from counselorsHash)

### DIFF
--- a/plugin-hrm-form/src/components/search/CasePreview/CasePreview.tsx
+++ b/plugin-hrm-form/src/components/search/CasePreview/CasePreview.tsx
@@ -1,21 +1,23 @@
 /* eslint-disable react/prop-types */
 import React from 'react';
+import { Template } from '@twilio/flex-ui';
 
 import { Case } from '../../../types/types';
 import CaseHeader from './CaseHeader';
 import CaseSummary from './CaseSummary';
 import CaseTags from '../../case/CaseTags';
 import { Flex } from '../../../styles/HrmStyles';
-import { CaseWrapper } from '../../../styles/search';
+import { CaseWrapper, CaseFooter, CaseFooterText, CounselorText, SummaryText } from '../../../styles/search';
 
 type OwnProps = {
   currentCase: Case;
   onClickViewCase: () => void;
+  counselorsHash: { [sid: string]: string };
 };
 
 type Props = OwnProps;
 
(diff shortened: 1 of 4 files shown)


plugin-hrm-form/src/components/search/CasePreview/CasePreview.tsx
```diff
@@ -1,28 +1,31 @@
...
-const CasePreview: React.FC<Props> = ({ currentCase, onClickViewCase }) => {
+const CasePreview: React.FC<Props> = ({ currentCase, onClickViewCase, counselorsHash }) => {
   const { id, createdAt, updatedAt, connectedContacts, status, info } = currentCase;
 
   const orphanedCase = !connectedContacts || connectedContacts.length === 0;
   const firstContact = !orphanedCase && connectedContacts[0];
   const { name } = ((firstContact || {}).rawJson || {}).childInformation || {};
   const { categories, callSummary } = ((firstContact || {}).rawJson || {}).caseInformation || {};
   const summary = info?.summary || callSummary;
+  const counselor = counselorsHash[firstContact?.twilioWorkerId];
```
GPaoloni, line 28: Is there any reason why you are not using the `twilioWorkerId` from the case itself?
Screenshot from 2021-03-25 19-41-32

Asking because there will be cases like #1101 that won't have any connected case (voided) resulting in an empty counselor, but they do have a `twilioWorkerId`:
Screenshot from 2021-03-25 19-49-07
Screenshot from 2021-03-25 19-48-27

The only scenario I imagine can cause troubles is:
- A creates a case for a later use (not available functionality yet).
- B inputs a contact and connects it as first connected contact to the above case.

My suggestion will display A as counselor, which I think is correct, but may be worth asking.

GPaoloni (approved):
Lookin' great!

plugin-hrm-form/src/components/search/CasePreview/CasePreview.tsx
andvirga: Good catch! since that functionality isn't available yet I'm pretty confident that using your suggestion won't break anything for now.

PR #400, opened 2021-03-25
[WIP] Add customization support for case statuses

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-567

This PR:
- Adds a new property to the case reducer: `prevStatus`, which stores the case status as it comes from the DB.
- Adds case statuses as part of customization (each status describes the possible transitions to other statuses).
- Using above information, decide when and to which statuses the UI allows to transition at a given time.


GPaoloni:
Closed in favor of github.com/tech-matters/flex-plugins/pull/412

PR #404, opened 2021-03-30
Canned Responses feature for South Africa

andvirga:
Jira: [CHI-533](https://bugs.benetech.org/browse/CHI-533)

Investigation Doc: https://benetech.app.box.com/file/792929253957

Changes description:
- Created `CannedResponses` components that render a dropdown with the list of canned responses.
- Used `setInputText` action to put a canned response in the chat window (this response can be edited).
- Added ZA Canned Responses to `CannedResponses.json`.
- Added types.

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -191,7 +191,8 @@ const setUpComponents = setupObject => {
   }
 
   Components.setUpStandaloneSearch();
-  Components.setupCannedResponses();
+
+  if (featureFlags.enable_canned_responses) Components.setupCannedResponses();
 };
 
 /**
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import * as Flex from '@twilio/flex-ui';
 import { FlexPlugin, loadCSS } from 'flex-plugin';
 import SyncClient from 'twilio-sync';
@@ -13,6 +14,7 @@ import setUpMonitoring from './utils/setUpMonitoring';
 import * as TransferHelpers from './utils/transfer';
 import { changeLanguage } from './states/configuration/actions';
 import { issueSyncToken } from './services/ServerlessService';
+import CannedResponses from './components/CannedResponses';
 
 const PLUGIN_NAME = 'HrmFormPlugin';
 export const PLUGIN_VERSION = '0.10.0';
@@ -190,6 +192,7 @@ const setUpComponents = setupObject => {
   }
 
   Components.setUpStandaloneSearch();
+  Components.setupCannedResponses();
 };
 
 /**
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -1,4 +1,3 @@
-import React from 'react';
 import * as Flex from '@twilio/flex-ui';
 import { FlexPlugin, loadCSS } from 'flex-plugin';
 import SyncClient from 'twilio-sync';
--- a/plugin-hrm-form/src/components/CannedResponses.tsx
+++ b/plugin-hrm-form/src/components/CannedResponses.tsx
@@ -2,9 +2,18 @@
 import React from 'react';
 import { connect, ConnectedProps } from 'react-redux';
 import { Actions, Template, withTheme } from '@twilio/flex-ui';
-import { FormControl, InputLabel, Select, MenuItem } from '@material-ui/core';
+import {} from '@material-ui/core';
 
-import { CannedResponsesContainer } from '../styles/HrmStyles';
+import {
+  Box,
+  CannedResponsesContainer,
+  FormLabel,
+  FormSelect,
+  FormSelectWrapper,
+  FormError,
+  FormOption,
+  Row,
+} from '../styles/HrmStyles';
 import { RootState, namespace, configurationBase } from '../states';
 
 type OwnProps = {
@@ -29,20 +38,20 @@ const CannedResponses: React.FC<Props> = props => {
 
   return (
     <CannedResponsesContainer>
-      <FormControl className="form">
-        <InputLabel className="input-label" htmlFor="canned_response">
-          <Template code="CannedResponses" />
-        </InputLabel>
-        <Select value="" onChange={handleChange} name="canned_response">
+      <FormSelectWrapper fullWidth={true}>
+        <FormSelect id="canned_response" name="canned_response" onChange={handleChange} value="" fullWidth={true}>
+          <FormOption disabled selected isEmptyValue={true} value="">
+            Canned Responses
+          </FormOption>
           {cannedResponses.map(r => {
             return (
-              <MenuItem key={r.label} value={r.text}>
+              <FormOption key={r.label} value={r.text} isEmptyValue={r.text === ''}>
                 {r.label}
-              </MenuItem>
+              </FormOption>
             );
           })}
-        </Select>
-      </FormControl>
+        </FormSelect>
+      </FormSelectWrapper>
     </CannedResponsesContainer>
   );
 };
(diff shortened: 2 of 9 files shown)


plugin-hrm-form/src/formDefinitions/za-v1/CannedResponses.json
```diff
@@ -0,0 +1,106 @@
...
+  },
+  {
+    "label": "Corporal Punishment",
+    "text": "Well, in the South African Schools Act (1998) has banned all forms of corporal punishment in schools and places of education. However, in many schools, corporal punishment is still practiced. Educators who still use corporal punishment are committing an offence, has the matter been addressed with the school?"
+  },
+  {
+    "label": "COVID19",
+    "text": "It’s understandable that you are worried about getting the COVID19 Virus, however, here are some tips to help you stay safe • Wash your hands regularly / use sanitizer • Avoid handshakes, hugs and kisses. • Don’t touch your face unnecessarily • Keep a 1.5-meter distance from people, especially if they are symptomatic • Avoid going out into public and crowds. • Use this additional time to catch up on Term 2 Work • Check all info, especially info that seems fake. For all verified news Whats App - 060 012 3456, Type “HI” and follow the prompts • If you feel that you show symptoms or have come in touch with a COVID19 positive person, please also call the COVID19 Helpline on 08000 29999 • Stay calm and don't panic"
```
nickhurlburt, line 28: Can we add newlines before the bullets so that it shows up in the chat as:
```
It’s understandable that you are worried about getting the COVID19 Virus, however, here are some tips to help you stay safe 
 • Wash your hands regularly / use sanitizer
 • Avoid handshakes, hugs and kisses.
 • Don’t touch your face unnecessarily 
[...]
```

plugin-hrm-form/src/components/CannedResponses.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  if (!cannedResponses) return null;
+
+  return (
+    <CannedResponsesContainer>
+      <FormControl className="form">
+        <InputLabel className="input-label" htmlFor="canned_response">
+          <Template code="CannedResponses" />
+        </InputLabel>
+        <Select value="" onChange={handleChange} name="canned_response">
```
nickhurlburt, line 36: I think we want to move more toward using standard HTML elements for the dropdown, as we are with selects within the customization framework.  Then it is more accessible and the keyboard can be used to go through the list -- for example, using the up/down arrow keys and typing the first letters of the option to pull it up.  But if there are problems with that option, we can discuss.

plugin-hrm-form/src/components/CannedResponses.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ConnectedProps<typeof connector>;
+
+// eslint-disable-next-line react/display-name
+const CannedResponses: React.FC<Props> = props => {
+  const { channelSid, cannedResponses } = props;
+
+  const handleChange = event => {
+    Actions.invokeAction('SetInputText', {
```
nickhurlburt, line 22: Oh, of course... this is great.  So simple!

nickhurlburt (changes requested):
Just a couple questions/comments inline.

There are a few larger points this PR touches on.  I'll mention them here, but they are not blocking for this:
1. We're putting a lot into our configuration `Definition Versions`.  One aspect of those is that they are versioned so that we can pull things out of the database properly.  So a question is then -- if we make a change to something like canned responses that doesn't affect saved database records, do we need an all-new definition version, such as `za-v2`?  I am not sure how to handle this... there is some overhead to creating a new Definition Version, but also overhead to separating it out.
2. As you mention in the investigation document, we want a way to have a counselor/supervisor edit these.
3. Creating separate flex plugins.  I don't think we need to do that now, but this may be a good thing to separate eventually.

plugin-hrm-form/src/formDefinitions/za-v1/CannedResponses.json
andvirga: Good catch! I've added new lines to other responses as well.

plugin-hrm-form/src/components/CannedResponses.tsx
andvirga: Got it! I've replaced these material UI components with native HTML controls.

andvirga:
> Just a couple questions/comments inline.
> 
> There are a few larger points this PR touches on. I'll mention them here, but they are not blocking for this PR:
> 
> 1. We're putting a lot into our configuration `Definition Versions`.  One aspect of those is that they are versioned so that we can pull things out of the database properly.  So a question is then -- if we make a change to something like canned responses that doesn't affect saved database records, do we need an all-new definition version, such as `za-v2`?  I am not sure how to handle this... there is some overhead to creating a new Definition Version, but also overhead to separating it out.
> 2. As you mention in the investigation document, we want a way to have a counselor/supervisor edit these.
> 3. Creating separate flex plugins.  I don't think we need to do that now, but this may be a good thing to separate eventually.

About comment #1:
Good observation. I'm also not quite sure which is the best approach here. Strictly speaking, `CannedResponses` shouldn't be part of the Form Definitions (because it's not being used to build forms, only to populate a dropdown). However, there a lot of benefits to include that information there (easy to access from Flex-Plugins, easy to modify, etc.)
To answer your question I believe that when we modify something that is not related to the forms themselves (i.e: CannedResponses, OfficeInformation, etc.) **we don't need to create a new definition version**. However, we must create a new definition version if we have modified an entity that is used to build the forms (i.e: ChildInformation, CallerInfo, etc.)

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -190,6 +191,7 @@ const setUpComponents = setupObject => {
   }
 
   Components.setUpStandaloneSearch();
+  Components.setupCannedResponses();
```
nickhurlburt, line 194: Following our discussion today... can we wrap this in a feature flag?  This means we will be able to deploy the code before we want users to see it.  You can see an example of using feature flags at line 181 above.

plugin-hrm-form/src/formDefinitions/za-v1/CannedResponses.json
```diff
@@ -0,0 +1,106 @@
+[
```
nickhurlburt, line 1: I'd expect that Dee will want to make more edits to this, but we can get it on to Development as it is now.

nickhurlburt (approved):
Looks good!  As mentioned inline, it'd be good to wrap this in a feature flag

plugin-hrm-form/src/HrmFormPlugin.js
```diff
@@ -13,6 +13,7 @@ import setUpMonitoring from './utils/setUpMonitoring';
 import * as TransferHelpers from './utils/transfer';
 import { changeLanguage } from './states/configuration/actions';
 import { issueSyncToken } from './services/ServerlessService';
+import CannedResponses from './components/CannedResponses';
```
GPaoloni, line 16: Not used I believe?

plugin-hrm-form/src/components/CannedResponses.tsx
```diff
@@ -0,0 +1,67 @@
+/* eslint-disable react/prop-types */
+import React from 'react';
+import { connect, ConnectedProps } from 'react-redux';
+import { Actions, Template, withTheme } from '@twilio/flex-ui';
+import {} from '@material-ui/core';
```
GPaoloni, line 5: Delete this line?

plugin-hrm-form/src/components/CannedResponses.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import {
+  Box,
+  CannedResponsesContainer,
+  FormLabel,
+  FormSelect,
+  FormSelectWrapper,
+  FormError,
+  FormOption,
```
GPaoloni, line 14: Good reuse of this styles!

plugin-hrm-form/src/components/CannedResponses.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const mapStateToProps = (state: RootState) => {
+  return {
+    state,
+    cannedResponses: state[namespace][configurationBase].currentDefinitionVersion?.cannedResponses,
+  };
+};
+
+const connector = connect(mapStateToProps);
+export default connector(withTheme(CannedResponses));
```
GPaoloni, line 67: What purpose is serving withTheme here?

GPaoloni (commented):
Looks great! Random comments that maybe you already addressed in following PRs.

PR #409, opened 2021-04-01
Update IssueCategorizationTab.json

dee-luo:
ZA Updated Categories April 1


andvirga (approved):
The code looks good! Also, I've tested this functionality in my local environment and everything is working ok!

PR #412, opened 2021-04-06
Add customization support for case statuses

GPaoloni:
This PR is a reopen of https://github.com/tech-matters/flex-plugins/pull/400

Jira: https://bugs.benetech.org/browse/CHI-567

This PR:

- Adds a new property to the case reducer: prevStatus, which stores the case status as it comes from the DB.
- Adds case statuses as part of customization (each status describes the possible transitions to other statuses).
- Using above information, decide when and to which statuses the UI allows to transition at a given time. This also takes into considerations permissions of the current counselor (please suggest changes / point to mistakes on this!).

WIP:
- ~CI.~
- ~Reflect changes on the backend.~ Done in https://github.com/tech-matters/hrm/pull/107

### DIFF
--- a/plugin-hrm-form/src/components/case/Case.tsx
+++ b/plugin-hrm-form/src/components/case/Case.tsx
@@ -176,6 +176,21 @@ const Case: React.FC<Props> = props => {
     }
   }, [definitionVersions, updateDefinitionVersion, version]);
 
+  // Memoize can function so is not re-created on every render of case view, but only when relevan case info changes
+  const { can } = React.useMemo(
+    () =>
+      getPermissionsForCase(
+        props.connectedCaseState.connectedCase.info.definitionVersion || 'za-v1',
+        props.connectedCaseState.connectedCase.twilioWorkerId,
+        props.connectedCaseState.connectedCase.status, // should this be prevStatus instead?
+      ),
+    [
+      props.connectedCaseState.connectedCase.info.definitionVersion,
+      props.connectedCaseState.connectedCase.status,
+      props.connectedCaseState.connectedCase.twilioWorkerId,
+    ],
+  );
+
   const toggleCaseMenu = e => {
     e.persist();
     setAnchorEl(e.currentTarget || e.target);
@@ -287,7 +302,7 @@ const Case: React.FC<Props> = props => {
 
   const { task, form, counselorsHash } = props;
 
-  const { connectedCase, caseHasBeenEdited } = props.connectedCaseState;
+  const { connectedCase, caseHasBeenEdited, prevStatus } = props.connectedCaseState;
 
   const getCategories = firstConnectedContact => {
     if (firstConnectedContact?.rawJson?.caseInformation) {
@@ -382,7 +397,7 @@ const Case: React.FC<Props> = props => {
     contact: firstConnectedContact,
   };
 
-  const { can } = getPermissionsForCase(connectedCase);
+  // const { can } = getPermissionsForCase(connectedCase);
 
   switch (subroute) {
     case 'add-note':
@@ -422,6 +437,7 @@ const Case: React.FC<Props> = props => {
                 caseId={connectedCase.id}
                 name={fullName}
                 status={status}
+                prevStatus={prevStatus}
                 can={can}
                 counselor={caseCounselor}
                 categories={categories}
@@ -434,7 +450,8 @@ const Case: React.FC<Props> = props => {
                 handleInfoChange={onInfoChange}
                 handleStatusChange={onStatusChange}
                 handleClickChildIsAtRisk={onClickChildIsAtRisk}
-                definitionVersion={connectedCase.info.definitionVersion}
+                definitionVersion={definitionVersion}
+                definitionVersionName={connectedCase.info.definitionVersion}
                 isOrphanedCase={!firstConnectedContact}
               />
             </Box>
--- a/plugin-hrm-form/src/components/case/Case.tsx
+++ b/plugin-hrm-form/src/components/case/Case.tsx
@@ -227,7 +227,7 @@ const Case: React.FC<Props> = props => {
       const contact = await submitContactForm(task, form, connectedCase);
       await updateCase(connectedCase.id, { ...connectedCase });
       await connectToCase(contact.id, connectedCase.id);
-      props.markCaseAsUpdated(task.taskSid);
+      props.markCaseAsUpdated(task.taskSid); // (Gian): is this necessary?
       await completeTask(task);
     } catch (error) {
       console.error(error);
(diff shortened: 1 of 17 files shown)


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -176,6 +176,21 @@ const Case: React.FC<Props> = props => {
     }
   }, [definitionVersions, updateDefinitionVersion, version]);
 
+  // Memoize can function so is not re-created on every render of case view, but only when relevan case info changes
```
andvirga, line 179: Small typo: "relevant" 
Forgive me for my OCD

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -176,6 +176,21 @@ const Case: React.FC<Props> = props => {
...
+        props.connectedCaseState.connectedCase.twilioWorkerId,
+        props.connectedCaseState.connectedCase.status, // should this be prevStatus instead?
+      ),
+    [
+      props.connectedCaseState.connectedCase.info.definitionVersion,
+      props.connectedCaseState.connectedCase.status,
+      props.connectedCaseState.connectedCase.twilioWorkerId,
+    ],
+  );
+
```
andvirga, line 193: Great job including this in a useMemo

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -382,7 +397,7 @@ const Case: React.FC<Props> = props => {
     contact: firstConnectedContact,
   };
 
-  const { can } = getPermissionsForCase(connectedCase);
+  // const { can } = getPermissionsForCase(connectedCase);
```
andvirga, line 400: Remove this line

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -434,7 +450,8 @@ const Case: React.FC<Props> = props => {
                 handleInfoChange={onInfoChange}
                 handleStatusChange={onStatusChange}
                 handleClickChildIsAtRisk={onClickChildIsAtRisk}
-                definitionVersion={connectedCase.info.definitionVersion}
+                definitionVersion={definitionVersion}
+                definitionVersionName={connectedCase.info.definitionVersion}
```
andvirga, line 454: Are these two values different from each other?

plugin-hrm-form/src/components/case/CaseDetails.jsx
```diff
@@ -41,22 +38,45 @@ const CaseDetails = ({
...
-  const initialColor = (statusOptions.find(x => x.value === status) || {}).color || '#000000';
+      return true;
+    };
 
-  const [color, setColor] = useState(initialColor);
+    const renderStatusOptions = o => {
+      const disabled = !enableBasedOnPermissions(o);
+
+      return (
+        <FormOption key={o.value} value={o.value} style={{ color: '#000000' }} disabled={disabled}>
```
andvirga, line 63: It seems that we don't need to specify color here since some lines forward we're defining the color of each custom status. But I may be wrong.

plugin-hrm-form/src/components/common/forms/types.ts
```diff
@@ -169,6 +169,13 @@ export type LayoutVersion = {
   };
 };
 
+type StatusInfo = {
+  value: string;
+  label: string;
+  color: string; // color that represents this status in the UI
+  transitions: [string]; // possible statuses this one can transition to (further update may be to include who can make such a transition for a more granular control)
+};
```
andvirga, line 177: Amazing, you have built a state machine!

plugin-hrm-form/src/formDefinitions/v1/index.ts
```diff
@@ -39,6 +40,7 @@ const version: DefinitionVersion = {
     oneToOneConfigSpec: (oneToOneConfigSpec as unknown) as OneToOneConfigSpec,
     oneToManyConfigSpecs: oneToManyConfigSpecs as OneToManyConfigSpecs,
   },
+  caseStatus: (CaseStatus as unknown) as DefinitionVersion['caseStatus'],
```
andvirga, line 43: I think that you can replace `unknown` with `StatusInfo`. Am I right?

andvirga (approved):
Haven't tested the functionality but the code seems to be pretty solid!
Just left some minor comments.

nickhurlburt (approved):
I gave this a light review... looks good.  Does this work with the existing permissions on the backend to allow the transitions we want to allow?  For example, will a normal counselor be permitted to change 'open' to 'in progress'?

plugin-hrm-form/src/components/case/Case.tsx
GPaoloni: Thanks. I support this OCD haha.

plugin-hrm-form/src/components/case/Case.tsx
GPaoloni: Yes, the first one is the definition object (where the statuses are bundled), while the second one is the version name (a string).

plugin-hrm-form/src/formDefinitions/v1/index.ts
GPaoloni: Fixed in last commit, thanks!

GPaoloni:
> Does this work with the existing permissions on the backend to allow the transitions we want to allow? For example, will a normal counselor be permitted to change 'open' to 'in progress'?


@nickhurlburt  It did worked as it was before, but I've added a new rule to consider this cases in https://github.com/tech-matters/hrm/pull/107. Right now, only the creator can do that.

murilovmachado (approved):
Looks great, Gian! Just some points on `markCaseAsUpdated` and using `prevStatus` instead of `status`.

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -305,7 +320,7 @@ const Case: React.FC<Props> = props => {
 
     try {
       const updatedCase = await updateCase(connectedCase.id, { ...connectedCase });
-      props.markCaseAsUpdated(task.taskSid);
```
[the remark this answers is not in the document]
GPaoloni: You are right, but if I replace this action with `setConnectedCase(updatedCase, task.taskSid, false);`, won't that result in the correct behavior? I've replaced it because that fixes another bug that was about to be introduced in this PR.

Screenshot from 2021-04-07 10-39-55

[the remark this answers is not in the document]
murilovmachado: Oh, I didn't see the third attribute (`false`) on `setConnectedCase`. In this case, the code looks right!

PR #415, opened 2021-04-14
Click View Records

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-596
Primary reviewer: @GPaoloni 

Implements clicking on PreviousContactsBanner **View Records**.

image

P.S.: I've decided not to transform some files into Typescript just yet, because it would show up as more changes in Git, making it harder to review.

WIP:
- ~Bug: Blinks SearchForm before showing SearchResults~
- ~Bug: Refresh. Page > Click View Records > Click any other tabs: they have blank content~
- ~Change Checkbox color to blue~
- ~Change Back label to Categorize Contact~
- ~Bug: checkbox state gets incorrect when coming back from search results~
- ~Tests~

### DIFF
--- a/plugin-hrm-form/src/components/PreviousContactsBanner.tsx
+++ b/plugin-hrm-form/src/components/PreviousContactsBanner.tsx
@@ -9,6 +9,7 @@ import {
   searchContacts as searchContactsAction,
   searchCases as searchCasesAction,
   handleSearchFormChange as handleSearchFormChangeAction,
+  changeSearchPage as changeSearchPageAction,
 } from '../states/search/actions';
 import { namespace, searchContactsBase, configurationBase, RootState } from '../states';
 import { getNumberFromTask } from '../services/ContactService';
@@ -18,6 +19,7 @@ import { Bold } from '../styles/HrmStyles';
 import { StyledLink } from '../styles/search';
 import { ChannelTypes, channelTypes } from '../states/DomainConstants';
 import { changeRoute as changeRouteAction } from '../states/routing/actions';
+import { SearchPages } from '../states/search/types';
 
 type OwnProps = {
   task: CustomITask;
@@ -42,6 +44,7 @@ const PreviousContactsBanner: React.FC<Props> = ({
   searchCases,
   changeRoute,
   handleSearchFormChange,
+  changeSearchPage,
 }) => {
   useEffect(() => {
     if (task && task.attributes && !task.attributes.isContactlessTask && previousContacts === undefined) {
@@ -66,8 +69,9 @@ const PreviousContactsBanner: React.FC<Props> = ({
     const searchParams = { contactNumber };
     searchContacts(searchParams, counselorsHash, CONTACTS_PER_PAGE, 0);
     searchCases(searchParams, counselorsHash, CASES_PER_PAGE, 0);
-    changeRoute({ route: 'tabbed-forms', subroute: 'search' });
+    changeSearchPage(SearchPages.resultsContacts);
     handleSearchFormChange('contactNumber', contactNumber);
+    changeRoute({ route: 'tabbed-forms', subroute: 'search' });
   };
 
   const { contactsCount, casesCount } = previousContacts;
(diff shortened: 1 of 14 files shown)


plugin-hrm-form/src/components/PreviousContactsBanner.tsx
```diff
@@ -52,13 +64,23 @@ const PreviousContactsBanner: React.FC<Props> = ({
 
   if (!shouldDisplayBanner) return null;
 
+  const handleClickViewRecords = () => {
+    const contactNumber = getNumberFromTask(task as ITask);
```
GPaoloni, line 68: If you want to avoid using this typecastings, you could do something like
```
const handleClickViewRecords = () => {
  if (isOfflineContactTask || isInMyBehalfITask) return;
  // code .. in the subsequent execution branch TS will know that this is an ITask
}
```

plugin-hrm-form/src/components/PreviousContactsBanner.tsx
```diff
@@ -52,13 +64,23 @@ const PreviousContactsBanner: React.FC<Props> = ({
...
+    const searchParams = { contactNumber };
+    searchContacts(searchParams, counselorsHash, CONTACTS_PER_PAGE, 0);
+    searchCases(searchParams, counselorsHash, CASES_PER_PAGE, 0);
```
GPaoloni, line 71: Isn't this information already in the state (due to a search already being done when we retrieve the count of the contacts and cases for this `contactNumber`)?

plugin-hrm-form/src/components/tabbedForms/TabbedForms.tsx
```diff
@@ -59,6 +62,8 @@ const mapTabsToIndex = (task: CustomITask, contactForm: TaskEntry): TabbedFormSu
       : ['search', 'contactlessTask', 'childInformation', 'categories', 'caseInformation'];
   }
 
+  if (isEmptyCallType(contactForm.callType)) return ['search'];
```
GPaoloni, line 65: Just out of curiosity, when this might happen?

plugin-hrm-form/src/components/tabbedForms/TabbedForms.tsx
```diff
@@ -148,14 +153,16 @@ const TabbedForms: React.FC<Props> = ({ dispatch, routing, contactForm, currentD
...
 
   return (
     <FormProvider {...methods}>
       <div role="form" style={{ height: '100%' }}>
         <TabbedFormsContainer>
-          <TopNav>
-            <TransparentButton onClick={handleBackButton}>&lt; BACK</TransparentButton>
-          </TopNav>
+          <Box marginTop="10px" marginBottom="10px">
+            <SearchResultsBackButton handleBack={handleBackButton} text={strings['TabbedForms-BackButton']} />
```
GPaoloni, line 164: Can't we pass down a template to the `text` prop? Maybe making it to be of type `string | React.Element[]` allows us to pass down a `Template` component (is just more clean than using the strings object, not a really need of changing anything).

GPaoloni (approved):
Great!! Just come questions

GPaoloni:
Another thing I've been thinking about is that I don't really like having this `FormXXXX` components as part of the styles, we could do something more like an internal components library and start importing them from a place that feels more part of the components than the styles, (`components/common/inputs` or something?). What do you think? Just an idea as future refactoring work.

plugin-hrm-form/src/components/PreviousContactsBanner.tsx
murilovmachado: Yes, but there's a scenario where:
1) Yellow Banner Show (search is done)
2) User don't click **View Records**, instead, he does a custom search
3) At some point he decides to click **View Records**, at this time, the redux state contains his custom search (not the initial one)

That's why we fetch again. Surely we can optimize it so that the initial search stores the results (not just the count), and clicking on View Records would just play with redux (without needing to fetch again).

plugin-hrm-form/src/components/tabbedForms/TabbedForms.tsx
murilovmachado: This just happens after a page refresh. We wouldn't get into that in the usual flows.

murilovmachado:
> Another thing I've been thinking about is that I don't really like having this `FormXXXX` components as part of the styles, we could do something more like an internal components library and start importing them from a place that feels more part of the components than the styles, (`components/common/inputs` or something?). What do you think? Just an idea as future refactoring work.

I agree @GPaoloni, import components from styles is really weird. It's a good place to improve our code.

PR #425, opened 2021-04-26
(CHI-612) Hiding 'Added By' if the value is the same as Counselor

andvirga:
I forgot to include this change in CHI-612:
If the user who added the record is the same as the counselor assigned to the record, we do not need to show the "Added by" row.

Primary reviewer: anyone 😃


murilovmachado (approved):
Looks good to me!

PR #427, opened 2021-04-27
Case status Changes

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-613

This PR:
- Introduces `PermissionActions.CASE_STATUS_TRANSITION` to handle transitions that are not `CLOSE_CASE` nor `REOPEN_CASE` (e.g. `open -> inProgress` transition), and treats them same way as `canEditGenericField` type rules.
- Adapts case status UI to above change.
- Changes `inProgress` status color to be the same as `open` in za-v1 specs.


murilovmachado (approved):
Code and behavior look good to me!

PR #428, opened 2021-04-27
Removing translations included in serverless (CHI-528)

andvirga:
Jira: [CHI-528](https://bugs.benetech.org/browse/CHI-528)

These translations are not handled by Flex-plugins, instead of that, they are included in `serverless`.

Related to:
- [serverless](https://github.com/tech-matters/serverless/pull/51)
- [webchat](https://github.com/tech-matters/webchat/pull/28)


GPaoloni (approved):

nickhurlburt (approved):
I'm fine with this being merged and deployed on the basis of @GPaoloni 's review.  Though, why did we move this to serverless from Flex plugins?  Was that for chatbot integration?

GPaoloni:
@nickhurlburt when we ask for something to be translated, we call this that are services to fetch from Serverless, that's why translations weren't working when bundling them with the code. Only English is bundled so we can default to something, the rest of them are fetched on demand (no caching at the moment, something we could improve).
Screenshot from 2021-04-29 10-44-15

PR #429, opened 2021-04-28
Improve permissions

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-590
Primary reviewer: @GPaoloni 

Before, we were choosing which permission's rules to use based on the cases' `definition version`.
Now, we fetch this info from each environment (as a Service Configuration attribute, name `permissionConfig`).

P.S.: I've already updated all environments with `permissionConfig`.


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -180,15 +180,10 @@ const Case: React.FC<Props> = props => {
...
-        props.connectedCaseState?.connectedCase.info.definitionVersion || 'za-v1',
         props.connectedCaseState?.connectedCase.twilioWorkerId,
         props.connectedCaseState?.prevStatus,
       ),
-    [
-      props.connectedCaseState?.connectedCase.info.definitionVersion,
-      props.connectedCaseState?.connectedCase.twilioWorkerId,
-      props.connectedCaseState?.prevStatus,
-    ],
+    [props.connectedCaseState?.connectedCase.twilioWorkerId, props.connectedCaseState?.prevStatus],
```
andvirga, line 186: Nice refactor!

andvirga (approved):
Looks good!

plugin-hrm-form/src/permissions/index.ts
```diff
@@ -18,30 +18,28 @@ export const PermissionActions = {
...
+
+export const getPermissionsForCase = (twilioWorkerId: t.Case['twilioWorkerId'], status: t.Case['status']) => {
+  const { workerSid, isSupervisor, permissionConfig } = getConfig();
+  if (!permissionConfig || !twilioWorkerId || !status) return { can: undefined };
 
-  const { workerSid, isSupervisor } = getConfig();
   const isCreator = workerSid === twilioWorkerId;
   const isCaseOpen = status === 'open';
-  const rules = rulesMap[version];
+  const rules = rulesMap[permissionConfig] || fallbackRules;
```
GPaoloni, line 42: Should we silently continue here using ZA rules? Maybe a console.error at least to raise the issue?

GPaoloni (approved):
Works fine and code is good!

nickhurlburt (approved):
> I've already updated all environments with permissionConfig.

Thank you :)

plugin-hrm-form/src/permissions/index.ts
murilovmachado: Good idea. I'll keep the fallback so it won't ever crash, but call `console.error` so we can track it.

PR #437, opened 2021-05-11
Humairaa json update 061121

Humairaa127:
Mapping blank for non-required fields to value Unknown for Insights (Both ZA and ZM), adding sub category to Physical Health for ZM

### DIFF
--- a/plugin-hrm-form/src/formDefinitions/v1/caseForms/HouseholdForm.json
+++ b/plugin-hrm-form/src/formDefinitions/v1/caseForms/HouseholdForm.json
@@ -16,7 +16,7 @@
     "type": "select",
     "label": "Relationship to child",
     "options": [
-      { "value": "Unknown", "label": "" },
+      { "value": "", "label": "" },
       { "value": "Friend", "label": "Friend" },
       { "value": "Grandparent", "label": "Grandparent" },
       { "value": "Neighbour", "label": "Neighbour" },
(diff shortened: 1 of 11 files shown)


GPaoloni (approved):
Looks fine to me. However, I'd like to raise a brief discussion. This PR is almost entirely about adding "Unknown" to dependent selects (like `District`), that needs to be replicated once for every "option dependent list", introduces more complexity and is more error prone. This could be automatically added by the customization framework based on a property we can specify, like 
```javascript
{
  "name": "district",
  "label": "District",
  "type": "dependent-select",
  "dependsOn": "province",
  "includeUnknown": true, // <<<<<<< new property that will attach { "value": "Unknown", "label": "Unknown" } to every option
  ...
}
```

Do you consider is it worth adding this "improvement" to the framework, or we should just go with the changes in this PR? @dee-luo @nickhurlburt

dee-luo:
@Humairaa127 few comments: 

ZA Forms: 
- Perpetrator and Household forms where Municipality/District are not required, I think we want to store the "Unknown" value as the value for the blank default option (similar to how you set it up for Province). 
- Incident location - looks like we are missing an explicit Unknown option. 

ZM Forms: 
- Perpetrator and Household forms Relationship to Child - looks like we are missing an explicit Unknown option

I think that's it 😅

@GPaoloni I do like that idea. Maybe we can make this update on our next implementation with a fresh set of customization forms?

dee-luo (approved):

PR #438, opened 2021-05-11
Created Flex workflow for Staging Test

alebarquin:


murilovmachado (changes requested):
My only concern right now is related to the inputs since we're still running the deployment regardless of what the user types. See comments below. Besides that, it's looking good!

PR #442, opened 2021-05-12
[Breaking change] Moved offline contact routing info to serverless

GPaoloni:
This PR is related to https://github.com/techmatters/serverless/pull/58.

IMPORTANT: This PR contains a breaking change. Do not merge until above change in serverless has been deployed to all environments.

This PR removes the `initialAttributes` from the frontend when creating an offline contact, leaving the responsibility for the offline contact routing info to `assignOfflineContact` serverless function.

### DIFF
--- a/plugin-hrm-form/src/services/formSubmissionHelpers.ts
+++ b/plugin-hrm-form/src/services/formSubmissionHelpers.ts
@@ -43,8 +43,7 @@ export const submitContactForm = async (task: CustomITask, contactForm: Contact,
 
   if (isOfflineContactTask(task)) {
     const targetSid = contactForm.contactlessTask.createdOnBehalfOf as string;
-    const initialAttributes = { helpline, channelType: 'default', isContactlessTask: true, isInMyBehalf: true };
-    const finalAttributes = buildInsightsData(initialAttributes, contactForm, caseForm);
+    const finalAttributes = buildInsightsData({}, contactForm, caseForm);
     const inBehalfTask = await assignOfflineContact(targetSid, finalAttributes);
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/services/formSubmissionHelpers.ts
```diff
@@ -43,8 +43,7 @@ export const submitContactForm = async (task: CustomITask, contactForm: Contact,
 
   if (isOfflineContactTask(task)) {
     const targetSid = contactForm.contactlessTask.createdOnBehalfOf as string;
-    const initialAttributes = { helpline, channelType: 'default', isContactlessTask: true, isInMyBehalf: true };
```
nickhurlburt, line 46: There is no harm in keeping this in for now, right?  Serverless will just overwrite these fields?

I am reluctant to commit this until serverless is updated everywhere with techmatters/serverless#58.  If we commit it, then we create a dependency that we have to remember.

nickhurlburt (commented):
See comment -- I'm wondering if we can avoid this change for now.

plugin-hrm-form/src/services/formSubmissionHelpers.ts
GPaoloni: Yes you are right, I'm tagging this as a breaking change and we can merge later on. Thanks for the catch!

nickhurlburt:
@GPaoloni is this safe to merge now?  Per previous discussion, it can be pulled to `master` rather than `staging`.

GPaoloni:
@nickhurlburt  Yes it is safe now. I'm merging, thanks!

PR #443, opened 2021-05-17
Show Ip address on Previous Contacts banner

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-630

This PR enhances the Previous Contacts Banner to show:
- IP Address or;
- Phone Number or;
- Facebook Messenger ID

image

### DIFF
--- a/plugin-hrm-form/src/components/PreviousContactsBanner.tsx
+++ b/plugin-hrm-form/src/components/PreviousContactsBanner.tsx
@@ -3,7 +3,7 @@ import React, { useEffect } from 'react';
 import { connect } from 'react-redux';
 import { Template } from '@twilio/flex-ui';
 
-import { CustomITask, isITask } from '../types/types';
+import { CustomITask, isTwilioTask } from '../types/types';
 import {
   viewPreviousContacts as viewPreviousContactsAction,
   searchContacts as searchContactsAction,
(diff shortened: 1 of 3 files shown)


nickhurlburt (approved):
Looks good.  One random comments from someone who hasn't been looking at the code much recently... so don't take this too seriously:  `isITalk` doesn't seem like a clear name to me. Is the goal is to distinguish between tasks we've made up vs tasks that involve an active conversation?  `isInboundTask`?  `isConversationalTask`?  Not sure.

murilovmachado:
> Looks good. One random comments from someone who hasn't been looking at the code much recently... so don't take this too seriously: `isITalk` doesn't seem like a clear name to me. Is the goal is to distinguish between tasks we've made up vs tasks that involve an active conversation? `isInboundTask`? `isConversationalTask`? Not sure.

@nickhurlburt You're right, `isITask` is not a clear name. I was actually struggling to come with a good name. The thing is that our task type is `type CustomITask = ITask | OfflineContactTask | InMyBehalfITask;`.

We already had these two type checkers `isOfflineContactTask` and `isInMyBehalfITask`. I wanted a type checker for the first type and couldn't figure out a good name. So I've just followed the pattern as the previously existing one **is**+**Type**. But I agree the name is not descriptive enough.

I've tried `isNotOfflineContactTaskOrInMyBehalfITask` but it didn't seem a good name. And keep repeating the check as `!isOfflineContactTask(task) && !isInMyBehalfITask(task)` wasw no ideal too. So I appreciate suggestions on what would be a better name. Your suggestion `isInboundTask ` seems like a good idea. What do you think @GPaoloni ?

GPaoloni (approved):
Great! 
I think `isInboundTask` or `isTwilioTask` are good names.

murilovmachado:
> Great!
> I think `isInboundTask` or `isTwilioTask` are good names.

Thanks, I think `isTwilioTask` is very clear

PR #444, opened 2021-05-17
Pre-survey information and helpline passed through to Flex on non-data calls

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-279.

This PR:
- Maps caller age, caller gender, task helpline attribute, child language, child age and child gender to corresponding insights fields for non data & data call types (previously only data call types behave like this). This is achieved by moving this updates on the insights object to be part of the `coreAttributes` of `baseUpdates` function.
- Removed above updates from custom insights mappings.
- Sanitized empty string values to be `null` instead as suggested in https://github.com/techmatters/flex-plugins/pull/444#issuecomment-842662511.
- Added some type information to make code easier to follow as suggested in https://github.com/techmatters/flex-plugins/pull/444#discussion_r634470439.

Questions for @dee-luo (SOLVED):
- ~Because of the way the forms works, if there is no "caller age" (for example), this will impact in insights as empty string in `conversation_attribute_3`. Do we want it to be `null` instead? I ignore if empty string vs `null` affects insights, but maybe you know.~
- ~Previously only ZA was recording `helpline` in `conversation_attribute_8`, but we want all of the helplines to record it, right?~

Note:
Because of this new fields being added to `coreAttributes`, if any subsequent update of insights object changes one of the already assigned fields, it will be overwritten. For example, if in three months from now we add "is child at risk" to be stored in `conversation_attribute_3` from one of our "customizable insights maps", caller age won't be part of any `conversation_attribute_X`. This is not an issue as long as we remember what fields should not be customizable for the insights mapping.

### DIFF
--- a/plugin-hrm-form/src/___tests__/services/InsightsService.test.js
+++ b/plugin-hrm-form/src/___tests__/services/InsightsService.test.js
@@ -36,6 +36,7 @@ test('saveInsightsData for non-data callType', async () => {
     conversations: {
       content: 'content',
     },
+    helpline: 'helpline',
   };
 
   const twilioTask = {
@@ -45,20 +46,35 @@ test('saveInsightsData for non-data callType', async () => {
 
   const contactForm = {
     callType: 'Abusive',
+    callerInformation: {
+      age: '26',
+      gender: 'Girl',
+    },
+    childInformation: {
+      age: '18',
+      gender: 'Boy',
+      language: 'language',
+    },
   };
 
   await saveInsightsData(twilioTask, contactForm);
 
   const expectedNewAttributes = {
-    taskSid: 'task-sid',
-    channelType: 'sms',
+    ...previousAttributes,
     conversations: {
       content: 'content',
       conversation_attribute_2: 'Abusive',
       conversation_attribute_5: null,
       communication_channel: 'SMS',
+      conversation_attribute_3: contactForm.callerInformation.age,
+      conversation_attribute_4: contactForm.callerInformation.gender,
+      conversation_attribute_8: previousAttributes.helpline,
+      language: contactForm.childInformation.language,
     },
-    customers: {},
+    customers: {
+      year_of_birth: contactForm.childInformation.age,
+      gender: contactForm.childInformation.gender,
+    },
   };
 
   expect(twilioTask.setAttributes).toHaveBeenCalledWith(expectedNewAttributes);
@@ -87,6 +103,11 @@ test('saveInsightsData for data callType', async () => {
     childInformation: {
       age: '13-15',
       gender: 'Boy',
+      language: 'language',
+    },
+    callerInformation: {
+      age: '26',
+      gender: 'Girl',
     },
     caseInformation: {},
     categories: [
@@ -148,6 +169,11 @@ test('Handles contactless tasks', async () => {
     childInformation: {
       age: '3',
       gender: 'Unknown',
+      language: 'language',
+    },
+    callerInformation: {
+      age: '26',
+      gender: 'Girl',
     },
     caseInformation: {},
     categories: ['categories.Violence.Unspecified/Other'],
--- a/plugin-hrm-form/src/___tests__/services/InsightsService.test.js
+++ b/plugin-hrm-form/src/___tests__/services/InsightsService.test.js
@@ -80,6 +80,57 @@ test('saveInsightsData for non-data callType', async () => {
   expect(twilioTask.setAttributes).toHaveBeenCalledWith(expectedNewAttributes);
 });
 
+test('saveInsightsData for non-data callType (test that fields are sanitized)', async () => {
+  const previousAttributes = {
+    taskSid: 'task-sid',
+    channelType: 'sms',
+    conversations: {
+      content: 'content',
+    },
+    helpline: 'helpline',
+  };
+
+  const twilioTask = {
+    attributes: previousAttributes,
+    setAttributes: jest.fn(),
+  };
+
+  const contactForm = {
+    callType: 'Abusive',
+    callerInformation: {
+      age: '',
+      gender: '',
+    },
+    childInformation: {
+      age: '',
+      gender: '',
+      language: '',
+    },
+  };
+
+  await saveInsightsData(twilioTask, contactForm);
+
+  const expectedNewAttributes = {
+    ...previousAttributes,
+    conversations: {
+      content: 'content',
+      conversation_attribute_2: 'Abusive',
+      conversation_attribute_5: null,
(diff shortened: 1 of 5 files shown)


dee-luo:
@GPaoloni 

1. I believe empty string and null are the same in Insights - will just show as (empty value). Either should be fine!
2. Yes, let's record for all helplines. 

Thanks!

nickhurlburt:
I'm not convinced this is the right design, for the following reasons:
- While I could certainly be wrong, I don't think `null` and `""` are the same in Insights.  It _feels_ wrong to add `""` for every single record whether or not these attributes are present.  I could be convinced otherwise.  There's just a feeling that it's a bad idea.
- `baseUpdates` are things we think every helpline will use.  Helplines that don't use a chatbot won't need these attributes. We should only add something to `baseUpdates` if we're pretty sure that every future helpline will want it.

It's between you and @dee-luo how exactly to implement this and if you decide to merge this as is, that's okay.  Here are some possible suggestions I'll make:
- If we have nothing to say about an attribute, we should use `null` rather than `""`.
- Consider separating the non-data values into a new function separate from `baseUpdates`.  Even if they are still hard-coded, that will make the code more clear.  The `InsightsService` is designed to be a series of layers.
- A step further from the above is to change our JSON configs to allow it to specify values that get stored even in the event of non-data contacts.  That may be too much work now, but worth a thought.

Insights is one of the trickier parts of our system, so we want to take additional care to how well it's being factored, and be especially wary of quick fixes or hacks.  The customer needs are only going to get more complicated, so we need to make a point to make the code simpler.

GPaoloni:
> While I could certainly be wrong, I don't think null and "" are the same in Insights. It feels wrong to add "" for every single record whether or not these attributes are present. I could be convinced otherwise. There's just a feeling that it's a bad idea.

I agree with this, changing to `null`.

> baseUpdates are things we think every helpline will use. Helplines that don't use a chatbot won't need these attributes. We should only add something to baseUpdates if we're pretty sure that every future helpline will want it.

About this, I may be wrong, but I understand that this updates will be for every helpline, that's what motivated me to include them in `baseUpdates`. Do you think is not a good place?

> A step further from the above is to change our JSON configs to allow it to specify values that get stored even in the event of non-data contacts. That may be too much work now, but worth a thought.

This is a great idea, will address it as a future improvement!

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -97,13 +95,26 @@ export const baseUpdates: InsightsUpdateFunction = (
        */
       conversation_attribute_5: null,
       communication_channel,
+      /**
+       * Fields that should always be populated, whether it is a data contact or not.
+       */
+      conversation_attribute_3: contactForm.callerInformation.age.toString(),
+      conversation_attribute_4: contactForm.callerInformation.gender.toString(),
+      conversation_attribute_8: taskAttributes.helpline,
```
murilovmachado, line 103: At some point in future we should use `consts` instead of `conversation_attribute_3 `, `conversation_attribute_4 ` etc.

Something like:
```
  [CALLER_AGE]: contactForm.callerInformation.age.toString(),
  [CALLER_GENDER]: contactForm.callerInformation.gender.toString(),
  [HELPLINE]: taskAttributes.helpline,
```

It will improve readability and make it less error prone.

--------------
**EDIT:** I'm actually not sure if javascript syntax `[varName]` is the best way to go. A distracted developer could easily miss the brackets, making things not to be saved on insights. 

Let's keep this way for now.

--------------
**EDIT 2:** Actually, if we make `type InsightsAttributes` more strict, we could go with the syntax `[varName]`, and lint will alert in case the developer misses the brackets.

`CALLER_AGE: contactForm.callerInformation.age.toString(),` would throw lint error, where `[CALLER_AGE]: contactForm.callerInformation.age.toString(),` would work fine.

But again, we can do it in a later PR.

murilovmachado (approved):

plugin-hrm-form/src/services/InsightsService.ts
GPaoloni: I really like this idea!

plugin-hrm-form/src/___tests__/services/InsightsService.test.js
```diff
@@ -45,20 +46,86 @@ test('saveInsightsData for non-data callType', async () => {
...
+    customers: {
+      year_of_birth: contactForm.childInformation.age,
+      gender: contactForm.childInformation.gender,
+    },
+  };
+
+  expect(twilioTask.setAttributes).toHaveBeenCalledWith(expectedNewAttributes);
+});
+
+test('saveInsightsData for non-data callType (test that fields are sanitized)', async () => {
```
GPaoloni, line 83: This case tests that the empty string values are indeed nullified for the core attributes.

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -69,14 +69,65 @@ type InsightsUpdateFunction = (
...
+const HELPLINE = 'conversation_attribute_8';
+const LANGUAGE = 'language';
+const CHILD_AGE = 'year_of_birth';
+const CHILD_GENDER = 'gender';
+
+/**
+ * This are the core attributes that should be present for all kind of contacts, returned by baseUpdates function.
+ * Note this is a specialization of the InsightsAttributes type
+ */
+type CoreAttributes = {
```
GPaoloni, line 94: Here I decided to declare a new type `CoreAttributes` instead of restricting `InsightsAttributes` because those fields should be used only within `baseUpdates` (those updates that hold for all calltypes and all helplines). If we restrict `InsightsAttributes`, that will lead to more changes in the code for those functions that types its return as `InsightsAttributes` while returnin `{ }` (empty object). More updates can follow this PR, if you agree, where we can make three types:
- `CoreAttributes` those that should be present for all tasks.
- `CustomAttributes` those that can be customized (that will be the same as `InsightsAttributes` EXCLUDING the fields of `CoreAttributes`).
- `InsightsAttributes`: the union of the two above (will have same structure as `CoreAttributes`, as the fields of `CustomAttributes` will be optional, but that way we can distinct the semantic meaning of each type).

Do you think that way this code will look clearer? If so, is it future work or should I address that before merging this PR?

nickhurlburt (approved):
Looks good

plugin-hrm-form/src/services/InsightsService.ts
murilovmachado: I think it looks really good the way it is right now, you can definitely merge it. You're suggestion is also great, but we can leave it for later.

PR #446, opened 2021-05-25
Make focus visible on CallTypeButtons, TabbedForms and Case screens

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-641

All interactable elements are reachable by tab navigation, and they will display an outline when theyér focused.


nickhurlburt (approved):
One thing I notice is that the list vs grid view buttons on categories don't seem to respond to hitting 'enter'... it would be good to check on that.

Otherwise, I think this is good to get onto the development account so that people can play with it.  I am not sure whether we should merge to `staging` or just deploy off the branch.  It's possible that we'll decide not to include this in a release until more work in this area is complete.  So you could either 1) merge to `staging` and deploy, keeping in mind that you may have to revert the PR, _or_ 2) deploy this branch to Dev, in which case we need to remember that state.

Also good for @GPaoloni to review the code in more detail.

plugin-hrm-form/src/components/Section.jsx
```diff
@@ -1,24 +1,29 @@
...
     <SectionTitleContainer color={color}>
-      <ButtonBase style={{ width: '100%', padding: 0 }} onClick={handleExpandClick} data-testid={buttonDataTestid}>
+      <SectionTitleButton onClick={handleExpandClick} data-testid={buttonDataTestid}>
         <SectionTitleText>{sectionTitle}</SectionTitleText>
         {!hideIcon && (expanded ? <ArrowUpIcon /> : <ArrowDownIcon />)}
-      </ButtonBase>
+      </SectionTitleButton>
     </SectionTitleContainer>
-    <Collapse in={expanded} timeout="auto">
+    <SectionCollapse in={expanded} timeout="auto">
```
GPaoloni, line 24: Maybe just name this prop `expanded` instead of `in`? Describes better what it does imo.

GPaoloni (approved):
Looks great! One further update we could do (based on Monday's a11y meeting) is define a single CSS class that adds the outline (color, distance from the element like a margin, etc) and use that class on every focusable React component (not sure if that works tho, maybe Styled overrides it).

murilovmachado:
> One thing I notice is that the list vs grid view buttons on categories don't seem to respond to hitting 'enter'... it would be good to check on that.

I've seen that issue.  I can work on that, but keep in mind that there are other components that are not fully responding to keyboard interaction either (such as the drop downs on search tab). Also, in the next iteration of this screen, these buttons will be removed. That's why I didn't spend more time making it work, but I can do it. Let me know if it's not needed.

> Otherwise, I think this is good to get onto the development account so that people can play with it. I am not sure whether we should merge to `staging` or just deploy off the branch. It's possible that we'll decide not to include this in a release until more work in this area is complete. So you could either 1) merge to `staging` and deploy, keeping in mind that you may have to revert the PR, _or_ 2) deploy this branch to Dev, in which case we need to remember that state.

I think it's better to go with 2: Deploy from the branch and not merge it yet.

murilovmachado:
> Looks great! One further update we could do (based on Monday's a11y meeting) is define a single CSS class that adds the outline (color, distance from the element like a margin, etc) and use that class on every focusable React component (not sure if that works tho, maybe Styled overrides it).

I like the idea, but there's a big chance that different components will have different outline styles (bottom bar buttons vs call type buttons vs tabs vs back button). Let's wait until we have decided on the designs (how they will look), before making these code changes.

murilovmachado:
@nickhurlburt @GPaoloni This is ready for review. Since this include vey old code that has already been reviewed, I believe you can just review this [commit](https://github.com/techmatters/flex-plugins/pull/446/commits/4d6622aa63ac063484fcc6439ff6a533ee52eb94) (and maybe subsequent ones).

Besides styling changes, the main thing this PR does for focus management is:
1. When calling `changeRoute` we now pass another parameter: `autoFocus` (could be true or false);
2. When calling `createFormFromDefinition`(to create a form) we pass the `autoFocus` parameter (usually get from the route);
3. Internally, when calling `getInputType` to build the elements from the definition, we pass `autoFocus=true` to the first element of the form. (Here is where we really set an input to be autoFocused, leveraging HTML's `autoFocus` property)

nickhurlburt (approved):
Looks good! Did some light QA on this and it works well for the contact and case forms (the search form is different than the others, but we know we need to convert that). 

Looked at the code and I have the general idea. It's still unclear how we default to focusing on the first element in the form... does that just come from calling `useRef(null)`?

murilovmachado:
> Looked at the code and I have the general idea. It's still unclear how we default to focusing on the first element in the form... does that just come from calling `useRef(null)`?

Let's take `Child Info` tab as an example:
```
const shouldFocusFirstElement = display && autoFocus;
const firstElementRef = useFocus(shouldFocusFirstElement);
```
This first piece tells:
1. The condition to have focus: `const shouldFocusFirstElement = display && autoFocus;`
2. An **elementRef** that will be triggered focus when the condition is **true**: `const firstElementRef = useFocus(shouldFocusFirstElement);`

But until now this **elementRef** is not bound to any input yet. Next, this component makes a call to create the form, passing this **elementRef**:
```
const tab = createFormFromDefinition(formDefinition)(['contactlessTask'])(initialForm, firstElementRef)(
      updateCallBack,
    );
```
The form generated by this call will have its **first input** bound to the given **elementRef**. So it "knows" when to be focused (when the above condition is true).

PR #448, opened 2021-05-27
[Don't merge] Initialize plugin with garbled strings

GPaoloni:
This PR initializes the plugin with translation instead than usual English. The purpose is to test if we can properly translate the UI avoiding the async behavior of fetching translations.


nickhurlburt:
@GPaoloni can this be closed?

PR #450, opened 2021-06-01
Create CannedResponses.json

Humairaa127:
ZM canned responses v1

### DIFF
--- a/plugin-hrm-form/src/formDefinitions/v1/CannedResponses.json
+++ b/plugin-hrm-form/src/formDefinitions/v1/CannedResponses.json
@@ -1,0 +1,62 @@
+[
+  {
+    "label": "How do people contract HIV?",
+    "text": "HIV (human immunodeficiency virus) is a virus that attacks cells that help the body fight infection, making a person more likely to get other infections and diseases. The most common way that someone can get HIV is by having unprotected sex (anal, vaginal, or oral) with a person who has HIV or sharing drug needles or other drug equipment with someone who has HIV. If you're worried about HIV you need to get help quickly and the best place to go is to your local health clinic where they have special safe spaces for teenagers and young adults."
+  },
+  {
+    "label": "What are the Early signs and symptoms?",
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/formDefinitions/v1/CannedResponses.json
```diff
@@ -0,0 +1,62 @@
...
+  },
+  {
+    "label": "What is a Condom and how is it used?",
+    "text": "A condom is a sheath-shaped barrier device used during sexual intercourse to reduce the probability of pregnancy or a sexually transmitted infection. There are both male and female condoms. ; •	Using a condom properly every time you have sex is the best way of stopping you getting HIV and preventing pregnancy. There are different condoms for men and women and not everyone knows how to use them correctly, which is very important. So if you’re not sure or need free condoms the best place to go is to your local health clinic where you can talk to a service provider."
```
murilovmachado, line 12: This string contains a **tab** within it at `There are both male and female condoms. ; • (hidden tab here) Using a condom`
This is causing issues.

Can I just remove the tab part and it will be: `There are both male and female condoms. ; Using a condom`

murilovmachado (changes requested):
There's a hidden **tab** in one of the strings. Let me know if you or I can remove it.

murilovmachado:
Also, in order for these changes to be visible, we'll also need to add canned responses to `v1` definitions.
(I can do this on a separate PR)

On `src/formDefinitions/v1/index.ts`, we'll need to add:
```
...
cannedResponses: CannedResponses as CannedResponsesDefinitions,
...
```

plugin-hrm-form/src/formDefinitions/v1/CannedResponses.json
murilovmachado: I'll fix this tomorrow and merge it.

Humairaa127: fixed

murilovmachado (approved):

PR #452, opened 2021-06-09
Localization of strings

jennxu23:
Jira: https://bugs.benetech.org/secure/RapidBoard.jspa?rapidView=19&view=detail&selectedIssue=CHI-657

This PR cleans up some of the non-translated strings.

In this PR:
- Localized strings listed in Jira
- Fixed bug relating to pluralization in search menu (i.e 1 contact returned vs. 1 contacts returned)

Notes:
- As mentioned during the standup, instead of translating the placeholder when adding a note, we're taking out the placeholder entirely. That'll be in a separate PR, so the placeholder is untranslated currently.
- In the worker role, the case summary placeholder says "No case summary". In the supervisor role, the case summary placeholder says "Add case summary here". This is true both when searching for cases and when creating new cases.

Screen Shot 2021-06-09 at 3 21 28 PM
Screen Shot 2021-06-09 at 3 16 17 PM
Screen Shot 2021-06-09 at 3 17 38 PM


plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -134,6 +135,7 @@ const SearchResults: React.FC<Props> = ({
   const { count: casesCount, cases } = searchCasesResults;
   const contactsPageCount = Math.ceil(contactsCount / CONTACTS_PER_PAGE);
   const casesPageCount = Math.ceil(casesCount / CASES_PER_PAGE);
+  const { strings } = getConfig();
```
GPaoloni, line 138: Can you remove this, as it is not used? (and the above import that will be unused once you remove this).

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -183,14 +185,33 @@ const SearchResults: React.FC<Props> = ({
...
+                ) : (
+                  <>
+                    {casesCount}{' '}
+                    {casesCount === 1 ? (
+                      <Template code="PreviousContacts-Case" />
+                    ) : (
+                      <Template code="PreviousContacts-Cases" />
+                    )}
+                  </>
+                )}
```
GPaoloni, line 210: This is great! But the order is the inverse. We want to show "There are x cases" while looking at the contacts tab, as the contacts count is shown in the line below, and the same goes for cases tab.
Screenshot from the current state in Development environment:
Screenshot from 2021-06-11 12-12-02

GPaoloni (approved):
Looks very good! Just left some comments, but I'm approving the PR as no need for a second review once you address them. Also, can you leave out the `appConfig.js` file? I'm surprised it got pushed, as it is listed in the `.gitignore`.

plugin-hrm-form/src/components/case/AddPerpetrator.tsx
```diff
@@ -91,7 +91,11 @@ const AddPerpetrator: React.FC<Props> = ({
     const perpetrator = transformValues(PerpetratorForm)(temporaryCaseInfo.info);
     const createdAt = new Date().toISOString();
     const { workerSid } = getConfig();
-    const newPerpetrator = { perpetrator, createdAt, twilioWorkerId: workerSid };
+    const newPerpetrator = {
+      perpetrator,
+      createdAt,
+      twilioWorkerId: workerSid,
+    };
```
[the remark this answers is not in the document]
jennxu23: I think that change came from ESLint automatically modifying it.

GPaoloni (approved):
Thanks! :rocket:

PR #454, opened 2021-06-10
Add Helpline Select on ZM offline contact form

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-633

This PR adds a **Helpline** select on **Contact Info** Tab when creating an **offline contact** on **ZM**.

P.S.: The easiest way to test this is to edit **HrmFormPlugin** line 68 to be `definitionVersion: 'v1',` because **dev** environment is configured on Twilio to use ZA definitions. 

image

### DIFF
--- a/plugin-hrm-form/src/formDefinitions/v1/tabbedForms/ContactlessTaskTab.json
+++ b/plugin-hrm-form/src/formDefinitions/v1/tabbedForms/ContactlessTaskTab.json
@@ -1,0 +1,13 @@
+[
+  {
+    "name": "helpline",
+    "label": "Helpline",
+    "type": "select",
+    "options": [
+      { "value": "Unknown", "label": "" },
+      { "value": "ChildLine", "label": "ChildLine" },
+      { "value": "LifeLine", "label": "LifeLine" }
+    ],
+    "required": { "value": true, "message": "RequiredFieldError" }
+  }
+]
--- a/plugin-hrm-form/src/formDefinitions/v1/tabbedForms/ContactlessTaskTab.json
+++ b/plugin-hrm-form/src/formDefinitions/v1/tabbedForms/ContactlessTaskTab.json
@@ -3,8 +3,8 @@
     "name": "helpline",
     "label": "Helpline",
     "type": "select",
+    "defaultOption": "ChildLine",
     "options": [
-      { "value": "Unknown", "label": "" },
       { "value": "ChildLine", "label": "ChildLine" },
       { "value": "LifeLine", "label": "LifeLine" }
     ],
--- a/plugin-hrm-form/src/services/ContactService.ts
+++ b/plugin-hrm-form/src/services/ContactService.ts
@@ -16,7 +16,7 @@ import type {
   FormDefinition,
   FormItemDefinition,
 } from '../components/common/forms/types';
-import { InformationObject, ContactRawJson, SearchContactResult, isOfflineContactTask, CustomITask } from '../types/types';
+import { InformationObject, ContactRawJson, SearchContactResult, isOfflineContactTask } from '../types/types';
 
 /**
  * Un-nests the information (caller/child) as it comes from DB, to match the form structure
--- a/plugin-hrm-form/src/services/ContactService.ts
+++ b/plugin-hrm-form/src/services/ContactService.ts
@@ -16,7 +16,7 @@ import type {
   FormDefinition,
   FormItemDefinition,
 } from '../components/common/forms/types';
-import { InformationObject, ContactRawJson, SearchContactResult, isOfflineContactTask } from '../types/types';
+import { InformationObject, ContactRawJson, SearchContactResult, isOfflineContactTask, CustomITask } from '../types/types';
 
 /**
  * Un-nests the information (caller/child) as it comes from DB, to match the form structure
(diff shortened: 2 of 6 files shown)


plugin-hrm-form/src/formDefinitions/v1/tabbedForms/ContactlessTaskTab.json
```diff
@@ -0,0 +1,13 @@
+[
```
nickhurlburt, line 1: It feels a little weird to me that this tab is part fixed (via `ContactlessTaskTabDefinition`) and part based on a configurable `formDefinition`.  Is that going to cause trouble for us?  I am okay with it if both of you are, but I want to ask.

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -193,13 +193,15 @@ export async function saveToHrm(task, form, workerSid, helpline, uniqueIdentifie
    */
   const formToSend = transformForm(rawForm);
 
+  const helplineToSend = (isOfflineContactTask(task) && form.contactlessTask.helpline) || helpline;
```
nickhurlburt, line 196: Should we rename the `helpline` function parameter to be `workerHelpline` or something similar to make it clear where it's coming from?

nickhurlburt, line 196: Is there any risk that `form.contactlessTask` would be `undefined`, causing an error here?  If maybe we're using the `za-v1` `formDefinition`?  Would `form.contactlessTask?.helpline` be safer?

plugin-hrm-form/src/formDefinitions/v1/tabbedForms/ContactlessTaskTab.json
```diff
@@ -0,0 +1,13 @@
...
+  {
+    "name": "helpline",
+    "label": "Helpline",
+    "type": "select",
+    "options": [
+      { "value": "Unknown", "label": "" },
+      { "value": "ChildLine", "label": "ChildLine" },
```
nickhurlburt, line 8: Do we have a way to set a default value?  If so, then check with Dee if we can set the default to "ChildLine", at least while we're not running LifeLine on Aselo.

nickhurlburt (commented):
Good, with comments.  I am a little nervous about creating our first hybrid tab that has some fixed fields and some helpline-specific configuration, since it's a new pattern.  I'm fine with it if you both can't see any problems with it.

plugin-hrm-form/src/formDefinitions/v1/tabbedForms/ContactlessTaskTab.json
murilovmachado: The main difference between the `ContactlessTaskTab` and the other tabs is that we have **Counselors** select, which is dynamic. Because of that, we can't use plain JSON files to define this tab, we need typescript code. This way, we could have the file `ContactlessTaskTab.ts` instead of `ContactlessTaskTab.json` (per helpline, ex: v1, za-v1) defining all fields for this tab.

The tradeoff with having typescript files:
- May be harder for other people not familiar with the code to edit it;
- May impact our idea of having these JSON definition files living outside of our codebase;

That's why I've ended up implementing a hybrid approach:
- We still have a typescript that handles counselors select (and other common fields)
- But other custom fields are defined at `ContactlessTaskTab.json`, so other people can edit it easily;

The drawbacks of the current implementation are:
- It's a new approach for this tab only, which hurts standardization;
- Currently, we can't change the order of fields for each helpline;

One idea that may solve our problems of having a dynamic field being defined in static JSONs would be to create a new form type called `counselorSelect`, which we would use like this:
```
{
      name: 'counselors',
      type: 'counselorSelect',
      label: 'Counsellor',
},
```
Another idea that may bit a little bit harder to implement would be to have `initialOptions` the same way we have `initialValues`, like this:
```
const counselorOptions = getCounselorOptions();
const initialOptions = { counselors: counselorOptions };

<ContactlessTaskTab
      task={task}
      display={subroute === 'contactlessTask'}
      definition={currentDefinitionVersion.tabbedForms.ContactlessTaskTab}
      initialValues={contactForm.contactlessTask}
      initialOptions={initialOptions}
/>
```
With this in hand, we can have any number of dynamic select fields.

-----

In summary there're, at least, 5 different ways to handle this:
- Customizing with typescript files (instead of JSON);
- Hybrid: only the custom fields defined in JSON (this PR);
- Create `counselorSelect` specific type
- Implement `initialOptions`;
- Hardcode ;

What do you guys think would be better for now? @nickhurlburt @GPaoloni

nickhurlburt: This is an excellent description of the issue and the possible solutions, thank you!  Given this description, which we can refer back to later, using a hybrid model for this tab sounds good to me.  If new needs come up in the future, we can revisit this description and decide if the other options make more sense at that time.

nickhurlburt (approved):
Looks great, thanks!

nickhurlburt:
@murilovmachado looks like this got merged to `staging`.  Would you mind merging staging into master and then just deleting `staging`?  I think we've all agreed to drop staging, so let's do that.

murilovmachado:
@nickhurlburt Sure, I'll do that right now.

PR #455, opened 2021-06-10
removed placeholder in add note screen

jennxu23:
Jira: part of CHI-657 - https://bugs.benetech.org/secure/RapidBoard.jspa?rapidView=19&view=detail&selectedIssue=CHI-657

This PR removes the placeholder in the Add Notes section, since we found that it would require some workaround to translate the placeholder.

In this PR:
- Removed the default field from the FormTextArea component (in the case of textArea)


GPaoloni (approved):
Great!

nickhurlburt:
@GPaoloni could you add some more information to this description, so that we can remember the technical reason for this change, and what we might need to do to bring back placeholders?  The current description here will probably not give us the information we need in the future, when we will ask ourselves what it would take to have localizable placeholders and why we didn't do it now.

PR #458, opened 2021-06-15
twitter icon UI 

jennxu23:
Jira (for twitter UI): https://bugs.benetech.org/browse/CHI-671
Jira (for twitter handle edits): https://bugs.benetech.org/browse/CHI-679

In this PR:
- Twitter tasks are twitter-colored
- Twitter tasks show the twitter Icon
- Twitter Icon shown in case timeline
- User's twitter handle shows up on task item and in the previous contacts banner
- User's twitter handle shows up in task info section

Screen Shot 2021-06-16 at 11 29 26 AM
Screen Shot 2021-06-16 at 11 29 37 AM
Screen Shot 2021-06-17 at 4 48 24 PM


plugin-hrm-form/src/components/search/SearchForm/index.jsx
```diff
@@ -126,6 +127,8 @@ class SearchForm extends Component {
     const { task } = this.props;
 
     const contactNumberFromTask = getNumberFromTask(task);
+    const checkBoxName =
+      task.channelType === channelTypes.twitter ? `@${task.attributes.twitterUserHandle}` : contactNumberFromTask;
```
GPaoloni, line 131: Given that this is used twice (and probably will be used in other places), you could make it reusable by writing a function like 
```
const getContactInfoFromTask = (task: CustomITask) =>  task.channelType === channelTypes.twitter ? `@${task.attributes.twitterUserHandle}` : contactNumberFromTask; 
```
(or whatever name makes sense).
I'm not sure where this functions should live tho, `src/utils/formatters.ts` seems like a good place.

GPaoloni (approved):
Looks very good! Just some small commens.
Also, can you please add `'twitter'` channel to `mapChannel` function in `src/utils/mappers.ts`? The idea is that if `channel` is `channelTypes.twitter`, we want to return `Twitter` (uppercase).

plugin-hrm-form/src/utils/setUpComponents.js
```diff
@@ -410,3 +412,59 @@ export const removeActionsIfTransferring = () => {
...
+    },
+    Supervisor: {
+      TaskCanvasHeader: {
+        title: 'TaskHeaderLineTwitter',
+      },
+      TaskOverviewCanvas: {
+        firstLine: 'TaskHeaderLineTwitter',
+      },
+    },
+  };
```
GPaoloni, line 451: I've just realized that this "overrides" the `TwitterChatChannel.templates`, not providing `secondLine` (among other) values. Can you change it to be like
`TwitterChatChannel.templates.CallCanvas.firstLine = 'TaskHeaderLineTwitter';`
and same for all the other changes?

nickhurlburt (approved):
I don't see any further issues than those that Gian raised.

plugin-hrm-form/src/components/TwitterIcon.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
```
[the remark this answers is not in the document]
jennxu23: I think there was already an icons folder, so I moved it into that one.

plugin-hrm-form/src/components/common/icons/TwitterIcon.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+type Props = {
+  width: string;
+  height: string;
+  color: string;
+};
+
+// eslint-disable-next-line react/display-name
```
murilovmachado, line 10: Instead of ignoring this warning, it's better to have this line just below:
`TwitterIcon.displayName = 'TwitterIcon';`

plugin-hrm-form/src/components/common/icons/TwitterIcon.tsx
```diff
@@ -0,0 +1,32 @@
...
+      />
+    </svg>
+  );
+};
+
+TwitterIcon.propTypes = {
+  width: PropTypes.string.isRequired,
+  height: PropTypes.string.isRequired,
+  color: PropTypes.string.isRequired,
+};
```
murilovmachado, line 31: Since we're already using typescript in this file and we have the definitions in `type Props`, we don't need to define PropTypes, so we can remove them. Lint will warn, but in this case we're ignoring this rule with:
`/* eslint-disable react/prop-types */`
at the top of the file.

Later on it would be a good idea to have this rule ignored by default for all typescript files. I'll look into that later.

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -1,4 +1,6 @@
 {
+  "TaskHeaderLineTwitter": "@{{task.attributes.twitterUserHandle}}",
```
murilovmachado, line 2: Nice, so the task already comes into flex with the `twitterUserHandle ` attribute? That's good!

murilovmachado (approved):
That looks great! I've left some very minor comments on the code. One thing that I want to ask @nickhurlburt  is if we want this PR also to show the Twitter icon on the timeline, like this for example:
image

If the answer is yes, this can be done by adding this to `TimelineIcon.tsx`:
```
case channelTypes.twitter:
      return <TwitterIcon width="24px" height="24px" color={twitterColor} />;
```

jennxu23:
> This looks great! I've left some very minor comments on the code. One thing that I want to ask @nickhurlburt is if we want this PR also to show the Twitter icon on the timeline, like this for example:
> image
> 
> If the answer is yes, this can be done by adding this to `TimelineIcon.tsx`:
> 
> ```
> case channelTypes.twitter:
>       return <TwitterIcon width="24px" height="24px" color={twitterColor} />;
> ```

PR #460, opened 2021-06-18
added twitter channel to contacts waiting window

jennxu23:
Jira: https://bugs.benetech.org/browse/CHI-457

In this PR:
- Added a box for twitter in the contacts waiting window

Screen Shot 2021-06-18 at 12 43 35 PM


plugin-hrm-form/src/components/queuesStatus/QueueCard.jsx
```diff
@@ -23,13 +23,15 @@ class QueuesCard extends React.PureComponent {
...
     whatsapp: PropTypes.number.isRequired,
+    twitter: PropTypes.number.isRequired,
     longestWaitingDate: PropTypes.string,
     colors: PropTypes.shape({
       voiceColor: PropTypes.shape({ Accepted: PropTypes.string }),
       webColor: PropTypes.shape({ Accepted: PropTypes.string }),
       facebookColor: PropTypes.shape({ Accepted: PropTypes.string }),
       smsColor: PropTypes.shape({ Accepted: PropTypes.string }),
       whatsappColor: PropTypes.shape({ Accepted: PropTypes.string }),
+      twitterColor: PropTypes.string,
```
GPaoloni, line 34: I don't like the difference on how we are passing the colors. Given that each color is defined by a string, I think having them like the `twitterColor` is the correct thing to do. Could you change the old colors to be plain strings instead of objects? That will imply that you pass `channelColor.Accepted` to `QueuesStatus` in `setUpComponents`.

GPaoloni (approved):
Looks great! :rocket:

PR #463, opened 2021-06-25
fixed remaining localization strings

jennxu23:
Jira: [https://bugs.benetech.org/browse/CHI-701](https://bugs.benetech.org/browse/CHI-701)

In this PR, the Cases and Contacts tabs in the search results are localizable. The results are also correctly singular/plural depending on the number of results.

Screen Shot 2021-06-25 at 12 01 37 PM

### DIFF
--- a/plugin-hrm-form/src/components/search/SearchResults/index.tsx
+++ b/plugin-hrm-form/src/components/search/SearchResults/index.tsx
@@ -285,7 +285,8 @@ const SearchResults: React.FC<Props> = ({
                     <Template code="PreviousContacts-Case" />
                   ) : (
                     <Template code="SearchResultsIndex-Cases" />
-                  )}&nbsp;
+                  )}
+                  &nbsp;
                 </StyledCount>
                 <StyledFormControlLabel
                   control={<StyledSwitch checked={closedCases} onChange={handleToggleClosedCases} />}
--- a/plugin-hrm-form/src/components/search/SearchResults/index.tsx
+++ b/plugin-hrm-form/src/components/search/SearchResults/index.tsx
@@ -186,45 +186,42 @@ const SearchResults: React.FC<Props> = ({
         <ScrollableList>
           <StyledResultsContainer>
             <StyledResultsText>
-              <BoldText data-testid="SearchResultsCount">
-                {/* Intentionally we must show the option different at the one currently selected */}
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -183,26 +186,38 @@ const SearchResults: React.FC<Props> = ({
...
+                        &nbsp;
+                        {contactsCount} <Template code="PreviousContacts-Contact" />
+                      </>
                     ) : (
-                      <Template code="PreviousContacts-Contacts" />
+                      <>
+                        <Template code="PreviousContacts-ThereAre" />
+                        &nbsp;
+                        {contactsCount} <Template code="PreviousContacts-Contacts" />
+                      </>
```
GPaoloni, line 220: Just highlighting this, please read below comment.

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -226,7 +241,12 @@ const SearchResults: React.FC<Props> = ({
             <>
               <StyledContactResultsHeader>
                 <StyledCount data-testid="ContactsCount">
-                  {contactsCount} <Template code="SearchResultsIndex-Contacts" />
+                  {contactsCount}{' '}
+                  {contactsCount === 1 ? (
+                    <Template code="PreviousContacts-Contact" />
+                  ) : (
+                    <Template code="SearchResultsIndex-Contacts" />
+                  )}
```
GPaoloni, line 249: I see that there are two different approaches for formatting this with the blank spaces, one being `{' '}` and the other using `&nbsp;` character. Can we standardize and use the same? I don't have a strong preference but maybe `{' '} ` is more intuitive (as it's just a string, no special characters). Maybe it does not render accordingly tho, in that case ignore this.

GPaoloni (approved):
Looks good!

plugin-hrm-form/src/components/search/SearchResults/index.tsx
jennxu23: I tried standardizing everything with `&nbsp;`, but it causes some of the checks to fail. I think we can revisit this once I figure out what's going on with the tests, but at the current state the code is functioning and passes all the checks.

PR #466, opened 2021-06-30
solving offline contact 'undefined' channel bug

jennxu23:
Jira: [CHI-709](https://bugs.benetech.org/browse/CHI-709)

In this PR:
- 'isOfflineContact` and default channel type are passed directly to InsightsService, which leads to the channel information for offline contacts to be sent to Insights correctly


GPaoloni (approved):
Awesome! Thanks for fixing this.

As a reference, [this is the PR where this bug was introduced](https://github.com/techmatters/flex-plugins/pull/442). The cause is that `isContactlessTask` task attribute is used in `buildInsightsData` function to determine where to pull the proper channel from.

nickhurlburt (approved):
Great work!  Once you've included Gian's comments, ship it!

Also, if you are unaware -- `.DS_Store` will appear in a folder any time you open it with Finder.  I don't know if there's a way to disable that.

PR #470, opened 2021-07-05
Customization updates for July 6 deploy

dee-luo:
- Updated 0800 number to 116 in canned responses
- Removes `LifeLine` option from Helpline select, and edits stored value of ChildLine


nickhurlburt:
@GPaoloni I've edited the `defaultOption` so that it will work.  Now shows:
image

Also, Gian, I believe that this change will cause any records that have been saved on Development or ZM Staging with the helpline field added as 'ChildLine' or 'LifeLine' to crash when attempting to load them via search.  Is that right?

Once approved, please go ahead and squash/merge.

GPaoloni (approved):
Looks good to me. I've tested with older contacts and it still works fine. Also I think is worth @murilovmachado to glance this, he is more familiar with this part of the code.

PR #471, opened 2021-07-07
Added issuer helpline as quickfix for the blank helpline bug in Insights

GPaoloni:


nickhurlburt (approved):
Looks good.

Note to self -- the `helpline` from `getConfig()` comes from `manager.workerClient.attributes`, so this is sending the helpline of the reporting worker.  (As mentioned on a call, it ought to be the credited worker, but that's a later fix.)

PR #472, opened 2021-07-08
Unify helpline attribute to HRM and Insights

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-719
This PR is related to https://github.com/techmatters/serverless/pull/80

This PR:
- Adds the option to extend offline contact task attributes, making them more "Twilio Task like".
- Adds `getWorkerAttributes` serverless service.
- Using above, unifies the source of truth for helpline value sent to HRM and Insights. This is done by updating the task to contain the expected `helplineToSend`, then grabbing this value from task attributes in both places (HRM & Insights saving functions).

### DIFF
--- a/plugin-hrm-form/src/services/formSubmissionHelpers.ts
+++ b/plugin-hrm-form/src/services/formSubmissionHelpers.ts
@@ -1,13 +1,13 @@
 /* eslint-disable import/no-unused-modules */
 import { Actions, ITask, Manager } from '@twilio/flex-ui';
 
-import { getConfig, reRenderAgentDesktop } from '../HrmFormPlugin';
+import { getConfig } from '../HrmFormPlugin';
 import { TaskEntry as Contact } from '../states/contacts/reducer';
 import { Case, CustomITask, isOfflineContactTask, offlineContactTaskSid } from '../types/types';
 import { channelTypes } from '../states/DomainConstants';
 import { buildInsightsData } from './InsightsService';
 import { saveToHrm } from './ContactService';
-import { assignOfflineContact } from './ServerlessService';
+import { assignOfflineContact, getWorkerAttributes } from './ServerlessService';
 import { removeContactState } from '../states/actions';
 
 /**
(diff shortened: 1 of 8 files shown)


plugin-hrm-form/src/services/formSubmissionHelpers.ts
```diff
@@ -38,16 +38,35 @@ export const completeContactlessTask = async (task: CustomITask) => {
...
+  return thisWorkerHelpline || task.attributes.helpline || '';
+};
+
 export const submitContactForm = async (task: CustomITask, contactForm: Contact, caseForm: Case) => {
-  const { workerSid, helpline } = getConfig();
+  const { workerSid } = getConfig();
+
+  const helplineToSave = await getHelplineToSave(task, contactForm, caseForm);
+  // Add helplineToSave so it's grabbed when saving to Insights (either in buildInsightsData for offline contacts or sendInsightsData for live contacts)
+  /* const updatedTask = */ await task.setAttributes({ ...task.attributes, helplineToSave });
```
nickhurlburt, line 62: This is a little confusing, because `task.setAttributes` normally requires an API call back to Twilio.  But if I'm guessing right, you have some polymorphism here where this could either be a normal Twilio task (which makes an API call), or an `OfflineContactTask`, which is just a local object.  I'm not sure if there's anything we can do better, other than just comment it.

If I understand right, this PR adds one additional synchronous API call to Twilio when saving a contact.  If it's an offline contact, we are probably calling `/getWorkerAttributes`.  Or if it's a live contact, we are calling the API for `task.setAttributes`.  For South Africa and Zambia, this is an additional trip over the Atlantic and back to save something.  We can keep this for now, but we want to keep in mind that we slowing counselors down, and may need to address this.  Unfortunately we don't have great access to timing data right now to be able to quantitatively measure this.

One more additional issue here is that there's no error handling in case `task.setAttributes` fails.

plugin-hrm-form/src/services/formSubmissionHelpers.ts
```diff
@@ -38,16 +38,35 @@ export const completeContactlessTask = async (task: CustomITask) => {
...
+    if (contactForm.contactlessTask.helpline) return contactForm.contactlessTask.helpline;
+
+    const targetWorkerSid = contactForm.contactlessTask.createdOnBehalfOf as string;
+    const targetWorkerAttributes = await getWorkerAttributes(targetWorkerSid);
+    return targetWorkerAttributes.helpline;
+  }
+
+  const { helpline: thisWorkerHelpline } = getConfig();
+  return thisWorkerHelpline || task.attributes.helpline || '';
+};
```
nickhurlburt, line 55: @dee-luo this one block is worth you looking at so you know the logic.  The helpline we save on the contact will be:
```
if offline contact:
  if there's a helpline input on the form, use that
  else, grab the helpline attribute on the counselor selected in the form
else (live contact):
  use this worker's helpline if it's present, else use the task helpline if present, else use an empty string

File: plugin-hrm-form/src/utils/setUpActions.js [hunk 3]
```diff
@@ -306,7 +306,7 @@ const decreaseChatCapacity = setupObject => async payload => {
  * @returns {import('@twilio/flex-ui').ActionFunction}
  */
 export const beforeCompleteTask = setupObject => async payload => {
-  await sendInsightsData(setupObject)(payload);
+  await sendInsightsData(setupObject)(payload); // Having this behavior in here makes very opaque to the live task workflow what's saved into Insights. Should we move it?
```
nickhurlburt, line 309: Yes, I've become confused on the flow of how we get everything saved correctly.  Are we certain to have `helplineToSend` set before this call is made?

I don't have any good recommendations on how to change things, though.  This PR just seems like it's going to need good testing, and if you're confident it's working, then go ahead with it.  We may need to revisit changes later.

nickhurlburt (approved):
I left some comments, but ultimately I think this looks good.  We should get this on to dev and staging early tomorrow so that it can be well-tested.

Then I think it's still good if @murilovmachado does an after-merge review on Monday.

plugin-hrm-form/src/services/formSubmissionHelpers.ts
GPaoloni: Yes, I'm aware of your concern and agree. One thing that can be better here is, after we move `sendInsightsData` for live contacts out of the `beforeCompleteTask` event, and put it here in the `formSubmissionHelpers`, then we can easily pass the helpline as a parameter to all the functions. The issue right now is that `sendInsightsData` is in a complete different part of the code, and has access to the task but not to the scope of this function.

> If it's an offline contact, we are probably calling /getWorkerAttributes.

We can't avoid this if we intend to "see other worker's helpline". We can do different with the task attributes, as mentioned above.

On the error handling comment, is true that I'm not adding any logic for that, but we don't for any service, instead we handle the errors from the components calling those services. I can add something in here tho.

plugin-hrm-form/src/utils/setUpActions.js
GPaoloni: Yes we are sure of that, all the logic related to submitting the form and completing the task is contained in `formSubmissionHelpers`. The only thing left to move in here is the Insights saving process for live contacts.

Screenshot from 2021-07-09 11-35-54

plugin-hrm-form/src/services/formSubmissionHelpers.ts
murilovmachado, line 55: Great to see the source of truth centralized in a function! It'll be very useful

murilovmachado (commented):
Looks good!

PR #473, opened 2021-07-09
Factour out saveInsightsData for live contacts

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-722

Note: better reviewed commit by commit.

This PR moves the "sending data to Insights for live contacts" logic out of the `beforeCompleteTask` event (addressing [this](https://github.com/techmatters/flex-plugins/pull/472#discussion_r666515823) and [this](https://github.com/techmatters/flex-plugins/pull/472#discussion_r666534390) comments). Instead this process will be in the same function that it is for offline contacts (`submitContactForm` function under `src/services/formSubmissionHelpers.ts`). This way, all of the process of saving to HRM and to Insights is handled in a single place, reducing the complexity, making easier to be consistent in both places and eliminating the need of API calls to keep in sync.
Also the `setAttributes` method is reverted (and removed) from offline contacts as now it has no utility.

All the combinations mentioned [here](https://github.com/techmatters/flex-plugins/pull/472#discussion_r666517839)  were tested, but please test them again in case I'm missing something.

### DIFF
--- a/plugin-hrm-form/src/types/types.ts
+++ b/plugin-hrm-form/src/types/types.ts
@@ -122,7 +122,6 @@ export type OfflineContactTask = {
     helplineToSave?: string;
   };
   channelType: 'default';
-  setAttributes: (attributes: {}) => Promise<OfflineContactTask>;
 };
 
 export type InMyBehalfITask = ITask & { attributes: { isContactlessTask: true; isInMyBehalf: true } };
@@ -143,3 +142,7 @@ export function isInMyBehalfITask(task: CustomITask): task is InMyBehalfITask {
 export function isTwilioTask(task: CustomITask): task is ITask {
   return task && !isOfflineContactTask(task) && !isInMyBehalfITask(task);
 }
+
+export type SubmissionContext = {
+  helplineToSave: string;
+};
--- a/plugin-hrm-form/src/types/types.ts
+++ b/plugin-hrm-form/src/types/types.ts
@@ -142,7 +142,9 @@ export function isInMyBehalfITask(task: CustomITask): task is InMyBehalfITask {
 export function isTwilioTask(task: CustomITask): task is ITask {
   return task && !isOfflineContactTask(task) && !isInMyBehalfITask(task);
 }
-
-export type SubmissionContext = {
+/**
+ * This type is used within the context of a form submission, to share values between HRM & Insights
+ */
+export type ExtraParameters = {
   helplineToSave: string;
 };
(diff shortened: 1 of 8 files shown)


plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -55,6 +53,16 @@ const getStateCaseForms = taskSid => {
...
 
+/**
+ * @param {import('../types/types').CustomITask} task
+ */
+export const shouldSendInsightsData = task => {
+  const { featureFlags } = getConfig();
+  const hasTaskControl = !featureFlags.enable_transfers || TransferHelpers.hasTaskControl(task);
+
+  return hasTaskControl && featureFlags.enable_save_insights && !task.attributes?.skipInsights;
+};
```
GPaoloni, line 64: The only purpose of this function is to maintain all the future flags inside of the set up files (not deep inside of the code), but maybe `HrmFormPlugin.js` is a better place for this.

plugin-hrm-form/src/types/types.ts
```diff
@@ -143,3 +142,7 @@ export function isInMyBehalfITask(task: CustomITask): task is InMyBehalfITask {
 export function isTwilioTask(task: CustomITask): task is ITask {
   return task && !isOfflineContactTask(task) && !isInMyBehalfITask(task);
 }
+
+export type SubmissionContext = {
+  helplineToSave: string;
+};
```
murilovmachado, line 148: I was just a little bit confused with the name `SubmissionContext`. 

At first, I thought it would be a React Context Component (because of naming convention), but it's a simple object. Also, when I see `SubmissionContext` I don't immediately know what this object intends to hold. Looking at it, we see it holds the final helpline value. Do we intend to add new values in the future? Because if not, I think `helplineToBeSave` or `finalHelpline` is a more descriptive name than `SubmissionContext`.

But that was just my first impression while reading through the code. You guys may have a different opinion.

murilovmachado (approved):
Looks good to me, Gian. Just one question on `SubmissionContext`. 

P.S.: I didn't run the code this time, I've just read the PR. Let me know if you think it's risky to be merged without really running it, so I can spend more time on this if necessary.

plugin-hrm-form/src/types/types.ts
nickhurlburt: I agree -- it's hard to understand what `SubmissionContext` means, and it sounds like a React Context.

nickhurlburt (approved):
I've skimmed over this, but do not understand it in detail.  I think you should move forward anyway.  I trust you guys, and it's good to get this out on Development sooner rather than later so that we have time to spot issues if they arise.  It'll be great to have an automatic Insights/HRM testing tool for things like this.

I hope to review this later as I'd like to understand it, but don't wait for me.

PR #475, opened 2021-07-14
adjusted Insights download tool to be able to download multiple reports

jennxu23:
[Jira: CHI-715](https://bugs.benetech.org/browse/CHI-715)

In this PR:
- Bash script to download multiple Insights Reports
- Script generates CSVs of reports instead of printing report to terminal

Notes:
- This PR updates the process for downloading Insights reports. Create a `.txt` file with the Object Ids of the reports you want to download (one ID on each line). Then, run the following command: `./insights.sh {username} {password} {workspaceID} < {file}.txt`.


nickhurlburt (commented):
Is this a duplicate of #474?

jennxu23:
> Is this a duplicate of #474?

I think what happened was I was working of the same branch of #474, and it included the changes in #474. The only difference between the two is that in this PR was supposed to be focused on the bash script, but eventually the bashscript got included in #474 anyways. Thanks for pointing this out! I'll remove this PR and push the changes to #474 instead.

PR #477, opened 2021-07-14
Lifeline data entry

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-650
Primary reviewer: @GPaoloni 

This PR adds Categories customization based on the current helpline.
The main difference to customization is that `DefinitionVersion.IssueCategorizationTab` was changed to be a function that receives the helpline:
image


plugin-hrm-form/src/components/tabbedForms/IssueCategorizationTab.tsx
```diff
@@ -46,6 +46,7 @@ const IssueCategorizationTab: React.FC<Props> = ({
       updateForm(task.taskSid, 'categories', categories);
     };
 
+    if (IssueCategorizationTabDefinition === null || IssueCategorizationTabDefinition === undefined) return {};
```
murilovmachado, line 49: Maybe we don't need this guard clause, but it feels safe to leave it there.

plugin-hrm-form/src/formDefinitions/v1/tabbedForms/ContactlessTaskTab.json
```diff
@@ -5,7 +5,8 @@
     "type": "select",
     "defaultOption": "ChildLine Zambia (ZM)",
     "options": [
-      { "value": "ChildLine Zambia (ZM)", "label": "ChildLine" }
+      { "value": "ChildLine Zambia (ZM)", "label": "ChildLine" },
+      { "value": "LifeLine Zambia (ZM)", "label": "LifeLine" }
```
murilovmachado, line 9: Should we merge with this option available? Or we want to hide it for now?

plugin-hrm-form/src/formDefinitions/v1/tabbedForms/IssueCategorizationTab.json
```diff
@@ -1,132 +1,185 @@
...
+  "ChildLine Zambia (ZM)": {
+    "Missing children": {
+      "color": "#BBE3FF",
+      "subcategories": [
+        "Child abduction",
+        "Lost, unaccounted for or otherwise missing child",
+        "Runaway",
+        "Unspecified/Other"
+      ]
+    },
```
murilovmachado, line 11: Notice the added helpline level.

nickhurlburt (commented):
I ran this locally and did a little testing, and briefly looked over the code.  I checked that search will correctly pull out the appropriate categories.  Looks good, though Gian should do a full review.

Also, @GPaoloni do you think this requires a new definition version?

nickhurlburt:
@dee-luo note that currently, there is no way to see the helpline from past records, for example in search results.  I don't know if this has been planned for.

plugin-hrm-form/src/components/common/forms/types.ts
```diff
@@ -193,7 +193,7 @@ export type DefinitionVersion = {
     CallerInformationTab: FormDefinition;
     CaseInformationTab: FormDefinition;
     ChildInformationTab: FormDefinition;
-    IssueCategorizationTab: CategoriesDefinition;
+    IssueCategorizationTab: (helpline: string) => CategoriesDefinition;
```
GPaoloni, line 196: Not saying this needs to change, but I want to say something about this. If we change the type to be a function, then we won't be able to drop form definitions in a remote location and fetch them, we'll need to always bundle them with the code, because the TS index file of the definitions is what is defining this function. My intention with those TS index files was just a quick way to support multiple helplines but ideally that was going to be removed.

If we decide to preserve the definitions bundled with the code this will do fine. If we don't maybe changing the type of the definition to just an object
`IssueCategorizationTab: { [helpline: string]: CategoriesDefinition }`
and then using some function defined within the React app (and not inside the form definition) like 
```
const getCategoriesForHelpline = (helpline, def: IssueCategorizationTab): CategoriesDefinition => {
  return def[helpline] || def.default; // all helplines should provide a default in their definitions just in case
};
```
can solve this.

GPaoloni (approved):
Code looks good to me, just some comments. I haven't tested the behavior yet.

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -52,11 +52,18 @@ import ViewPerpetrator from './ViewPerpetrator';
...
+  CustomITask,
+  isOfflineContactTask,
+} from '../../types/types';
 import CasePrintView from './casePrint/CasePrintView';
 import { getPermissionsForCase, PermissionActions } from '../../permissions';
 
-const isStandaloneITask = (task): task is StandaloneITask => {
+export const isStandaloneITask = (task): task is StandaloneITask => {
   return task.taskSid === 'standalone-task-sid';
 };
```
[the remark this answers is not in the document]
GPaoloni: Also can you remind me the purpose of a standalone task?

plugin-hrm-form/src/components/common/forms/types.ts
nickhurlburt: Good point.  I think we will need to move the form definitions to a remote location eventually.  Whether we build for that right now or incur a small amount of debt that we have to fix later, I'll leave that up to what you think is best.

plugin-hrm-form/src/components/case/Case.tsx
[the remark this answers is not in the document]
murilovmachado: Standalone Task is used on Standalone Search. When we click on search from the left menu, it will create a (fake) task in redux, that's the standalone task.

plugin-hrm-form/src/components/case/ViewContact.tsx
```diff
@@ -39,6 +41,18 @@ type OwnProps = {
...
+  useEffect(() => {
+    const fetchHelpline = async () => {
+      if (!isStandaloneITask(task)) {
+        const helplineToSave = await getHelplineToSave(task, form);
+        setHelpline(helplineToSave);
+      }
+    };
+
+    fetchHelpline();
+  }, [task, form]);
```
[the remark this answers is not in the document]
murilovmachado: I've talk to Nick about it. I was thinking of moving this call to fetch the helpline up in the hierarchy, and the child components wouldn't need to fetch it by themselves. I may be doing this on the part 2 of this PR.

plugin-hrm-form/src/components/common/forms/types.ts
murilovmachado: Right now the code is almost like this, what the function in the TS index does is really what you suggested on `getCategoriesForHelpline` (behaves the same). Also, I think we can make those JSON files remote, but keep the TS index file in the code (not one per definition as it is now, but a single one). I feel confident it won't get in the way of making those JSON files remote.

plugin-hrm-form/src/components/case/Case.tsx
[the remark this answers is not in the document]
GPaoloni: Thanks!

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -191,12 +191,13 @@ export async function saveToHrm(task, form, workerSid, uniqueIdentifier, shouldF
   // This might change if isNonDataCallType, that's why we use rawForm
   const timeOfContact = getDateTime(rawForm.contactlessTask);
 
+  const { helplineToSave: helpline } = task.attributes;
```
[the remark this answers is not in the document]
murilovmachado: In the second PR, I'm using the helper method here.

murilovmachado:
Closing this as I've merged part 2 (that contains all of this) into master.
https://github.com/techmatters/flex-plugins/pull/482

PR #481, opened 2021-07-20
Script to setup Twilio account

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-729

This PR adds a script to automate the entire step 2 of [this guide](https://benetech.app.box.com/services/box_for_office_online/4881/772893818136/1c0778.fd6b89fbc9665fdf8c58766af90d6d01715dee1c97316ac157d6947c10a6c71d?node_type=file). 
TS added and eslint changed to apply the rules correctly.

If everything went fine, the script will log the final state object which contains a lot of useful resources SIDs that will be required in further steps of the guide.
If something goes wrong, the script will delete the resources created so far, except the Parameter Store params. This is intentional as in order to clean those the AWS script-user should have param deletion permission, and seems to me dangerous (one error there can be irreversible if we don't backup the params).

WIP:
- Document the script.
- Share the env vars to run it.
- Add npm scripts to compile & run the script
- (Optional) Add yargs to take input instead of env vars.


scripts/src/createTwilioResources.ts
```diff
@@ -0,0 +1,330 @@
...
+      createSyncService,
+      createAPIKey,
+      fetchChatService,
+      saveWorkspaceToSSM,
+      saveWorkflowToSSM,
+      saveSyncServiceToSSM,
+      saveAPIKeyToSSM,
+      saveAPISecretToSSM,
+      saveChatServiceToSSM,
+    ].reduce(async (accumPromise, func) => {
```
nickhurlburt, line 310: Very cool!

nickhurlburt (approved):
This is fantastic.  I agree with requiring manual deletion of parameters when needed -- the annoyance of doing that should be rare, and the alternatives are either risky or expensive.  Well done balancing that tradeoff!

Some long term thoughts, which you don't need to worry about now:
* This `scripts` folder is getting to be a mess, especially with all the packages we have to include to make everything work.  At some point we'll need to separate it out into multiple npm packages.
* One weakness we have about claiming to be open source is that with a true open source project, someone should be able to use the available source code to set up their own accounts.  But people can't do that now because there's so much configuration that needs to be done that the code doesn't show.  But this script gets us a little bit closer to that 'true open source' model, which is good.  Achieving that 'true open source' situation is *not* our priority right now, but as we build more of this, it's good to keep in mind how a person outside of Tech Matters could use it to set up their own accounts.

scripts/src/createTwilioResources.ts
```diff
@@ -0,0 +1,330 @@
...
+    configuration: JSON.stringify({
+      task_routing: {
+        filters: [
+          {
+            filter_friendly_name: `${state.helpline}`,
+            expression: `helpline==${state.helpline}`,
+            targets: [
+              {
+                expression:
+                  "(worker.waitingOfflineContact != true AND ((task.channelType == 'voice' AND worker.channel.chat.assigned_tasks == 0) OR (task.channelType != 'voice' AND worker.channel.voice.assigned_tasks == 0)) AND ((task.transferTargetType == 'worker' AND task.targetSid == worker.sid) OR (task.transferTargetType != 'worker' AND worker.sid != task.ignoreAgent))) OR (worker.waitingOfflineContact == true AND task.targetSid == worker.sid AND task.isContactlessTask == true)",
```
nickhurlburt, line 143: I want to flag this for @dee-luo @Humairaa127 and @ray-hc.  This is part of Gian's script to (partially) automate setting up new Twilio accounts.  This includes setting up the initial routing expressions on TaskRouter.  I wanted to raise this for two reasons:
1. If we update the initial expression that all helplines should have, please update it here as well.
1. Eventually, I would love to not just have a script that creates the initial account (and then any edits after that are manually done in the Twilio Console), but also have a script do the later updates.  That means we can deploy TaskRouter changes at the same time as code changes that rely on them, and we have a version history for the changes.  Still a long way to go to achieve that.  Baby steps!

murilovmachado (approved):
This is great, Gian! Code is very easy to follow.

PR #485, opened 2021-07-28
Et staging customisations

Humairaa127:
Added form customisations in et-v1
Updated index.js to reference et-v1
Updated permissions to include et.ts


murilovmachado (approved):
Looks good to me!

@GPaoloni can you review the Insights related definitions? I'm not super familiar with that. Also, when this is merged and deployed, we'll need to update these `serviceConfiguration` properties: 
```
"permissionConfig": "za-v1",
"definitionVersion": "za-v1",
```
to:
```
"permissionConfig": "et",
"definitionVersion": "et-v1",
```

And a last question @GPaoloni : I see that the current `permissionConfig` is set to `za-v1`, which doesn't exist. The acceptable values are `type PermissionConfig = 'zm' | 'za' | 'et';`. Ethipia may have issues currently related to permissions.

plugin-hrm-form/src/formDefinitions/et-v1/HelplineInformation.json
```diff
@@ -0,0 +1,15 @@
+{
+  "label": "Helpline",
+  "helplines": [
+    {
+      "label": "Adama ChildLine",
+      "value": "ChildLine Ethiopia (ET)",
```
murilovmachado, line 6: Here the value I believe should be `Adama Child Helpline` as well.
@GPaoloni can you double-check that?

plugin-hrm-form/src/formDefinitions/et-v1/tabbedForms/IssueCategorizationTab.json
```diff
@@ -0,0 +1,133 @@
+{
+  "ChildLine Ethiopia (ET)": {
```
murilovmachado, line 2: Here too, I suspect the helpline is set to be `Adama ChildLine` on Ethiopia Twilio Console.
Or should we update the Twilio Console to be `ChildLine Ethiopia (ET)`?

plugin-hrm-form/src/formDefinitions/et-v1/HelplineInformation.json
Humairaa127: change made

plugin-hrm-form/src/formDefinitions/et-v1/tabbedForms/IssueCategorizationTab.json
Humairaa127: change made

PR #491, opened 2021-08-12
Fix Case Helpline

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-741
Primary reviewer: @GPaoloni 

When creating an **offline contact** with a **case**, the case's helpline was not the same as the offline contact.
This PR fixes it by getting the `helpline` value from **redux**.

### DIFF
--- a/plugin-hrm-form/src/___tests__/services/InsightsService.test.js
+++ b/plugin-hrm-form/src/___tests__/services/InsightsService.test.js
@@ -24,9 +24,9 @@ const zambiaUpdates = (attributes, contactForm, caseForm) => {
   return attsToReturn;
 };
 
-const expectWithV1Zambia = (attributes, contactForm, caseForm, extraParameters) =>
+const expectWithV1Zambia = (attributes, contactForm, caseForm) =>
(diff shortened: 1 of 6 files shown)


nickhurlburt (commented):
It looks good to me but @GPaoloni better review.

GPaoloni (approved):
I think this is great, thanks for the fix!
One question on this topic: Now we are using this value comming from Redux almost everywhere but when we submit a contact, right? Can we change that too and always rely on the value stored in Redux? That way we achieve our final goal of "a unique source of thrush" (for one more iteration of course, not in this PR).

murilovmachado:
> I think this is great, thanks for the fix!
> One question on this topic: Now we are using this value comming from Redux almost everywhere but when we submit a contact, right? Can we change that too and always rely on the value stored in Redux? That way we achieve our final goal of "a unique source of truth" (for one more iteration of course, not in this PR).

I agree, we won't need to pass `ExtraParameters` anymore, since we have that updated/correct value on redux (contact form).

PR #492, opened 2021-08-13
Setup ET and MW front end permissions

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-764
Primary reviewer: @GPaoloni 

This PR fixes ET permissions and adds MW permissions.
When this gets deployed to MW Staging, we should update its **serviceConfiguration** to use MW permissions, because right now it's using ZM.


GPaoloni (approved):
Great! :rocket: 
We need to replicate this on the hrm repo, right?

murilovmachado:
> Great! 🚀
> We need to replicate this on the hrm repo, right?

Yep, backend PR will be up soon.

PR #495, opened 2021-08-24
Humairaa zm lifeline staging

Humairaa127:


murilovmachado (approved):
Code looks good. I've also downloaded and checked that everything worked fine.

My only question is about the categories color, I've noticed most of them were repeated:
image

But it's something that we can easily change later if we decide to.

PR #500, opened 2021-09-09
Upload Case Document

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-744
Primary reviwer: @GPaoloni


plugin-hrm-form/src/services/ServerlessService.ts
```diff
@@ -137,3 +137,38 @@ export const getWorkerAttributes = async (workerSid: string) => {
   const response = await fetchProtectedApi('/getWorkerAttributes', body);
   return response;
 };
+
+const toBase64 = file =>
```
nickhurlburt, line 141: Is this still in use?

nickhurlburt (approved):
I gave this a very quick review.  Looks good to me.  I don't see any security issues since all of that is handled in serverless functions, which is good.

There are a couple things that I've realized are O(N) in the number of helplines we support, whenever we add a new feature like this:
1. Adding a new JSON file to each configuration (like `DocumentForm.json`)
2. Adding translations for new strings (not an issue now because it's still all English)

Nothing to do about that right now, but when these start to become heavier tasks, we will want to invest in making these easier to support as N gets bigger.

plugin-hrm-form/src/services/ServerlessService.ts
murilovmachado: Good catch, removed.

plugin-hrm-form/src/components/case/AddDocument.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+    const onDeleteFile = async (fileName: string) => {
+      await deleteFile(fileName);
+    };
+
+    const generatedForm = createFormFromDefinition(DocumentForm)([])(initialForm)(
+      updateCallBack,
+      onFileChange,
+      onDeleteFile,
+    );
```
murilovmachado, line 124: Here's probably the more important tech debt of this PR.

Notice that the way it's currently implemented, if we have multiple fields of type `file-upload`, (e.g.: `document1` and `document2`), the **upload** and **delete** behavior of both documents will be the same.

Example: We can't have `document1` to be uploaded to AWS S3, but `document2` to a different place. Currently they share the upload and delete functions.

nickhurlburt:
One thing I forgot to ask: accessibility.  Does this decrease or increase our accessibility debt, or keep it the same?  Are actionable elements focusable, and non-actionable elements not focusable?  Do the focus for new views follow a sensible sequence?

The task to show focus is not until later, but this should be built so that focus works correctly when we start to show focus.

plugin-hrm-form/src/components/case/AddDocument.tsx
nickhurlburt: This seems reasonable to me.

murilovmachado:
> One thing I forgot to ask: accessibility. Does this decrease or increase our accessibility debt, or keep it the same? Are actionable elements focusable, and non-actionable elements not focusable? Do the focus for new views follow a sensible sequence?
> 
> The task to show focus is not until later, but this should be built so that focus works correctly when we start to show focus.

The thing that can impact accessibility the most on this PR is the `UploadFileInput` component. The native `<input type="file">` is very poorly customizable, and could not meet our designs at all. So, instead, we're using a `<input type="button">` which the user interacts with, that basically forwards its click to a **hidden** `<input type="file">`. I think this is a good topic to discuss with James/Sina, and see what they have to say about this component.

GPaoloni (approved):
Code looks great, behavior is the expected. Really great job!! :rocket: 

Just one minor UI bug. When you try to save with no document, the error message pops up indicating the issue.Then I added a document but the UI didn't notice this change and the form state remains in error, not being able to save the document in HRM (even if it's already saved in AWS).
Screenshot from 2021-09-14 16-20-40

plugin-hrm-form/src/formDefinitions/et-v1/caseForms/DocumentForm.json
```diff
@@ -0,0 +1,17 @@
+[
```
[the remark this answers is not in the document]
murilovmachado: Yes, similar to other forms, we have `const document = methods.getValues();` to get all values defined in the form. So, if the form definitions have more fields, all of them will be saved in the `document` object.

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -91,8 +95,12 @@ const bindCreateSelectOptions = (path: string) => (o: SelectOption) => (
  * @param {() => void} updateCallback Callback called to update form state. When is the callback called is specified in the input type.
  * @param {FormItemDefinition} def Definition for a single input.
  */
-// eslint-disable-next-line sonarjs/cognitive-complexity
-const getInputType = (parents: string[], updateCallback: () => void) => (def: FormItemDefinition) => (
+const getInputType = (
+  parents: string[],
+  updateCallback: () => void,
+  onFileChange?: (event: any) => Promise<string>,
+  onDeleteFile?: (fileName: string) => Promise<void>,
```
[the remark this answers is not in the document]
murilovmachado: Great idea, thanks!

PR #504, opened 2021-09-21
Et prod

Humairaa127:


dee-luo:
@Humairaa127 will we have two fields named "Region" if we rename District to Region?

dee-luo (approved):

plugin-hrm-form/src/formDefinitions/et-v1/caseForms/HouseholdForm.json
```diff
@@ -35,8 +35,8 @@
     "type": "input"
   },
   {
-    "name": "district",
-    "label": "District",
+    "name": "region",
```
GPaoloni, line 38: This will cause issues with the already existing records in staging db, as they will have empty value for region. Is that ok?

GPaoloni (approved):
Besides my comment, looks good.

PR #506, opened 2021-09-23
Helpline Deployment Scalability: generate deploy file for Serverless

GPaoloni:
This PR is the first step to automate the setup of the Serverless environment for a new account.
This PR:
- Adds a script that consumes the input values and generates a `.yml` file ready to be dropped in Serverless repo. This file is the deploy workflow ready to consume the Parameter Store values.
- Modify how the scripts are run. Now a single script called `setupNewAccount` can be executed when setting a new helpline. This script will prompt which subscripts should be executed (in case we want to opt-out some of them to perform a partial setup).


scripts/src/helpers/log.ts
```diff
@@ -1,12 +1,10 @@
 /* eslint-disable no-console */
```
GPaoloni, line 1: The changes in this file are just to reduce duplicated code and prevent typing errors.

scripts/src/setupNewAccount/setupTwilioServerless.ts
```diff
@@ -0,0 +1,72 @@
+import { readFileSync, writeFileSync } from 'fs';
+import fetch from 'node-fetch';
+import { inspect } from 'util';
+import { logError, logSuccess } from '../helpers/log';
+import type { ScriptsInput } from './types';
+// import { Octokit } from '@octokit/core';
```
GPaoloni, line 6: Leaving this commented as this library is a wrapper around GitHub APIs. It can be useful later on opening PRs.

scripts/src/setupNewAccount/setupTwilioServerless.ts
```diff
@@ -0,0 +1,72 @@
...
+    throw new Error(
+      'The generated workflow file differs from the one being used by Aselo Development.',
+    );
+};
+
+export const setupTwilioServerless = async (input: ScriptsInput) => {
+  try {
+    await checkWorkflowInSync();
+
+    const fileName = 'new-workflow.yml'; // eventually this will be part of the input
```
GPaoloni, line 58: This can be an input later on. It will be needed for opening new PRs against Serverless repo from this script.

scripts/templates/serverless-workflow-template
```diff
@@ -0,0 +1,85 @@
+# This is a basic workflow to publish a Twilio function with Github Actions
+
+name: <HELPLINE> <ENVIRONMENT> release
```
GPaoloni, line 3: Note that this will differ from what we are currently using. For example, to run the script for Malawi, the `HELPLINE` value was "Tizandizine Helpline", not "Malawi". This will change the naming style we're currently using in the actions tab (instead of "Malawi Staging" it would be "Tizandizine Helpline Staging"). Let me know if you prefer this to be an input.

scripts/README.md
```diff
@@ -1,38 +1,53 @@
...
 - [createTwilioResources](#createTwilioResources)
+- [setupTwilioServerless](#setupTwilioServerless)
+
+To run the script:
+- Create a `.env` file and fill it with the proper values (see next section).
+- Install dependencies with `npm install`.
+- Run `npm run setupNewAccount` on the root folder.
 
-## createTwilioResources
-This script performs the all the steps described in the section 2 of the [Twilio account setup guide](https://benetech.app.box.com/services/box_for_office_online/4881/772893818136/1c0778.fd6b89fbc9665fdf8c58766af90d6d01715dee1c97316ac157d6947c10a6c71d?node_type=file).
```
nickhurlburt, line 7: I think that link may be ephemeral. Use https://benetech.app.box.com/file/772893818136

scripts/src/setupNewAccount/index.ts
```diff
@@ -0,0 +1,73 @@
...
+      const twilioResources = await createTwilioResources(input);
+    }
+
+    if (promptResult.setupTwilioServerless === 'y') {
+      logSuccess('Running setupTwilioServerless script..');
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const twilioServerless = await setupTwilioServerless(input);
+    }
+
+    logSuccess('Setup script completed succesfully!');
```
nickhurlburt, line 65: nit: successfully

scripts/templates/serverless-workflow-template
nickhurlburt: I'm not sure I fully understand the implications. Overall, naming is still a question we haven't figured out. We tend to talk about helplines by their country, but in many cases the helpline's actual name doesn't mention their country. We don't have a good naming convention yet. You are welcome to suggest one.

scripts/README.md
```diff
@@ -1,38 +1,53 @@
...
 
-To run this script we need to provide the following environment
-| Variable               | Description |
-|------------------------|-------------|
-| AWS_ACCESS_KEY_ID      | AWS script-user access key (or any other user with Parameter Store access) |
-| AWS_SECRET_ACCESS_KEY  | AWS script-user access secret (or any other user with Parameter Store access) |
-| TWILIO_ACCOUNT_SID     | Target Twilio account sid |
-| TWILIO_AUTH_TOKEN      | Target Twilio account auth token |
-| HELPLINE               | Helpline's friendly name (e.g. South Africa Helpline) |
-| SHORT_HELPLINE         | Short code for this helpline (e.g. ZA) |
```
nickhurlburt, line 37: Will this still work if we make the short code 3 characters?

scripts/src/setupNewAccount/createTwilioResources.ts
```diff
@@ -422,81 +414,65 @@ const cleanupPartialResources = async (state: State): Promise<void> => {
 };
 
 /**
- * Naming convenyion we use for the envionments.
+ * Secuence of functions to fetch/create Twilio resources and save the required AWS Parameter Store values.
```
nickhurlburt, line 417: nit: sequence

nickhurlburt (approved):
This is cool! One step closer to automated helpline deployment.

scripts/README.md
```diff
@@ -1,38 +1,53 @@
...
-| ENVIRONMENT            | Target environment, one of Development, Staging or Production |
\ No newline at end of file
+As of 2021/09/22, this script:
+- Creates a `new-workflow.yml` workflow file in the root folder, with the targeted account. It will be configured ready to consume the parameters from AWS Parameter Store following the convention. If the file `templates/serverless-workflow-template` is out of sync with the Aselo Development deployment workflow file, the script will abort in order to prevent generating old version of workflows (manually disable this check if you are sure that you want the template version).
```
murilovmachado, line 53: Very nice that it aborts if the template is out of sync!

scripts/src/setupNewAccount/createTwilioResources.ts
```diff
@@ -131,7 +123,7 @@ const saveHrmStaticKeyToSSM = saveStateKeyToSSM('hrmStaticApiKeySecret');
 /**
  * Twilio resources related functions
  */
-const fetchFlexWorkspace = async (state: State): Promise<State> => {
+const fetchFlexWorkspace = async (state: State) => {
```
murilovmachado, line 126: Notice that you've removed `: Promise<State>` from the returning type of several functions. Are they still returning `: Promise<State>` implicitly? or are they returning something else now?

scripts/templates/serverless-workflow-template
murilovmachado: This `<HELPLINE>` value is only used for naming the GitHub action, correct?

If that's the case, I think it's easier for us to read `Malawi Staging release` than `Tizandizine Helpline Staging release`, so we don't have to memorize the internal name of the helpline.

Also, some helplines such as Zambia has two sub-helplines `ChildLine ZM` and `LifeLine ZM`.

murilovmachado (approved):
Looks great!

scripts/templates/serverless-workflow-template
GPaoloni: > This <HELPLINE> value is only used for naming the GitHub action, correct?

No, it's also used for the workflow filter (the value the counselor needs to receive tasks).
I agree that it may be hard to remember all the names, but also having the country code results in the issue Nick mentioned. I'm not sure what the best option is, but is easy to ask this "deploy name" as an input for the scripts.

PR #508, opened 2021-09-24
Fix on contact initialization (before accept task)

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-782

### DIFF
--- a/plugin-hrm-form/src/utils/setUpActions.js
+++ b/plugin-hrm-form/src/utils/setUpActions.js
@@ -91,7 +91,7 @@ export const initializeContactForm = payload => {
   const { currentDefinitionVersion } = Manager.getInstance().store.getState()[namespace][configurationBase];
 
   Manager.getInstance().store.dispatch(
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -91,7 +91,7 @@ export const initializeContactForm = payload => {
   const { currentDefinitionVersion } = Manager.getInstance().store.getState()[namespace][configurationBase];
 
   Manager.getInstance().store.dispatch(
-    GeneralActions.initializeContactState(currentDefinitionVersion.tabbedForms)(payload.task.taskSid),
+    GeneralActions.initializeContactState(currentDefinitionVersion)(payload.task.taskSid),
```
GPaoloni, line 94: This call end in this function being invoked, which consumes the `DefinitionVersion` object (not only `tabbedForms`).

Screenshot from 2021-09-24 19-17-27

Everything was working fine with the forms because when a task is selected and the `TaskView` is mounted, this effect is run.

Screenshot from 2021-09-24 19-19-36

nickhurlburt (approved):
Nice fix!

PR #512, opened 2021-10-01
Added fullstory monitoring support

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-812

This PR:
- Modifies deploy script to include the Fullstory app id in `secret.js`.
- Adds Fullstory monitoring behind `enable_fullstory_monitoring` feature flag.


nickhurlburt (approved):
This looks great as a first cut. Later, we will need to figure out how to keep track of which account is being recorded, since it's all the same id.

PR #513, opened 2021-10-01
[FULLSTORY] Added Fullstory id to all deploy files

GPaoloni:
This PR adds Fullstory id to all the deploy actions except for Aselo Development (done at https://github.com/techmatters/flex-plugins/pull/512).

IMPORTANT: This PR should be merged only after https://github.com/techmatters/flex-plugins/pull/512 and when we are sure we want to start using Fullstory in the rest of the environments.


nickhurlburt (approved):
Looks good, and thanks for putting this together now.

PR #522, opened 2021-10-14
Copy flow script upgrade

gpaoloni:
This PR
- Moves copy flow main logic into `/src` folder to bundle with TS on compile. Original file is kept as JS for simplicity but should be upgraded if possible.
- Simplifies the usage of `copyFlow` script by fetching most of the required values using Twilio APIs.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -2,6 +2,7 @@
 
 ## Scripts index
 - [setupNewAccount](#setupNewAccount)
+- [copyFlow](#copyFlow)
 
 ## setupNewAccount
 This scripts automates some of the steps to setup a new account for helpline, prompting the user asking which scripts should execute. The scripts executed by setupNewAccount are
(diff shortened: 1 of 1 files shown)


scripts/copyFlow.js
```diff
@@ -57,72 +12,24 @@ function recReplaceIds(flow, oldIdDict, newIdDict) {
...
  * @param {boolean} update - Whether to update a flow or create a new one
  */
-async function copyAndPostFlow(oldIdsFile, newIdsFile, flowFile, update = false) {
-  const oldIdStr = await fs.readFile(oldIdsFile, 'utf8');
-  const oldIds = JSON.parse(oldIdStr);
-  // eslint-disable-next-line global-require
-  const client = require('twilio')(oldIds.AccountSid, oldIds.AuthToken);
+async function copyAndPostFlowFromFiles(oldIdsFile, newIdsFile, flowFile = null, update = false) {
+  const prompt = new Confirm(
+    'This is a deprecated script. Use the updated version instead. Are you sure you want to continue?',
```
nickhurlburt, line 17: If this is deprecated, is there any reason not to just remove it? Or if we don't want to remove it, can we leave instructions in the comments about when we can remove it? As it is now, if I get this message I'm not sure what to do next (other than ask you about it).

murilovmachado (approved):
Looks good to me. This will be very helpful. This is one of the most annoying steps when setting up a new helpline.

PR #526, opened 2021-10-18
Modified age mapper to get values above >25

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-826

This PR changes the way the age mapper returns the values to allow "max age" to be greater that 25.

### DIFF
--- a/plugin-hrm-form/src/utils/mappers.ts
+++ b/plugin-hrm-form/src/utils/mappers.ts
@@ -49,11 +49,14 @@ export const mapChannelForInsights = (channel: string) => {
   }
 };
 
-export const mapAge = (age: string) => {
+export const mapAge = (ageOptions: string[]) => (age: string) => {
   const ageInt = parseInt(age, 10);
 
-  if (ageInt >= 0 && ageInt <= 25) return age;
-  if (ageInt > 25) return '>25';
+  const maxAge = ageOptions.find(e => e.includes('>'));
+  const maxAgeInt = parseInt(maxAge.replace('>', ''), 10);
+
+  if (ageInt >= 0 && ageInt <= maxAgeInt) return age;
+  if (ageInt > maxAgeInt) return maxAge;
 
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/utils/mappers.ts
```diff
@@ -49,11 +49,14 @@ export const mapChannelForInsights = (channel: string) => {
...
 };
 
-export const mapAge = (age: string) => {
+export const mapAge = (ageOptions: string[]) => (age: string) => {
   const ageInt = parseInt(age, 10);
 
-  if (ageInt >= 0 && ageInt <= 25) return age;
-  if (ageInt > 25) return '>25';
+  const maxAge = ageOptions.find(e => e.includes('>'));
+  const maxAgeInt = parseInt(maxAge.replace('>', ''), 10);
```
GPaoloni, line 56: For now this code assumes that there is a `>xx` value in the age dropdown, where `xx` is the greater age allowed.

plugin-hrm-form/src/utils/mappers.ts
```diff
@@ -49,11 +49,14 @@ export const mapChannelForInsights = (channel: string) => {
...
+
+  if (ageInt >= 0 && ageInt <= maxAgeInt) return age;
+  if (ageInt > maxAgeInt) return maxAge;
```
nickhurlburt, line 59: Maybe we want to make this less fragile. Right now, if `>xx` isn't specified in configuration, `maxAge.replace(...)` will throw an error and fail. Could we do something like:
```suggestion
  const maxAge = ageOptions.find(e => e.includes('>'));
  if (maxAge) {
    const maxAgeInt = parseInt(maxAge.replace('>', ''), 10);
    if (ageInt > maxAgeInt) return maxAge;
  }
  if (ageInt >= 0 && ageInt <= maxAgeInt) return age;
```
Does that have problems that I'm missing?

PR #527, opened 2021-10-18
In customizations

Humairaa127:


plugin-hrm-form/src/formDefinitions/in-v1/tabbedForms/CallerInformationTab.json
```diff
@@ -0,0 +1,117 @@
+[
+    {
+      "name": "Name",
```
GPaoloni, line 3: I just checked the code and the same issue happens here: some places of our code expects a first and last names in this form. I'm not sure if this can cause troubles tho, I need to give a deeper look.

plugin-hrm-form/src/formDefinitions/in-v1/HelplineInformation.json
```diff
@@ -0,0 +1,15 @@
+{
+  "label": "Helpline",
+  "helplines": [
+    {
+      "label": "Trustline",
+      "value": "Aarambh Trustline (IN)",
```
GPaoloni, line 6: @dee-luo Be aware that this is not the configuration I've given to the task queue. Something to be addressed I believe.

PR #535, opened 2021-11-01
Fix Add Another Task button to wait for next task

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-841
Primary Reviewer: @GPaoloni 

This PR fixes the Add Another Task Button to be disabled (and shows a spinner) after clicking on it. It will only be clickable again after one of the following happens:
- A reservation to the user gets created, or;
- A reservation fails to be created after 5 seconds.

### DIFF
--- a/plugin-hrm-form/src/components/ManualPullButton/index.tsx
+++ b/plugin-hrm-form/src/components/ManualPullButton/index.tsx
@@ -42,7 +42,15 @@ const ManualPullButton: React.FC<Props> = ({ queuesStatusState, chatChannelCapac
 
   const { maxMessageCapacity } = worker.attributes;
   const maxCapacityReached = chatChannelCapacity >= parseInt(maxMessageCapacity, 10);
-  const disabled = maxCapacityReached || !isAnyChatPending(queuesStatusState.queuesStatus) || isWaitingNewTask;
+  const { isAvailable } = worker.worker;
+  const noTasks = worker.tasks.size === 0;
+
+  const disabled =
+    !isAvailable ||
+    noTasks ||
+    maxCapacityReached ||
+    !isAnyChatPending(queuesStatusState.queuesStatus) ||
+    isWaitingNewTask;
 
   return (
     <AddTaskButton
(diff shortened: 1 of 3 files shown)


nickhurlburt:
Behavior looks great!

image

nickhurlburt:
Though, interesting, we still got one of these errors somehow:
image

I did *lots* of clicking so it's still a good sign that there's only one, but I'm not sure how it can still happen.

murilovmachado:
@nickhurlburt I'm not really sure how to reproduce your scenario. The **+ Another Task** for me is **not** available when I'm at a scenario similar to yours:

image

What I did was:
- Make **user 1** online, and a task appears to **user 1** accept/reject
- Make **user 2** online. This user will see there's **1 task in queue**, but it's not gonna be assignable to him
- **+Another Task** is **not** available to **user 2**

nickhurlburt:
@murilovmachado for me, there was just one task waiting in the queue and I was the only person available. However I had changed the task queue so that I wasn't in it.

The difference seems to be that the 'Add Another Task' button doesn't get disabled just because the worker has zero tasks assigned. In the case you showed, I would guess that the logic for determining whether there are tasks in the queue does not include tasks that are in the 'reserved' state, which is the state when the worker has been given the task but hasn't accepted it.

We have probably never implemented the "disable the button if the worker has zero tasks assigned" logic. I _think_ we should, but there might be a reason why we didn't, or it might be unusually difficult. In that case we can leave it be for now. @GPaoloni may have thoughts here.

GPaoloni (approved):
Looks awesome! 
One question, do you think is more reusable to split `loading` from `disabled` in `AddTaskButton`? Cause maybe we could always disable when `loading` is true, but that might make the behavior more opaque.

Regarding to what @nickhurlburt has mentioned, if the button should be disabled if there are no tasks for the counselor, one of the props of `ManualPullButton` is `worker: WorkerState` which contains a map of the active tasks. You could use `worker.tasks.size`, it should be zero if no tasks are assigned, but I'm not 100% of the behavior of this map.
Also, if it should be disabled if the worker is not available, `worker.worker.isAvailable` can be useful too.

murilovmachado:
> One question, do you think is more reusable to split `loading` from `disabled` in `AddTaskButton`? Cause maybe we could always disable when `loading` is true, but that might make the behavior more opaque.

It's already split, I'm passing both `disabled` and `isLoading` as separate values to `AddTaskButton`. But I'm not making any assumption such as making it disabled by default if `isLoading` is true.

PR #539, opened 2021-11-08
Clear categories redux state when selecting a different helpline

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-754

This PR:
- Cleans categories form state on `UPDATE_HELPLINE` action. I think this is the cleanest way to do this, as it's part of the same action. The alternative would be to dispatch a second action to clear the categories after helpline changes.
- Renamed `fetchHelpline` to `setHelpline` (`TaskView` component) to empathize that it changes some state.
- Rework on `TaskView` effect that triggers ~`fetchHelpline`~ `setHelpline`: 
  While working on this issue, I've realized:
  - This effect was called each time _any_ value of `contactForm.contactlessTask` changed (cause it's in the dependencies array). While this is fine, the function ~`fetchHelpline`~ `setHelpline` was also invoked each time the effect executed. This is not a good behavior IMO as  ~`fetchHelpline`~ `setHelpline` is an async function that also dispatches a Redux action, we should call it when required only. 
  - The initial form state included `defaultHelpline` as the `contactForm.helpline` value. This was later on modified by the sequential calls to ~`fetchHelpline`~ `setHelpline` which lead to the right value, but I think is cleaner to start empty (`''`) and then set the proper value.
  
  Both of the above are addressed by:
  - `contactForm.helpline` initializes as empty string, then is properly set. Should we add some kind of check here to at least warn us if `helpline` is ever empty when the form is submitted?
  - Including conditions that need to be met in order to call ~`fetchHelpline`~ `setHelpline`.
  - ~`fetchHelpline`~ `setHelpline` will actually dispatch the `UPDATE_HELPLINE` action only if really needed (`helpline !== helplineToSave`).

Please take a deep review of this PR, as `helpline` is a critical piece of our system. Feel free to provide feedback if you think any of this changes is unnecessary/could be easier/whatever suggestion.

Note: You can checkout to [this commit](https://github.com/techmatters/flex-plugins/pull/539/commits/5a1f8a8438eb105c876d1217c41414339480bc75) that includes some `console.log`s prefixed with `>>>>>` to make easier spotting when ~`fetchHelpline`~ `setHelpline` runs and when it changes some value.
What to test in this PR?:
-  The initial issue is solved (categories are cleaned when helpline changes).
- `contactForm.helpline` starts empty but then is always set for all kind of tasks.
- ~`fetchHelpline`~ `setHelpline` runs only when required. That is if a) `contactForm.helpline` is not set or b) if the task is an offline contact and `contactForm.contactlessTask.helpline` changed.
- The correct `contactForm.helpline` is carried all the lifetime of the contact, being the one submitted with the form when the counselor saves it.

### DIFF
--- a/plugin-hrm-form/src/components/TaskView.tsx
+++ b/plugin-hrm-form/src/components/TaskView.tsx
@@ -56,15 +56,9 @@ const TaskView: React.FC<Props> = props => {
   // Set contactForm.helpline for all contacts on the first run. React to helpline changes for offline contacts only
   React.useEffect(() => {
     const setHelpline = async () => {
-      console.log('>>>>> 1) setHelpline called');
       if (task && !isStandaloneITask(task)) {
-        console.log('>>>>> 2) conditions met, checking helpline to save');
         const helplineToSave = await getHelplineToSave(task, contactlessTask || {});
-        console.log('>> helpline is: ', helpline);
-        console.log('>> helplineToSave is: ', helplineToSave);
-        console.log('>> contactlessTask.helpline is: ', contactlessTask?.helpline);
         if (helpline !== helplineToSave) {
-          console.log('>>>>> 3) Updating helpline!!');
           updateHelpline(task.taskSid, helplineToSave);
         }
       }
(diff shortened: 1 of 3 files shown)


murilovmachado (approved):
Thanks for all the description explaining the details of it. It looks good to me, the code is clear and the behavior looks correct!

nickhurlburt:
Otherwise, I think the explanation you gave and when you have chosen to update things makes sense.

nickhurlburt (approved):
Great!

PR #542, opened 2021-11-16
Prototype on dual write contacts

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-847
Design Document: https://benetech.box.com/s/k2jiz0c4reoarvtq1rt3re2lnfd28ya6

In order to test this easily, modify getConfig() to return:
```
return {
  ...,
  serverlessBaseUrl: 'https://serverless-9971-dev.twil.io',
  helpline: 'SaferNet',
  definitionVersion: 'br-v1',
  permissionConfig: 'br',
}
```

### DIFF
--- a/plugin-hrm-form/src/dualWrite/index.ts
+++ b/plugin-hrm-form/src/dualWrite/index.ts
@@ -1,0 +1,24 @@
+import { ITask } from '@twilio/flex-ui';
+
+import { getConfig } from '../HrmFormPlugin';
+import saveContactToSaferNet from './br';
+
+type DualWriteFn = (task: ITask, payload: any) => Promise<void>;
+
+type SaveContactByDefinitionVersion = {
+  [definitionVersion: string]: DualWriteFn;
+};
+
+const saveContactByDefinitionVersion: SaveContactByDefinitionVersion = {
+  'br-v1': saveContactToSaferNet,
+};
+
+export const saveContactToExternalBackend = async (task: ITask, payload: any) => {
+  const { featureFlags, definitionVersion } = getConfig();
+  if (!featureFlags.dualWrite) return;
+
+  const saveContact = saveContactByDefinitionVersion[definitionVersion];
+  if (saveContact) {
+    await saveContact(task, payload);
+  }
+};
--- a/plugin-hrm-form/src/services/ContactService.ts
+++ b/plugin-hrm-form/src/services/ContactService.ts
@@ -239,9 +239,13 @@ export const saveContact = async (
   shouldFillEndMillis = true,
 ) => {
   const response = await saveContactToHrm(task, form, workerSid, uniqueIdentifier, shouldFillEndMillis);
-  await saveContactToExternalBackend(task, response.requestPayload);
 
-  return response.responseJson;
+  // TODO: add catch clause to handle saving to Sync Doc
+  try {
+    await saveContactToExternalBackend(task, response.requestPayload);
+  } finally {
+    return response.responseJson;
+  }
 };
 
 export async function connectToCase(contactId, caseId) {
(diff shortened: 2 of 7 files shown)


plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -221,8 +228,21 @@ export async function saveToHrm(task, form, workerSid: string, uniqueIdentifier:
...
+
+export const saveContact = async (
+  task,
+  form,
+  workerSid: string,
+  uniqueIdentifier: string,
+  shouldFillEndMillis = true,
+) => {
+  const response = await saveContactToHrm(task, form, workerSid, uniqueIdentifier, shouldFillEndMillis);
+  await saveContactToExternalBackend(task, response.requestPayload);
```
nickhurlburt, line 242: The real trick of the whole matter is here. What happens if one call fails and the other succeeds?

Also note that this makes the two calls serially, meaning the counselor will wait longer for the contact to save.

murilovmachado: The idea of having it serial and in this order is that we will only try to save at their backend if it works fine saving on HRM first.

But if it fails when saving on their backend then there's a couple of things we could do:
- Keeping both backends in sync:
  - Delete the contact from HRM and warn the user with 'Error from backend';
  - Delete the contact from HRM and warn the user with 'Error from SaferNet backend';
- Don't keep the backends in sync:
  - Don't delete the contact and warn the user that 'Contact was not saved at SaferNet Backend'
  - Don't delete the contact and not warn him, but log the issue somewhere else;

I'm not sure if we should decide what happens in this scenario or if SaferNet gets to decide it. Also, in the future, could a second helpline want a different behavior for the error scenario? In the design document I've left this scenario in the section **What's Missing**.

murilovmachado: At first, choosing one of **Don't keep the backends in sync** options might seen a bad idea. But if we think from the user perspective, it could be very frustrating to not be able to save the contact at all (in case there's an issue with SaferNet backend). That's why I've kept them listed as options.

nickhurlburt: There are pros and cons to all of those options. Those should be called out in the design document so we can make a choice.

GPaoloni: Idempotence using task id on hrm helps us here. I'm only unsure about the offline contacts.

nickhurlburt (approved):
looks good to me!

plugin-hrm-form/src/dualWrite/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import { getConfig } from '../HrmFormPlugin';
+import saveContactToSaferNet from './br';
+
+type DualWriteFn = (task: ITask, payload: any) => Promise<void>;
+
+type SaveContactByDefinitionVersion = {
+  [definitionVersion: string]: DualWriteFn;
+};
+
+const saveContactByDefinitionVersion: SaveContactByDefinitionVersion = {
```
GPaoloni, line 12: Do you think is better to use definition version instead of some attribute in service configuration for this? I'm just thinking what can happen if we ever change the version and forget to "copy the behavior" for the new version or implement the new one.

plugin-hrm-form/src/dualWrite/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  'br-v1': saveContactToSaferNet,
+};
+
+export const saveContactToExternalBackend = async (task: ITask, payload: any) => {
```
GPaoloni, line 16: This is not something for now but could be added to our tech debt. I realize that `payload` is the very same `body` that is sent to hrm. I'm fine with leaving it as `any`, but imo it will be great if we start refining the types around `ContactService` to match what our api expects.

GPaoloni (approved):
Looks very good!

plugin-hrm-form/src/dualWrite/index.ts
murilovmachado: This is a good question. At first, I did that exactly, but then I started to think that our app may keep requiring more and more properties on **serviceConfig** that feel redundant (at least now that we don't actually have any **v2**, **v3** etc). Let's wait until we have those scenarios to decide if it's better to create another attribute, or re-use definition version in some way, eg.: getting a substring of it.

PR #546, opened 2021-11-24
Scrub Rollbar telemetry inputs

nickhurlburt:
See https://docs.rollbar.com/docs/rollbarjs-telemetry

@GPaoloni let's talk about this when we next get a chance.


GPaoloni (approved):
After reading the docs, seems like this is the option we need. Let me know if should help with testing!

nickhurlburt:
Tested locally, and I think this is working:
Screen Shot 2021-11-26 at 5 31 36 PM

So I can merge if there are no concerns.

PR #548, opened 2021-12-01
Changed to use helplineLanguage in service configuration

GPaoloni:
This PR removes the old `configuredLanguage` property and changes code to use `helplineLanguage` instead.
From this PR, the order of language that will be used is:
- `counselorLanguage` from worker attributes.
- `helplineLanguage` from service configuration.
- `defaultLanguage` hardcoded.

For messages sent to the person in contact, an attempt to retrieve `task.attributes.language` will be performed prior to the above.

The later commits changes the plugin `init` function to be `async` and allows fetching translations from remote source. In order for this to work properly we should add a component that indicates a loading state until the translation is complete. I can either finish this or merge only the first commit for the original changes, leaving the rest for a later one (if desired).

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -247,7 +247,7 @@ export default class HrmFormPlugin extends FlexPlugin {
    * @param flex { typeof import('@twilio/flex-ui') }
    * @param manager { import('@twilio/flex-ui').Manager }
    */
-  init(flex, manager) {
+  async init(flex, manager) {
     loadCSS('https://use.fontawesome.com/releases/v5.15.1/css/solid.css');
 
     if (process.env.NODE_ENV !== 'development')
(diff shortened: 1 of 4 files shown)


murilovmachado (approved):
Great, helplineLanguage is really a better name.

After this gets merged, we should update serviceConfiguration for all environments, correct? Or the approach we're following is: only update the serviceConfiguration of the helplines which default language is not english (eg.: SaferNet environments only)

stephenhand (approved):
LGTM

GPaoloni:
> After this gets merged, we should update serviceConfiguration for all environments, correct? Or the approach we're following is: only update the serviceConfiguration of the helplines which default language is not english (eg.: SaferNet environments only)

I think the later approach makes sense: default to English unless explicitly told.

> Are we going to remove the support for non-bundled translations going forward then?

I don't know to be honest, I just pushed this to have stored it somewhere.

GPaoloni:
Closing in favor of https://github.com/techmatters/flex-plugins/pull/559.

PR #550, opened 2021-12-01
Save pending contacts at the Sync Client

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-879
Serverless: [WIP](Change in the error500 function)
Primary reviewer: @GPaoloni 

When a contact fails to be saved on the external backend, it will be saved at Twilio's Sync Client, in a list called `pending-contacts`. The format of a pending contact is:
- **retries**: number of retries
- **task**: the Twilio Task
- **payload**: the payload used to save the contact at Aselo backend
- **error**: the error received when trying to save at the external backend

### DIFF
--- a/plugin-hrm-form/src/utils/sharedState.js
+++ b/plugin-hrm-form/src/utils/sharedState.js
@@ -1,4 +1,8 @@
 import { getConfig } from '../HrmFormPlugin';
+import { saveContactToExternalBackend } from '../dualWrite';
+
+const isSharedStateClientConnected = sharedStateClient =>
+  sharedStateClient && sharedStateClient.connectionState === 'connected';
 
 /**
  * Saves the actual form into the Sync Client
@@ -11,7 +15,7 @@ export const saveFormSharedState = async (form, task) => {
   if (!featureFlags.enable_transfers) return null;
 
   try {
-    if (sharedStateClient === undefined || sharedStateClient.connectionState !== 'connected') {
+    if (!isSharedStateClientConnected(sharedStateClient)) {
       console.error('Error with Sync Client conection. Sync Client object is: ', sharedStateClient);
       window.alert(strings.SharedStateSaveFormError);
       return null;
@@ -43,7 +47,7 @@ export const loadFormSharedState = async task => {
   if (!featureFlags.enable_transfers) return null;
 
   try {
-    if (sharedStateClient === undefined || sharedStateClient.connectionState !== 'connected') {
+    if (!isSharedStateClientConnected(sharedStateClient)) {
       console.error('Error with Sync Client conection. Sync Client object is: ', sharedStateClient);
       window.alert(strings.SharedStateLoadFormError);
       return null;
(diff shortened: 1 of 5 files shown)


plugin-hrm-form/src/utils/sharedState.js
```diff
@@ -11,7 +15,7 @@ export const saveFormSharedState = async (form, task) => {
   if (!featureFlags.enable_transfers) return null;
 
   try {
-    if (sharedStateClient === undefined || sharedStateClient.connectionState !== 'connected') {
+    if (!isSharedStateClientConnected(sharedStateClient)) {
       console.error('Error with Sync Client conection. Sync Client object is: ', sharedStateClient);
       window.alert(strings.SharedStateSaveFormError);
```
stephenhand, line 20: Should we be firing alerts & swallowing exceptions this low down the stack? Would it not be better to let the exception be thrown and let the caller  decide how to handle it?

plugin-hrm-form/src/utils/sharedState.js
```diff
@@ -66,3 +70,72 @@ export const loadFormSharedState = async task => {
...
+ * @param {*} error error returned when trying to save contact to external backend
+ */
+export const savePendingContactToSharedState = async (task, payload, error) => {
+  const { featureFlags, sharedStateClient, strings } = getConfig();
+
+  if (!featureFlags.enable_dual_write) return null;
+  if (!task || !payload) return null;
+
+  try {
+    if (!isSharedStateClientConnected(sharedStateClient)) {
```
stephenhand, line 120: Same here, should we be swallowing exceptions in low level client / utility functions? Whilst this error handling might be appropriate for the current use case, it's possible we might start calling it in other scenarios that want to handle the error differently?

plugin-hrm-form/src/utils/sharedState.js
```diff
@@ -66,3 +70,72 @@ export const loadFormSharedState = async task => {
...
+  ...(error.cause && { stack: error.cause }),
+});
+
+/**
+ * Saves a pending contact into the Sync Client
+ * @param {import("@twilio/flex-ui").ITask} task task used to save the contact
+ * @param {*} payload payload used to save the contact
+ * @param {*} error error returned when trying to save contact to external backend
+ */
+export const savePendingContactToSharedState = async (task, payload, error) => {
```
stephenhand, line 113: Dumb question time: What happens to the contacts after they are saved? Does an external system pick them up or is there something in the web app that does it? Or is it TODO?

stephenhand (approved):
Just a couple of questions / things to consider going forward, but LGTM

plugin-hrm-form/src/utils/sharedState.js
murilovmachado: I agree, throwing an exception and handling alerts at a higher step of the hierarchy would be a  better way to do this.

plugin-hrm-form/src/utils/sharedState.js
murilovmachado: A quick overview of saving contacts:
- The contact is saved to our backed (usually, that's all)
- Optionally, a contact can be saved in a third-party (external) backend
  - If it fails to be saved in the external backend, then it will be saved in the pending-contacts list (this PR)
  - Later on, we will implement a way to retry saving these pending contacts to the external backend

plugin-hrm-form/src/utils/sharedState.js
GPaoloni: I agree with this, but what would be a better approach? The Sync Client can be initially OK but fail in further calls. Should we check for it's state prior to try any call that involves it? It's been a while since I haven't seen this code but I remember doing this as a way to at least know if something went wrong without "blocking the counselor workflow".

plugin-hrm-form/src/utils/sharedState.js
```diff
@@ -66,3 +70,72 @@ export const loadFormSharedState = async task => {
...
+  const { featureFlags, sharedStateClient, strings } = getConfig();
+
+  if (!featureFlags.enable_dual_write) return null;
```
GPaoloni, line 116: Nothing wrong with this, but I think is redundant as this function is only called from dual write.

GPaoloni (approved):
Looks good to me! Nice to finally see sync lists in actions.

plugin-hrm-form/src/utils/sharedState.js
murilovmachado: In case it goes inside this error clause, the function currently:
- logs the error
- calls window.alert
- returns null

It could be changed to:
- logs the error (which I think it's the correct place to do it)
- but not call window.alert
- and throw an error (e.g.: SyncClientConnectionError) instead of return null

At setupActions file, where we use this function, I think it would be a better place to window.alert.

current setupActions code:
```
...
const documentName = await saveFormSharedState(form, payload.task);
...
```
Could be changed to:
```
...
try {
  const documentName = await saveFormSharedState(form, payload.task);
  ...
} catch (err) {
  if (err instanceof SyncClientConnectionError) {
    window.alert();
  }
}
// maybe a finally clause if we need
```

Something like this I think would be ideal. But since this is not the most straightforward change to test, I'll focus on getting the functionality ready and thid PR merged, before refactoring the way we handle these error alerts.

nickhurlburt (approved):
Looks good. Nothing further to add after other reviews.

PR #551, opened 2021-12-06
CHI-863: Add Full Story element identifiers

stephenhand:
* Added requested identifiers as `data-fs-id` attributes, to keep them separate from HTML IDs and reduce the chance of developers breaking them with updates
* Used naming convention `Category-ElementFunction-ControlType` where Category would be a high level functional area like 'Contact' or 'Case', ElementFunction would be what the element is supposed to do 'EndCallOrChat', 'Transfer', and ControlType is what the control is, 'Button', 'DropDown' etc.
* Changed some linter settings to not get upset about windows line endings.

There are 2 elements specified in CHI-863 that have not had attributes explicitly added to them: 
* The 'Accept' button, this is not part of our plugin but part of the Twlio core UI. There is not a simple reliable wauy of adding extra attributes to this markup as far as we are aware. The suggested alternative would be to use a class based CSS selector for the event capturing clicks of these buttons in Full Story - `button.Twilio-TaskButton-Accept` and `button.Twilio-IncomingTask-Accept` for the All Tasks panel and Incoming Task panel versions of this button respectively 
TODO - add data element to 'Accept' button, which is defined in flex rather than the plugin

### DIFF
--- a/plugin-hrm-form/.eslintrc.json
+++ b/plugin-hrm-form/.eslintrc.json
@@ -17,6 +17,7 @@
     "no-alert": "off",
     "import/no-extraneous-dependencies": "off",
     "jsx-a11y/tabindex-no-positive": "off",
-    "sonarjs/no-duplicate-string": "off"
+    "sonarjs/no-duplicate-string": "off",
+    "prettier/prettier": ["error", {"endOfLine": "auto"}]
   }
 }
--- a/plugin-hrm-form/.eslintrc.json
+++ b/plugin-hrm-form/.eslintrc.json
@@ -17,7 +17,6 @@
     "no-alert": "off",
     "import/no-extraneous-dependencies": "off",
     "jsx-a11y/tabindex-no-positive": "off",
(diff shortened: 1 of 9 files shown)


nickhurlburt (approved):
Looks good to me, though it'd be good to get a review from Murilo or Gian.

(Note: we will often list a Primary Reviewer in the PR description to make it clear who needs to review, and this is a good practice now that we have more people.)

One thing to note is that one missing thing from the issue is the End Chat button:
image

This is a Twilio button like the Accept button, so we probably just need the Twilio ID for it, rather than adding our own.

stephenhand:
Looks like we can select the 'End Call / Chat' button with `button.Twilio-TaskCanvasHeader-EndButton`

nickhurlburt:
Let's try to get this merged and deployed to Development before the team meeting tomorrow, to enable Jana to start working with it. A second review would be good to do before then, but not strictly necessary.

GPaoloni:
Looks good to me!

PR #552, opened 2021-12-07
Added custom fullstory events for most alerts in Aselo

stephenhand:
* Switched Aselo over to using the fullstory NPM package rather than injecting a script tag
* Dedicated helpers for logging backend errors and form validation warnings. 
* Details of exactly what form validations failed are provided in Form Error event payloads
* Error messages & stacks are provided in backend error payloads
* Added a second local jest config JS to run jest standalone - so you can run tests that don't need to run in Twilio fast locally

Not sure who should be 'primary' reviewer for this, so making it @murilovmachado


plugin-hrm-form/src/fullStory/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+/**
+ * Sends form validation errors to fullstory.com, under the event name 'Form Error'
+ * Specvific form error information is sent as a JSON blob in a single property for now because I'm not sure how much fullstory likes dynamic payloads
+ * @param origin - A string identifying the form the error originated from
+ * @param formError - The error object produced by the form validation (i.e. the first parameter of the error handler)
+ */
+export function recordFormValidationError(origin: string, formError: FieldValues): void {
```
GPaoloni, line 36: I see that `recordFormValidationError`  is imported in many files but not used outside `recordingErrorHandler`. Is this intentional?

plugin-hrm-form/src/fullStory/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+}
+
+/**
+ * Wrapping function, converts a standard form error handler into one which records the error in fullstory first
+ * @param origin - A string identifying the form the error originated from
+ * @param handler - The core error handling logic you want to augment
+ * @return - the augmented handler, ready to use in a form
+ */
+export function recordingErrorHandler(
```
GPaoloni, line 60: This is the function we the devs should use around `react-hook-forms` to build the `onError` callback, right?

plugin-hrm-form/src/components/tabbedForms/BottomBar.jsx
```diff
@@ -97,10 +99,10 @@ class BottomBar extends Component {
     }
   };
 
-  onError = () => {
+  onError = recordingErrorHandler('Tabbed HRM Form', () => {
```
GPaoloni, line 102: I realize that we have many different error identifiers like `Tabbed HRM Form`. Is a good idea to group this in a sort of errors dictionary to have them in a single place? Maybe in the FS helpers file? Something like
```
const TabbedHRMFormError = 'Tabbed HRM Form';
const UpdateCaseError = 'Update Case';
// .... more error identifiers
```

GPaoloni (approved):
This is great! I haven't tested if the events are being sent to FS, but the code looks amazing and the API being exposed feels like it is exactly what we need.

plugin-hrm-form/src/fullStory/index.ts
stephenhand: Yes, this is a convenience function to cut down on boilerplate a little, but you can still call `recordFormValidationError` directly in your handler if it this wrapper function doesn't suit your requirements

plugin-hrm-form/src/fullStory/index.ts
stephenhand: The `recordFormValidationError` imports are probably redundant now, I will remove them

plugin-hrm-form/src/components/tabbedForms/BottomBar.jsx
stephenhand: I prefer to have these things defined close to the code that uses it rather than in a central list which I need to jump out of the code to refer back to. Do you have a use case in mind where it would be beneficial for them to be centralised? If we start hitting scenarios where the same errors are being recorded from multiple places, we can define an `enum` for them in `/fullStory/index.ts`?

I could define them as consts in the file where they are being referenced, although this seems a little redundant in cases where the string is only used once.

plugin-hrm-form/jest-standalone.config.js
```diff
@@ -0,0 +1,10 @@
+// For rapidly running non flex UI dependent tests locally
```
murilovmachado, line 1: Great, currently running any tests locally indeed takes a long time!

plugin-hrm-form/src/fullStory/index.ts
murilovmachado: I was going to ask exactly that, so it's our call if we use `recordingErrorHandler` passing the handler, or just call `recordFormValidationError` separately and do not use `recordingErrorHandler `, correct?

plugin-hrm-form/babel.config.js
```diff
@@ -1,5 +1,5 @@
 // babel.config.js
 module.exports = {
-  presets: ['@babel/preset-env', '@babel/preset-react'],
+  presets: ['@babel/preset-env', '@babel/preset-react', '@babel/preset-typescript'],
```
murilovmachado, line 3: nice adittion.

plugin-hrm-form/jest-standalone.config.js
```diff
@@ -0,0 +1,10 @@
...
+
+const { defaults } = require('jest-config');
+
+module.exports = config => {
+  config = config || { ...defaults, rootDir: '.' };
+  config.transformIgnorePatterns = ['/node_modules/(?!wavesurfer.js)'];
```
murilovmachado, line 7: I don't understand what this line does, particularly.

plugin-hrm-form/src/utils/setUpMonitoring.js
```diff
@@ -81,12 +82,10 @@ function setUpRollbarLogger(plugin, workerClient, monitoringEnv) {
 }
 
 function setUpFullStory() {
+  FullStory.init({
+    orgId: fullStoryId,
+  });
```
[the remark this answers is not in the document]
murilovmachado: Yeah!

murilovmachado (approved):
Looks good to me!

plugin-hrm-form/src/fullStory/index.ts
stephenhand: Exactly, either approach is valid.

plugin-hrm-form/jest-standalone.config.js
stephenhand: It's removed as part of the jest testing PR

PR #553, opened 2021-12-07
Update CaseInformationTab.json

Humairaa127:
Fix default value of 'hasConversationEvolved' field in Summary form to not default to 'Não'.

### DIFF
--- a/plugin-hrm-form/src/formDefinitions/br-v1/tabbedForms/CaseInformationTab.json
+++ b/plugin-hrm-form/src/formDefinitions/br-v1/tabbedForms/CaseInformationTab.json
@@ -32,7 +32,7 @@
       "label": "Houve uma evolução da conversa?",
       "type": "select",
       "options": [
-        { "value": " ", "label": " " },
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/formDefinitions/br-v1/tabbedForms/CaseInformationTab.json
```diff
@@ -32,6 +32,7 @@
       "label": "Houve uma evolução da conversa?",
       "type": "select",
       "options": [
+        { "value": " ", "label": " " },
```
nickhurlburt, line 35: Elsewhere, such as in Line 13 above, we have an empty value defined as an empty string. That is, we have used `""` rather than the `" "` that we have here. Is there a reason to include a space here instead?

Humairaa127: No reason, changed it, thanks Nick.

nickhurlburt:
Thanks, merging

PR #555, opened 2021-12-08
Chi 863 fix data attributes

stephenhand:
Some follow on fixes for https://github.com/techmatters/flex-plugins/pull/551 to places where the data-fs-id attribute wasn't being passed down the React component hierarchy correctly

Primary reviewer @murilovmachado


murilovmachado (approved):
Looks good to me!

PR #556, opened 2021-12-08
Use Jest directly for CI

stephenhand:
Converts CI to use jest directly, rather than `twilio flex:plugins:test` - cuts build times down from 14 to 8 minutes and makes running tests locally not far off twice as fast too.

Thoughts?


plugin-hrm-form/src/___tests__/services/ContactService.test.js
```diff
@@ -10,6 +10,9 @@ import { offlineContactTaskSid } from '../../types/types';
 
 const helpline = 'ChildLine Zambia (ZM)';
 
+// eslint-disable-next-line no-empty-function
```
stephenhand, line 13: This is hacky, running under vanilla Jest 'fetch' was not defined for some reason. I tried to add a polyfill in the `setup.js` file I added, but `global` was not defined there for some reason?

But there probably is a more elegant workaround than this if we go in this direction,

plugin-hrm-form/src/___tests__/services/ContactService.test.js
```diff
@@ -10,6 +10,9 @@ import { offlineContactTaskSid } from '../../types/types';
...
+global.fetch = global.fetch ? global.fetch : () => {};
```
stephenhand, line 14: This is hacky, but for some reason 'fetch' was undefined when running in vanilla jest. I tried polyfilling it in `setup.js` but `global` itself was undefined there (as was `window`) 😕 

Needless to say this workaround can probably be made more elegant / general if we go in this direction.

plugin-hrm-form/src/___tests__/setup.js
```diff
@@ -0,0 +1,5 @@
+// setup file
```
GPaoloni, line 1: I see that there are small differences between this and `src/setupTests.js`, but could they be merged/unified?

GPaoloni (approved):
I totally support this! :rocket: 

Watch mode can be executed with `npm run test:jest -- --watch`

Screenshot from 2021-12-08 12-35-34

Thanks for these improvements!

GPaoloni:
Also, given that we have no Flex UI dependent tests, should we move our CI to use this instead?

plugin-hrm-form/src/___tests__/setup.js
stephenhand: Yes, sorry I didn't know about the existing setup file because it wasn't explicitly referenced in any configuration - I will move my stuff into the pre-existing file

stephenhand:
> Also, given that we have no Flex UI dependent tests, should we move our CI to use this instead?

This PR includes updating the workflows to use 'regular' jest - but up to you guys if we make the swoitch, I can remove the workflow update before I merge if you'd prefer to run the twilio runner in CI

GPaoloni:
> This PR includes updating the workflows to use 'regular' jest - but up to you guys if we make the swoitch, I can remove the workflow update before I merge if you'd prefer to run the twilio runner in CI

Oh sorry, I didn't realized you was already doing it. I prefer to use this instead, I see no benefit including the Twilio wrapper.

plugin-hrm-form/package.json
```diff
@@ -11,6 +11,7 @@
     "predeploy": "npm run build",
     "deploy": "flex-plugin deploy",
     "test": "UNBUNDLED_REACT=true twilio flex:plugins:test",
+    "test:jest": "node ./node_modules/jest/bin/jest.js --config jest-standalone.config.js",
```
murilovmachado, line 14: Can't we just call jest directly, like: `"test:jest": "jest --config jest-standalone.config.js"`.
It feels weird to me to **explicitly** call jest from within `node_modules` folder.
(maybe we need to add jest as a devDependency?)

murilovmachado (approved):
The current tests running time are indeed annoying. Hopefullly this changes will save us a lot of time.

plugin-hrm-form/package.json
stephenhand: I will try installing jest as a dev dep and see if that works. I just didn't want the task relying on jest being installed globally

plugin-hrm-form/jest-standalone.config.js
```diff
@@ -0,0 +1,15 @@
+// For rapidly running non flex UI dependent tests locally
+const { defaults } = require('jest-config');
+
+module.exports = config => {
+  config = config || {
+    ...defaults,
+    rootDir: '.',
+    transformIgnorePatterns: ['/node_modules/(?!wavesurfer.js)'],
```
nickhurlburt, line 8: Note that this line was a workaround proposed by Twilio Support in December 2020 in order to address a compatibility issue in Flex UI 1.24.0 that was causing test failures. By now, they've _probably_ resolved that issue, and made this line unnecessary. There's no urgency on removing it, but you're welcome to test tests without it and remove it if they don't fail.

The error at the time was:
```
> flex-plugin test --env=jest-environment-jsdom-sixteen

FAIL src/___tests__/utils/transfer.test.js
● Test suite failed to run

Jest encountered an unexpected token

This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

...

Details:

/home/runner/work/flex-plugins/flex-plugins/plugin-hrm-form/node_modules/wavesurfer.js/src/util/index.js:1
({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){export { default as getId } from './get-id';
^^^^^^

SyntaxError: Unexpected token 'export'

at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:537:17)
at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:579:25)
at Object.<anonymous> (node_modules/@twilio/player/dist/index.js:11:12)
```

nickhurlburt (approved):
I am trying to think of anything we might miss if we're not running the Twilio jest testing. I can't think of anything. The speedup is significant, so I'm fine with moving forward with this.

plugin-hrm-form/jest-standalone.config.js
stephenhand: Works fine without that config line, removed

plugin-hrm-form/package.json
stephenhand: Works, changed as you suggest

PR #557, opened 2021-12-08
Sends taskLanguage when calling initPostSurvey

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-829
This PR is related to https://github.com/techmatters/serverless/pull/168

This PR adds an optional parameter (`taskLanguage`) to the `postSurveyInit` call.
To test this, point to `dev` environment of serverless, start a webchat task and edit it's attributes to contain a property like `"language": "pt-BR"`.  Any other language should result in an error log and the default English message being sent.

Screenshot from 2021-12-08 18-42-33

### DIFF
--- a/plugin-hrm-form/src/utils/setUpActions.js
+++ b/plugin-hrm-form/src/utils/setUpActions.js
@@ -315,16 +315,21 @@ export const setUpPostSurvey = setupObject => {
 };
 
 /**
- * @type {import('@twilio/flex-ui').ActionFunction}
+ * @param {ReturnType<typeof getConfig> & { translateUI: (language: string) => Promise<void>; getMessage: (messageKey: string) => (language: string) => Promise<string>; }} setupObject
+ * @returns {import('@twilio/flex-ui').ActionFunction}
  */
-const triggerPostSurvey = async payload => {
+const triggerPostSurvey = setupObject => async payload => {
   const { task } = payload;
 
   // if (featureFlags.enable_post_survey) {
   if (TaskHelper.isChatBasedTask(task)) {
+    const { taskSid } = task;
     const channelSid = TaskHelper.getTaskChatChannelSid(task);
+    const taskLanguage = getTaskLanguage(setupObject)(payload);
 
-    await postSurveyInit({ channelSid, taskSid: task.taskSid });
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -315,16 +315,21 @@ export const setUpPostSurvey = setupObject => {
...
 /**
- * @type {import('@twilio/flex-ui').ActionFunction}
+ * @param {ReturnType<typeof getConfig> & { translateUI: (language: string) => Promise<void>; getMessage: (messageKey: string) => (language: string) => Promise<string>; }} setupObject
+ * @returns {import('@twilio/flex-ui').ActionFunction}
  */
-const triggerPostSurvey = async payload => {
+const triggerPostSurvey = setupObject => async payload => {
   const { task } = payload;
 
   // if (featureFlags.enable_post_survey) {
```
stephenhand, line 324: Remove the commented out code?

plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -315,16 +315,21 @@ export const setUpPostSurvey = setupObject => {
 };
 
 /**
- * @type {import('@twilio/flex-ui').ActionFunction}
+ * @param {ReturnType<typeof getConfig> & { translateUI: (language: string) => Promise<void>; getMessage: (messageKey: string) => (language: string) => Promise<string>; }} setupObject
+ * @returns {import('@twilio/flex-ui').ActionFunction}
  */
-const triggerPostSurvey = async payload => {
+const triggerPostSurvey = setupObject => async payload => {
```
stephenhand, line 321: Could the naming of the function now be confusing, since it no longer triggers a post survey, but generates a function that triggers a post survey? Maybe something like `createPostSurveyTrigger` would be more descriptive? (sorry, I'm fussy about naming). 

That said, is there any need to curry this function? It's not exported, might it not be simpler just to have a first order function with 2 parameters? Does it need to be an Twilio `ActionFunction` type if we only use it internally?

stephenhand (approved):

murilovmachado (approved):

PR #559, opened 2021-12-09
Changed to use helplineLanguage in service config

GPaoloni:
This PR is an isolation of commit https://github.com/techmatters/flex-plugins/pull/548/commits/8bd94bfb25fc0fcf7581755e9b3344bdabe94931 from https://github.com/techmatters/flex-plugins/pull/548. 

This PR removes the old configuredLanguage property and changes code to use helplineLanguage instead.
From this PR, the order of language that will be used is:
- `counselorLanguage` from worker attributes.
- `helplineLanguage` from service configuration.
- `defaultLanguage` hardcoded.

For messages sent to the person in contact, we will look for the language like:
- `task.attributes.language`.
- `helplineLanguage`.

Note that this PR differs from https://github.com/techmatters/flex-plugins/pull/548:
Instead than doing
```
const getTaskLanguage = ({ counselorLanguage, helplineLanguage }) => ({ task }) =>
  task.attributes.language || counselorLanguage || helplineLanguage;
```

now we do 
```
const getTaskLanguage = ({ helplineLanguage }) => ({ task }) => task.attributes.language || helplineLanguage;
```

This is to not interfere with the work done in https://github.com/techmatters/flex-plugins/pull/557. @dee-luo Is this ok or we should use the `counselorLanguage` when calculating the task language?

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -32,13 +32,13 @@ export const getConfig = () => {
   const logoUrl = manager.serviceConfiguration.attributes.logo_url;
   const chatServiceSid = manager.serviceConfiguration.chat_service_instance_sid;
   const workerSid = manager.workerClient.sid;
-  const { helpline, counselorLanguage, helplineLanguage } = manager.workerClient.attributes;
+  const { helpline, counselorLanguage } = manager.workerClient.attributes;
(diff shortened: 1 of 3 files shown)


stephenhand (approved):

dee-luo:
@GPaoloni helpline language works great! To my knowledge, I don't believe counselorLanguage is being used anywhere, so no concerns.

PR #543, opened 2021-11-19
[DON'T REVIEW] CSAM IWF reporting from Aselo: UI

GPaoloni:
~NOTE: I'm splitting this in smaller PRs. Don't review yet.~
This PR is now divided in:
- https://github.com/techmatters/flex-plugins/pull/561
- https://github.com/techmatters/flex-plugins/pull/562
- https://github.com/techmatters/flex-plugins/pull/563
- https://github.com/techmatters/flex-plugins/pull/565

Jira: https://bugs.benetech.org/browse/CHI-703
Mockups: https://marvelapp.com/prototype/h32j53e/screen/82016151

This PR is related to:
- https://github.com/techmatters/serverless/pull/142
- https://github.com/techmatters/hrm/pull/131

This PR adds the UI to support CSAM reports.

~WIP:~
- ~Wrap under feature flag.~
- ~Interact with HRM (on development).~
- ~Change Contact types to include the `csamReports` association.~
- ~UI to display already submitted reports.~
- ~CI.~
- ~Work on focus.~
- Unit tests
  - ~Components~
  - ~csam-report reducer & actions~

To test the behavior you should use the above linked branch of HRM.

PR #561, opened 2021-12-09
CSAM: Adds routes types and rework Case components routing

GPaoloni:
This PR is part of "splitting https://github.com/techmatters/flex-plugins/pull/543 in smaller chunks".
Note: is easier to review commit-per-commit.

This PR 
- Isolates the routes where a `Case` can be present.
- Refines the routes being passed to the `Case` component.


plugin-hrm-form/src/states/routing/types.ts
```diff
@@ -44,6 +44,16 @@ export type AppRoutes =
...
 
+export type CSAMReportRoute = {
+  route: 'csam-report';
+  subroute: 'form' | 'loading' | 'status';
+};
+
+type OtherRoutes = CSAMReportRoute;
+
+// The different routes we have in our app
+export type AppRoutes = AppRoutesWithCase | OtherRoutes;
```
murilovmachado, line 55: Great! Later on we should split the other route types, too.

murilovmachado (approved):
Looks good!

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -529,7 +539,6 @@ const Case: React.FC<Props> = props => {
                 Icon={CancelIcon}
                 text={<Template code="BottomBar-CancelNewCaseAndClose" />}
                 onClick={handleCancelNewCaseAndClose}
-                data-fs-id="Case-SaveAndEnd-Button"
```
stephenhand, line 532: Not my fullstory IDs!! Do you just need to merge in from master?

stephenhand (approved):
A bit lost in this one I have to admit, I'll be able to review PRs in this area better when I understand the internal routing / redux stuff in the plugins a little more.

plugin-hrm-form/src/components/case/Case.tsx
GPaoloni: Yes this was solved after merging master into the branch. Thanks for pointing!

PR #565, opened 2021-12-09
CSAM: UI components

GPaoloni:
This PR is part of "splitting https://github.com/techmatters/flex-plugins/pull/543 in smaller chunks".
Notes: 
- Is easier to review commit-per-commit.
- I know this is still a big PR if you count lines, but reviewing it per-commit should be quite easy. I can split in more PRs if you consider this to be worthy tho.

This PR (ordered by commit):
- Adds one more type of input to our `formGenerators`: the `email` type input. It has a rudimentary validation to only allow email-like strings. Also every form generated input now has a `data-testid` (equals to it's `path`) so we can refer to them in tests.
- Adds the components to submit a new csam report:
  - `CSAMReportFormScreen`:  the form generated by consuming `CSAMReportFormDefinition`. This is not the final version (not identical to the mocks). Here's a lot of room for improvement torwards a more dynamic approach to use different csam report forms, but for now this is the easiest way to match the mocks (if we ever want dynamics forms, is likely that the mocks will need to be simplified).
  - `CSAMReportStatusScreen`: a screen to display the result of the submission request. Here we'll share the code obtained (i.e. csam report id).
  - `CSAMReport`: the component where we'll render the above two, plus all the logic for routing. Displays a spinner while the submission is being performed to the external source (it might fail, but there were no mocks to handle this case, so for now I'm just sending back to the form).
- Adds to `TabbedForms`: 
  - A button to open csam form (wrapped in a feature flag).
  - Renders the csam report attachments under contact summary tab (if there's any).
- Displays the csam reports attached to a contact in `ContactDetails`.

### DIFF
--- a/plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
+++ b/plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
@@ -1,0 +1,179 @@
+/* eslint-disable react/prop-types */
+import React from 'react';
+import { useForm, FormProvider } from 'react-hook-form';
+import { CircularProgress } from '@material-ui/core';
+import { connect, ConnectedProps } from 'react-redux';
+
+import CSAMReportStatusScreen from './CSAMReportStatusScreen';
+import CSAMReportFormScreen from './CSAMReportFormScreen';
+import { CSAMReportContainer, CSAMReportLayout, CenterContent } from '../../styles/CSAMReport';
+import { FormItemDefinition } from '../common/forms/types';
+import { getInputType } from '../common/forms/formGenerators';
+import { definitionObject, keys, initialValues } from './CSAMReportFormDefinition';
+import type { CustomITask } from '../../types/types';
+import { getConfig } from '../../HrmFormPlugin';
+import * as actions from '../../states/csam-report/actions';
+import * as routingActions from '../../states/routing/actions';
+import * as contactsActions from '../../states/contacts/actions';
+import { RootState, csamReportBase, namespace, routingBase, configurationBase } from '../../states';
+import { reportToIWF } from '../../services/ServerlessService';
+import { createCSAMReport } from '../../services/CSAMReportService';
+import useFocus from '../../utils/useFocus';
+
+type OwnProps = {
+  taskSid: CustomITask['taskSid'];
+};
+
+const mapStateToProps = (state: RootState, ownProps: OwnProps) => ({
+  csamReportState: state[namespace][csamReportBase].tasks[ownProps.taskSid],
+  routing: state[namespace][routingBase].tasks[ownProps.taskSid],
+  counselorsHash: state[namespace][configurationBase].counselors.hash,
+});
+
+const mapDispatchToProps = {
+  updateFormAction: actions.updateFormAction,
+  updateStatusAction: actions.updateStatusAction,
+  clearCSAMReportAction: actions.clearCSAMReportAction,
+  changeRoute: routingActions.changeRoute,
+  addCSAMReportEntry: contactsActions.addCSAMReportEntry,
+};
+
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ConnectedProps<typeof connector>;
+
+// exported for test purposes
+export const CSAMReportScreen: React.FC<Props> = ({
+  updateFormAction,
+  updateStatusAction,
+  clearCSAMReportAction,
+  changeRoute,
+  addCSAMReportEntry,
+  taskSid,
+  csamReportState,
+  routing,
+  counselorsHash,
+}) => {
+  const [initialForm] = React.useState(csamReportState.form); // grab initial values in first render only. This value should never change or will ruin the memoization below
+  const methods = useForm();
+  const firstElementRef = useFocus();
+
+  const currentCounselor = React.useMemo(() => {
+    const { workerSid } = getConfig();
+    return counselorsHash[workerSid];
+  }, [counselorsHash]);
+
+  const formElements = React.useMemo(() => {
+    const onUpdateInput = () => {
+      const values = methods.getValues(Object.values(keys));
+      updateFormAction(values, taskSid);
+    };
+
+    const generateInput = (e: FormItemDefinition, index: number) => {
+      const generatedInput = getInputType([], onUpdateInput)(e);
+      const initialValue = initialForm[e.name] === undefined ? initialValues[e.name] : initialForm[e.name];
+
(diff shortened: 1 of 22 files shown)


plugin-hrm-form/src/components/CSAMReport/CSAMReportStatusScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+    // how should we handle possible IWF API error here? Show a screen, an alert & go back to form?
+    <CSAMReportContainer data-testid="CSAMReport-StatusScreen">
+      <CSAMReportLayout>
+        <ButtonBase onClick={onClickClose} style={{ marginLeft: 'auto' }} data-testid="Case-CloseCross">
+          <HiddenText>
+            <Template code="Case-CloseButton" />
+          </HiddenText>
+          <Close />
+        </ButtonBase>
+        <Box marginTop="15%" marginBottom="auto">
```
stephenhand, line 51: Dumb question from a guy who hasn't done frontend dev for a while (and then not using React), - are inline styles considered ok to use in react? I thought styles were best kept in CSS classes rather than in the markup, but if inline styles are the react way then cool.

stephenhand (approved):

plugin-hrm-form/src/components/CSAMReport/CSAMReportStatusScreen.tsx
GPaoloni: Yes inline styles are not preferred, but this are not styles in themselves but props that the `Box` accepts and consumes. `Box` is simply a wrapper to add margins an paddings without modifying the inner component styles (the childrens of the `Box`). 
Screenshot from 2021-12-16 11-50-38

murilovmachado (approved):
Looks good to me. Just a couple of questions.

plugin-hrm-form/src/styles/CSAMReport/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+`;
+
+export const CSAMAttachmentText = styled(FontOpenSans)`
+  font-style: italic;
+  font-size: 13px;
+  color: ${props => props.theme.colors.defaultButtonColor};
+`;
+CSAMAttachmentText.displayName = 'CSAMAttachmentText';
+
+/*
```
nickhurlburt, line 82: If this isn't used, can it be removed?

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
```diff
@@ -0,0 +1,179 @@
...
+      return (
+        <CSAMReportStatusScreen
+          reportStatus={csamReportState.reportStatus}
+          onClickClose={onClickClose}
+          onSendAnotherReport={onSendAnotherReport}
+        />
+      );
+    }
+    default: {
+      window.alert('Invalid route reached!');
```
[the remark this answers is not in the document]
nickhurlburt: Agreed. And -- @stephenhand do we need to do anything in order to get this error picked up by full story?

plugin-hrm-form/src/components/ContactDetails.tsx
```diff
@@ -94,6 +94,12 @@ const Details: React.FC<Props> = ({
   const definitionVersion = definitionVersions[version];
   const addedBy = counselorsHash[createdBy];
 
+  const csamReportsAttached =
+    csamReports &&
+    csamReports
+      .map(r => `CSAM on ${format(new Date(r.createdAt), 'yyyy MM dd h:mm aaaaa')}m\n#${r.csamReportId}`)
```
nickhurlburt, line 100: Does this match how our dates are localized and displayed in other places?

nickhurlburt (approved):
Very little to add in addition to what other reviewers wrote. Great to see this!

plugin-hrm-form/src/components/ContactDetails.tsx
GPaoloni: In general we don't localize the strings, or at least, I couldn't find places in our code where we do (only a chunk of it, like the year/month/day). I've just changed to use a more general method we have been using. I can change this to be whatever format makes sense tho.

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -88,8 +88,6 @@
 
   "SharedStateSaveFormError": "The information stored in the form couldn't be saved. Task will be transferred anyway.",
   "SharedStateLoadFormError": "The information stored in the form by previous counsellor couldn't be retrieved. Starting current task with clear contact form.",
-  "SharedStateSaveContactError": "Could not save pending contact at the Shared State.",
```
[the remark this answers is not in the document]
GPaoloni: It was not intended, already reverted thanks for pointing.

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
```diff
@@ -0,0 +1,179 @@
...
+      accum: { [k in keyof typeof definitionObject]: JSX.Element },
+      [k, e]: [string, FormItemDefinition],
+      index: number,
+    ) => ({
+      ...accum,
+      [k]: generateInput(e, index),
+    });
+
+    return Object.entries(definitionObject).reduce(reducerFunc, null);
+  }, [firstElementRef, initialForm, methods, taskSid, updateFormAction]);
```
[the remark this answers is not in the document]
GPaoloni: Because the `createFormFromDefinition` returns an array that is then displayed, at most, split in half. In this case, to follow the mocks I needed to reference to different fields on the definition, hence just an array would make things harder. If we want to do keep the mocks, this won't be so customizable (unless we introduce "dependent-render type inputs", that only appears in screen when a condition on other field is matched).

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -189,6 +192,44 @@ const getInputType = (parents: string[], updateCallback: () => void, customHandl
...
+                {error && (
+                  <FormError>
+                    <Template id={`${path}-error`} code={error.message} />
+                  </FormError>
+                )}
+              </FormLabel>
+            );
+          }}
+        </ConnectForm>
+      );
```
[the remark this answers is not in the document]
GPaoloni: Addressed in https://github.com/techmatters/flex-plugins/pull/561/commits/a3c72908ccaaf58368d0246bcc15d82e3ca0716d. Thanks for pointing!

PR #566, opened 2021-12-10
Revert "Changed to use helplineLanguage in service config"

GPaoloni:
Reverts techmatters/flex-plugins#559


GPaoloni:
Closing as the CI failing means that https://github.com/techmatters/flex-plugins/pull/559 was not the cause of it.

PR #567, opened 2021-12-10
Custom goodbye message for SaferNet

murilovmachado:
Jira: https://bugs.benetech.org/browse/CHI-926
Primary reviewer: @GPaoloni 

This PR changes the Goodbye message to a custom one for when successfully saving a contact at SaferNet's backend. When it fails, it will fallback to the default goodbye message.

This should not affect other helplines

~~[WIP]: I was getting `Twilio: Error while setting LastConsumedMessageIndex` when saving a contact. Not sure if it was due to the Twilio Flex instability on Friday evening, or if it's something to fix. Gonna double check that on Monday.~~ This is happening in Aselo development as well, so it's not related to this PR.

### DIFF
--- a/plugin-hrm-form/src/utils/setUpActions.js
+++ b/plugin-hrm-form/src/utils/setUpActions.js
@@ -149,19 +149,28 @@ const sendMessageOfKey = messageKey => setupObject => async payload => {
   });
 };
 
+let customGoodbyeMessage;
+export const setCustomGoodbyeMessage = message => (customGoodbyeMessage = message);
+
 /**
  * @param {string} messageKey
  * @returns {(setupObject: ReturnType<typeof getConfig> & { translateUI: (language: string) => Promise<void>; getMessage: (messageKey: string) => (language: string) => Promise<string>; }) => import('@twilio/flex-ui').ActionFunction}
  */
-const sendSystemMessageOfKey = messageKey => setupObject => async payload => {
+const sendSystemMessageOfKeyOrCustom = messageKey => setupObject => async payload => {
+  if (customGoodbyeMessage) {
+    const message = customGoodbyeMessage;
+    customGoodbyeMessage = null; // Clear customGoodbyeMessage
+    return sendSystemMessage({ taskSid: payload.task.taskSid, message, from: 'Bot' });
+  }
+
   const { getMessage } = setupObject;
   const taskLanguage = getTaskLanguage(setupObject)(payload);
   const message = await getMessage(messageKey)(taskLanguage);
-  await sendSystemMessage({ taskSid: payload.task.taskSid, message, from: 'Bot' });
+  return sendSystemMessage({ taskSid: payload.task.taskSid, message, from: 'Bot' });
 };
 
 const sendWelcomeMessage = sendMessageOfKey('WelcomeMsg');
-const sendGoodbyeMessage = sendSystemMessageOfKey('GoodbyeMsg');
+const sendGoodbyeMessage = sendSystemMessageOfKeyOrCustom('GoodbyeMsg');
 
 /**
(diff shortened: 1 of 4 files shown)


plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -149,19 +149,28 @@ const sendMessageOfKey = messageKey => setupObject => async payload => {
...
+    customGoodbyeMessage = null; // Clear customGoodbyeMessage
+    return sendSystemMessage({ taskSid: payload.task.taskSid, message, from: 'Bot' });
+  }
+
   const { getMessage } = setupObject;
   const taskLanguage = getTaskLanguage(setupObject)(payload);
   const message = await getMessage(messageKey)(taskLanguage);
-  await sendSystemMessage({ taskSid: payload.task.taskSid, message, from: 'Bot' });
+  return sendSystemMessage({ taskSid: payload.task.taskSid, message, from: 'Bot' });
 };
```
GPaoloni, line 170: Given that this is only custom for goodbye message, I don't think this name is appropriate, as I would infer that is a more abstract function. Also, the not-custom part could be defined in terms of `sendMessageOfKey` rather than duplicating that chunk, unless of course is required for the goodbye message to be sent by `Bot`.

plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -149,19 +149,28 @@ const sendMessageOfKey = messageKey => setupObject => async payload => {
   });
 };
 
+let customGoodbyeMessage;
+export const setCustomGoodbyeMessage = message => (customGoodbyeMessage = message);
```
GPaoloni, line 153: I don't love this being mutated each time a contact is finished and then each time the message is consumed, but I don't think Redux is a better fit, and can't see a way to use callbacks for this piece, so it may be the best option.

GPaoloni (approved):
Looks good. I have not tested the behavior, how should I? Change the definition version to be `br-v1`?

murilovmachado:
> Looks good. I have not tested the behavior, how should I? Change the definition version to be `br-v1`?

Testing the behaviour would be not so straightforward, because the integration with SaferNet is not working 100% just yet. Just the code review for this time is fine.

plugin-hrm-form/src/utils/setUpActions.js
murilovmachado: Yeah, this was tricky. It was the only way I've found to make it work without having to drastically change the way we send goodbye messages.

stephenhand:
Looks god for this particular use case, but if we wanted to make custom copy for various parts of the UI generally supported, what we could do is:

Have an 'custom-text' folder, with a folder underneath for each helpline that wants custom copy, then localisation folders underneath that, and then using the webpack fakey environment variables we could specify which custom text folder(s) a build is using.

When the translations are loaded, the app could read this variable and simply load these up and overwrite the base translations using an `Object.assign` call. That way we could override any translatable copy in the app for any helpline without any specific code used where the custom text is used.

Thoughts?

murilovmachado:
> Looks god for this particular use case, but if we wanted to make custom copy for various parts of the UI generally supported, what we could do is:
> 
> Have an 'custom-text' folder, with a folder underneath for each helpline that wants custom copy, then localisation folders underneath that, and then using the webpack fakey environment variables we could specify which custom text folder(s) a build is using.
> 
> When the translations are loaded, the app could read this variable and simply load these up and overwrite the base translations using an `Object.assign` call. That way we could override any translatable copy in the app for any helpline without any specific code used where the custom text is used.
> 
> Thoughts?

Yep, separating custom text into its own folder seems good so we don't get our `messages.json ` files filled with unrelated custom strings. Let's keep an eye to see if the number of custom copies increases, so we can apply this change when needed.

PR #569, opened 2021-12-13
CHI-838: Script tweaks from SaferNet Production environment setup

stephenhand:
## Description

* createTwilioResources: An AWS Parameter already existing does not fail the script, it just moves on.
* createTwilioResources: Does not attempt to delete chatbots because that blocks the unique name from being reused for 30 days. Instead, on creation, if the bot is already found, it uses that instead
* createTwilioResources: Remove functions now attempt to remove resources even if they weren't created on this script run. This means they can fix things when previous cleanup attempts failed
* createTwilioResources: Refactored functioins to remove some copy & pasta
* Made Aselo Beta the baseline environment to verify generated workflows against. rather than Aselo Development,
* Various spelling & markdown corrections.

…script. Tries to remove resources even if not created that run. No longer removes bots and uses existing if they already exist. Uses Aselo Beta rather than Dev as compare baseline

### Checklist
[ X ] Corresponding issue has been opened
[ N/A ] New tests added
[ N/A ] Strings are localized

### Verification steps

Run these scripts setting up a new environment.

### DIFF
--- a/scripts/src/setupNewAccount/createTwilioResources.ts
+++ b/scripts/src/setupNewAccount/createTwilioResources.ts
@@ -13,10 +13,25 @@
  *  DATADOG_ACCESS_TOKEN=<Datadog Access Token>
  */
 import twilio from 'twilio';
+import { KeyInstance } from 'twilio/lib/rest/api/v2010/account/key';
+import { ServiceInstance } from 'twilio/lib/rest/sync/v1/service';
+import { WorkflowInstance } from 'twilio/lib/rest/taskrouter/v1/workspace/workflow';
+import { TaskQueueInstance } from 'twilio/lib/rest/taskrouter/v1/workspace/taskQueue';
+import { TaskChannelInstance } from 'twilio/lib/rest/taskrouter/v1/workspace/taskChannel';
 import { saveSSMParameter } from '../helpers/ssm';
 import { createS3Bucket } from '../helpers/s3';
 import { logError, logSuccess, logWarning } from '../helpers/log';
 import { ScriptsInput } from './types';
+import { removeResource, removeWorkspaceResource } from './removeTwilioResource';
+
+const PRE_CHATBOT_UNIQUE_NAME = 'demo_chatbot';
+const POST_CHATBOT_UNIQUE_NAME = 'post_survey_bot';
+
+const WORKFLOW_FRIENDLY_NAME = 'Master Workflow';
+const SYNC_SERVICE_FRIENDLY_NAME = 'Shared State Service';
+const HRM_API_KEY_FRIENDLY_NAME = 'hrm-static-key';
+
+const SURVEY_RESOURCE_FRIENDLY_NAME = 'Survey';
 
 const client = twilio(process.env.TWILIO_ACCOUNT_SID, process.env.TWILIO_AUTH_TOKEN);
 
@@ -131,9 +146,15 @@ const saveStateKeyToSSM = (key: keyof DynamicState) => async (state: State) => {
   const description = getSSMDescriptionFunction[key](state);
   const tags = getSSMTags(state);
 
-  await saveSSMParameter(name, value, description, tags);
+  try {
+    await saveSSMParameter(name, value, description, tags);
 
-  logSuccess(`AWS Parameter Store: Succesfully saved ${name}`);
+    logSuccess(`AWS Parameter Store: Successfully saved ${name}`);
+  } catch (error) {
+    if ((<any>error).code === 'ParameterAlreadyExists') {
+      logWarning(`Parameter ${key} already exists, moving on.`);
+    } else throw error;
+  }
   return state;
 };
 
@@ -185,7 +206,7 @@ const createTaskQueue = async (state: State) => {
     targetWorkers: `helpline=='${state.helpline}'`,
   });
 
-  logSuccess(`Twilio resource: Succesfully created task queue ${q.sid}`);
+  logSuccess(`Twilio resource: Successfully created task queue ${q.sid}`);
   return { ...state, taskQueueSid: q.sid };
 };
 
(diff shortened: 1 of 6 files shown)


scripts/src/setupNewAccount/workflowGenerator.ts
```diff
@@ -31,22 +31,25 @@ type WorkflowConfig = {
 };
 
 export const checkWorkflowInSync = async (remoteUrl: string, templatePath: string) => {
-  const aseloDevConfig = {
-    environment: 'Development',
-    shortEnvironment: 'DEV',
+  const aseloBetaConfig = {
```
stephenhand, line 34: @GPaoloni - Is this correct now after your changes? It looked to be in local tests.

murilovmachado: I thought it should be: 
```
{
    environment: 'Staging', // Staging here instead of Beta
    shortEnvironment: 'STG',
    helpline: 'Aselo',
    shortHelpline: 'AS',
}
```

But plz @GPaoloni confirm.

murilovmachado (approved):
Looks good to me! Just on question on `aseloBetaConfig`

scripts/src/setupNewAccount/workflowGenerator.ts
stephenhand: I tried that initially but it didn't match in the comparison, which is why I wanted to query it

scripts/src/setupNewAccount/removeTwilioResource.ts
```diff
@@ -0,0 +1,57 @@
...
+  resourceGetter: () => { list: ResourceListFunction<T> },
+): Promise<void> => {
+  const resource = (await resourceGetter().list({ limit: 20 })).find(
+    (key) => key.friendlyName === resourceFriendlyName,
+  );
+  if (resource) {
+    await resource.remove();
+    logWarning(`Twilio resource: Successfully removed ${resourceTypeDescription} ${resource.sid}`);
+  }
+};
```
GPaoloni, line 32: I'm aware this script is used for new accounts only and can be convinced, but rn I think we shouldn't rely on friendly names to do this operation, but in sids instead. Friendly names can be (and in development) are repeated. They can also repeat if for some reason (a bug?) one of this scripts fails deleting something.

scripts/src/setupNewAccount/removeTwilioResource.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @param client
+ * @param workspaceSid
+ * @param resourceFriendlyName - the 'friendly name' used in the Twilio GUI - used to try to locate the resource.
+ * @param resourceTypeDescription - a string just used in logs to differentiate the resource from others in the script.
+ * @param resourceGetter - A function to return the resource type provided a workspace instance, e.g. workspace => workspace.workflows
+ */
+export const removeWorkspaceResource = async <T extends TwilioResource>(
+  client: Twilio,
+  workspaceSid: string | undefined,
+  resourceFriendlyName: string,
```
GPaoloni, line 45: Same as above comment.

scripts/src/setupNewAccount/workflowGenerator.ts
GPaoloni: It was not merged against master maybe? It should match this config now.

scripts/src/setupNewAccount/workflowGenerator.ts
```diff
@@ -31,22 +31,25 @@ type WorkflowConfig = {
...
+    environment: 'Beta',
```
GPaoloni, line 35: This is not the correct (nor a valid env for this scripts), but `Staging` is.

scripts/src/setupNewAccount/removeTwilioResource.ts
stephenhand: Unfortunately the SID is not known unless the resource was created that run. If anything goes wrong and resources are left on the system after the script fails, subsequent runs will never clean them up - you have to go in and start cleaning things up manually.

The change is intended to make the script a bit more robust and repeatable.  There's a small chance of friendly name conflicts, but I think the above problem is more likely, which is why I raised the change. We can leave it as was if people think friendly name clashes are a bigger problem though

scripts/src/setupNewAccount/createTwilioResources.ts
```diff
@@ -131,9 +146,15 @@ const saveStateKeyToSSM = (key: keyof DynamicState) => async (state: State) => {
...
+  try {
+    await saveSSMParameter(name, value, description, tags);
 
-  logSuccess(`AWS Parameter Store: Succesfully saved ${name}`);
+    logSuccess(`AWS Parameter Store: Successfully saved ${name}`);
+  } catch (error) {
+    if ((<any>error).code === 'ParameterAlreadyExists') {
+      logWarning(`Parameter ${key} already exists, moving on.`);
+    } else throw error;
+  }
```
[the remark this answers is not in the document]
stephenhand: Aborting in this scenario is what it currently does. it causes issues trying to run it a second time which is why I changed the behaviour. If we could update the values then that would be ideal, but the API doesn't seem to allow it when I tried, something about the type of parameter they are?

scripts/src/setupNewAccount/removeTwilioResource.ts
GPaoloni: Makes sense. I just wanted to point out that potential issue.

This raises another concern to me. If someone accidentally runs this script on the wrong account, we delete everything :boom:. I can imagine that accident occurring if the script user forgets to update the env vars (either in bash or .env file). Maybe asking the credentials in a prompt/a confirmation are a good idea?

stephenhand:
To be honest, most of this PR might become redundant if the terraform approach looks like a good way to go - so maybe we don't need too much time debating about the best approaches for creating / removing resources :-)

scripts/src/setupNewAccount/removeTwilioResource.ts
stephenhand: I think that's a good idea, we could even just read back the friendly name of the account back and ask to confirm that this is correct.

But I think I'll put this PR on ice until we decide scripts vs terraform as a team - we can always revisit it

PR #571, opened 2021-12-15
Terraform Twilio Environment Provisioning POC

stephenhand:
## Description

A proof of concept for provisioning our Twilio infrastructure via terraform

Exactly what is covered and what is not is described in the README.md that is part of the PR

### Checklist
[ NO ] Corresponding issue has been opened
[ N/A ] New tests added
[ N/A ] Strings are localized

### Verification steps

Use the terraform scripts on a new environment


twilio-iac/aselo-internal/serverless.tf
```diff
@@ -0,0 +1,19 @@
...
+  service_sid = twilio_serverless_services_v1.serverless.sid
+  unique_name = "dev"
+  domain_suffix = "dev"
+}
+
+resource "twilio_serverless_services_environments_v1" "production" {
+  service_sid = twilio_serverless_services_v1.serverless.sid
+  unique_name = "production"
+  domain_suffix = "production"
+}
```
murilovmachado, line 19: Do we have to explicitly declare these two environments? Just by creating `resource "twilio_serverless_services_v1" "serverless"` doesn't it give us those two environments?

twilio-iac/aselo-internal/service-configuration.json
```diff
@@ -0,0 +1,28 @@
...
+    "hrm_base_url": "https://hrm-#.tl.techmatters.org",
+    "logo_url": "https://tl-public-chat-#-#.s3.amazonaws.com/SaferNet-logo.png",
+    "serverless_base_url": "https://serverless-#-#.twil.io",
+    "configuredLanguage": "#-#",
+    "definitionVersion": "#-#",
+    "permissionConfig": "#",
+    "pdfImagesSource": "https://tl-public-chat-#-#.s3.amazonaws.com",
+    "multipleOfficeSupport": true
+  }
+}
```
murilovmachado, line 28: As I understood, the twilio provider doesn't support setting the ServiceConfiguration. Are we using this by another script or is it just here to be tracked in the repo? Should we have one file like this for each environment?

twilio-iac/aselo-internal/studioFlow.tf
```diff
@@ -0,0 +1,288 @@
...
+          }
+        ],
+        "type": "trigger",
+        "name": "Trigger",
+        "properties": {
+          "offset": {
+            "y": -70,
+            "x": 0
+          }
+        }
```
murilovmachado, line 24: The worst thing about these Studio Flow json files is that they keep both the components logic and UI positioning in the same file. We can live with that, but I guess we can have a lot of diffs around this.

I believe future enhancement to Studio Flows will continue to be made directly from Twilio console UI, instead of editing this file.

And a third thing to mention is that different helplines may have different flows. So we may need to have one file for each environment.

twilio-iac/aselo-internal/taskRouter.tf
```diff
@@ -0,0 +1,100 @@
...
+            "expression": "helpline=='${var.helpline}'",
+            "targets": [
+              {
+                "expression": "(worker.waitingOfflineContact != true AND ((task.channelType == 'voice' AND worker.channel.chat.assigned_tasks == 0) OR (task.channelType != 'voice' AND worker.channel.voice.assigned_tasks == 0)) AND ((task.transferTargetType == 'worker' AND task.targetSid == worker.sid) OR (task.transferTargetType != 'worker' AND worker.sid != task.ignoreAgent))) OR (worker.waitingOfflineContact == true AND task.targetSid == worker.sid AND task.isContactlessTask == true)",
+                "queue": twilio_taskrouter_workspaces_task_queues_v1.helpline_queue.sid
+              }
+            ]
+          }
+        ]
+      }
```
murilovmachado, line 57: This files is also another one that may be different from one helpline to another.

murilovmachado (approved):
Overall looks good. I have a couple of question, but my main question is:

Some of these files that would be different from one helpline to another, such as `chatbots.tf`, `studioFlow.tf`, `taskRouter.tf`.
- Should we create one of each for each environment?
  - So we could track the current state as code in the repo
  - And also track the tfstate correctly?
- Or they will be used just as a basic bootstrap?

twilio-iac/aselo-internal/service-configuration.json
stephenhand: So I think configuration can be done with provisioners that run an NPM script that patches the config. This would be done in a TF file so the appropriate parts can be injected in - this file was just some early experimentation, it's not used in the POC, I will delete it,

twilio-iac/aselo-internal/studioFlow.tf
stephenhand: There's nothing stopping us having a process where people are using the UI to edit flows, but I think we need to get the definitions under source control so it can be versioned, reviewed etc. The process might be:
* Edit in GUI in Staging
* Test it out
* Import updated state to Terraform
* Export Json & Update Terraform definition
* Terraform Apply In Staging to line up the checked in definition with the state
* Raise a PR to merge the updated flow definition

The terraform script can then be applied to production - we never have to manually update prod in the GUI

stephenhand: As far as customisations per helpline are concerned, I'm still figuring out exactly how that would look, but I think we can have a common core, then override certain definitions based on the terraform workspace they are part of. This might involve refactoring some of the resources into modules - but I think once that is done, you can conditionally replace the standard studioFlows.tf with a custom version based on the workspace or variables passed in

twilio-iac/aselo-internal/serverless.tf
stephenhand: I don't know, perhaps it does, I don't know if you get the normal defaults if you create the resources via terraform? We'd still need these defined if we wanted to reference them elsewhere in Terraform (like setting up environment variables), but they are possibly redundant right now

twilio-iac/aselo-internal/aws.tf
```diff
@@ -0,0 +1,66 @@
...
+  docs_s3_location = "tl-aselo-docs-${lower(var.short_helpline)}-${lower(var.environment)}"
+  chat_s3_location = "tl-public-chat-${lower(var.short_helpline)}-${lower(var.short_environment)}"
+}
+
+resource "aws_s3_bucket" "docs" {
+  bucket = local.docs_s3_location
+  cors_rule {
+    allowed_headers = ["*"]
+    allowed_methods = ["GET", "POST", "PUT"]
+    allowed_origins = ["https://flex.twilio.com", "http://localhost:3000", "https://localhost:3000"]
```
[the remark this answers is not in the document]
stephenhand: > Overall looks good. I have a couple of question, but my main question is:
> 
> Some of these file would be different from one helpline to another, such as `chatbots.tf`, `studioFlow.tf`, `taskRouter.tf`.
> 
> * Should we create one of each for each environment?
>   
>   * So we could track the current state as code in the repo
>   * And also track the tfstate correctly?
> * Or they will be used just as a basic bootstrap?

At the moment I'm thinking we'd using Terraform workspaces to track the .tfstate for different accounts / environments, but I suppose putting everything into modules and just having a different root module for each environment could also work?

If we use workspaces, we can put the resources we want to customise into modules, then use the `count` meta-attribute to include / exclude different versions of resources e.g. `module "safernet_chatbot" { ... count="var.chatbot_type == "safernet_chatbot"" ... }` or something like that

If we used separate root modules, we could just put different modules in each of the root modules. 

I think either could work, thoughts?

twilio-iac/aselo-internal/studioFlow.tf
nickhurlburt: Note that Twilio has just launched subflows as GA: https://www.twilio.com/docs/studio/subflows  This can be helpful in defining core, shared subflows that are the same across environments (environments can't actually share the same subflow entity, but we could have shared, common terraform code across them).

nickhurlburt: For per-account config, we may be able to use similar methods as we have in the [https://github.com/techmatters/infrastructure-config/tree/master/terraform](AWS terraform modules)

twilio-iac/aselo-terraform/aarambh-production.tfvars
```diff
@@ -0,0 +1,8 @@
+helpline    = "Aarambh"
+short_helpline = "IN"
+short_environment = "PROD"
+environment = "Production"
+account_sid = "AC64efffad72dccd812ca0c156355217ed"
+datadog_app_id = "9869f8b7-106c-42ea-9f74-1e3ab74daee9"
```
nickhurlburt, line 6: We should keep these ids out of public repos

nickhurlburt (commented):
We will face key questions around how we provide a default base for every helpline while also allowing for individual configurations that remain manageable. Overall this looks good.

twilio-iac/aselo-terraform/aarambh-production.tfvars
stephenhand: The Twilio account was added as a safety check - basically if it uses the account ID from tfvars it will never accidentally run the wrong terraform against the wrong account because you had the wrong creds in your environment - it would just fail to authenticate because you were using the wrong auth token. 

If we took out this and relied entirely on the environment variables, it would remove that extra safety so it would be good to think of an alternative - perhaps using this: https://registry.terraform.io/providers/bwoznicki/assert/latest/docs/data-sources/test

That said, you would have to be really not paying attention to the plan output saying it was going to change EVERYTHING - or running it suppressing the confirmation, which you should never do manually, so maybe it's ok

twilio-iac/aselo-internal/serverless.tf
GPaoloni: As far as I remember, the production environment is NOT created automatically, this is the correct think imo.

twilio-iac/aselo-terraform/README.md
```diff
@@ -0,0 +1,89 @@
...
+* API Keys - Whilst API Keys can be created using the twilio terraform provider, it's a bit useless because it provides no way of accessing the secret to record somewhere, so a key created in terraform can never be accessed as far as I can tell.
+* Okta - doesn't set up anything in Twilio or Okta for this right now. No sign of any support for setting up single sign on in the Twilio provider, but there is an Okta provider.
+* DataDog - it puts the keys you provide via tfvars in the AWS Parameter Store, but it won't provision the application in DataDog for you (but it could!).
+* Default Studio Flows - it doesn't clean up the original 'Messaging Flow', because it's never under control of terraform. Could be removed with a provisioner possibly, but this might result in unexpected behaviour if somebody duplicates the terraform managed Messaging Flow to test with and then terraform goes and deletes it...
+...
+
+### What it shouldn't do
+
+* Workflows - It won't generate new serverless & flex github workflows. The aim is to refactor these into single workflows, so in the meantime use the existing scripts in /scripts/setupNewAccount for these
```
GPaoloni, line 87: I'm not sure I get this one. This is for the deploy workflows on GH actions? If so, I don't think we want that, but again, I may be misunderstanding.

twilio-iac/aselo-terraform/aarambh-production.tfvars
```diff
@@ -0,0 +1,6 @@
+helpline    = "Aarambh"
+short_helpline = "IN"
+short_environment = "PROD"
+environment = "Production"
+operating_info_key = "aselo-dev"
+serverless_url = "https://serverless-3078-production.twil.io"
```
GPaoloni, line 6: While not strictly a secret, we've been trying to not expose all of this (specially urls to particular subdomains). I think this entire scripts folder does not belongs to this repo anymore, but to the private one instead. What do you think?

twilio-iac/aselo-terraform/aws.tf
```diff
@@ -0,0 +1,68 @@
...
+  block_public_policy = true
+  restrict_public_buckets = true
+}
+
+resource "aws_s3_bucket" "chat" {
+  bucket = local.chat_s3_location
+  cors_rule {
+    allowed_headers = ["*"]
+    allowed_methods = ["GET", "POST", "PUT"]
+    allowed_origins = ["https://flex.twilio.com"]
```
GPaoloni, line 32: Do we ever want to PUT/POST against this bucket? :thinking:

twilio-iac/aselo-terraform/aws.tf
```diff
@@ -0,0 +1,68 @@
...
+    // API Key secrets are not accessible from the twilio terraform provider
+    // HRM_STATIC_KEY = jsonencode(["TWILIO", "NOT_SET", "Twilio account - HRM static secret to perform backend calls"])
+    S3_BUCKET_DOCS = jsonencode(["TWILIO", local.docs_s3_location, "Twilio account - Post Survey bot chat url"])
+    POST_SURVEY_BOT_CHAT_URL = jsonencode(["TWILIO", "https://channels.autopilot.twilio.com/v1/${var.account_sid}/${twilio_autopilot_assistants_v1.post_survey.sid}/twilio-chat", "Twilio account - Post Survey bot chat url"])
+    OPERATING_INFO_KEY = jsonencode(["TWILIO", var.operating_info_key, "Twilio account - Operating Key info"])
+    APP_ID = jsonencode(["DATADOG", var.datadog_app_id, "Datadog - Application ID"])
+    ACCESS_TOKEN = jsonencode(["DATADOG", var.datadog_access_token, "Datadog - Access Token"])
+
+  }
+  name  = "${var.short_environment}_${jsondecode(each.value)[0]}_${var.short_helpline}_${each.key}"
```
GPaoloni, line 64: Maybe a comment can help cause I struggled a bit understanding it, but what an awesome trick! :raised_hands:

twilio-iac/aselo-terraform/chatbots.tf
```diff
@@ -0,0 +1,404 @@
...
+      "collect" : {
+        "validate_on_failure" : "task://collect_fallback"
+      }
+    }
+  })
+  log_queries = true
+}
+
+//Post Survey
+resource "twilio_autopilot_assistants_v1" "post_survey" {
```
GPaoloni, line 52: No resource is ever linked to this bot. Is that part of the subsequent PR? And it may be better to group the steps by resource (everything related to pre bot, then everything for post bot).

twilio-iac/aselo-terraform/flex.tf
```diff
@@ -0,0 +1,14 @@
+resource "twilio_flex_flex_flows_v1" "messaging_flow" {
+  channel_type  = "sms"
+  chat_service_sid = twilio_chat_services_v2.flex_chat_service.sid
+  friendly_name = "Flex Messaging Channel Flow"
+  integration_type = "studio"
+  integration_flow_sid = twilio_studio_flows_v2.messaging_flow.sid
```
GPaoloni, line 6: Does this points to the newly created flow? Can it eventually get mixed with the default one if not cleared?

twilio-iac/aselo-terraform/survey.tf
```diff
@@ -0,0 +1,28 @@
...
+// Workflow
+resource "twilio_taskrouter_workspaces_workflows_v1" "survey_workflow" {
+  friendly_name = "Survey"
+  workspace_sid = twilio_taskrouter_workspaces_v1.flex_task_assignment.id
+  configuration = jsonencode({
+  "task_routing": {
+        "filters": [
+          {
+            "filter_friendly_name": "Survey Filter",
+            "expression": "helpline=='${var.helpline}'",
```
GPaoloni, line 17: Not a big issue but I don't think this expression is right in here.

twilio-iac/aselo-terraform/taskRouter.tf
```diff
@@ -0,0 +1,88 @@
...
+variable "environment" {}
+variable "short_environment" {}
+variable "operating_info_key" {}
+
+// Workspaces
+resource "twilio_taskrouter_workspaces_v1" "flex_task_assignment" {
+  friendly_name      = "Flex Task Assignment"
+  multi_task_enabled = true
+  event_callback_url = "${var.serverless_url}/webhooks/taskrouterCallback"
+  events_filter = "task.created,task.canceled,task.completed" //Ignore the docs where it implies this should be 'EventsFilter=task.created, task.canceled, task.completed'
```
GPaoloni, line 15: I don't get the comment in here, but the events filters seems fine.

GPaoloni (approved):
I have not tested it yet, but the idea seems correct to me and the amount of setup this performs is amazing, thanks! Hope we can go in this direction. 
A few things to consider, as already mentioned above we'll need custom configs for chatbots and studio flows, but also for task router filters (there are some helplines that split the chat from the calls etc). Having this split into modules where each helpline has it's own configuration seems mostly appropriate to me, specially if we move this folder to the infra repo.

twilio-iac/aselo-terraform/flex.tf
stephenhand: It could be, but the account creation doc tells you to create a new one, so I thought best to replicate that approach for a first attempt, but nothing to stop us refining it. 

All we need to do is import the default flow into the terraform state in the 'import defaults' script and terraform will update that rather than creating another

twilio-iac/aselo-terraform/chatbots.tf
stephenhand: This is based on what I saw in Aselo Beta - post survey existed but just appeared to be an empty bot?

scripts/src/terraformSupplemental/importDefaultTwilioResourcesToTerraform.ts
```diff
@@ -0,0 +1,130 @@
...
+import { config } from 'dotenv';
+import { logError, logSuccess, logWarning } from '../helpers/log';
+
+config();
+
+const TERRAFORM_WORKING_DIRECTORY = '../twilio-iac/aselo-terraform';
+
+logSuccess(fs.realpathSync(TERRAFORM_WORKING_DIRECTORY));
+logSuccess(process.env.TWILIO_ACCOUNT_SID);
+logSuccess(process.env.TWILIO_AUTH_TOKEN);
```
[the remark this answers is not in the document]
stephenhand: All these initial logs can go, were mainly for debugging the script

twilio-iac/aselo-terraform/aarambh-production.tfvars
stephenhand: We would lose a lot moving this into a separate repo going forward. 

At the moment we can easily build a pipeline where we can make changes to the Twilio provisioning and the plugins in the same PR, then terraform apply > deploy plugin. In separate repos it makes coordinating provisioning & deployment a lot more challenging. 

I was actually going to suggest we consider consolidating the serverless, plugin and webchat repos into one so the entire 'communications domain' can be managed under a unified CI / CD process, making rolling out breaking changes between these systems much cheaper - but that's a bigger conversation.

We can move more stuff out of the checked in tfvars files, or move the files wholesale into a secure store (s3, or even a private git repo just for these files), I would prefer either option to moving the whole terraform config into a private repo but obvs that's something we need to decide between us.

The specific issue of the serverless domain should be addressed if Twilio implement https://github.com/twilio/terraform-provider-twilio/issues/85 - having it as a variable if a workaround anyway. Until then it could just be another environment variable you need to have configured before terraforming

twilio-iac/aselo-terraform/README.md
```diff
@@ -0,0 +1,89 @@
+# Twilio Terraform Scripts
+
+These are scripts for provisioning some of the Aselo Twilio infrastructure.
+
+## Prerequisites
+
+You will require the following installed locally:
+
+* Terraform - https://www.terraform.io/downloads
+* Go - https://www.terraform.io/downloads. This is required to build the Twilio Terraform Plugin. You require version 17+ to build the plugin. Ubuntu users, do not use the version on the Debian package manager, it is too old to build the plugin, follow the instructions on the link instead
```
GPaoloni, line 10: Wrong link in here perhaps?

PR #574, opened 2021-12-17
[WIP] Flex UI v2

GPaoloni:
## Description
This PR is an attempt to upgrade Flex UI to v2 (alpha state rn).
It includes many changes, being the biggest ones:
- Many updated dependencies (read Flex v2 docs for more information).
- `Theme` now does not have a `colors` property, so all of our styles are reworked to consume colors from a hardcoded object (listing all the styles as they are in our current working version of the UI).
- Changes in several places to make TS compile just fine.

### Checklist
- [ :heavy_check_mark: ] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-839)
- [ N/A] New tests added
- [ N/A] Strings are localized

### Verification steps
To run this:
- Make sure you have Twilio cli installed, and @twilio-labs/plugin-flex at least at v4.1.3 (mine is 4.2.2).
- Delete `node_modules` folder.
- Run `npm ci --legacy-peer-deps`. [The reason](https://github.com/linsight/react-keyboard-event-handler/issues/29) for this flag being needed.
- Run `npm start` (or `npm run dev`).

## Potential Refactors
- `getConfig` is not safe to use, we should move to better ways of getting this info: Redux selectors, custom hooks.
- `getConfig` is used for the new form definitions when a contact is initialized. Now we can use Redux thunks. Any operation in the store that requires another part of the state should be moved to be a thunk for sure.
- As mentioned above `Theme` have changed, this means the colors are now hardcoded, and we've lost some styles when using Flex UI components. We should decide if we want to move and start using [Paste](https://paste.twilio.design/) or an alternative for our styles. https://paste.twilio.design/customization/customization-provider/
- Many tests are failing. This will take some amount of effort to decide if we a) move all tests to RTL or b) find a way to make fix them and keep as they are.

## Bugs / Feedback
- UI bugs
  - Availability status dropdown is horizontal.
  - "Send message" button for chats is weird.
- Canned responses not working fine (fails with a not very descriptive error).
- The migration guide works fine but encountered many issues with the flex plugin for twilio-cli. They should have a requirements section of some kind.
- Still no way to update the `strings` context? Can't we consume the strings with a HoC context consumer like `withStrings`?
- Can't we "wrap the entire Flex in a custom component of our own" yet?
- Using `Flex.Manager.getInstance()` leads to a Redux error sometimes. Not all the things we want from here are in Redux. Can they be consumed in some other way (like a `Context`)? Should we drop them in a reducer after initialization?
- Do they know any alternatives to re-writing all the tests that depends on enzyme adapter 16?


GPaoloni:
@nickhurlburt @murilovmachado @stephenhand You can run Flex 2.0 with this PR. Lemme know if it works for you.

nickhurlburt:
@GPaoloni I tried to run this using `npm run dev` and I'm getting `Module not found: Can't resolve 'slate-history' in '/Users/nick/code/flex-plugins/plugin-hrm-form/node_modules/@udecode/plate-core/dist'`. I've tried using `npm install --legacy-peer-deps`, `npm ci --legacy-peer-deps` and `UNBUNDLED_REACT=true npm ci --legacy-peer-deps`. I get that error in every case. I have `@twilio-labs/plugin-flex` `v4.1.3`.

Any ideas?

GPaoloni:
@nickhurlburt Can you try deleting `node_modules` and then doing `npm ci` with the `package-lock` as it is in this branch? That worked for me. It didn't when doing `npm install` tho so there might some dependency update that's not working as expected. Also it might be worth trying with `@twilio-labs/plugin-flex 4.2.2`.

PR #575, opened 2021-12-20
CHI-959: Terraform Importer

stephenhand:
Should probably look to merge this after https://github.com/techmatters/flex-plugins/pull/615 - I think the TF importer will prove to be a key tool for aligning terraform with account configurations

## Description

Importer script (currently only a very limited POC) that could scrape Terraform HCL files to automatically import existing Twilio resources from established environments into the terraform state for managing going forward

### Checklist
- [ N/A ] New tests added

### Verification steps

Use on existing twilio account not currently managed in terraform. It shouldn't pose any risk to the twilio environment but use a staging account to be safe.


scripts/src/terraformSupplemental/importDefaultTwilioResourcesToTerraform.ts
```diff
@@ -1,82 +1,74 @@
...
-      `${flexFlow.sid}`,
-      `module.flex.twilio_flex_flex_flows_v1.${terraformResourceName}`,
-      `${flexFlowFriendlyName} flex flow`,
-    );
-  } else {
-    logWarning(`${flexFlowFriendlyName} flex flow not found to import`);
+  if (dryRun) {
+    logInfo('================== DRY RUN ==================');
+    logInfo('No imports will be performed, command outputs are what would have been run.');
   }
```
GPaoloni, line 58: Nice!

scripts/src/terraformSupplemental/importHCLTwilioResourcesToTerraform.ts
```diff
@@ -0,0 +1,139 @@
...
+  const withoutKey = `${resourceType}.${resourceName}`;
+
+  switch (typeof resourceKey) {
+    case 'number':
+      return `${withoutKey}[${resourceKey}]`;
+    case 'string':
+      return `${withoutKey}["${resourceKey}"]`;
+    default:
+      return withoutKey;
+  }
```
GPaoloni, line 44: Just out of curiosity, what's this depending on?

scripts/src/terraformSupplemental/twilioResources.ts
```diff
@@ -0,0 +1,129 @@
...
+      alias: 'dryRun',
+      global: true,
+      describe:
+        'Do a dry run, it will log the terraform import commands it would have otherwise run to stdout instead for you to review / copy & run manually',
+    })
+    .option('v', {
+      type: 'string',
+      alias: 'varFile',
+      global: true,
+      describe: 'Specify a tfvars file relative to the account directory',
```
GPaoloni, line 23: This file should be set prior to running the script right? And it should like the ones we set for from-scratch-helplines provided by Terraform?

scripts/src/terraformSupplemental/twilioResources.ts
```diff
@@ -0,0 +1,129 @@
...
+        argv.positional('tfFilePath', {
+          describe:
+            'The path of the *.tf file with defines the resources you want to import, relative to the twilio-iac directory',
+          type: 'string',
+        });
+        argv.option('t', {
+          type: 'array',
+          alias: 'type',
+          describe:
+            "Specify a resource type to scan for by its Terraform name, e.g. 'twilio_autopilot_assistants_field_types_v1'. Can be specified multiple times. Omitting this will scan for all supported resource types.",
```
GPaoloni, line 43: This is amazing

scripts/src/terraformSupplemental/twilioResources.ts
```diff
@@ -0,0 +1,129 @@
...
+        argv.option('v', {
+          type: 'string',
+          alias: 'varFile',
+          describe: 'Specify a tfvars file relative to the account directory',
+        });
+        argv.option('s', {
+          type: 'array',
+          alias: 'sid',
+          describe:
+            'Specify an sid referenced in the tf file which would otherwise come from an external source, e.g. --sid var.sid=1234 or --sid twilio_taskrouter_workspaces_v1.from_other_tf_file.sid=4321',
```
GPaoloni, line 60: This one is to override stuff from the imports and use an external one? When does makes sense to use this?

scripts/src/terraformSupplemental/twilioResources.ts
```diff
@@ -0,0 +1,129 @@
...
+        });
+        argv.option('m', {
+          type: 'string',
+          alias: 'modulePath',
+          describe: 'Specify a dot separated path for the module, e.g. top_module1.sub_module2',
```
GPaoloni, line 65: Same question as above.

scripts/src/terraformSupplemental/twilioResources.ts
```diff
@@ -0,0 +1,129 @@
...
+            tfvarsFile: argv.varFile as string,
+            dryRun: argv.dryRun as boolean,
+            modulePath,
+            sids,
+          });
+        }
+      },
+    )
+    .command(
+      'import-account-defaults <accountDirectory>',
```
GPaoloni, line 105: This command loads the defaults only, while the previous command (`'import-tf <accountDirectory>  <tfFilePath>'`) imports all the supported resources, right?

GPaoloni (approved):
Looks really good to me! (tho I'm not sure I got 100% of all that's going on).
Some thoughts:
- A small README might be really helpful, as I don't really know what should I start filling the vars with.
- I really hope we have all accounts in Terraform before I ever need to edit this hahaha (just kidding).

Also, what does HCL stands for?

stephenhand:
> Looks really good to me! (tho I'm not sure I got 100% of all that's going on). Some thoughts:
> 
> * A small README might be really helpful, as I don't really know what should I start filling the vars with.
> * I really hope we have all accounts in Terraform before I ever need to edit this hahaha (just kidding).
> 
> Also, what does HCL stands for?

Thanks as always for your thorough review

I was hoping the --help would be enough but probably something in the scripts readme for context is needed as well.

HCL I think just stands for HashiCorp Language. It's the JSONish syntax used in Terraform configurations

scripts/src/terraformSupplemental/twilioResources.ts
stephenhand: So `import-account-defaults ...` just takes a hardcoded set of resources that normally get created by default when a new Twilio account is created, checks to see if they exist on the target account, and imports them if they are.

It is a required step prior to running `terraform apply` on a brand new account, because otherwise terraform will encounter a load of stuff from its configuration that is not in the tfstate but does exist on the account (because we reuse these default resources in our Aselo configuration), which upsets Terraform. This script will line up the tfstate of a fresh account with the reality, allowing the first `terraform apply` to run correctly.

`twilioResources import-tf ...` does the same process, but instead of a hardcoded set of resources, you point it at a `*.tf` file and it scans the resources in there. It's use case is more around trying to import large amounts of resources that are defined in configuration, already exist on the account, but aren't in tfstate. Importing existing, non-terraform managed managed accounts is one use case, but also importing significant additions made to staging accounts outside of terraform would be another (I used it to import the SafeSpot 'which district?' autopilot question, which would have been time consuming otherwise).

We could, in theory, reimplement `twilioResources import-account-defaults` as a wrapper for `twilioResources import-tf` that points at a special tf file defining all the default resources, since other than the initial input, you are correct, they essentially do the same job.

scripts/src/terraformSupplemental/twilioResources.ts
stephenhand: This is primarily used to specify ids that might come from variables or be defined in another file.

So for example, you could be importing `autopilot-tasks.tf`, and in there is a task resource `my_task`. However, the TF resource ID for a task is `<assistant_sid>/<task_sid>`, so the script needs to grab both sids in order to build up the correct `terraform import` command. If the assistant that owns `my_task` is defined in `autopilot-tasks.tf` that's ok, the script will query the assistant SID automatically. However, if the assistant is defined in `autopilot-assistants.tf`, then the script can't query it because it's not being pointed at a file that defines the resource. To work around this, you would pass in the assistant SID explicitly with a parameter something like `--sid twilio_autopilot_v1.my_assistant.sid=5678`

If the assistant SID was passed in as a variable instead of being in a different file in the same module, you would do something like `--sid var.my_assistant_sid=5678`. Note that you STILL need to pass this, even if `my_assistant_sid` is defined in a tfvars file or as a TF_VARS_* environment variable. The script doesn't scan provided terraform variables automatically for its own use, it just passes the tfvars file you specify down to the `terraform import` command. It could be enhanced to do this if needed, but doesn't right now.

scripts/src/terraformSupplemental/twilioResources.ts
stephenhand: It isn't required, and if it is provided, I think the only think it will use from them is the AWS creds needed to interact with the remote `tfstate` (which can be passed as environment variables anyway), the Twilio creds need to be set via env vars anyway, because that's where the script reads from to use the Twilio client. I might remove the option because it's possibly causing more confusion than being helpful.

FYI in the new multi-directory world, the checked in .tfvars files are gone. The things they used to set are now just default values for each account's `varaibles.tf` file. The only tfvars file you use now is a local only `private.tfvars` that you can use to hold each account's sensitive variable values, and that's optional as you can set them via TF_VARS_* environment variables (but the private.tfvars is easier imho)

scripts/src/terraformSupplemental/importHCLTwilioResourcesToTerraform.ts
stephenhand: The resourceKey is a string if the resource definition has a `for_each` clause with a set of strings (like in /twilio-iac/terraform-modules/chatbots/main.tf), it will be a number if the resource definition has a `for_each` with a set of numbers or a `count` clause (not currently used in Aselo TF config) or undefined if it's just a regular singleton resource without either of these set

scripts/src/terraformSupplemental/twilioResources.ts
stephenhand: This is to give context of where the module is in relation to the route so the script can generate the correct syntax for the `terraform import` command.

If you want to import a Twilio Autopilot Assistant defined in the root module, you just do:
```
terraform import twilio_autopilot_assistant_v1.my_assistant UA4321
```

However, if you wanted to import the same assistant defined in module `sub_module_2`, which is a depended on by `top_,module1` which is in turn depended onm by the configuration's root module, you need to do:
```
terraform import module.top_module1.module.sub_module2.twilio_autopilot_assistant_v1.my_assistant UA4321
```

In order for the script to know it needs to use the above syntax to import the resource to the correct location, you need to call the script with `-m top_module1.sub_module2` - it does not have enough context to work this out itself, since it is only provided with a single tf file. It has to assume the tf file is part of the root module unless instructed otherwise

PR #576, opened 2021-12-20
CHI-958: Terraform chatbot module

stephenhand:
Primary Reviewer @GPaoloni 

## Description

Migrate the terraform chatbot resources into a module, both because it needs doing but also to demo how we could move other resources into modules as required

### Checklist
- [ N/A ] New tests added

### Related Issues
Story - CHI-958

### Verification steps

Set up a new account with the version of the terraform resources prior to these changes
Update to this version
Do a terraform apply, reviewing the plan - it should NOT be deleting & creating chatbot resources if the refactor block is working correctly.


twilio-iac/terraform-modules/chatbots/default/variables.tf
```diff
@@ -0,0 +1 @@
+variable "serverless_url" {}
```
murilovmachado, line 1: Here, we can also describe more the variables, such as setting a `type` and `description`:

```
variable "serverless_url" {
  description = "Serverless URL"
  type        = string
}
````

File: twilio-iac/aselo-terraform/main.tf [hunk 2]
```diff
@@ -14,3 +14,7 @@ terraform {
   }
 }
 
+module "chatbots" {
+  source = "../terraform-modules/chatbots/default"
+  serverless_url = var.serverless_url
+}
```
murilovmachado, line 20: Right now, we're using only `serverless_url` as input. But I can see there's a lot of hardcoded data in the resources created. I believe we will start moving these hardcoded data to be input as well, is that correct?

murilovmachado (approved):
The code looks good to me. I didn't run it to check the behavior, but it seems correct. I have only one question on the current hardcoded data.

twilio-iac/aselo-terraform/main.tf
stephenhand: I was assuming we would move hard coded stuff to variables as and when we needed to set them to different values for environments. Sometimes the actual structure of the resources will change, in which cases we would need a different module completely, so I'm not trying to preempt what will need to be passed in in the future at this point.

GPaoloni (approved):
Looks great to me! Questions around what is possible/best, and a few more:
- [Read review comments first] This `main.tf` is now providing a single example of chatbot configured as our development one I assume. If we want to change it to be the JM one for example, that uses almost identical structure but with more questions on the survey  task, where and how can that be specified? (assuming my below question about the `jsonencode` is not useful).
- If a helpline needs localized chatbots (like ZM), we have two options:
  - Use templated messages that are retrieved from a dedicated Serverless Function (ideal).
  - Use more than one chatbot, one for each language, all identical but the messages are translated accordingly (current).
  In order to keep this current approach till we improve, we'll need to define one presurvey chatbot per-language, and each with it's own JSON spec. That can only be achieved with a customized module, right?
- >Sometimes the actual structure of the resources will change, in which cases we would need a different module completely

  How will this work? Each environment (Twilio account) will have it's own variables file where we specify "grab default module for X but this custom module for Y"? And where will those variables live?
  
Sorry if some of this makes no sense, I'm trying to make my head around what this will look like hehe.

twilio-iac/terraform-modules/chatbots/default/main.tf
```diff
@@ -0,0 +1,413 @@
+terraform {
+  required_providers {
+    twilio = {
+      source  = "twilio/twilio"
+      version = "0.9.2"
+    }
```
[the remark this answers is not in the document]
stephenhand: I thought each module needed to specify it's dependencies separately? Perhaps the version doesn't need to be set / can be set via a variable.

Either way it's probably best to keep versions aligned, even if each module might end up using different providers

stephenhand:
> Looks great to me! Same as Mu, I have not tried running the code, but I would love if you can show me/us how to correctly do this, at least until the `plan` step.
> 
> Questions around what is possible/best, and a few more:
> 
> * [Read review comments first] This `main.tf` is now providing a single example of chatbot configured as our development one I assume. If we want to change it to be the JM one for example, that uses almost identical structure but with more questions on the survey  task, where and how can that be specified? (assuming my below question about the `jsonencode` is not useful).
> * If a helpline needs localized chatbots (like ZM), we have two options:
>   
>   * Use templated messages that are retrieved from a dedicated Serverless Function (ideal).
>   * Use more than one chatbot, one for each language, all identical but the messages are translated accordingly (current).
>     In order to keep this current approach till we improve, we'll need to define one presurvey chatbot per-language, and each with it's own JSON spec. That can only be achieved with a customized module, right?
> * > Sometimes the actual structure of the resources will change, in which cases we would need a different module completely
>   
>   
>   How will this work? Each environment (Twilio account) will have it's own variables file where we specify "grab default module for X but this custom module for Y"? And where will those variables live?
> 
> Sorry if some of this makes no sense, I'm trying to make my head around what this will look like hehe.

So for localisations we could define a single module but pull multiple copies of it into the parent `main.tf`, passing in a switch variable to get it to load up different data depending on the localisation, meaning we could generate multiple chatbots from a single definition. We could load up those resources into the Terraform config from elsewhere in the repo, rather than them having to be embedded in the terraform stuff 

So if we were to continue with this setup, what we would have to do it is set up switches for each environment that uses a custom module (basically use the 'count' attribute, setting the custom module count to '1' and the default module count to '0' for environments that use a custom module, vs the opposite for those using the default

However

After https://github.com/techmatters/flex-plugins/pull/615 is completed & merged every helpline will have their own 'main.tf' as an entry point, and we can simply pull in whatever modules we need for each helpline - it might increase duplication a little but it makes customising helplines simple.

twilio-iac/terraform-modules/chatbots/default/main.tf
```diff
@@ -0,0 +1,413 @@
...
+resource "twilio_autopilot_assistants_v1" "post_survey" {
+  unique_name       = "post_survey_bot"
+  friendly_name     = "A bot that collects a post-survey"
+  log_queries = true
+}
+
+resource "twilio_autopilot_assistants_tasks_v1" "survey" {
+  unique_name   = "survey"
+  assistant_sid = twilio_autopilot_assistants_v1.pre_survey.sid
+  actions       = jsonencode({
```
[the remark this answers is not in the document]
stephenhand: Yes I think we could separate out the language content from the resource structure, so chatbots that only vary in terms of language, phrasing, or answer synonyms don't need to have a whole separate module defined

For helplines that need more customisation, we could also consider composing the chatbots module from further submodules, so they can reuse parts that are common.

I think Terraform gives us most of the tools we need to avoid too much duplication, and we can always investigate what extra https://terragrunt.gruntwork.io/ could give us as suggested by @josebui

PR #580, opened 2021-12-21
CHI-854: Refactored form definitions to load as dynamic modules

stephenhand:
Primary Reviewer: @murilovmachado 

Looks like a monster PR ( 40ish files 😬 ) - but most of those are test fixes, and most of the remainder are changing the form definition identifier from a string to an enumeration type. There is only really 1 file where the substantive changes are - `/plugin-hrm-form/src/formDefinitions/index.ts`

## Description

* Refactored form definitions to load as dynamic modules
* Removed redundant duplicate index.ts files in each set of form definitions.
* Changed the form definition version identifier to be an enum rather than a string for more static checking

### Checklist
- [ X ] Tests fixed / updated

### Verification steps

* Deploy version & verify form pages load correctly


plugin-hrm-form/src/formDefinitions/index.js
```diff
@@ -1,15 +0,0 @@
-import v1 from './v1';
-import zaV1 from './za-v1';
-import etV1 from './et-v1';
-import mwV1 from './mw-v1';
-import brV1 from './br-v1';
-import inV1 from './in-v1';
-
-export default {
-  v1, // Zambia V1
```
nickhurlburt, line 9: We may want to preserve these comments in the new `DefinitionVersionId ` enum? At least the Zambia one.

nickhurlburt (approved):
Someone who's worked more directly with this code should also take a look, but this looks great to me, and removes boilerplate code!

plugin-hrm-form/src/formDefinitions/index.js
stephenhand: We could even give the enum aliases more descriptive names, i.e. `ZambiaV1`, `SaferNetV1` ? Restoring comments for now

plugin-hrm-form/src/services/ServerlessService.ts
```diff
@@ -105,9 +105,10 @@ export const listWorkerQueues = async (body: {
  * Function that mimics the fetching of a version definition for all the forms used within the app.
  * Later on this will be fetched in async way.
  */
-export const getDefinitionVersion = async (version: string): Promise<DefinitionVersion> => definitionVersions[version];
+export const getDefinitionVersion = async (version: DefinitionVersionId): Promise<DefinitionVersion> =>
```
GPaoloni, line 108: Maybe this is not the correct place for this anymore, since we are (likely) not moving the definitions to Serverless. Not needed to move it now tho.

plugin-hrm-form/src/formDefinitions/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    /* webpackMode: "eager" */ `./${version}/insights/oneToOneConfigSpec.json`
+  );
+  const oneToManyConfigSpecsModule = await import(
+    /* webpackMode: "eager" */ `./${version}/insights/oneToManyConfigSpecs.json`
+  );
+  const caseStatusModule = await import(/* webpackMode: "eager" */ `./${version}/CaseStatus.json`);
+
+  const { helplines } = helplineInformationModule.default;
+  const defaultHelpline =
+    helplineInformationModule.default.helplines.find(helpline => helplines).value || helplines[0].value;
```
GPaoloni, line 59: What's the purpose of `.find(helpline => helplines)` here? _I think_ this will always result in `true` cause `helplines` is a truthy object reference. Also, shouldn't `helpline.default` property be used here rather than `helpline.value`?

plugin-hrm-form/src/formDefinitions/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  brV1 = 'br-v1', // Safernet Brasil v1
+  etV1 = 'et-v1', // Ethiopia v1
+  inV1 = 'in-v1', // Aarambh Trustline v1
+  mwV1 = 'mw-v1', // Malawi v1
+  zaV1 = 'za-v1', // South Africa v1
+}
+
+export async function loadDefinition(version: DefinitionVersionId) {
+  /*
+   * Unfortuntately I could not get lazy loading to work, which would have been nice because only the form definitions used by that helpline would be loeded.
```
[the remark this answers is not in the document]
GPaoloni: Also, I'm fine with this being eager, the idea is mostly wrap the definition load in an async function. But I could make this work like
`const layoutDefinitionsModule = require(`./${version}/LayoutDefinitions.json`);` (and analogous for the rest of the modules). Not sure if there's any difference tho. If you try that be careful to remove the `.default` part from each ~imported~ required module.

GPaoloni (approved):
Looks great! I'm approving cause I consider this to be mostly in place, but please take a look at my question before merging.

plugin-hrm-form/src/formDefinitions/index.ts
stephenhand: The find function is wrong - should be `helpline => helpline.default` - thanks Gian! The original implementations are using helpline.value though.

GPaoloni: Yeah I noticed, thanks for the fix!

plugin-hrm-form/src/formDefinitions/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  const callTypeButtonsModule = await import(/* webpackMode: "eager" */ `./${version}/CallTypeButtons.json`);
+  const helplineInformationModule = await import(/* webpackMode: "eager" */ `./${version}/HelplineInformation.json`);
+  const cannedResponsesModule = await import(/* webpackMode: "eager" */ `./${version}/CannedResponses.json`);
+  const oneToOneConfigSpecModule = await import(
+    /* webpackMode: "eager" */ `./${version}/insights/oneToOneConfigSpec.json`
+  );
+  const oneToManyConfigSpecsModule = await import(
+    /* webpackMode: "eager" */ `./${version}/insights/oneToManyConfigSpecs.json`
+  );
+  const caseStatusModule = await import(/* webpackMode: "eager" */ `./${version}/CaseStatus.json`);
```
murilovmachado, line 55: One thing that will be a concern in the future (when these JSONs are fetched from somewhere else) is this sequence of 17 awaits. We should find a way to **await Promise.all([...])**.

Just using **Promise.all** directly, in this case, might be bad for readability because we would have to match the ordering of parameters and the const names in the returned array:
``` 
const [moduleA, moduleB, moduleC, moduleX] = await Promise.all(['pathA', 'pathB', 'pathC', 'pathX']);
```

I would suggest extracting a method that could be used like this:
```
const modules = await fetchAllModules({
    moduleA: 'pathA',
    moduleB: 'pathB',
    moduleC: 'pathC',
    moduleX: 'pathX',
  });
```

And the returned object would be:
```
{
    moduleA: {/* loaded JSON from pathA */},
    moduleB: {/* loaded JSON from pathB */},
    moduleC: {/* loaded JSON from pathC */},
    moduleX: {/* loaded JSON from pathD */},
  }
```

This method would **Promise.all** under the hood, but its interface makes it much more readable for our scenario.

Does it make sense? It doesn't need to be done in this PR, but I think it's something that we will have to deal with as soon as we start fetching these JSONs from somewhere else.

murilovmachado (approved):
Looks good to me. I've also tested it locally and it worked fine. Just have one comment on the way we currently load the JSON files.

plugin-hrm-form/src/formDefinitions/index.ts
murilovmachado: wow, this was confusing at first because there are **two different** `default` properties: one from the module and another one from the JSON itself.😂

plugin-hrm-form/src/formDefinitions/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const { helplines } = helplineInformationModule.default;
+  const defaultHelpline =
+    helplineInformationModule.default.helplines.find(helpline => helpline.default).value || helplines[0].value;
```
murilovmachado, line 59: line **58-59** could just use `helplines` defined on **57**:
```
 const { helplines } = helplineInformationModule.default;
 const defaultHelpline = helplines.find(helpline => helpline.default).value || helplines[0].value;
```

plugin-hrm-form/src/formDefinitions/index.ts
stephenhand: Makes sense and definitely worth doing if we moved to 'real' async loading (eagerly loading a dynamic module means the json is preloaded at startup anyway).

The reason I backed away from trying to do anything fancy here is that the webpack dynamic module loader seems to be veeerry fussy about how you dynamically load modules, so I wanted to keep the loading code as simple as possible to keep from tripping it up. For example, when I changed the path where the definitions are loaded from in a following PR, I tried to put the root part of the path into a constant. This confused webpack and prevented the modules from loading at runtime, so I had to revert it back to repeated hardcoded root paths directly in the `module(...)` call.

If we mode to pulling the json from a remote location, we will be free of those limitations and can load them concurrently along the lines you suggest. I'm sure we could do something that would be both concurrent & webpack friendly now but to me it seems better to wait until it's necessary & more straightforward.

PR #585, opened 2021-12-28
Aarambh updates 12282021

dee-luo:
## Description
A number of new Aarambh customization changes. Most notable thing to check - they wanted to replace "Counselor" with "Case Worker" across the platform. I've done this by creating a new translation folder "en-IN" and the flexUI.json only contains strings for "Counselor".


GPaoloni (approved):
The changes looks good, and the trick for the translation is great! How will you refer to it, via service config?

PS: I can improve the age select converting into a "range", so we will only need to provide 2 numbers, `start` and `end` instead than 99. Let me know if you think this is worth.

GPaoloni:
@dee-luo Broadcasting the PR to the rest of the team so they are also aware.

stephenhand (approved):
Approving because it gets the immediate requirement sorted, but I'm a little nervous about blending UI customisations & translations going forward. One thing is that 'en-IN' isn't a valid ISO localisation which might trip up 3rd party localisation related software we might use in the future. More importantly, it isn't obvious what the purpose of the 'en-IN' localisation is. A new engineer tasked with setting up Childline India, for instance, might reasonably expect that this is the correct localisation to use for them, which would be wrong. It adds to the non-intuitive 'tribal knowledge' required to maintain the system.

More generally, at the moment I feel we're taking a piecemeal approach to UI customisation, addressing each requirement with a specific fix as it comes in. I think we need to try and anticipate which broad categories of customisation requests we expect to see from helplines and come up with more systematic way to cater for them.

GPaoloni:
@stephenhand While you are right and I adhere to your concern on "blending UI customisations & translations", I think this is a completely valid usage of this "strings", since they are more of a [templating system than a strictly translation one](https://www.twilio.com/docs/flex/developer/ui/localization-and-templating), that we abuse to provide an easy way to translate our app. However, is true that the name is not specific enough to clarify that this are very particular requirements that only apply to Aarambh helpline, so maybe refining the name can prevent confusion down the road. Seems like a topic to raise as tech discussion with the rest of the team!

dee-luo:
@GPaoloni Yes, if we could set the service config for Aarambh to use these translations that would be great! And I love the idea for the age range selector (makes these files less cumbersome!), although we want to make sure we are still able to include other options other than the range (e.g. Unborn, Other, ">max", etc.). If this is a quick change, would be great to include as part of our form definition improvements.

stephenhand:
> @stephenhand While you are right and I adhere to your concern on "blending UI customisations & translations", I think this is a completely valid usage of this "strings", since they are more of a [templating system than a strictly translation one](https://www.twilio.com/docs/flex/developer/ui/localization-and-templating), that we abuse to provide an easy way to translate our app. However, is true that the name is not specific enough to clarify that this are very particular requirements that only apply to Aarambh helpline, so maybe refining the name can prevent confusion down the road. Seems like a topic to raise as tech discussion with the rest of the team!

It's not the mechanism being used (which looks like the only way to do it really), as you say, it's more about how we organise things - it's more than just naming though. Designing our system such that we can keep all the per helpline customisations we do (whether they are cosmetic or functional changes) together and separate from the core application will set us up to scale much better than if they need to be implemented in lots of different places. Definitely something I'd like for us to explore more in the New Year.

GPaoloni:
> It's not the mechanism being used (which looks like the only way to do it really), as you say, it's more about how we organise things - it's more than just naming though. Designing our system such that we can keep all the per helpline customisations we do (whether they are cosmetic or functional changes) together and separate from the core application will set us up to scale much better than if they need to be implemented in lots of different places. Definitely something I'd like for us to explore more in the New Year.

Totally agreed with this! Now that you mention it by the way, as you've been working with the form definitions recently, you surely noticed that there are UI configurations bundled alongside system configuration stuff. That's probably another piece that we want to tackle in order to do better in separation of concerns. Totally agreed with all you said in this thread!

GPaoloni:
As discussed with @dee-luo, while this PR contains some changes that should trigger a new definition version (`value` property is changed for some options in dropdowns), we are updating this one because this is not used in production yet, only in staging, so is fine if the "data already recorded in the DB" is not 100% accurate. For subsequent changes, if the changes on a definition that is being used in production "trigger a new version", we won't be able to do this or we'll end up having bugs in the UI.

PR #587, opened 2021-12-30
CHI-853: Form Definitions Module

stephenhand:
## Description

Moves form definiitons types and related code into separate module. Adds script to generate an empty set of forms

This PR does not refactor the plugins to use the new module, it just creates it and adds the creation script. I was thinking refactoring the plugin could be a follow on, to keep the PR sizes down

### Checklist
- [ X ] New tests added
- [ N / A ] Strings are localized

### Verification steps

Run tests, general regression


scripts/package.json
```diff
@@ -22,7 +24,8 @@
     "prompt": "^1.2.0",
     "prompt-confirm": "^2.0.4",
     "twilio": "^3.66.0",
-    "yargs": "^15.4.1"
+    "yargs": "^15.4.1",
+    "hrm-form-definitions": "file:../hrm-form-definitions"
```
nickhurlburt: I see... so this does not require pushing new NPM package versions to a remote repo like https://www.npmjs.com/ ? This means there is less friction than I thought. That is good news for being able to further modularize our code.

I wonder if we should employ this same strategy for our helper package in serverless? @GPaoloni

hrm-form-definitions/form-definitions/br-v1/CallTypeButtons.json
```diff
@@ -0,0 +1,38 @@
+[
```
nickhurlburt, line 1: Would it be a good idea to `git mv` these form definitions? That would maintain the file history, make the PR clearer and also ensure we don't have duplicated files. On the other hand, we would have to also change the plugin to read from here at the same time, increasing the PR size.

nickhurlburt (commented):
Good to see that this is easier than I thought. I definitely like the idea of creating more base modules that can hold shared code. My mental model of managing internal modules/libraries comes from Java and C, and not from JS/TS, so I'd be interested in understanding how they're different.

hrm-form-definitions/form-definitions/br-v1/CallTypeButtons.json
stephenhand: When the time comes to remove the old ones, can I delete the new copies, then `git mv` the originals? Or would the previous existence of the copies mess that up?

As for duplication, I don't intend to keep the 2 sets around long, as soon as this PR gets merged I will get the follow up refactoring out the ones in the plugin source tree raised

nickhurlburt: My guess is that if new copies are deleted first, then `git mv` would work fine. I don't *think* previous existence of files with the same name matters. That would probably require that the deleting and moving are separate, unsquashed commits.

I'm fine with the fast-follow approach regarding duplication. It just requires sufficient communication with everyone who edits form definitions so they know where to make edits and what will happen to them.

hrm-form-definitions/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
```
GPaoloni, line 1: I'm fine with whatever style we use, but I just wanna point that this are not consistent with the plugin-hrm-form ones, and same goes for eslint. Maybe there are reasons for this.

stephenhand: I think it's just the printWidth that's different here? The 120 limit was making some of the code in this repo ugly AF and I don't think limiting lines to be that narrow is really necessary - but I can change it back for consistency's sake.

The eslint differences probably arise from the fact that I based it off the scripts module rather than the plugins. That seemed like the better one to base it off since neither of them have any react stuff in them. Do we need to align all 3 modules more?

murilovmachado (approved):
Looks good to me. I've mostly glanced over this PR, but I get the idea that there are two main things here:
1. Move all **form-definitions** related files into their own module
2. A script to generate a new set of form-definitions (eg.: "ie-v1", for Ireland)

I got that you didn't **git mv** files because you didn't delete the old definitions yet. My question is: couldn't they be deleted in this PR? or there's something left to do in order to safely delete them from `plugin-hrm-form`?

stephenhand:
> Looks good to me. I've mostly glanced over this PR, but I get the idea that there are two main things here:
> 
> 1. Move all **form-definitions** related files into their own module
> 2. A script to generate a new set of form-definitions (eg.: "ie-v1", for Ireland)
> 
> I got that you didn't **git mv** files because you didn't delete the old definitions yet. My question is: couldn't they be deleted in this PR? or there's something left to do in order to safely delete them from `plugin-hrm-form`?

The work to refactor `plugin-form-hrm`  to use the new module is in https://github.com/techmatters/flex-plugins/pull/589/files - as you can see there's a fair bit of tweaking beyond just moving the files, nothing complicated but a lot. I thought splitting the PRs at that point would be good to keep the PR sizes getting out of control. 

I deleted the new copies, then moved the originals in that PR, so hopefully that will wind up the same as if I'd just `git mv`ed them in the first place (see convo with nick above)

stephenhand:
@murilovmachado @GPaoloni @nickhurlburt  - This PR should be cleaner now the previous one has been merged

hrm-form-definitions/src/specification/generate.ts
```diff
@@ -0,0 +1,37 @@
...
+    )
+    .sort(
+      ([, spec1], [, spec2]) =>
+        (spec1.order ?? Number.MAX_VALUE) - (spec2.order ?? Number.MAX_VALUE),
+    )
+    .map(([name, itemDefinitionSpecification]) =>
+      generateDefaultItem(<DefinitionSpecification>itemDefinitionSpecification, {
+        type: 'input',
+        name,
+        label: name,
```
GPaoloni: Question: isn't this losing the upper case first letter for labels?

hrm-form-definitions/src/specification/generate.ts
```diff
@@ -0,0 +1,37 @@
...
+  fallback: T,
+): T {
+  if (specification.default) {
+    return specification.default;
+  }
+  console.warn('No default was specified, using fallback');
+  return fallback;
+}
+
+export function generateDefaultForm(specification: FormDefinitionSpecification): FormDefinition {
```
GPaoloni: This is great! All the types and helpers setup is totally worth it, super easy to read. Is it more readable if the callbacks are factored out (the ones passed to `.filter`, `.sort` and `.map`)? Or that ends up being more complicated?

hrm-form-definitions/src/specification/specification.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export type FormItemDefinitionSpecification = DefinitionSpecification<
+  FormItemDefinition,
+  FormAndItemDefinition
+> & { order?: number };
+
+export type FormDefinitionSpecification = {
+  items: Record<string, FormItemDefinitionSpecification>;
+  // If validation for the entire form is required, not tied to individual elements
+  validator?: (form: FormDefinition) => void;
```
GPaoloni: Nice! Will this override if the item has it's own `validator` instead?

scripts/src/formDefinitions/generateNewHelplineFormDefinitions.ts
```diff
@@ -0,0 +1,91 @@
...
+    .option('r', {
+      alias: 'root',
+      describe:
+        "Root folder to place the definitions in. Will default to '../hrm-form-definitions/form-definitions', ready for use in the app if run using the npm script",
+    })
+    .option('f', {
+      alias: 'force',
+      describe:
+        'Auto confirm any prompts. WARNING: This will mean existing form definitions is the target directory will be overwritten without warning',
+    })
```
GPaoloni: D: Don't feel like this is needed while it's dangerous.

twilio-iac/aselo-terraform/aselo-beta.tfvars
```diff
@@ -0,0 +1,6 @@
+helpline    = "Aarambh"
```
GPaoloni: I'm sure this is not correct, was this unintentional bundled?

GPaoloni (commented):
Really nice to see this, thanks! Mostly questions around, maybe they are answered in the follow up PR.

hrm-form-definitions/reackHookFormTypes.ts
```diff
@@ -0,0 +1,23 @@
+
```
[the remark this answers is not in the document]
stephenhand: I'll try that, so adding stuff as a dev dependency will use the types to compile but not bring in the code? That's really useful to know..

hrm-form-definitions/src/specification/generate.ts
stephenhand: I find those calls harder to read when the functions aren't inline speaking for myself

hrm-form-definitions/src/specification/specification.ts
stephenhand: This is supposed to be in addition to the per item validators, if you need to do something like validate the order, or check that one element must be present if another one is etc. Basically any check that goes beyond the scope of a single item.

scripts/src/formDefinitions/generateNewHelplineFormDefinitions.ts
stephenhand: Was just thinking in case we need to run it in an automated script or something, but that's probably not going to be the case

twilio-iac/aselo-terraform/aselo-beta.tfvars
stephenhand: Correct, it's a WIP intended for another PR down the line, will remove.

hrm-form-definitions/src/specification/generate.ts
stephenhand: It isn't supposed to be something that gets used as is, just a placeholder to get you started. With correct configuration you shouldn't see this, anything marked as required should be given a default really

hrm-form-definitions/.gitignore
```diff
@@ -0,0 +1 @@
+dist/
```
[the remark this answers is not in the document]
stephenhand: Addressed in https://github.com/techmatters/flex-plugins/pull/595

twilio-iac/aselo-terraform/aselo-beta.tfvars
stephenhand: Addressed in https://github.com/techmatters/flex-plugins/pull/595

hrm-form-definitions/reackHookFormTypes.ts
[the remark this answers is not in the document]
stephenhand: That worked great! Addressed in https://github.com/techmatters/flex-plugins/pull/595

PR #591, opened 2022-01-04
Fix es-IN translations for Aarambh

GPaoloni:
## Description
This PR adds the missing `en-IN` translation in our map of bundled translations (missed this step before).

### Verification steps
- Edit the `helplineLanguage` to be 'en-IN' in `HrmFormPlugin.js` file (line 62).
- Start local server with `npm run dev`.
- Check that the intended translations are displayed in the UI.

Question for @dee-luo: I see that the strings in `messages.json` are still using the counsellor word. Should this be updated? If so, feel free to do in this very same PR.


dee-luo:
@GPaoloni We don't have to worry about the messages now as they are just doing data entry. Thanks for catching!

nickhurlburt (approved):
Looks good, thanks!

PR #592, opened 2022-01-04
Aarambh language template update

dee-luo:
Missed one instance of Counsellor in my last PR - sorry!

### DIFF
--- a/plugin-hrm-form/src/translations/en-IN/flexUI.json
+++ b/plugin-hrm-form/src/translations/en-IN/flexUI.json
@@ -1,8 +1,9 @@
 {
   "Case-Counsellor": "Case Worker",
-  "Case-ActionHeaderCounselor": "Case Worker",
+  "Case-ActionHeaderCounselor": "Case Worker: ",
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/translations/en-IN/flexUI.json
```diff
@@ -1,8 +1,9 @@
 {
   "Case-Counsellor": "Case Worker",
-  "Case-ActionHeaderCounselor": "Case Worker",
+  "Case-ActionHeaderCounselor": "Case Worker: ",
```
GPaoloni, line 3: Is the colon intended here?

GPaoloni (approved):
Looks great, just a question.

plugin-hrm-form/src/translations/en-IN/flexUI.json
dee-luo: @GPaoloni yes - this is for case section (incident, household member, etc.) details at the top where we show who added the entry. In the en-US translation file, we have:  "Case-ActionHeaderCounselor": "Counsellor:"

PR #594, opened 2022-01-05
Auto retry saving pending contacts

murilovmachado:
Primary reviewer: @GPaoloni 
Jira: https://bugs.benetech.org/browse/CHI-880

## Description
This PR implements auto-retry for the pending contacts:
- The auto-retry is triggered when any contact succeeds to be saved on the external backend.
- When a pending contact is successfully saved on the external backend, it's removed from the pending contacts list.
- When a pending contact fails to be saved on the external backend, its 'retries' attribute is incremented.
- `pending-contacts-lock` is used to ~prevent~ minimize concurrency issues

### Checklist
- [ ] New tests added [WIP]

### Verification steps
Testing this PR is not trivial because it involves a lot of setups. I'm writing unit tests to make sure `autoRetrySavingPendingContacts()` works as intended.

### DIFF
--- a/plugin-hrm-form/src/dualWrite/index.ts
+++ b/plugin-hrm-form/src/dualWrite/index.ts
@@ -4,7 +4,7 @@ import { getConfig } from '../HrmFormPlugin';
 import { savePendingContactToSharedState, autoRetrySavingPendingContacts } from '../utils/sharedState';
 import saveContactToSaferNet from './br';
 
-type DualWriteFn = (task: ITask, payload: any) => Promise<void>;
+type DualWriteFn = (task: ITask, payload: any, isAutoRetry?: boolean) => Promise<void>;
 
 type SaveContactByDefinitionVersion = {
   [definitionVersion: string]: DualWriteFn;
--- a/plugin-hrm-form/src/dualWrite/index.ts
+++ b/plugin-hrm-form/src/dualWrite/index.ts
@@ -1,7 +1,7 @@
 import { ITask } from '@twilio/flex-ui';
 
 import { getConfig } from '../HrmFormPlugin';
-import { savePendingContactToSharedState } from '../utils/sharedState';
+import { savePendingContactToSharedState, autoRetrySavingPendingContacts } from '../utils/sharedState';
 import saveContactToSaferNet from './br';
 
 type DualWriteFn = (task: ITask, payload: any) => Promise<void>;
@@ -22,6 +22,7 @@ export const saveContactToExternalBackend = async (task: ITask, payload: any) =>
   if (saveContact) {
     try {
       await saveContact(task, payload);
+      autoRetrySavingPendingContacts(saveContact);
     } catch (err) {
       savePendingContactToSharedState(task, payload, err);
     }
--- a/plugin-hrm-form/src/utils/sharedState.js
+++ b/plugin-hrm-form/src/utils/sharedState.js
@@ -1,5 +1,4 @@
 import { getConfig } from '../HrmFormPlugin';
-import { saveContactToExternalBackend } from '../dualWrite';
 import { recordBackendError } from '../fullStory';
 
 const isSharedStateClientConnected = sharedStateClient =>
@@ -145,7 +144,7 @@ export const savePendingContactToSharedState = async (task, payload, error) => {
 
 /**
  * Loops through the pending contacts and try saving them to the external backend.
- * When a pending contact is successfully saved to the external backend, it's removed from the pending contacts list.
+ * When a pending contact is successfully saved on the external backend, it's removed from the pending contacts list.
  * When a pending contact fails to be saved on the external backend, its 'retries' attribute is incremented.
  * @param {*} saveContactFn Function that saves a contact to the external backend
  */
--- a/plugin-hrm-form/src/utils/sharedState.js
+++ b/plugin-hrm-form/src/utils/sharedState.js
@@ -180,7 +180,7 @@ export const autoRetrySavingPendingContacts = async saveContactFn => {
       const { task, payload } = value;
 
       try {
-        await saveContactFn(task, payload);
+        await saveContactFn(task, payload, true);
         successfulRetriesIndexes.push(index);
       } catch (e) {
         failedRetriesIndexes.push(index);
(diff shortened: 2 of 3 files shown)


plugin-hrm-form/src/utils/sharedState.js
```diff
@@ -142,3 +141,64 @@ export const savePendingContactToSharedState = async (task, payload, error) => {
...
+
+  try {
+    const list = await sharedStateClient.list('pending-contacts');
+    const successfulRetriesIndexes = [];
+    const failedRetriesIndexes = [];
+
+    const pendingContacts = (await list.getItems()).items;
+
+    if (pendingContacts.length === 0) {
+      await pendingContactsLock.set({ isLocked: false });
```
stephenhand, line 174: Redundant, `finally` block already handles this

plugin-hrm-form/src/utils/sharedState.js
```diff
@@ -142,3 +141,64 @@ export const savePendingContactToSharedState = async (task, payload, error) => {
...
+  const { sharedStateClient, strings } = getConfig();
+
+  if (!isSharedStateClientConnected(sharedStateClient)) {
+    console.error('Error with Sync Client conection. Sync Client object is: ', sharedStateClient);
+    console.error(strings.SharedStateSaveContactError);
+    return;
+  }
+
+  // Lock to prevent concurrency isues
+  const pendingContactsLock = await sharedStateClient.document('pending-contacts-lock');
```
stephenhand, line 161: Is the write operation to the SaferNet backend not idempotent? Would save some headaches if it was!

As it is I'm not sure if the lock is helping. Since the read & write of the lock aren't atomic, it doesn't guarantee the state won't be accessed concurrently - 2 clients could read the same version of the lock before it can be updated just like they could read 2 versions of the pending contacts list before it is updated. 

I suppose there are  2 read hurdles that 2 browser sessions have to get through concurrently now to produce duplicate records, but we have also introduced a possibility of a 'hanging lock', if a browser process gets torn down violently or hangs in the wrong place, we could be stuck with a lock that nothing will release.

If you want to keep the lock I would make it a timestamp rather than a bool, then you can at least disregard and reset it if it was set more than a few minutes ago, removing that risk

plugin-hrm-form/src/utils/sharedState.js
```diff
@@ -142,3 +141,64 @@ export const savePendingContactToSharedState = async (task, payload, error) => {
...
+    };
+    await Promise.all(failedRetriesIndexes.map(itemIndex => list.mutate(itemIndex, incrementRetries)));
+    await Promise.all(successfulRetriesIndexes.map(itemIndex => list.remove(itemIndex)));
+
+    console.log(`${successfulRetriesIndexes.length} pending contacts were saved to external backend`);
+    console.log(`${failedRetriesIndexes.length} pending contacts could not be saved`);
+  } catch (e) {
+    console.error('Error while auto retrying to save pending contacts', e);
+  } finally {
+    await pendingContactsLock.set({ isLocked: false });
```
stephenhand, line 202: Is `await` in `finally` blocks ok? I assumed it would be a nono, but if tsc and the linters aren't complaining it must be ok.

plugin-hrm-form/src/dualWrite/index.ts
```diff
@@ -22,6 +22,7 @@ export const saveContactToExternalBackend = async (task: ITask, payload: any) =>
   if (saveContact) {
     try {
       await saveContact(task, payload);
+      autoRetrySavingPendingContacts(saveContact);
```
stephenhand, line 25: Could there be a possibility of the counsellor being sent messages with links for contacts that other counsellors saved in this scenario? Is that desirable or do we need to suppress the message?

murilovmachado: There's actually a chance of changing the goodbye message, which is not desirable. Fixing it.

plugin-hrm-form/src/utils/sharedState.js
murilovmachado: Actually, the code never reaches `finally` in this case, because there's an early `return`.

plugin-hrm-form/src/utils/sharedState.js
murilovmachado: > Is the write operation to the SaferNet backend not idempotent? Would save some headaches if it was!

We cannot assume it's idempotent, because it's not in our control. Also, although SaferNet is currently the only helpline using this feature,  this will be called for any future helpline that requires saving to their external backend.

> Since the read & write of the lock aren't atomic, it doesn't guarantee the state won't be accessed concurrently

Correct, the lock is not preventing if two different counselors click at the very exact millisecond (or say in 30ms-ish delta). But it will definitely cover 99.9% of the scenarios.
We could think of:
- creating a more robust lock strategy:
  - such as using timestamps as you suggested)
- or using a queue for triggering trying auto-retries
  - so we wouldn't have concurrency issues
- or looking for another solution (I'm not very experienced with concurrency issues)

But it would be very rare for us to fall into these issues with the lock as it is right now. I'd suggest coming back to it only if we ever feel the need.

>  we could be stuck with a lock that nothing will release

It's very simple to release the lock manually, but I agree it could pass many days before someone realizes pending contacts are not being auto-retried. But in the end, it falls in the scenario above (very unlikely to ever happen) and the harm of the pending contacts not being auto-retried for some time is almost none.

plugin-hrm-form/src/utils/sharedState.js
murilovmachado: It's ok to use `await` in a `finally` block, but I could use an inner try-catch inside this finally block, if there's a need to handle `await pendingContactsLock.set({ isLocked: false });` errors. Maybe logging? I can't think of something else to do.

plugin-hrm-form/src/utils/sharedState.js
stephenhand: Code in a `finally` block always runs irrespective of where or how you exit the code in the `try` block. The `finally` code should execute after you hit the `return` statement in this case.

murilovmachado: I didn't know that! Thanks

murilovmachado:
Closing this PR, since this functionality was implemented as a Twilio Serverless Function.

PR #596, opened 2022-01-12
Aarambh staging updatev4

Humairaa127:
- Add "Don't Know" option to dropdowns on Perpetrator and Incident Forms. 
- Replace 'Child at Risk' with 'Child in imminent physical danger'
- Rename 'Specify other service' to 'Specify other'
- Replace actionTaken dropdown with individual checkboxes
- Update insights mapping on oneToOneConfigSpec and oneToManyConfigSpecs

### DIFF
--- a/hrm-form-definitions/form-definitions/in-v1/caseForms/IncidentForm.json
+++ b/hrm-form-definitions/form-definitions/in-v1/caseForms/IncidentForm.json
@@ -65,7 +65,8 @@
       "options": [
         { "value": "Unknown", "label": "" },
         { "value": "Yes", "label": "Yes" },
-        { "value": "No", "label": "No" }
+        { "value": "No", "label": "No" },
+        { "value": "DontKnow", "label": "Don't Know" }
       ]
     },
     {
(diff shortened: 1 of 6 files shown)


hrm-form-definitions/form-definitions/in-v1/caseForms/IncidentForm.json
```diff
@@ -65,7 +65,8 @@
       "options": [
         { "value": "Unknown", "label": "" },
         { "value": "Yes", "label": "Yes" },
-        { "value": "No", "label": "No" }
+        { "value": "No", "label": "No" },
+        { "value": "DontKnow", "label": "Don't Know" }
```
GPaoloni, line 69: I'm fine with this, just asking, isn't it better to use `"Unknown"` value and give a different label to it? Or is there any reason why they want a separate option?

GPaoloni (approved):
Looks good! Just a question.

GPaoloni:
@nickhurlburt @stephenhand Tagging the rest of the team so they are also aware of this changes, and is always good for all of us to check the use cases of this forms, as new ideas regarding the scripts to generate or validate them may occur.

PR #597, opened 2022-01-12
[PoC] First implementation Worker Availability

GPaoloni:
## Description
This PR is a proof of concept of the Option 1 for worker's availability, as described in https://benetech.app.box.com/notes/902771185700.
 
### Checklist
- [ ] Corresponding issue has been opened: https://bugs.benetech.org/browse/CHI-982
[N/A] New tests added
[N/A] Strings are localized

### Verification steps
Keep an eye to the worker status activity at every time while doing this. On top of that, you can open the browser console and find the logs prefixed with `>>>>` that are intentionally included here, to check when the calculation is run and what the values are.
- Run this branch locally with `npm run dev`.
- Try chat based contacts and manual pull for extra tasks.

Required work on top of this:
- Proper handling for transferred tasks to a particular worker (direct transfers). This should be handled in the corresponding Serverless function (set `keepAvailable`) when the transferred task is created and in Flex UI (unset `keepAvailable`) when the transferred task is accepted or rejected.
- Proper handling for offline contacts. This should be handled in the corresponding Serverless function (set & unset `keepAvailable`).
- Define a new flag to indicate when a worker is "ready" to accept tasks vs when it's not, probably using task attributes (something like `isReady`). This flag should be taken into consideration to set Available to allow tasks come in, but it's probably not required to be part of the task assignment rules.
- Define if we want more statuses other than "Ready" and "Not Ready" (analogous to Available and Offline).
- UI (this will partially depend on above item):
  - Top right corner of worker view.
  
    Screenshot from 2022-01-12 15-34-01

  - No tasks view of worker's view.
  
    Screenshot from 2022-01-12 15-34-54

  - Workers statuses of supervisor's view.
  
    Screenshot from 2022-01-12 15-36-27

- How can we deal with the lost information of worker's reports in Insights? Custom timestamps? Fullstory?

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.js
+++ b/plugin-hrm-form/src/HrmFormPlugin.js
@@ -198,6 +198,9 @@ const setUpComponents = setupObject => {
   Components.setUpStandaloneSearch();
 
   if (featureFlags.enable_canned_responses) Components.setupCannedResponses();
+
+  // set feature flag here
+  if (true) Components.setUpWorkerStatusHandler();
 };
 
 /**
--- a/plugin-hrm-form/src/components/ManualPullButton/index.tsx
+++ b/plugin-hrm-form/src/components/ManualPullButton/index.tsx
@@ -18,26 +18,44 @@ type Props = OwnProps & ReturnType<typeof mapStateToProps>;
 const ManualPullButton: React.FC<Props> = ({ queuesStatusState, chatChannelCapacity, worker, workerClient }) => {
   const [isWaitingNewTask, setWaitingNewTask] = useState(false);
 
+  const [, availableActivity] = React.useMemo(() => {
+    return Array.from(workerClient.activities).find(([k, v]) => v.name === 'Available');
+  }, [workerClient.activities]);
+
+  // move to utils for reusability
+  const setKeepAvailable = async () => {
+    const attributes = { ...workerClient.attributes, keepAvailable: true };
+    await workerClient.setAttributes(attributes);
+  };
+
+  // move to utils for reusability
+  const unsetKeepAvailable = async () => {
+    const { keepAvailable, ...attributes } = workerClient.attributes;
+    await workerClient.setAttributes(attributes);
+  };
+
   // Increase chat capacity, if no reservation is created within 5 seconds, capacity is decreased and shows a notification.
   const increaseChatCapacity = async () => {
     setWaitingNewTask(true);
     let alertTimeout = null;
 
-    const cancelTimeout = () => {
+    const cancelTimeout = async () => {
       setWaitingNewTask(false);
       clearTimeout(alertTimeout);
+      await unsetKeepAvailable();
     };
 
     alertTimeout = setTimeout(async () => {
       setWaitingNewTask(false);
       workerClient.removeListener('reservationCreated', cancelTimeout);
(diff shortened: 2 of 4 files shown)


stephenhand:
We should put some thought into how we robustly & deterministically test this if we implement it. It seems like there's lots of corner cases that could trip things up if we aren't careful. Would unit tests be sufficient or do we need to test some scenarios running against a real Twilio backend?

murilovmachado (approved):
Looks good! I've tried locally and I've seen the availability being handled automatically.

One question: Right now, in some scenarios, the user cannot set herself as unavailable. In the final implementation, I suppose the user will be able to set herself as unavailable, right?

GPaoloni:
@murilovmachado Yes, absolutely. But for that to happen, we need to change the "Availability dropdown" and handle the status changes a bit differently. The worker will then set it's status as "Ready" or "Not Ready", and the internals will be automatically handled.

dee-luo:
Some answers to the open questions above: 

* We will want to allow other statuses from the counselor's perspective (similar to how Twilio allows you to set different states at the moment). Each status will map to either "Ready" or "Not Ready" on the backend. 
* Transfers will be important to have, so I do think it will be important to do the worker directory list. 
* For Insights: I don't think FullStory will work unfortunately, since Full Story does not really do much time tracking. I think the best approach is for us to log timestamps of activity changes for now, so we have the data if we need to generate a report. 
* I'm not sure I understand the issue with concurrent independent processes that depend on the keepAvailable flag. Does it matter if the flag is already set? I do think we want to still allow the concurrent processes to happen instead of blocking them, but want to understand what the implications are.

GPaoloni:
> I'm not sure I understand the issue with concurrent independent processes that depend on the keepAvailable flag. Does it matter if the flag is already set? I do think we want to still allow the concurrent processes to happen instead of blocking them, but want to understand what the implications are.

Based on this related piece https://github.com/techmatters/serverless/pull/181/files.
Imagine the following scenario:
- Someone starts submitting an offline contact in my behalf while I'm working (`A`).
- In parallel, I'm submitting an offline contact in my own behalf (`B`).
- The person submits this report. Here the flag is set to `keepAvailable` in my worker attributes, and saves my `previousAttributes`(A) (that is, the same but without `keepAvailable` and `waitingOfflineContact`).
- Almost at the same time, I submit mine. `keepAvailable` is already set for me, so the process continues, but here `previousAttributes`(B) will contain `keepAvailable` set to `true` (that was just set above).
- `A` completes it's flow, setting my worker attributes to `previousAttributes`(A).
- `B` completes it's flow, setting my worker attributes to `previousAttributes`(B). Now my worker is stuck with `keepAvailable` set to `true`.
Also there is a window of time between `A` and `B` where this flags are not set, when they should, because of `B` has started.
Now that I'm thinking, this potential bug already exists for the `waitingOfflineContact` flag, something we need to address I think.
But same kind of issues may arise if two actions using `keepAvailable` takes place at almost the same time, which is imo more likely than the same for `waitingOfflineContact`.

A possible solution to this two issues: keep track of the entire list of processes using this flags, and only clean them when all of this are done. For this, as I've said, we may need to use Sync. What do the rest of the eng team think? @murilovmachado @stephenhand @nickhurlburt

stephenhand (commented):
.

PR #598, opened 2022-01-12
Dee jm updates 011222

dee-luo:
Jamaica updates: 
- Category updates
- Added "Adult Contact" to Call Type buttons (for adults who call in looking for help, they do not want to capture any information and will treat like non-data call)
- Added four fields to Note form 
- Remove City from Caller Information
- Update gender options on all forms
- Update layout form splits


Humairaa127:
Looks good, question on the Note form fields, should there be 3 subjective and one objective?

dee-luo:
@Humairaa127 thanks for catching that! Updated the fields - should be subjective, objective, assessment, plan.

GPaoloni (approved):

PR #601, opened 2022-01-17
Prepopulate form: chatbot to form definition mappings

GPaoloni:
## Description
This PR introduces a new customization file available for the helplines: `PrepopulateKeys.json`. It should be part of the root of the definition, and it's an object with two arrays of strings where each key represents an answer we want to grab from the bot's memory and drop n the form, one represents the values to grab if the contact is "child calling about self" type, the other if it's "caller". Obviously this keys must exist in the form definition for the according tab (child or caller). This file is not required, so if a definition version does not have it, nothing goes wrong, but if it does the file must be an array. An example to address the JM needs (grab the `parish` from the survey) is like `{ "ChildInformationTab": ["parish"], "CallerInformationTab": ["parish"] }` as the entire content of this `PrepopulateKeys.json` file.

### Checklist
- [x] Corresponding issue has been opened: https://bugs.benetech.org/browse/CHI-1003
- [WIP] New tests added
- [N/A] Strings are localized

### Verification steps
- Create a new `PrepopulateKeys.json` file in the corresponding form definition you are using (remember this now lives in `flex-plugins/hrm-form-definitions/` so the entire path for `jm-v1` will look like `flex-plugins/hrm-form-definitions/form-definitions/jm-v1/PrepopulateKeys.json`). Fill it with the fields you want to include in the prepopulate process (like `{ "ChildInformationTab": ["phone1"], "CallerInformationTab": ["phone2"] }`).
- Send a new chat task and reserve it but don't accept it yet. Note that in this step, the survey must be answered (at least the first question with `Yes` or `No`).
- Go to Twilio console, locate the task and edit it's attributes to mimic the bot answers to contain one that matches what you expect, under the `collected_data` property of the memory. You can add something like `"phone1":{"answer":"123"}`.
- Now accept the task, and the answer from above should be pasted in the corresponding form field.
- Repeat the above to check the other call type.

@dee-luo Does this format makes sense? What do the rest of the team thinks? I don't consider this behavior to belong anywhere else in the form definitions so that's why a new file is introduced, but let me know if you think differently.

### DIFF
--- a/hrm-form-definitions/src/formDefinition/loadDefinition.ts
+++ b/hrm-form-definitions/src/formDefinition/loadDefinition.ts
@@ -1,3 +1,5 @@
+/* eslint-disable global-require */
+/* eslint-disable import/no-dynamic-require */
 import {
   CallTypeButtonsDefinitions,
   CannedResponsesDefinitions,
@@ -22,7 +24,7 @@ export enum DefinitionVersionId {
 // Using a variable for the root of the dynamic import confuses webpack :-(
 // const DEFINITION_JSON_ROOT = '../../form-definitions/';
 
-export async function loadDefinition(version: DefinitionVersionId) {
+export async function loadDefinition(version: DefinitionVersionId): Promise<DefinitionVersion> {
   /*
    * Unfortuntately I could not get lazy loading to work, which would have been nice because only the form definitions used by that helpline would be loeded.
    * It appears to be an issue with the module loader, when the bundle was split across multiple files for lazy loading, the additional bundle JS files were not loaded at runtime.
@@ -80,6 +82,13 @@ export async function loadDefinition(version: DefinitionVersionId) {
     /* webpackMode: "eager" */ `../../form-definitions/${version}/CaseStatus.json`
   );
 
+  let prepopulateKeys;
+  try {
+    prepopulateKeys = require(`../../form-definitions/${version}/PrepopulateKeys.json`);
+  } catch (err) {
+    prepopulateKeys = [];
+  }
+
   const { helplines } = helplineInformationModule.default;
   const defaultHelpline =
     helplineInformationModule.default.helplines.find((helpline: HelplineEntry) => helpline.default)
@@ -110,5 +119,6 @@ export async function loadDefinition(version: DefinitionVersionId) {
       oneToManyConfigSpecs: oneToManyConfigSpecsModule.default as OneToManyConfigSpecs,
     },
     caseStatus: caseStatusModule.default as DefinitionVersion['caseStatus'],
+    prepopulateKeys,
   };
 }
--- a/hrm-form-definitions/src/formDefinition/loadDefinition.ts
+++ b/hrm-form-definitions/src/formDefinition/loadDefinition.ts
@@ -86,7 +86,7 @@ export async function loadDefinition(version: DefinitionVersionId): Promise<Defi
   try {
     prepopulateKeys = require(`../../form-definitions/${version}/PrepopulateKeys.json`);
   } catch (err) {
-    prepopulateKeys = [];
+    prepopulateKeys = { ChildInformationTab: [], CallerInformationTab: [] };
   }
 
   const { helplines } = helplineInformationModule.default;
--- a/hrm-form-definitions/src/formDefinition/loadDefinition.ts
+++ b/hrm-form-definitions/src/formDefinition/loadDefinition.ts
@@ -84,7 +84,9 @@ export async function loadDefinition(version: DefinitionVersionId): Promise<Defi
 
   let prepopulateKeys;
   try {
-    prepopulateKeys = require(`../../form-definitions/${version}/PrepopulateKeys.json`);
+    prepopulateKeys = await import(
+      /* webpackMode: "eager" */ `../../form-definitions/${version}/PrepopulateKeys.json`
+    );
   } catch (err) {
     prepopulateKeys = { ChildInformationTab: [], CallerInformationTab: [] };
   }
--- a/plugin-hrm-form/src/utils/prepopulateForm.ts
+++ b/plugin-hrm-form/src/utils/prepopulateForm.ts
@@ -11,7 +11,10 @@ const getUnknownOption = (key: string, definition: FormDefinition) => {
   const inputDef = definition.find(e => e.name === key);
 
   if (inputDef && inputDef.type === 'select') {
-    return inputDef.unknownOption || inputDef.options.find(e => e.value === 'Unknown').value;
+    const unknownOption = inputDef.unknownOption || inputDef.options.find(e => e.value === 'Unknown');
+    if (unknownOption && unknownOption.value) return unknownOption.value;
+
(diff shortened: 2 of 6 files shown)


plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -11,7 +11,10 @@ const getUnknownOption = (key: string, definition: FormDefinition) => {
   const inputDef = definition.find(e => e.name === key);
 
   if (inputDef && inputDef.type === 'select') {
-    return inputDef.unknownOption || inputDef.options.find(e => e.value === 'Unknown').value;
+    const unknownOption = inputDef.unknownOption || inputDef.options.find(e => e.value === 'Unknown');
+    if (unknownOption && unknownOption.value) return unknownOption.value;
+
+    console.error(`getUnknownOption couldn't determine a valid unknown option for key ${key}.`);
```
stephenhand, line 17: If we carry on regardless is this more of a `console.warn`?

hrm-form-definitions/src/formDefinition/loadDefinition.ts
```diff
@@ -80,6 +82,13 @@ export async function loadDefinition(version: DefinitionVersionId) {
     /* webpackMode: "eager" */ `../../form-definitions/${version}/CaseStatus.json`
   );
 
+  let prepopulateKeys;
+  try {
```
stephenhand, line 86: Might it be easier to manage if we had an optional property on form elements called 'botAnswer' or something, which you could populate with the name of the bot question you want to prepopulate the field from? Then you could map answers to form items when the names don't exactly match up and I think from the perspective of trying to maintain the definitions it's easier than having the mappings off in a separate file.

plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -44,21 +46,59 @@ const getAnswerOrUnknown = (
...
+  task: ITask,
+  answers: any,
+  tabFormDefinition: FormDefinition,
+  prepopulateKeys: DefinitionVersion['prepopulateKeys'],
+) => {
+  // Get values from task attributes
+  const { firstName, language } = task.attributes;
+
+  // Get required values from bot's memory
+  const age = getAnswerOrUnknown(answers, 'age', tabFormDefinition, mapAge);
```
stephenhand, line 82: When we add form definition validation we should validate that the required answer mappings are present in the definitions rather than hardcoding them (ok we still need to hardcode 'age' because of the special mapping but it should still be required in the definition). It makes the behaviour more explicit to people working with the definitions.

But until we can validate this sort of thing hardcoding is the way to go.

plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -1,6 +1,6 @@
 import { ITask, Manager } from '@twilio/flex-ui';
 import { capitalize } from 'lodash';
-import { callTypes, FormDefinition } from 'hrm-form-definitions';
+import { callTypes, FormDefinition, DefinitionVersion } from 'hrm-form-definitions';
 
 import { mapAge, mapGenericOption } from './mappers';
```
stephenhand, line 5: These mapper functions only appear to be used in this module. Can we move their definitions into this file and make them private to the module?

GPaoloni:
> Might it be easier to manage if we had an optional property on form elements called 'botAnswer' or something, which you could populate with the name of the bot question you want to prepopulate the field from? Then you could map answers to form items when the names don't exactly match up and I think from the perspective of trying to maintain the definitions it's easier than having the mappings off in a separate file.

Hmm this makes a lot of sense and I really like the idea. The only issue I see is that the people working on the definitions will need to remember to add the `botAnswer` property to the same fields in both form definitions (child and caller) or we might miss capturing some information. I'm fine with moving to this approach, but I want to hear what @dee-luo and @Humairaa127 think. What approach did you prefer/find more ergonomic to use? Please reach me if the thread is not clear.

GPaoloni:
After discussing with @dee-luo and @Humairaa127, they prefer to go with a new file (`PrepopulateKeys.json`) that will have the following shape `{ "ChildInformationTab": string[ ], "CallerInformationTab": string[ ] }`. An example is like `{ "ChildInformationTab": ["phone1"], "CallerInformationTab": ["phone2"] }`: this will grab `phone1` if the contact is of type "child" and `phone2` if the contact is of type "caller". I will wait for @stephenhand's or @murilovmachado's approval before merging.

stephenhand (approved):

hrm-form-definitions/src/formDefinition/loadDefinition.ts
```diff
@@ -80,6 +82,13 @@ export async function loadDefinition(version: DefinitionVersionId) {
...
+    prepopulateKeys = require(`../../form-definitions/${version}/PrepopulateKeys.json`);
```
murilovmachado, line 87: Is the reason that we're not using `await import` here is because it's an optional file?

plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -44,21 +46,59 @@ const getAnswerOrUnknown = (
...
+) => {
+  // Get values from task attributes
+  const { firstName, language } = task.attributes;
+
+  // Get required values from bot's memory
+  const age = getAnswerOrUnknown(answers, 'age', tabFormDefinition, mapAge);
+  const gender = getAnswerOrUnknown(answers, 'gender', tabFormDefinition);
+
+  // This field is not required yet it's bundled here as if it were. Leaving it from now but should we move it to prepopulateKeys where it's used?
+  const ethnicity = getAnswerOrUnknown(answers, 'ethnicity', tabFormDefinition);
```
murilovmachado, line 86: Yep, we definitely should move it to the corresponding `prepopulateKeys.json`.

plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -44,21 +46,59 @@ const getAnswerOrUnknown = (
...
+  const customizableValues = prepopulateKeys.reduce((accum, key) => {
+    const value = getAnswerOrUnknown(answers, key, tabFormDefinition);
+    return value ? { ...accum, [key]: value } : accum;
+  }, {});
+
+  return {
+    ...(firstName && { firstName }),
+    ...(gender && { gender }),
+    ...(age && { age }),
+    ...(ethnicity && { ethnicity }),
```
murilovmachado, line 98: ...and then remove it from here.

murilovmachado (approved):
Looks good to me!

hrm-form-definitions/src/formDefinition/loadDefinition.ts
GPaoloni: Changed to `await import`. With the try catch block it works the same.

plugin-hrm-form/src/utils/prepopulateForm.ts
GPaoloni: @dee-luo Can you confirm if Zambia (form definition `v1`) is the only one using this?

PR #603, opened 2022-01-18
Update IssueCategorizationTab.json

Humairaa127:
Add Parent Category identifiers to LifeLine categories to allow for Insights distinguishing.


dee-luo:
@Humairaa127 we actually don't need the distinction between the ChildLine / LifeLine categories (e.g. Non-Counselling contacts doesn't need a separate label), as we can filter based on Queue. The ones that need distinction are the LifeLine categories that share the same substring (across SCT and KGS).

Humairaa127:
> @Humairaa127 we actually don't need the distinction between the ChildLine / LifeLine categories (e.g. Non-Counselling contacts doesn't need a separate label), as we can filter based on Queue. The ones that need distinction are the LifeLine categories that share the same substring (across SCT and KGS).

Thanks, fixed!

PR #607, opened 2022-01-21
Added adult as a possible non data calltype

GPaoloni:
## Description
Adds `adult` as a possible key for call types. This allows `jm-v1` custom button setup to work properly.

### Verification steps
- Change the used `definitionVersion` to be `jm-v1` by changing the return of `getConfig` (`src/HrmFormPlugin.js`) to be
```
  /**
   * @type {any}
   */
   definitionVersion: 'jm-v1',
```
- Start local server (`npm run dev`).
- Submit offline and webchat based contacts using the `Adult Contact` call type button.

### DIFF
--- a/hrm-form-definitions/src/formDefinition/callTypes.ts
+++ b/hrm-form-definitions/src/formDefinition/callTypes.ts
@@ -7,6 +7,7 @@ export const callTypes = {
   hangup: 'Hang up',
   wrongnumber: 'Wrong Number',
   abusive: 'Abusive',
+  adult: 'Adult Contact',
   test: 'test',
 } as const;
 


nickhurlburt (approved):
Looks good. Please create a Jira ticket for the longer-term improvements you want to make here -- please describe it carefully including acceptance criteria, so it will be ready to go in an estimation meeting.

stephenhand (approved):

GPaoloni:
> Please create a Jira ticket for the longer-term improvements you want to make here

@nickhurlburt  I don't really feel the urge to change stuff here, but if we want this piece of code to be really customizable (allowing for any kind of string to be a "non-data call type") then yes we need to change stuff. I can create a ticket for that, if that's intended.

PR #613, opened 2022-01-26
CHI-566: Refactor cases

stephenhand:
## Description

Refactors cases along more generic lines to reduce duplication and prepare for upcoming functionality enhancements. Branched from https://github.com/techmatters/flex-plugins/pull/606 with the edit case functionality enhancements removed

### Checklist
- [ X ] New tests added
- [ ] Strings are localized

### Related Issues
Related to CHI-566

### Verification steps

Regression test adding each type of item in an existing case
Regression test adding each type of item in a new case
Regression test editing top level info in an existing case
Regression test adding top level info in a new case
Regression test viewing each type of item in an existing case
Regression test viewing each type of item in a new case


plugin-hrm-form/src/states/case/types.ts
```diff
@@ -151,3 +187,42 @@ export type CaseDetails = {
...
   contact: any; // ToDo: change this
 };
+
+export type CaseUpdater = (
+  original: t.CaseInfo,
+  temporaryInfo: t.CaseItemEntry,
+  index: number | undefined,
+) => t.CaseInfo;
+
+export const updateCaseSectionListByIndex = (
```
stephenhand, line 197: I appreciate some of the logic in these functions is redundant without the edit case, but they aren't much less complex only supporting the add case, so I thought I would leave them as is in anticipation of the edit functionality

murilovmachado:
Some of the `Save and Add Another` buttons have incerrect labels:

image
image
image

The other ones seem fine: household, perpetrators, document

murilovmachado:
View Referral has two issues:
- Page title
- No data displayed

image

murilovmachado:
View Note Screen has incorrect title:

image

stephenhand:
> Some of the `Save and Add Another` buttons have incerrect labels:
> 
> image image image
> 
> The other ones seem fine: household, perpetrators, document

Thanks for catching this, sorry it slipped off my todo list - I don't think theses screens actually had 'Add Another' buttons before the generic form gave everything one.

Are they intentionally omitted - in other words do I need to hide them or just add a translation?

murilovmachado:
> Thanks for catching this, sorry it slipped off my todo list - I don't think theses screens actually had 'Add Another' buttons before the generic form gave everything one.
> 
> Are they intentionally omitted - in other words do I need to hide them or just add a translation?

I think you should hide them, for now. But it's a good opportunity for to tell Dee and Jana about this inconsistency. Maybe they'll decide it's good to include it.

stephenhand:
> > Thanks for catching this, sorry it slipped off my todo list - I don't think theses screens actually had 'Add Another' buttons before the generic form gave everything one.
> > Are they intentionally omitted - in other words do I need to hide them or just add a translation?
> 
> I think you should hide them, for now. But it's a good opportunity for to tell Dee and Jana about this inconsistency. Maybe they'll decide it's good to include it.

Spoke to @dee-luo - she elected to keep them so I added the required keys to the translations

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -116,12 +116,15 @@
   "BottomBar-AddThisContactToExistingCase": "Add this Contact to Existing Case",
   "BottomBar-CancelNewCaseAndClose": "Yes, Cancel New Case and Close",
   "BottomBar-SaveAndEnd": "Save and End",
+  "BottomBar-SaveAndAddAnotherNote": "Save and Add Another Note",
```
stephenhand, line 119: Hi @Humairaa127 - sorry, first time doing a PR that affects translations, how do we usually do this? Do we add translations in the same PR or a follow up?

Humairaa127: You can add it to the same PR. Just want to confirm, @GPaoloni  is this correct?

GPaoloni: Yes we usually just bundle the English (default) string in the same PR where it's used, and we add in different translations later on as needed.

murilovmachado:
@stephenhand The View Note screen should be displayed differently than the other screens.

How it should be (unless something has changed that I'm not aware):
image


How it is in the PR:
image

I agree that's an inconsistency, but I think it was designed to be different so that the users could read the text more easily.

murilovmachado (approved):
Looks great! The only last thing to handle is the view note screen.

stephenhand:
> @stephenhand The View Note screen should be displayed differently than the other screens.
> 
> How it should be (unless something has changed that I'm not aware): image
> 
> How it is in the PR: image
> 
> I agree that's an inconsistency, but I think it was designed to be different so that the users could read the text more easily.

What you say makes total sense. Rather than making a special case for Notes, I'm thinking I might make the generic viewer assume that if the form definition provided only has a single 'textarea' type item, it should display it at full width without the field heading. Does that seem reasonable?

murilovmachado:
> What you say makes total sense. Rather than making a special case for Notes, I'm thinking I might make the generic viewer assume that if the form definition provided only has a single 'textarea' type item, it should display it at full width without the field heading. Does that seem reasonable?

I'm fine with either approach.

stephenhand:
> > What you say makes total sense. Rather than making a special case for Notes, I'm thinking I might make the generic viewer assume that if the form definition provided only has a single 'textarea' type item, it should display it at full width without the field heading. Does that seem reasonable?
> 
> I'm fine with either approach.

Done - I also removed the time from the 'Added on ... ' text for referrals, since it's not specified and is always 00:00, which was bugging me.

plugin-hrm-form/src/states/case/types.ts
```diff
@@ -31,20 +31,56 @@ export type ViewReferral = {
...
+  return (
+    tci &&
+    (tci.screen === NewCaseSubroutes.ViewDocument ||
+      tci.screen === NewCaseSubroutes.ViewIncident ||
+      tci.screen === NewCaseSubroutes.ViewPerpetrator ||
+      tci.screen === NewCaseSubroutes.ViewHousehold ||
+      tci.screen === NewCaseSubroutes.ViewNote ||
+      tci.screen === NewCaseSubroutes.ViewReferral)
+  );
+}
```
GPaoloni, line 55: No need to change anything, but _I think_ this could be analogous and avoid repeating the keys (no need to update the custom guard).
```
const ViewTemporaryCaseInfoScreens = [
  NewCaseSubroutes.ViewDocument,
  NewCaseSubroutes.ViewIncident,
  NewCaseSubroutes.ViewPerpetrator,
  NewCaseSubroutes.ViewHousehold,
  NewCaseSubroutes.ViewNote,
  NewCaseSubroutes.ViewReferral,
] as const;

export type ViewTemporaryCaseInfo = {
  screen: typeof ViewTemporaryCaseInfoScreens[number]; // indexing the array with "number" type will give the mapped type with all the array values and because it's declared "as const" there's no repetition
  info: t.CaseItemEntry;
};

export function isViewTemporaryCaseInfo(tci: TemporaryCaseInfo): tci is ViewTemporaryCaseInfo {
  return tci && ViewTemporaryCaseInfoScreens.includes(tci.screen as any); // typecast needed on tci.screen
}
```

GPaoloni (approved):
I only reviewed the code and gotta say this is a master piece, thanks!!
Do you use any plugin to sort the imports alphabetically?

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
```diff
@@ -0,0 +1,221 @@
...
+        </BottomButtonBar>
+      </CaseActionLayout>
+    </FormProvider>
+  );
+};
+
+AddEditCaseItem.displayName = 'AddEditCaseItem';
+
+const mapStateToProps = (state: RootState, ownProps: AddEditCaseItemProps) => {
+  const caseState: CaseState = state[namespace][connectedCaseBase]; // casting type as inference is not working for the store yet
```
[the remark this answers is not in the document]
stephenhand: It's a copy & pasta from one of the previous type specific versions

plugin-hrm-form/src/components/case/ViewCaseItem.tsx
```diff
@@ -11,47 +11,55 @@ import { CaseState } from '../../states/case/reducer';
...
 
+const ViewCaseItem: React.FC<Props> = ({
+  counselorsHash,
+  temporaryCaseInfo,
+  onClickClose,
+  formDefinition,
+  itemType,
+}) => {
+  if (!isViewTemporaryCaseInfo(temporaryCaseInfo))
+    throw new Error('This component only supports temporary case info of the ViewTemporaryCaseInfo type');
```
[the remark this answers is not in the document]
stephenhand: Ok, will include it in an upcoming PR

plugin-hrm-form/src/states/case/types.ts
stephenhand: Very cool trick, thanks!

plugin-hrm-form/src/components/common/forms/UploadFileInput.tsx
```diff
@@ -53,11 +54,12 @@ const UploadFileInput: React.FC<UploadFileInputProps> = ({
   RequiredAsterisk,
   initialValue,
   htmlElRef,
+  // eslint-disable-next-line sonarjs/cognitive-complexity
 }) => {
   const [isLoading, setLoading] = useState(false);
   const fileUploadRef = useRef<HTMLButtonElement>();
+  const fileName = initialValue || watch(path);
```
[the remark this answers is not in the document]
stephenhand: Yes, the control wouldn't load up with a value already set previously (it never had to). Technically shouldn't go in this PR but it's  not a big deal since it will be used in PRs coming soon

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -259,44 +249,20 @@ const Case: React.FC<Props> = props => {
...
+  type CaseItemInfo<T extends TemporaryCaseInfo> = T['info']; // A bit redundant but looks cleaner than the anonymous subtype reference syntax
 
-  const onClickViewPerpetrator = (perpetrator: PerpetratorEntry) => {
-    props.updateTempInfo({ screen: 'view-perpetrator', info: perpetrator }, props.task.taskSid);
-    props.changeRoute({ route, subroute: 'view-perpetrator' }, props.task.taskSid);
+  const onCaseItemActionClick = <T extends TemporaryCaseInfo>(
+    targetSubroute: CaseItemRoute<T>,
+  ): ((entry: CaseItemInfo<T>) => void) => (entry: CaseItemInfo<T>) => {
+    props.updateTempInfo({ screen: targetSubroute, info: entry } as TemporaryCaseInfo, props.task.taskSid);
+    props.changeRoute({ route, subroute: targetSubroute } as AppRoutes, props.task.taskSid);
```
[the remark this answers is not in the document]
stephenhand: I think I see where you're coming from. You can have a go at doing it if you like. I'm still not convinced of the value of this route / info type pairing, probably because we never finished the conversation where you & Murilo explained it to me! We will have to pick that up again :-)

PR #616, opened 2022-01-28
CSAM reporting cleanup

GPaoloni:
Note: easier to review commit by commit.

## Description
This PR is a UI/UX cleanup of the CSAM report components.
Please read the below linked Jira card for the full detail on this changes. Other than those, the only really relevant change is the inclusion of a new input type radio.

Special considerations:
- One of the items of the ticket is "The italic font on the summary screen should pull the real italic font, not the oblique version of it". After investigating, it looks like we will need to import a new font family just to address this, and after a small talk with Dee she's ok leaving it as they are.
- The regex validation for url is not the most exact one, but I couldn't find a valid url that's not accepted. The inverse might be true tho (not restrictive enough). Happy to get some help if someone has more expertise in rexeg.
- I'm hiding the overall label for the radio button set to match the designs. This is not very accessible tho. Happy to change if the design is reviewed.

### Checklist
- [x] Corresponding issue has been opened https://bugs.benetech.org/browse/CHI-977
- [x] New tests added
- [x] Strings are localized

### Verification steps
- Run Flex locally (`npm run dev`).
- Open a new CSAM report (top right button of the contact form).
- Confirm that behavior is the expected, validations are as described in the ticket and the functionality to submit reports is still working.

### DIFF
--- a/plugin-hrm-form/src/___tests__/components/CSAMReport/CSAMReport.test.js
+++ b/plugin-hrm-form/src/___tests__/components/CSAMReport/CSAMReport.test.js
@@ -68,6 +68,117 @@ test("Form renders but can't be submitted empty", async () => {
   expect(alertSpy).toHaveBeenCalled();
 });
 
+test("Form renders but can't be submitted on invalid url", async () => {
+  const alertSpy = jest.spyOn(window, 'alert');
+
+  const updateFormAction = jest.fn();
+  const updateStatusAction = jest.fn();
+  const clearCSAMReportAction = jest.fn();
+  const changeRoute = jest.fn();
+  const addCSAMReportEntry = jest.fn();
+  const csamReportState = { form: initialValues };
+  const routing = { route: 'csam-report', subroute: 'form' };
+  const counselorsHash = { workerSid };
+
+  render(
+    <StorelessThemeProvider themeConf={themeConf}>
+      <CSAMReportScreen
+        taskSid={taskSid}
+        updateFormAction={updateFormAction}
+        updateStatusAction={updateStatusAction}
+        clearCSAMReportAction={clearCSAMReportAction}
+        changeRoute={changeRoute}
+        addCSAMReportEntry={addCSAMReportEntry}
+        csamReportState={csamReportState}
+        routing={routing}
+        counselorsHash={counselorsHash}
+      />
+    </StorelessThemeProvider>,
+  );
+
+  expect(screen.getByTestId('CSAMReport-FormScreen')).toBeInTheDocument();
+  expect(screen.queryByTestId('CSAMReport-Loading')).not.toBeInTheDocument();
+  expect(screen.queryByTestId('CSAMReport-StatusScreen')).not.toBeInTheDocument();
+
+  const webAddressInput = screen.getByTestId('webAddress');
+  expect(webAddressInput).toBeInTheDocument();
+
+  fireEvent.change(webAddressInput, {
+    target: {
+      value: 'this is not a valid url',
+    },
+  });
+
+  const submitButton = screen.getByTestId('CSAMReport-SubmitButton');
+  expect(submitButton).toBeInTheDocument();
+
+  fireEvent.click(submitButton);
+
+  expect(await screen.findAllByText('NotURLFieldError')).not.toHaveLength(0);
+  expect(alertSpy).toHaveBeenCalled();
+});
+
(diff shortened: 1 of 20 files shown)


murilovmachado:
Just a very minor detail: when there're more than 500 chars in the description, it correctly errors, but shows no error message below the component:
image

plugin-hrm-form/src/states/routing/types.ts
```diff
@@ -47,6 +47,7 @@ export type AppRoutesWithCase =
 export type CSAMReportRoute = {
   route: 'csam-report';
   subroute: 'form' | 'loading' | 'status';
+  previousRoute: AppRoutes;
```
murilovmachado, line 50: Now that we're using **previous route**, it's one more reason to investigate react-router again (where we get this for free, I believe)

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -237,6 +241,59 @@ export const getInputType = (parents: string[], updateCallback: () => void, cust
           }}
         </ConnectForm>
       );
+    case 'radio-input':
+      return (
+        <ConnectForm key={path}>
+          {({ errors, register, setValue, watch }) => {
+            const [isMounted, setIsMounted] = React.useState(false); // value to avoid setting the default in the first render.
+
```
murilovmachado, line 249: One thing that is not handled here is the autofocus thing. If in a JSON definition, the first input is a `radio-input` it won't be auto-focused. Not needed for this PR, but I think it would be good to have a `// TODO: autofocus` so we don't forget in the future.

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -237,6 +241,59 @@ export const getInputType = (parents: string[], updateCallback: () => void, cust
...
+                    <FormLabel htmlFor={`${path}-${value}`}>
+                      <Row>
+                        <FormRadioInput
+                          id={`${path}-${value}`}
+                          data-testid={`${path}-${value}`}
+                          name={path}
+                          type="radio"
+                          value={value}
+                          onChange={updateCallback}
+                          innerRef={register(rules)}
```
murilovmachado, line 279: more specifically, I think here would be the part where we set autofocus. But again, we can work on it later.

murilovmachado (approved):
Looks good to me!

plugin-hrm-form/src/___tests__/components/CSAMReport/CSAMReport.test.js
```diff
@@ -68,6 +68,117 @@ test("Form renders but can't be submitted empty", async () => {
...
+  expect(submitButton).toBeInTheDocument();
+
+  fireEvent.click(submitButton);
+
+  expect(await screen.findAllByText('NotURLFieldError')).not.toHaveLength(0);
+  expect(alertSpy).toHaveBeenCalled();
+});
+
+test("Form can't be submitted if anonymous value is undefined", async () => {
+  const alertSpy = jest.spyOn(window, 'alert');
```
stephenhand, line 122: Any possibility we could DRY these tests up a little? a lot of the setup looks to be repeated

stephenhand (approved):
One minor comment about tests, otherwise LGTM

plugin-hrm-form/src/states/routing/types.ts
GPaoloni: Time to go back to https://www.twilio.com/docs/flex/developer/ui/custom-views-and-routes and https://assets.flex.twilio.com/docs/releases/flex-ui/1.24.0/Actions.html. Looks like we can even use other libraries, like `react-router`.

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
GPaoloni: Addressed in https://github.com/techmatters/flex-plugins/pull/616/commits/7ceab033712126072f9e9cfdfab8725d16f2c8eb.

plugin-hrm-form/src/___tests__/components/CSAMReport/CSAMReport.test.js
GPaoloni: Addressed in https://github.com/techmatters/flex-plugins/pull/616/commits/0086653fe64db9b52099784ddb58fe6926d57e86.

PR #630, opened 2022-02-08
CHI-1014: Added filtering by Twilio account in FullStory

mmythily:
Description
- CHI-1014: I added a function to allow filtering with use of twilio account id to FullStory. This is recorded as `helpline` user property in FullStory. 
- Service configuration has been set to allow fullstory recording in Aselo Beta. This needs to be tested once this PR is merged.

### DIFF
--- a/plugin-hrm-form/src/utils/setUpMonitoring.js
+++ b/plugin-hrm-form/src/utils/setUpMonitoring.js
@@ -3,7 +3,12 @@ import Rollbar from 'rollbar';
 import { datadogRum } from '@datadog/browser-rum';
 import * as FullStory from '@fullstory/browser';
 
-import { rollbarAccessToken, datadogAccessToken, datadogApplicationID, fullStoryId } from '../private/secret';
+import {
+  rollbarAccessToken,
+  datadogAccessToken,
+  datadogApplicationID,
+  fullStoryId,
+} from '../private/secret';
 
 function setUpDatadogRum(workerClient, monitoringEnv) {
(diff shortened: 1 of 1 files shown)


stephenhand:
I think whatever linter you are using is configured differently to the linter that we run on CI - it looks like yours is enforcing a max 80 character line limit, whereas the one in CI allows 120, which is causing the linter check to fail.

Can you configure your linter to take it's settings from the project's eslint configuration?

mmythily:
Thanks @stephenhand.  I have yet to successfully configure my linter to take settings from the project's eslint configuration. Here is one with a manual fix.

stephenhand (approved):
LGTM

plugin-hrm-form/src/utils/setUpMonitoring.js
```diff
@@ -87,11 +87,22 @@ function setUpFullStory() {
   });
   console.log('Fullstory monitoring is enabled');
 }
+/**
+ * Identifies and usage by Twilio Account ID (accountSid) in FullStory
+ * @param workerClient
+ */
+function helplineIdentifierFullStory(workerClient) {
+  const { accountSid } = workerClient;
+  FullStory.setUserVars({ helpline: accountSid });
```
GPaoloni, line 96: This will be reflected as `helpline` property in Fullstory? I think that's fine, we should just clarify this to Dee, as `helpline` will mean different things in code than in FS.

GPaoloni (approved):
Looks good!

PR #632, opened 2022-02-09
Create ca.ts

Humairaa127:
## Description

Default permissions

### Checklist
- [ ] New tests added
- [ ] Strings are localized


GPaoloni:
@stephenhand Not sure if you already addressed this, but the changes on the permissions should also be reflected on the HRM side. It will be good to start thinking in dropping this rules in a share location (an s3 bucket maybe?) that could also be part of the Terraform provisioning. What do you think?

PR #636, opened 2022-02-11
CSAM Report UI cleanup II

GPaoloni:
## Description
This PR addresses the second round of feedback on the CSAM UI.

Note: easier to review commit by commit.

### Checklist
- [ ] Corresponding issue has been opened https://bugs.benetech.org/browse/CHI-1062
- [N/A] New tests added
- [ ] Strings are localized

### Verification steps
- Run local server like `npm run dev`.
- Confirm that the comments on the above Jira issue looks like expected (or that there is a comment explaining why it does not).

What to look at:
- Radio buttons shrinked
  Screenshot from 2022-02-11 16-50-28

- Reduced blank space between radio buttons and next elements
  Screenshot from 2022-02-11 16-51-01

- Clicking on the UK data protection act opens a new tab. Also added an icon at the end of the string. All of this is done in the same localizable string (wording, link and icon). This way, the style will be the same no matter the language (hence the phrase lenght).
  Screenshot from 2022-02-11 16-51-33

- Outline is black (`auto`) across the entire app. Agreed with Jana to standardize in black, later on we'll change the color of the outline app-wide (changing to #009DFF for this screen only will make the rest of the app inconsistent).
  Screenshot from 2022-02-11 16-53-10

- Success icon matches the mocks
  Screenshot from 2022-02-11 16-54-35

- Copy button icons shrinked
  Screenshot from 2022-02-11 16-55-00

Jana's comments not addressed in the PR:
> 6. When I typed “www.something” as the url, I was able to save it without an error message. Next I tried “2222” and got an error message, so it is doing some validation. I do think we should require a .com or a .org, or something of the sort to validate it.

After discussing with Jana and Dee, advanced validation attempts might exclude lots of valid urls. We will do basic validation and let the IWF one reject urls that makes no sense (if there's such a thing).

> 9. This is probably out of scope, but I was able to hit the button “send another report” without saving the original CSAM report I was working on. I’d like us to show the alert dialog when we have a chance to add this in.

Will be addressed after the dialog is implemented (in edit case feature).

### DIFF
--- a/plugin-hrm-form/src/translations/en-US/flexUI.json
+++ b/plugin-hrm-form/src/translations/en-US/flexUI.json
@@ -333,7 +333,7 @@
   "CSAMReportForm-ContactDetails": "Contact details",
   "CSAMReportForm-ContactDetailsDescription": "Reports can be filed anonymously or with contact details if the caller would like to follow up or be available to provide further details to IWF.",
   "CSAMReportForm-LearnMore": "Learn more",
-  "CSAMReportForm-ContactDetailsInfo": "Contact details will be recorded on the IWF database for 3 months, and then will then be deleted in accordance with the <a class=\"link-text-decoration-none\" href=\"https://www.legislation.gov.uk/ukpga/2018/12/contents/enacted\">UK Data Protection Act</a>",
+  "CSAMReportForm-ContactDetailsInfo": "Contact details will be recorded on the IWF database for 3 months, and then will then be deleted in accordance with the <a class=\"link-text-decoration-none\" href=\"https://www.legislation.gov.uk/ukpga/2018/12/contents/enacted\" target=\"_blank\" rel=\"noopener noreferrer\">UK Data Protection Act</a>",
   "CSAMReportForm-ReportSent": "CSAM Report Sent!",
   "CSAMReportForm-CopyCode": "Copy confirmation code for sharing",
   "CSAMReportForm-Attachment": "CSAM Report has been filled",
--- a/plugin-hrm-form/src/translations/en-US/flexUI.json
+++ b/plugin-hrm-form/src/translations/en-US/flexUI.json
@@ -333,7 +333,7 @@
   "CSAMReportForm-ContactDetails": "Contact details",
   "CSAMReportForm-ContactDetailsDescription": "Reports can be filed anonymously or with contact details if the caller would like to follow up or be available to provide further details to IWF.",
   "CSAMReportForm-LearnMore": "Learn more",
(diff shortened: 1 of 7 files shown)


plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -333,7 +333,7 @@
   "CSAMReportForm-ContactDetails": "Contact details",
   "CSAMReportForm-ContactDetailsDescription": "Reports can be filed anonymously or with contact details if the caller would like to follow up or be available to provide further details to IWF.",
   "CSAMReportForm-LearnMore": "Learn more",
-  "CSAMReportForm-ContactDetailsInfo": "Contact details will be recorded on the IWF database for 3 months, and then will then be deleted in accordance with the <a class=\"link-text-decoration-none\" href=\"https://www.legislation.gov.uk/ukpga/2018/12/contents/enacted\">UK Data Protection Act</a>",
+  "CSAMReportForm-ContactDetailsInfo": "Contact details will be recorded on the IWF database for 3 months, and then will then be deleted in accordance with the <a class=\"link-text-decoration-none\" href=\"https://www.legislation.gov.uk/ukpga/2018/12/contents/enacted\" target=\"_blank\" rel=\"noopener noreferrer\">UK Data Protection Act</a>.   <span class=\"ContactDetailsInfo-open-in-new-icon material-icons\">open_in_new</span>",
```
GPaoloni, line 336: This line uses temple strings to deliver HTML within a localizable string.

Also, if someone is interested, the reason for `rel="noopener noreferrer"`: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

mmythily (approved):
All the changes look great to me.

nickhurlburt:
@GPaoloni who's the primary reviewer for this? Looks like this has been open for awhile.

GPaoloni:
@murilovmachado  Can I ask you to take the primary review on this?

plugin-hrm-form/src/translations/en-US/flexUI.json
murilovmachado: That's very good to know! I use `_blank` all the time

murilovmachado (approved):
Looks great! I've run the code locally and the visual changes seem correct

plugin-hrm-form/src/translations/en-US/flexUI.json
stephenhand: HTML directly in translation strings feels a bit icky to me, seems like there's potential for issues around security (like this one) accessibility etc that could arise from having unsanitised HTML in these strings - also it could trip us up if we ever try to use these translations outside of a web environment, although that might be unlikely

Do the translations support custom tokens, or just the ones read from the task? If we could pass in custom tokens we could have something like 

```
Contact details will be recorded on the IWF database for 3 months, and then will then be deleted in accordance with the {{dataProtectionLink}}
```

And the text for that link could be another translation key. It would be less convenient but we would keep all the markup under the control of the application.

This is more yours, Murilo's & Mythily's area than mine though, so if you tell me I'm worrying unnecessarily I'll take your word for it! :-) I suppose if we don't go bananas and just keep it for simple stuff like links & line breaks, it's unlikely to cause problems

stephenhand (approved):

PR #641, opened 2022-02-15
Fullstory devMode configuration option set to true

mmythily:

## Description
- Bug fix that will stop throwing this Fullstory related error - `Failed to send Form Error event Error: FullStory is not loaded, please ensure the init function is invoked before calling FullStory API functions`
- FullStory recording can be configured to true or false for devMode usage. Currently, it is set as true.

### Related Issues
Fixes [https://bugs.benetech.org/browse/CHI-1027](https://bugs.benetech.org/browse/CHI-1027)

### Verification steps
1. FullStory recording will be turned off in development environment and the following message will logged in the console: `FullStory is in dev mode and is not recording: event method not executed`
2. Check that deployed versions with this change are able to record events for FullStory

### DIFF
--- a/plugin-hrm-form/src/utils/setUpMonitoring.js
+++ b/plugin-hrm-form/src/utils/setUpMonitoring.js
@@ -84,7 +84,7 @@ function setUpRollbarLogger(plugin, workerClient, monitoringEnv) {
 function setUpFullStory() {
   FullStory.init({
     orgId: fullStoryId,
-    devMode: true,
+    devMode: process.env.NODE_ENV === 'development',
   });
   console.log('Fullstory monitoring is enabled');
 }
--- a/plugin-hrm-form/src/utils/setUpMonitoring.js
+++ b/plugin-hrm-form/src/utils/setUpMonitoring.js
@@ -84,7 +84,7 @@ function setUpRollbarLogger(plugin, workerClient, monitoringEnv) {
 function setUpFullStory() {
   FullStory.init({
     orgId: fullStoryId,
-    devMode: false,
+    devMode: true,
   });
   console.log('Fullstory monitoring is enabled');
 }
--- a/plugin-hrm-form/src/utils/setUpMonitoring.js
+++ b/plugin-hrm-form/src/utils/setUpMonitoring.js
@@ -84,11 +84,13 @@ function setUpRollbarLogger(plugin, workerClient, monitoringEnv) {
 function setUpFullStory() {
   FullStory.init({
     orgId: fullStoryId,
+    devMode: false,
   });
   console.log('Fullstory monitoring is enabled');
 }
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/utils/setUpMonitoring.js
```diff
@@ -84,11 +84,13 @@ function setUpRollbarLogger(plugin, workerClient, monitoringEnv) {
 function setUpFullStory() {
   FullStory.init({
     orgId: fullStoryId,
+    devMode: true,
```
stephenhand, line 87: Will setting this to `true` all the time not disable Fullstory in deployed versions as well? Do we not need something like:

```suggestion
    devMode: process.env.NODE_ENV === 'development',
```

mmythily: According to the FullStory SDK docs, it sounds like setting devMode configuration to true is sufficient in order to turn off recording only in development environment. [https://github.com/fullstorydev/fullstory-browser-sdk](https://github.com/fullstorydev/fullstory-browser-sdk).

However, I also see that in one of the examples - Angular with devMode enabled - its contingent on env variable.  

If you think it is a better idea to implement devMode with `process.env.NODE_ENV === 'development'` as a condition, I will change it.

stephenhand (approved):

plugin-hrm-form/src/utils/setUpMonitoring.js
mmythily: Makes sense. Thank you, Steve.

plugin-hrm-form/src/utils/setUpMonitoring.js
```diff
@@ -84,11 +84,13 @@ function setUpRollbarLogger(plugin, workerClient, monitoringEnv) {
 function setUpFullStory() {
   FullStory.init({
     orgId: fullStoryId,
+    devMode: process.env.NODE_ENV === 'development',
```
GPaoloni, line 87: I thought that this condition will be based on the feature flag, so we wont record FS recordings unless the helpline agrees with it. As it is right now, all helplines will start sending FS events. I might be missing some more context tho.

GPaoloni (approved):
I left one comment, but otherwise looks good to me.

plugin-hrm-form/src/utils/setUpMonitoring.js
GPaoloni: Lol you are right, I've missed that. Thanks for clarifying!

mmythily: Thanks @GPaoloni and @stephenhand. I learned more about feature flags and FS event recording from this discussion.

PR #643, opened 2022-02-18
In stg update 02.18.2022

Humairaa127:
## Description
- Remove 'keepConfidential' and 'didDiscussRightsWithChild' from CaseInformationTab.json
- Adjust LayoutDefinitions.json to adjust Summary form, Perpetrator, Household and Incident forms.

### Checklist
- [ N/A ] Corresponding issue has been opened
- [ N/A ] New tests added
- [x] Strings are localized

### DIFF
--- a/hrm-form-definitions/form-definitions/in-v1/LayoutDefinitions.json
+++ b/hrm-form-definitions/form-definitions/in-v1/LayoutDefinitions.json
@@ -6,15 +6,12 @@
       "splitFormAt": 15
     },
     "caseInformation": {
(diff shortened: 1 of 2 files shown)


GPaoloni (approved):
Looks good. 
Question, is this already being used in production? The consequence of removing existing fields in the same definition version is that, if there are exiting records in the production DB that have this field populated with data, the counselors won't be able to access it (since we load "contact details" using the very same definition). If that's not an issue, good to go!

stephenhand (approved):

PR #646, opened 2022-02-21
Philippines Definition Files

murilovmachado:
primary reviewer: @stephenhand 

## Description
This PR adds the auto-generated files for Philippines helpline.
It also changes Philipinnes Terraform State to use `ph-v1` definitions. (But this change will only be applied after this PR is merged)


stephenhand (approved):
LGTM.

PR #650, opened 2022-02-22
Modified Fullstory to allow recording single field errors

mmythily:

## Description
- Added FullStory event recording for documenting single field errors. 

### Verification steps
- To test this functionality in development environment, set devMode to false. 
- In FullStory 'Dashboards', check for the 'Disaggregated Errors' card to filter through errors.

### DIFF
--- a/plugin-hrm-form/src/fullStory/index.ts
+++ b/plugin-hrm-form/src/fullStory/index.ts
@@ -56,7 +56,7 @@ function transformErrorObj(errorObj: any) {
  */
 export function recordFormValidationError(origin: string, formError: FieldValues): void {
   let errorJson: string = '{}';
-  let errorArr: any;
+  let errorArr: any[] = [];
   try {
     errorJson = JSON.stringify(serializableFormValidationError(formError));
     errorArr = transformErrorObj(serializableFormValidationError(formError));
--- a/plugin-hrm-form/src/fullStory/index.ts
+++ b/plugin-hrm-form/src/fullStory/index.ts
@@ -36,8 +36,10 @@ export function recordEvent(eventType: string, payload: any) {
 function getFieldErrorObj(tabName: string, fieldName: string, object: any): any[] {
   if (object.hasOwnProperty('type') && object.hasOwnProperty('message')) {
     return [{ [fieldName]: { tab: tabName, ...object } }];
+  } else if (typeof object === 'object') {
+    return Object.keys(object).flatMap(child => getFieldErrorObj(tabName, child, object[child]));
   }
-  return Object.keys(object).flatMap(child => getFieldErrorObj(tabName, child, object[child]));
+  return [];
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/fullStory/index.ts
```diff
@@ -49,6 +54,21 @@ export function recordFormValidationError(origin: string, formError: FieldValues
     // eslint-disable-next-line camelcase
     errors_str: errorJson,
   });
+
+  const errorArr = Object.values(errorObj)
```
stephenhand, line 58: Do we not need to keep the information about which tab the error occurred on?

For example if you forget a phone number on the 'Child Information' tab of 'TabbedForm' (as opposed to the 'Caller Information' tab), the error would be structured `{ childInformation: { phoneNumber: "RequiredFieldError" } }` (or something like that)

This code looks like it would store 'TabbedForm' on `form_str`, `phoneNumber` on `field_str` but wouldn't store the fact it was on the `childInformation` tab, should we have an extra field in the event for this?

plugin-hrm-form/src/fullStory/index.ts
```diff
@@ -49,6 +54,21 @@ export function recordFormValidationError(origin: string, formError: FieldValues
...
+    .flatMap(obj => Object.entries(obj))
+    .map(([key, value]) => ({ [key]: value }));
+
+  errorArr.map(error => {
+    return recordEvent('Custom Error', {
+      // eslint-disable-next-line camelcase
+      form_str: origin,
+      // eslint-disable-next-line camelcase
+      field_str: `${Object.keys(error)}`,
```
stephenhand, line 67: This will store an array with a single string item, rather than just the key, e.g. `["phoneNumber"]` not `"phoneNumber"` - is that the intention?

GPaoloni (approved):
Looks good to me (other than Steve's comments).

plugin-hrm-form/src/fullStory/index.ts
```diff
@@ -49,6 +54,21 @@ export function recordFormValidationError(origin: string, formError: FieldValues
     // eslint-disable-next-line camelcase
     errors_str: errorJson,
   });
+
+  const errorArr = Object.values(errorObj)
+    .flatMap(obj => Object.entries(obj))
+    .map(([key, value]) => ({ [key]: value }));
+
+  errorArr.map(error => {
```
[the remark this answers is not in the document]
mmythily: Thanks @stephenhand, will be implementing this.

plugin-hrm-form/src/fullStory/index.ts
mmythily: Thanks for bringing this up. yes, it should just be a string.

stephenhand (approved):
LGTM

PR #652, opened 2022-02-23
Autopopulate Contact Info Date/Time

GPaoloni:
## Description
This PR
- Adds a new option to `date-input` and `time-input` type definitions: `initializeWithCurrent`, an optional boolean that if present will use the current date or time.
- Sets `initializeWithCurrent` to `true` in the date and time pickers of the offline contact tab.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1024)
- [ ] New tests added
- [N/A] Strings are localized

### Verification steps
- `➜ UNBUNDLED_REACT=true npm ci`
- `➜ npm run dev`
- Start adding an offline contact and confirm that in the first form tab the date and time pickers are using the latest available option.

PR #654, opened 2022-02-24
Fixed wrong initial value for time input when missing leading zero

GPaoloni:

## Description
Fixes a silly bug introduced in https://github.com/techmatters/flex-plugins/pull/652:
When the hours or the minutes have a leading zero, `Date.getHours` and `Date.getMinutes` methods does not includes the zero (as they return a `number`).
Screenshot from 2022-02-24 17-04-19

This resulted in the following "initial value" for a time like `17:03`:
Screenshot from 2022-02-24 17-04-36

which is not a valid value for a time input (hh:mm format).
The fix is actually simple: always prefix a zero to the hours and the minutes, and only use the last two characters:
Screenshot from 2022-02-24 17-01-19

Giving as a result the correct initial value for a time input with times like `17:03`:
Screenshot from 2022-02-24 17-05-30

### Verification steps
- Run locally with `npm run dev`.
- Add an offline contact at a time that has leading zeros in the hour or minutes components (like 08:xx, 12:0x).

### DIFF
--- a/plugin-hrm-form/src/components/common/forms/formGenerators.tsx
+++ b/plugin-hrm-form/src/components/common/forms/formGenerators.tsx
@@ -59,7 +59,7 @@ export const getInitialValue = (def: FormItemDefinition) => {
       if (def.initializeWithCurrent) {
         const date = new Date();
         // Return the locale hh:mm
-        return `${`0${date.getHours()}`.slice(-2)}:${`0${date.getMinutes()}`.slice(-2)}`;
+        return `${date.getHours().toString().padStart(2, '0')}:${date.getMinutes().toString().padStart(2, '0')}`;
       }
 
       return '';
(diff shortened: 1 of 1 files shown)


murilovmachado (approved):
Looks good!

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -59,7 +59,7 @@ export const getInitialValue = (def: FormItemDefinition) => {
       if (def.initializeWithCurrent) {
         const date = new Date();
         // Return the locale hh:mm
-        return `${date.getHours()}:${date.getMinutes()}`;
+        return `${`0${date.getHours()}`.slice(-2)}:${`0${date.getMinutes()}`.slice(-2)}`;
```
stephenhand, line 62: Does `padStart` not work for this case?

stephenhand (approved):
LGTM, comment is not blocking

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
GPaoloni: Didn't knew about this function, thanks for the suggestion! Changing to it as it is clearer.

PR #656, opened 2022-02-24
Post survey logic not triggered for AseloCustomChannels

GPaoloni:

## Description
This PR avoid triggering post survey for what we call "Aselo Custom Channels" (i.e. `'twitter'` and `'instagram'` so far). The reason is that the messages sent thru this channels are not working properly with Twilio autopilot (channel webhooks not being raised), making the post survey useless in this cases.

### Checklist
- [x] Corresponding [issue](https://bugs.benetech.org/browse/CHI-1078) has been opened
- [ ] New tests added
- [N/A] Strings are localized

### Verification steps
As this is already deployed to -dev- environment, you can trigger usual chats and check that post survey works, but it should not work for Twitter or Instagram based tasks.

NOTE: there is an issue with Twitter in Aselo Dev right now, as we are bundling the credentials for the @AseloDev account but the linked one is my personal. Something already discussed with Dee and that we need to address.


plugin-hrm-form/src/states/DomainConstants.ts
```diff
@@ -7,6 +9,8 @@ export const channelTypes = {
   twitter: 'twitter',
 } as const;
 
+export const isAseloCustomChannelTask = (task: CustomITask) => ['twitter', 'instagram'].includes(task.channelType);
```
GPaoloni, line 12: I don't think this is the best place for this method but I don't know where else should I drop it, some ideas?

BTW, I don't believe this is the best place for this file either :p

GPaoloni: Also this is a TODO: once @mmythily's changes on Instagram stuff are in, I'm happy to define a better and more unified way of listing `AseloCustomChannelTask` (like an object) to make it more scalable than just `['twitter', 'instagram']`.

murilovmachado (approved):
I really don't understand anything on Twilio Chat Orchestrations, but the code changes make sense to me.

plugin-hrm-form/src/states/DomainConstants.ts
stephenhand: I normally find these questions have easier answers when code is grouped by functional area rather by type of module. Since this function is only currently used in one module does it not make sense to just define it in the module where it's used for now?

PR #660, opened 2022-02-28
Make Non Data Call Types customizable

GPaoloni:

## Description
This PR adds support for any `non-data` call type. After this PR is merged, all `non-data` call types PR must be defined in the `CallTypeButtons.json` file of the form definition.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1033)
- [ ] New tests added
- [ ] Strings are localized

### Verification steps
- Edit the form definition that you'll be using on the local server, adding/removing `non-data` call types.
- `➜ UNBUNDLED_REACT=true npm ci` to install the updated form definition.
- ➜ `npm run dev` to start a local server.
- Test saving and retrieving the contacts. `data` contacts should keep working as usual (pre-populate. copy functionality, etc).

### DIFF
--- a/hrm-form-definitions/src/formDefinition/callTypes.ts
+++ b/hrm-form-definitions/src/formDefinition/callTypes.ts
@@ -4,6 +4,7 @@ export const callTypes = {
   caller: 'Someone calling about a child',
 } as const;
 
-export type DataCallTypes = typeof callTypes['child' | 'caller'];
-export type CallTypeKeys = DataCallTypes | string; // This results in "strings" as it's a broader type. Leaving the DataCallTypes in intentionally to emphasize them.
-export type CallTypes = string;
+export type DataCallTypesKeys = keyof typeof callTypes;
+export type CallTypeKeys = DataCallTypesKeys | string; // This results in "strings" as it's a broader type. Leaving the DataCallTypesKeys in intentionally to emphasize them.
+export type DataCallTypes = typeof callTypes[keyof typeof callTypes];
+export type CallTypes = DataCallTypes | string; // This results in "strings" as it's a broader type. Leaving the DataCallTypes in intentionally to emphasize them.
--- a/hrm-form-definitions/src/formDefinition/callTypes.ts
+++ b/hrm-form-definitions/src/formDefinition/callTypes.ts
@@ -1,16 +1,9 @@
+// Data callTypes. Preserving name to avoid big refactor
 export const callTypes = {
   child: 'Child calling about self',
   caller: 'Someone calling about a child',
-  silent: 'Silent',
-  blank: 'Blank',
-  joke: 'Joke',
-  hangup: 'Hang up',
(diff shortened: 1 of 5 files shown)


hrm-form-definitions/src/formDefinition/callTypes.ts
```diff
@@ -1,16 +1,10 @@
...
-  test: 'test',
 } as const;
 
-export type CallTypeKeys = keyof typeof callTypes;
-export type CallTypes = typeof callTypes[keyof typeof callTypes];
-export type DataCallTypes = typeof callTypes['child' | 'caller'];
+export type DataCallTypesKeys = keyof typeof callTypes;
+export type CallTypeKeys = DataCallTypesKeys | string; // This results in "strings" as it's a broader type. Leaving the DataCallTypesKeys in intentionally to emphasize them.
+export type DataCallTypes = typeof callTypes[keyof typeof callTypes];
+export type CallTypes = DataCallTypes | string; // This results in "strings" as it's a broader type. Leaving the DataCallTypes in intentionally to emphasize them.
```
GPaoloni, line 10: I know this are just `string`s, but I left the type alias anyways (and the unions) to make clear the intentions of this, as some of them have semantic meaning.

hrm-form-definitions/src/formDefinition/callTypes.ts
```diff
@@ -1,16 +1,10 @@
...
-  abusive: 'Abusive',
-  adult: 'Adult Contact',
-  test: 'test',
 } as const;
 
-export type CallTypeKeys = keyof typeof callTypes;
-export type CallTypes = typeof callTypes[keyof typeof callTypes];
-export type DataCallTypes = typeof callTypes['child' | 'caller'];
+export type DataCallTypesKeys = keyof typeof callTypes;
+export type CallTypeKeys = DataCallTypesKeys | string; // This results in "strings" as it's a broader type. Leaving the DataCallTypesKeys in intentionally to emphasize them.
```
stephenhand, line 8: I wonder if it might be better to change it so that the category could be `child`, `caller` or `non-data`, or if `data` and `non-data` is all over the back end, add another attribute (e.g. `dataCategory`) that is only on data call types and can only be `caller` or `child`. 

Then the front end could drive it's logic off that attribute, and the name could just be a name, rather than having to try to express that the name 'might be meaningful but it might not'

GPaoloni: I entirely agree with you, this is a needed refactor. I suggest going with a two properties object like `{ type: 'child' | 'caller' | 'non-data', label: string }`, so we can drive the logic using the `type` and display in the UI the `label`, so each helpline can use the wording they want. This will involve some migration on the backend plus a rework on the UI so I'd prefer to have that in it's own ticket.

stephenhand (approved):

PR #667, opened 2022-03-08
CHI-1040: Basic Case Edit functionality

stephenhand:
Primary reviewer: @GPaoloni 

## Description

* Basic functionality for CHI-1040
* Requires updated HRM for editing notes
* Does not navigate back to view after edit is saved (this will come in a follow on PR)
* Still needs tests (will come in a follow on PR after some refactor PRs)
 
### Checklist
- [ NO ] New tests added
- [ ] Strings are localized

### Related Issues
Fixes #CHI-1040

### Verification steps

Regression only - test functionality in following PRs


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -470,6 +473,7 @@ const Case: React.FC<Props> = props => {
   };
   switch (subroute) {
     case NewCaseSubroutes.AddNote:
+    case NewCaseSubroutes.EditNote:
```
GPaoloni, line 476: This might be a dumb question, I may be missing something, but won't it be easier to follow if the difference between an addition vs an edit is given by prop here? Something like `<AddEditCaseItem isEditingItem={action===NewCaseSubroutes.EditNote} .... />`. That way we won't need to compute it by doing `isEditTemporaryCaseInfo(temporaryCaseInfo)` inside the `AddEditCaseItem` component.

GPaoloni (approved):
Code looks awesome!
I was not able to test it tho, as I'm getting an error when trying to run the new updates on hrm. When running the migration `20220301134500-migrate-notes-to-counsellor-notes`, the following error occurs:

Screenshot from 2022-03-10 20-32-41

This happens in my local DB and in the Docker one as well. I've tried using node v12 and reinstalling hrm dependencies (`npm ci`) but it persist. I did some research on Google but couldn't find anything really relevant :/ Do you have any idea on why this might be?

GPaoloni:
Thanks @stephenhand! As you pointed out, the issue was that the Postgress version was outdated. In my case, the container was building with the old version, to make it work I had to
- `docker-compose down`
- `docker-compose build`
- `docker-compose up`

Tested the behavior and it is correct! The only thing I've noticed is that the Edit and Close buttons have no spacing. Maybe because of my screen's dimensions (I'm working on a 14" laptop right now).
Screenshot from 2022-03-11 15-46-57

stephenhand:
@GPaoloni @murilovmachado - you both make good points about the routes, but there's some significant refactoring of the case routes in https://github.com/techmatters/flex-plugins/pull/679 that might address your concerns (or raise additional ones...)

plugin-hrm-form/src/components/case/Case.tsx
stephenhand: > Thanks @stephenhand! As you pointed out, the issue was that the Postgres version was outdated. In my case, the container was building with the old version, to make it work I had to
> 
> * `docker-compose down`
> * `docker-compose build`
> * `docker-compose up`
> 
> Tested the behavior and it is correct! The only thing I've noticed is that the Edit and Close buttons have no spacing. Maybe because of my screen's dimensions (I'm working on a 14" laptop right now). Screenshot from 2022-03-11 15-46-57

The styling issue was addressed in follow on PRs, but I've since backported it to this one, it should match Jana's mockups now

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
```diff
@@ -154,15 +181,22 @@ const AddEditCaseItem: React.FC<Props> = ({
...
+    },
+  );
 
   return (
     <FormProvider {...methods}>
       <CaseActionLayout>
         <CaseActionFormContainer>
-          <ActionHeader titleTemplate={`Case-Add${itemType}`} onClickClose={onClickClose} counselor={counselor} />
+          <ActionHeader
+            titleTemplate={isEditingExistingItem() ? `Case-Edit${itemType}` : `Case-Add${itemType}`}
```
murilovmachado, line 196: One thing that I forgot to mention is to create these `Case-EditSomething` strings. Right now, I'm getting titles like these:
image
image

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
```diff
@@ -88,12 +97,21 @@ const AddEditCaseItem: React.FC<Props> = ({
 
   const { temporaryCaseInfo } = connectedCaseState;
 
+  const isEditingExistingItem = (): boolean => isEditTemporaryCaseInfo(temporaryCaseInfo);
```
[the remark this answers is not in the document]
stephenhand: Probably, I think `isEditTemporaryCaseInfo` is removed in #679 - or rendered trivial - anyway

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
stephenhand: Good catch! How do we normally do the localisations, I did the Portuguese in Google Translate since they were small - if you'd be so kind as to check it? :-) What is the usual process for translations

murilovmachado: The Portuguese translations seem right to me.
I think we don't have a process yet for all translations. We get most of them during the initial conversation with a helpline, but later on, new strings come up, and we don't have a specific process for that.

PR #669, opened 2022-03-08
Rework permissions files + Implement Edit Case Permissions

GPaoloni:

## Description
This PR is a small rework on the way we declare the permissions for each helpline:
- Instead than `.ts` files, now the rules are defined in a `.json` file. For this, I've given the shape that I find most intuitive to work with, but will love the feedback from @dee-luo, @Humairaa127 or @adedoyino to check.
- Moved all the existing sets of rules for each helpline to this new format, following the same stuff that's on the backend and adding the more general "is supervisor or is case open and is creator" rule for those that does not exist in the backend (like "can edit child at risk").

Brief explanation on how the new files will define the permissions:
There are 4 static conditions that we can check: `isSupervisor`, `isCreator`, `isCaseOpen` and `everyone`. For each action, we can define sets (lists) of conditions that should be met in order to "allow the person to perform the action". If all the conditions under one set (a single list) are met (i.e. `true`), then the action can be performed.
Some examples:
- Lets say that we want to allow a user to `addReferral` if the user either a) is a supervisor or b) if is the creator and the case is still open. This can be expressed as `[["isSupervisor"], ["isCreator", "isCaseOpen"]]`. Note that this is "a list of lists". The external list can be thought as "the sets of conditions". Each internal list is "a set of conditions". If all the conditions within an internal list are true, then the user is allowed, in this case to `addReferral`.
- An action that can only be performed by a supervisor can be expressed as `[['isSupervisor']]`. Only one set of conditions, only one condition within.
- An action that can't be performed by anyone can be expressed as `[]` (an empty list). This means that there are no possible sets of conditions to allow someone to perform the action.
- An action that can be performed by everyone can be expressed as `[['everyone']]`. This is, a single set of conditions with one condition within: `everyone` will always evaluate to `true`.

The idea of this refactor is that moves us one step closer to having only one file that defines the permissions, and both, frontend and backend can read on this file to decide the permissions. 
This also introduces more work to be done if we go down this road:
- Move the `fetchRules` function to occur on plugin initialization, store it in global state and then consume from there (like we do with form definitions).
- Replicate this rework on the backend.
- Unify the source of truth for the `.json` files. When we do this, it will be worth discussing with the team where we should drop them, as if we put them somewhere different than HRM, we'll need to "cache" the definitions after fetching them, or we'll be performing many unnecessary calls. On the other hand, we could leave them inside of the backend and expose them from an API to the frontend. The later one has some benefits in my opinion:
  - No need to cache them, as they will be bundled with the code.
  - Rules will continue to be in source control, without the need of creating a new repo/structure to hold them, plus the workflow for the product team will be mostly the same.

Notes: 
- While this PR introduces a lot of new actions, only a few of them are actually being used in the code, being the "edit" ones not being used anywhere yet. I think @stephenhand will use this ones while working on the edit case stories.
- Some permissions might not be the right ones, I just copied over the ones from HRM assuming they are already in place.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1044)
- [x] New tests added
- [N/A] Strings are localized

### Verification steps
- `npm run dev`
- Test different rules configs if desired to see that they still work as intended.
- Check on own and others cases, if possible also check with a supervisor user, to see that the different sets of rules are behaving as expected.


plugin-hrm-form/src/permissions/index.ts
```diff
@@ -19,60 +13,70 @@ export const PermissionActions = {
...
+  };
+
+  const checkCondition = (condition: Condition) => conditionsState[condition];
+  const checkConditionsSet = (conditionsSet: ConditionSet) => conditionsSet.every(checkCondition);
+  const checkConditionsSets = (conditionsSets: ConditionSets) => conditionsSets.some(checkConditionsSet);
+
+  const rules = fetchRules(permissionConfig);
+
+  const rulesAreValid = Object.values(PermissionActions).every(action => rules[action]);
+  if (!rulesAreValid) throw new Error(`Rules file for ${permissionConfig} is incomplete.`);
```
GPaoloni, line 71: This is just a dumb validation to check that we did not forgot any action in the rules. More advanced validation can be done, I just didn't spent too much time on this now.

stephenhand (approved):
Looks much more manageable for larger permission sets.

dee-luo:
This makes sense to me! To clarify, once we go with this approach, when we make changes to a helpline's permissiojns we only have to edit their json file, correct? 

I'm curious what the "caseStatusTransition" permission does - I know we have customized case statuses and permissions based on how the case transitions, but would like to understand how that needs to get defined here.

GPaoloni:
@dee-luo 
> To clarify, once we go with this approach, when we make changes to a helpline's permissiojns we only have to edit their json file, correct?

Correct!

> I'm curious what the "caseStatusTransition" permission does - I know we have customized case statuses and permissions based on how the case transitions, but would like to understand how that needs to get defined here.

If you recall, the dropdown for case status can contain more options than just `open` and `closed` (the custom statuses). If a counselor does not have permissions to close the case, the option will not show at all. Same is true for all the status transitions that involves a custom status, and they are filtered  based on this permission: you can move the case to a custom state if `caseStatusTransition` action evaluates `true` fro you.

Humairaa127 (approved):
Looks great and very helpful! Just something to note for later on: If we add the ability to create custom case forms other than (perpetrator, incident, etc) i assume we will need to create new permissions?

PR #675, opened 2022-03-10
CHI1043 case details warning dialog

mmythily:

## Description
- In Case Details page, if Summary, Follow up date, Case Status, risk checkbox are changed and the user attempts to press Close button, a dialog window is triggered with a warning message to save or exit
- Strings for en-US and pt-BR have been added. 
- The dialog component will possibly need tests

### Checklist
- [ ] New tests added
- [x] Strings are localized

### Related Issues
Fixes # CHI-1043.

### Verification steps
Save Changes Dialog


stephenhand (approved):
LGTM - I will merge these up into the refactored version I have on my branch

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -701,9 +707,33 @@ const Case: React.FC<Props> = props => {
...
                 <Box marginRight="15px">
-                  <StyledNextStepButton secondary roundCorners onClick={props.handleClose}>
+                  <StyledNextStepButton
+                    secondary
+                    roundCorners
+                    onClick={caseHasBeenEdited === true ? () => setCloseDialog(true) : props.handleClose}
+                  >
                     <Template code="BottomBar-Close" />
                   </StyledNextStepButton>
+                  <CloseTaskDialog open={closeDialog} onClose={props.handleClose}>
```
stephenhand, line 717: I think we just want to return to the case if the user dismisses the dialog, rather than closing the case? I think we only want to close the case if the user explicitly clicks the don't save button

mmythily: Makes sense. Thanks for the catch. I will setCloseDialog to false.

murilovmachado (approved):
The code looks **very good** and the behavior is also **correct**.

There're two things that could be improved:
1) The mockups seem to have:
image
- A yellow warning Icon above the title
- The background color of the `Don't Save` Button is transparent

2) Tab Navigation goes out of the dialog. You could use the component `<TabPressWrapper>` for this. There're some instructions on how to use it in its comments:
image
Basically in your scenario, the two things to do would be:
- Wrap the content of your dialog inside `<TabPressWrapper>`
- Explicitly define your three buttons `tabIndex` (This is usually not a good practice, we can revisit this in the future).

Your code would be similar to:
image
(I've actually tried it, and it breaks the layout a bit. So you'd have to fix the layout again)

Besides those small details, everything looks very good. Good job!

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -701,9 +707,33 @@ const Case: React.FC<Props> = props => {
...
+                    <Row>
+                      <StyledNextStepButton secondary onClick={props.handleClose} margin="15px auto">
+                        <Template code="BottomBar-DontSave" />
+                      </StyledNextStepButton>
+                      <StyledNextStepButton disabled={null} onClick={handleUpdate} margin="15px auto">
+                        <Template code="BottomBar-Save" />
+                      </StyledNextStepButton>
+                    </Row>
+                    <Box marginBottom="25px" />
+                  </CloseTaskDialog>
```
GPaoloni, line 736: Not a requirement for this PR. Maybe in a followup PR it makes sense to factor out this as a separate component? I don't know if we'll reuse this dialog in some other places tho.

GPaoloni (approved):
Looks good, great work!

plugin-hrm-form/src/components/case/Case.tsx
mmythily: I will do that.

mmythily:
Looks like refactoring the CloseTaskDialog component is required which I will address in a future PR along with rest of the suggestion because the style changes are depending on it as well. For now, I will merge this PR.

PR #679, opened 2022-03-16
Chi 1040 case routing refactor

stephenhand:
Primary reviewer: @GPaoloni 

## Description

* Refactors routing for the editable 'sections' of cases to have a single subroute per section and then a `CaseItemAction` which can be `View`, `Add` or `Edit`
* Leverage the above to implement the 'back to view' requirement of the case edit feature
* Removes usage of activities endpoint, now builds activities from the case information it already has

NOTE: This version 'crosses the Rubicon' in terms of HRM compatibility, the new version is required. Previous PRs would work except for Note / Referral editing with an old version of HRM

TODO: Extra unit tests

### Checklist
- [ ] New tests added
- [ ] Strings are localized

### Related Issues
Fixes CHI-1040

### Verification steps

Section types: Note, Referral, Document, Perpetrator, Incident

* Edit an item each section type of a case as you are adding the new case as part of a contact.
* Open a case in the case list, add an item each section type then edit it.
* Open a case in the case list with existing items for each section and edit an existing item from each section.
* Repeat the previous 2 steps from the case search

In each case ensure that documents can only have comments edited, and that the document can still be downloaded after editing


plugin-hrm-form/src/states/routing/types.ts
```diff
@@ -9,54 +9,70 @@ export type TabbedFormSubroutes =
...
+      route: 'new-case';
+      subroute?: CaseSectionSubroute;
+      action: CaseItemAction;
+      autoFocus?: boolean;
+    }
+  | {
+      route: 'select-call-type';
+      subroute?: CaseSectionSubroute;
+      action: CaseItemAction;
+    };
```
murilovmachado, line 53: Great, this answers my question on the first PR.

plugin-hrm-form/src/states/routing/types.ts
```diff
@@ -9,54 +9,70 @@ export type TabbedFormSubroutes =
...
       route: 'new-case';
-      subroute?: typeof NewCaseSubroutes[keyof typeof NewCaseSubroutes];
+      subroute?: typeof NewCaseOtherSubroutes[keyof typeof NewCaseOtherSubroutes];
       autoFocus?: boolean;
     }
   | {
       route: 'select-call-type';
-      subroute?: typeof NewCaseSubroutes[keyof typeof NewCaseSubroutes];
+      subroute?: typeof NewCaseOtherSubroutes[keyof typeof NewCaseOtherSubroutes];
     };
```
murilovmachado, line 76: Are we still using the old type? I thought the new one would replace this completely.

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -306,135 +306,150 @@ const Case: React.FC<Props> = ({
...
+                ci.counsellorNotes = ci.counsellorNotes ?? [];
+                return ci.counsellorNotes;
+              },
+              temp => ({
+                note: temp.form.note.toString(),
+                createdAt: temp.createdAt,
+                twilioWorkerId: temp.twilioWorkerId,
+              }),
+            )}
+          />
```
murilovmachado, line 345: Here I got the feeling that it was not so clear/easy to read what `applyTemporaryInfoToCase` receives. I guess it receives an adapter because of the old way notes and referrals were being stored, right?

murilovmachado (approved):
Looks good to me! I could navigate between the screens just fine. I could not execute some edits because of permissions, but I guess that's another thing.

stephenhand:
> Looks good to me! I could navigate between the screens just fine. I could not execute some edits because of permissions, but I guess that's another thing.

Hmm, permissions should only affect this PR the same way it did the others, if you were able to edit in other PRs but not in this one that's a problem, I can look into it tomorrow

PR #682, opened 2022-03-18
Update Jamaica and Aselo Demo Permissions

dee-luo:
-Update Jamaica permissions for who can edit case info
-Create new demo.json file for permissions to be used in new Aselo Staging/Prod set up


murilovmachado:
Looks good, but this will require changing the permissions on the backend as well.

murilovmachado (approved):
Now that the backend is merged, this is good to go

PR #683, opened 2022-03-18
Make Offline Contact Channel options customizable

GPaoloni:

## Description
This PR adds support for custom channels in the channels dropdown of Contact Info tab (Offline contacts only).
- Removes the `otherContactChannels` object and all references to it. If the helpline wants more channels other than the contact ones (Facebook, web, voice, SMS, Whatsapp, Twitter, Instagram), then those should defined in the new format introduced here.
- The custom channels can be defined via an optional json file under `flex-plugins/hrm-form-definitions/form-definitions/<definition version>/tabbedForms/ContactlessTaskTab.json` (that is, a new file `ContactlessTaskTab.json` under the `/tabbedForms` folder of the definition version). 
- If the file is not present, nothing happens, the UI will look like
  Screenshot from 2022-03-18 15-12-30

- To add custom channels the new file should have a shape like this
  ```
  {
    "offlineChannels": ["Custom Channel 1", "Custom Channel 2"]
  }
  ```
  That is, an array (list) of strings where each one is a custom channel. The result of the above is an expanded list of the channels like
  Screenshot from 2022-03-18 15-03-11

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-967)
- [ ] New tests added
- [N/A] Strings are localized

### Verification steps
- `UNBUNDLED_REACT=true npm ci` to remove the old `node_modules` and install the updated version of `hrm-form-definitions`.
- `npm run dev` and confirm everything works fine (the list of channels of an offline contact should be only the ones that support an actual contact).
- Create a `/tabbedForms/ContactlessTaskTab.json` file as described above for the custom forms being used.
- `UNBUNDLED_REACT=true npm ci` to install the updated definitions.
- `npm run dev` again and confirm the custom channels are now being added to the dropdown.
- Case timeline icon should default to the `AssignmentInd` icon
  Screenshot from 2022-03-18 15-12-39

### DIFF
--- a/plugin-hrm-form/src/components/ContactDetails.tsx
+++ b/plugin-hrm-form/src/components/ContactDetails.tsx
@@ -76,7 +76,7 @@ const Details: React.FC<Props> = ({
   const childOrUnknown = formatName(childName);
   const childUpperCased = childOrUnknown.toUpperCase();
   const formattedChannel =
-    channel === 'default' ? mapChannelForInsights(details.contactlessTask.channel) : mapChannel(channel);
+    channel === 'default' ? mapChannelForInsights(details.contactlessTask.channel) : mapChannelForInsights(channel);
   const formattedDate = `${format(new Date(dateTime), 'MMM d, yyyy / h:mm aaaaa')}m`;
   const formattedDuration = formatDuration(conversationDuration);
 
--- a/hrm-form-definitions/src/formDefinition/loadDefinition.ts
+++ b/hrm-form-definitions/src/formDefinition/loadDefinition.ts
@@ -65,6 +65,16 @@ export async function loadDefinition(version: DefinitionVersionId): Promise<Defi
   const issueCategorizationTabModule = await import(
     /* webpackMode: "eager" */ `../../form-definitions/${version}/tabbedForms/IssueCategorizationTab.json`
   );
+
+  let contactlessTaskTabModule;
+  try {
+    contactlessTaskTabModule = await import(
+      /* webpackMode: "eager" */ `../../form-definitions/${version}/tabbedForms/ContactlessTaskTab.json`
+    );
+  } catch (err) {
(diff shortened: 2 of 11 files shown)


plugin-hrm-form/src/components/ContactDetails.tsx
```diff
@@ -76,7 +76,7 @@ const Details: React.FC<Props> = ({
   const childOrUnknown = formatName(childName);
   const childUpperCased = childOrUnknown.toUpperCase();
   const formattedChannel =
-    channel === 'default' ? mapChannelForInsights(details.contactlessTask.channel) : mapChannel(channel);
+    channel === 'default' ? mapChannelForInsights(details.contactlessTask.channel) : mapChannelForInsights(channel);
```
GPaoloni, line 79: I realized this difference while going thru the code. If you think this is not ok feel free to revert the commit (the last one is this change).

dee-luo:
@GPaoloni this change looks good to me! I have one request - can we rename custom channels to "offline channels"? I don't want to get this confused with our messaging custom channel support. 

@Humairaa127 should also review to confirm this will fulfill the helpline requests she's been getting

Humairaa127:
@GPaoloni yes this LGTM!

stephenhand (approved):

plugin-hrm-form/src/components/ContactDetails.tsx
murilovmachado: I feel like your change is correct. But I don't have much context on this part... Do we still need the `mapChannel` function? I confess I'm a little bit confused to both of them co-existing.

murilovmachado (approved):
Looks good to me! Both code and behavior look correct to me. I had just the same question as you about `mapChannel` and `mapChannelForInsights`, but it seems right.

PR #684, opened 2022-03-22
add warning dialog on close in Case details /tempInfo sections

mmythily:

## Description
This PR adds a warning dialog modal for content in the nested case details sections (+Note, +Perpetrators, etc.) triggered by changes by user

### Checklist
- [x] New tests added - older test changed
- [n/a] Strings are localized

### Related Issues
Fixes #CHI1054

### Verification steps
There will be a dialog warning modal at changes for content in the nested case details

### DIFF
--- a/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
+++ b/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
@@ -38,6 +38,7 @@ import { AppRoutesWithCase } from '../../states/routing/types';
 import useFocus from '../../utils/useFocus';
 import { recordingErrorHandler } from '../../fullStory';
 import { AddTemporaryCaseInfo, CaseUpdater, isAddTemporaryCaseInfo, TemporaryCaseInfo } from '../../states/case/types';
+import CloseCaseDialog from './CloseCaseDialog';
 
 type CaseItemPayload = { [key: string]: string | boolean };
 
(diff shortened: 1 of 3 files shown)


stephenhand (approved):
LGTM for the adding only case, but now the 'edit case section' functionality is merged, it will need updating it to work with that as well (see comment on `AddCaseEditItem.tsx`

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
```diff
@@ -87,6 +88,13 @@ const AddEditCaseItem: React.FC<Props> = ({
   const firstElementRef = useFocus();
 
   const { temporaryCaseInfo } = connectedCaseState;
+  const [tempInfoHasBeenEdited, setTempInfoHasBeenEdited] = React.useState(false);
```
stephenhand, line 91: I wonder if it might make more sense to add an `isEdited` flag to the `temporaryCaseInfo` object rather than having a separate top level state property? Both work but it might be neater to keep the temp info related state together?

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
```diff
@@ -87,6 +88,13 @@ const AddEditCaseItem: React.FC<Props> = ({
...
+  const [openDialog, setOpenDialog] = React.useState(false);
+  React.useEffect(() => {
+    if (temporaryCaseInfo.info !== null) {
```
stephenhand, line 94: Whilst this logic works for the Add case, it won't work for Edit. I would suggest you try hooking onto the `isDirty` state of the react form, but only to set the edited flag when true, NOT to unset is when false. 

This will mean the form stays dirty if the user switches tasks etc. It also means the form will stay 'edited' if the user sets the form back to it's starting state manually, but we decided as a team that that's acceptable.

You will have to unset the edited flag whenever temporaryCaseInfo is cleared, but if you make the flag part of the temporaryCaseInfo object itself as suggested in the other comment, this will happen automatically without you having to write any additional code

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
mmythily: Sounds good. I initially did this... similar functionality to `caseHasBeenEdited` for case, but realized they both have the same outcome. But, I will apply the state level changes in reducer and actions.

mmythily:
Closing this PR  - Suggestions are implemented in PR #690

PR #687, opened 2022-03-24
CHI-1041: Last Updated when & by whom on case sections

stephenhand:

## Description

* Saving a case section when editing now sets the updatedAt and updatedBy attributes of the case section to the actioning user and current time
* Viewing a case section that has been edited after it was created will show when the most recent edit was performed and by whom.
* Moved some helper functions to their own module to reduce dependencies (this is actually to fix a weird module loading issue in the CSAM reports tests, not sure why it started happening but moving these functions so you didn't need to pull in a whole clatter of irrelevant dependencies to use them fixed it).

### Checklist
- [ X ] New tests added
- [ X ] Strings are localized

### Related Issues
Fixes CHI-1041

### Verification steps

* Add a case section to a case list case
* View it (check 'Added' line is correct)
* Edit it
* Check 'Updated' line is correct.
* Repeat above for new case added with a contact


plugin-hrm-form/src/translations/pt-BR/flexUI.json
```diff
@@ -165,7 +165,8 @@
   "BottomBar-AddThisContactToExistingCase": "Adicionar este contato a um caso existente",
   "Case-Counsellor": "Atendente",
   "Case-CaseNumber": "Caso #",
-  "Case-ActionHeaderAdded": "Adicionado:",
+  "Case-ActionHeaderAdded": "Adicionado em {{date}} às {{time}} por {{counsellor}}",
+  "Case-ActionHeaderUpdated": "Atualizado em {{date}} às {{time}} por {{counsellor}}",
```
murilovmachado, line 169: Looks good to me!

murilovmachado (approved):
Looks great! Behaviour is also correct.

PR #689, opened 2022-03-24
CHI1054 POC/WIP PR isDirty redux modifications

mmythily:

## Description
This is just a POC / WIP PR
Scenario 1
- I made changes to redux to add tempInfoHasBeenEdited flag as false in updateTempInfo action
 - I am using isDirty API and when changes occur it updates tempInfoHasBeenEdited from false to true
 - Outcome:tempInfoHasBeenEdited is getting updated to true right away as soon as any use event occurs

Scenario 2
- I am using isDirty API and when changes occur it updates using useState and useEffect
 - Outcome: this is the same outcome as I have before where tempInfoEdited is indefinitely set to true once any other user event occurs

### Verification steps
- The first scenario is running in the current code. To check the second scenario, comment out line 193 and use line 194 for the click event.

### DIFF
--- a/plugin-hrm-form/src/___tests__/components/case/AddEditCaseItem.test.tsx
+++ b/plugin-hrm-form/src/___tests__/components/case/AddEditCaseItem.test.tsx
@@ -178,7 +178,7 @@ describe('Test AddHousehold', () => {
     expect(screen.getByTestId('Case-CloseButton')).toBeInTheDocument();
     screen.getByTestId('Case-CloseButton').click();
 
-    expect(screen.getByTestId('CloseCaseDialog')).toBeInTheDocument();
+    // expect(screen.getByTestId('CloseCaseDialog')).toBeInTheDocument();
   });
 
   test('a11y', async () => {
--- a/plugin-hrm-form/src/___tests__/states/case/actions.test.ts
+++ b/plugin-hrm-form/src/___tests__/states/case/actions.test.ts
@@ -62,6 +62,7 @@ describe('test action creators', () => {
       type: types.UPDATE_TEMP_INFO,
       value,
       taskId: task.taskSid,
+      tempInfoHasBeenEdited: undefined,
     };
 
     expect(actions.updateTempInfo(value, task.taskSid)).toStrictEqual(expectedAction);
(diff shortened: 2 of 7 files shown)


stephenhand:
I pulled this and took a look, it seems there are some internal behaviours around the 'Add' case causing problems. The the form state goes from an 'empty state' where all the input values are `null` to another 'empty state' where all the input values are empty strings, without any visible changes when the user first interacts with the form, which makes it 'dirty' even though nothing visible has changed.

TBH I wouldn't concern myself overly with the 'Add' scenario, this was just an afterthought for consistency and wasn't part of the original requirement, the important case to get right is the 'Edit' scenario.

The dialog always showing in the 'Add' scenario is justifiable to me, because adding an item at all, before you even change any information is an 'unsaved change', so being prompted about it is reasonable behaviour (although not prompting is also justifiable :-) )

Hopefully the 'Edit' scenario won't exhibit the same inconsistencies because the form is already populated when it loads up.

I would say merge the edit code and get that case behaving as expected and if the 'Add' scenario pops the dialog on dismissing an empty form, that's probably OK, as long as it isn't behaving completely bizarrely

@dee-luo does this sound reasonable?

In any event focus on the edit scenario first, that's the more important use case

mmythily:
Sounds good. I will work with the edit case scenario. I am wondering between the 3 of the approaches (1 previous PR and 2 in this), which one would be most feasible, in your opinion. 
Personally, I like the approach with isDirty API and no changes to redux. I say no redux changes because there there have been a lot of undesirable side effects in my experience and adds no value at the time.

stephenhand:
> Sounds good. I will work with the edit case scenario. I am wondering between the 3 of the approaches (1 previous PR and 2 in this), which one would be most feasible, in your opinion. Personally, I like the approach with isDirty API and no changes to redux. I say no redux changes because there there have been a lot of undesirable side effects in my experience and adds no value at the time.

We need to keep the state in Redux somewhere, because the isDirty flag resets every time you view a different task. You can watch the state of the flag, but you set your own flag when you see it set to 'true' but not set it back when isDirty is set to false, because this could be due to switching tasks.

I would recommend adding the property to the TemporaryCaseInfo object, rather than the connectedCaseState object like you did in this PR though, because the temporaryCaseInfo gets cleared when you close the AddEditCaseSection component, so you don't need to worry about resetting it

mmythily:
Closing this PR  - Suggestions are implemented in PR #690

PR #690, opened 2022-03-25
CHI1054 Warn Dialog on 'Add / Edit Case Section' form with unsaved changes

mmythily:

## Description
- This PR adds a warning dialog modal for 'Add / Edit Case Section' form when users try to cancel out of the nested case details sections (`+Note`, `+Perpetrators`, etc. for Adding; `View` for Editing) triggered by unsaved changes
- This is achieved by creating `isEdited` flag within `temporaryCaseInfo` object in `connectedCaseState`
- isDirty property of formState object from react hook form API was used to track modifications of any inputs. 
- Implemented code review based on PR #689 and #684

### Checklist
- [n/a] New tests added
- [n/a] Strings are localized

### Related Issues
Fixes #[1054](https://bugs.benetech.org/browse/CHI-1054)

### Verification steps
- Within case details, when adding or editing a new item (Note, Referral, Household Member, Perpetrator, Incident, Document), when any changes or additions to the fields are made, and attempting to exit without saving, a dialog modal is triggered prompting to either not save or save the changes
Warning Dialog Modal

### DIFF
--- a/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
+++ b/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
@@ -102,9 +102,10 @@ const AddEditCaseItem: React.FC<Props> = ({
 
   const [initialForm] = React.useState(getTemporaryFormContent(temporaryCaseInfo) ?? {}); // grab initial values in first render only. This value should never change or will ruin the memoization below
   const methods = useForm(reactHookFormOptions);
+  const [openDialog, setOpenDialog] = React.useState(false);
 
+  // The hook along with useEffect help render the DOM at first instance checking for user changes in the form. Without, atleast 2 more changes by the user are required when relying on methods.formState.isDirty directly
   const [isDirty, setDirty] = React.useState(false);
-
   React.useEffect(() => {
     setDirty(methods.formState.isDirty);
   }, [methods.formState.isDirty]);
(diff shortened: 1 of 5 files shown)


plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
```diff
@@ -102,6 +103,12 @@ const AddEditCaseItem: React.FC<Props> = ({
   const [initialForm] = React.useState(getTemporaryFormContent(temporaryCaseInfo) ?? {}); // grab initial values in first render only. This value should never change or will ruin the memoization below
   const methods = useForm(reactHookFormOptions);
 
+  const [isDirty, setDirty] = React.useState(false);
+
+  React.useEffect(() => {
```
stephenhand, line 108: Why do we need to track this with a hook rather than just using the property directly? Is it to work around some of the startup behaviours we see? Maybe put a comment explaining why because it seems redundant at first glance

stephenhand (approved):

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
mmythily: Yes, as you said, its a work around because the DOM requires 2 additional changes to track the changes without. I am leaving some comments. Thanks!

PR #691, opened 2022-03-28
All Cases Column Sorting

murilovmachado:

## Description
This PR changes the All Cases table to be sorted by the columns:
- Case#
- Opened
- Updated
- Follow Up

It does also changes fixes some visual changes to loading and the table headers.

### Checklist
- [ ] New tests added
- [x] Strings are localized

### Related Issues
Fixes https://bugs.benetech.org/browse/CHI-1116

### Verification steps
It needs the [backend PR](https://github.com/techmatters/hrm/pull/157) deployed or running locally in order to be tested.


plugin-hrm-form/src/components/caseList/CaseListTable.jsx
```diff
@@ -1,63 +0,0 @@
-import React from 'react';
```
stephenhand, line 1: Great work migrating these - is there any possibility we could try to get these to be renames in git, rather than delete & creates? Since most of the content is the same it would make reviewing this change a bit easier and preserve the file history going forward

plugin-hrm-form/src/types/types.ts
```diff
@@ -121,6 +121,14 @@ export type SearchCaseResult = {
   cases: Case[];
 };
 
+export type GetCasesParams = {
+  limit?: number;
+  offset?: number;
+  sortBy?: 'id' | 'createdAt' | 'updatedAt' | 'info.followUpDate';
+  order?: 'ASC' | 'DESC';
```
stephenhand, line 128: Minor nitpick, could we rename this property to something like 'sortDirection' - 'order' is a bit vague.

Also could we define these union sets as either a const object or an enum to reference in calling code, just so it's always obvious we are picking from a constrained range of values rather than using arbitrary strings?

plugin-hrm-form/src/components/caseList/CaseListTableHeadCell.tsx
```diff
@@ -0,0 +1,64 @@
...
+    );
+  };
+
+  const borderBottom = () => (sortBy === column ? '3px solid #009DFF' : 'none');
+
+  const handleClick = async () => {
+    if (!column) return;
+
+    const isDifferentColumn = column !== sortBy;
+    const updatedOrder = isDifferentColumn ? 'DESC' : changeOrder(order);
```
stephenhand, line 48: Could we set a default direction for each column? I think defaulting to a reverse sort for alphabetical columns could be confusing / annoying

stephenhand (approved):
Few minor things but LGTM

Should we consider throwing a temporary feature flag on this, since it won't work without an new version of HRM? I know it doesn't crash without a new HRM but the sorting indicators being there but having no effect isn't ideal

PR #705, opened 2022-04-08
ECPAT prod: change permissions

GPaoloni:

## Description
- Updates `zm` to be the permissions set used (updated via service config but updating here so they reflect the actual state).
- Small addition to the README on how to load the variables into a Linux terminal.


stephenhand (approved):
Thanks for the README improvements

PR #709, opened 2022-04-14
CHI1157: Case Information fields in Case Print are customizable

mmythily:
Primary reviewer:@stephenhand 

## Description
- Within Case Printing, refactored call for Case Information fields to be customized in form definitions with **highlightedAtCasePrinting** flag.
- Removed use of hardcoded definitions for 'Keep Confidential' and 'May social worker call'.

### Checklist
- [n/a] New tests added
- [n/a] Strings are localized

### Related Issues
- [Case Printing:Remove hardcoded fields](https://bugs.benetech.org/browse/CHI-1157)

### Verification steps
- Add `"highlightedAtCasePrinting": true` to any field within `hrm-form-definitions/.../caseInformationTab.json` to show the definition and value in Case Printing.

### DIFF
--- a/plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
+++ b/plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
@@ -30,28 +30,6 @@ type OwnProps = {
 };
 type Props = OwnProps;
 
-const extraFieldDefinitions = (strings: ReturnType<typeof getConfig>['strings']): FormDefinition => {
-  return [
-    {
-      name: 'keepConfidential',
-      label: strings['ContactDetails-GeneralDetails-KeepConfidential'],
-      type: 'checkbox',
-    },
-    {
-      name: 'okForCaseWorkerToCall',
-      label: strings['ContactDetails-GeneralDetails-OKToCall'],
-      type: 'mixed-checkbox',
-    },
-  ];
-};
-
-const addExtraValues = (caseInformation: ContactRawJson['caseInformation']) => {
-  return {
-    keepConfidential: Boolean(caseInformation?.keepConfidential),
-    okForCaseWorkerToCall: caseInformation?.okForCaseWorkerToCall,
-  };
-};
-
 const CasePrintView: React.FC<Props> = ({ onClickClose, caseDetails, definitionVersion, counselorsHash }) => {
   const { pdfImagesSource, strings } = getConfig();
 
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
```diff
@@ -102,6 +80,25 @@ const CasePrintView: React.FC<Props> = ({ onClickClose, caseDetails, definitionV
     loadImagesInMemory(imageSources);
   }, [logoSource, chkOnSource, chkOffSource]);
 
+  const caseInfoDefinitions = [...definitionVersion.tabbedForms.CaseInformationTab].filter(definition => {
```
stephenhand, line 83: Spreading these into a different array seems redundant, as `filter` produces a new array anyway (rather than copying the original)

plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
```diff
@@ -102,6 +80,25 @@ const CasePrintView: React.FC<Props> = ({ onClickClose, caseDetails, definitionV
...
+    // eslint-disable-next-line
+    return definition['highlightedAtCasePrint'] ? definition : null;
+  });
+
+  const addExtraValues = (caseInformation: ContactRawJson['caseInformation']) => {
```
stephenhand, line 88: Why is this function required? Presumably working around some sort of typing issue? Problem is that this function limits  the items that can potentially be put in the case section to an arbitrary subset of names. 

If a helpline customises it's CaseInformation tab, any items they add with a different name won't be put on that section of the print view, even if they set the flag. We really need it to work with whatever is in the caseInformation map, whatever it's named

plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
mmythily: Thanks Steve, I think I only need to perform this pattern of spreading when mutating state, which is not we are doing. It's a habit.

plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
mmythily: I was using the pattern left in the code base, but you are right. It works regardless.

stephenhand (approved):
LGTM! - @dee-luo do we need to specify anything in this PR for setting the new flag on existing helplines or will that be done in a separate PR?

dee-luo:
@stephenhand @mythilytm just let me and @Humairaa127 know when this is merged and we will set the flags properly!

PR #711, opened 2022-04-14
CHI-1165: Enabled custom form definitions for notes & referrals to be supported

stephenhand:
Primary reviewer: @GPaoloni 

## Description

* All data is now fed through from timeline activities to note & referral forms, not just a predefined set.
* Added an optional 'notes' section of the layout definition json, with a `previewFields` item where the notes fields that should be used in the timeline summary for notes can be specified, in line with what's in place for incidents. Referrals still uses hardcoded fields for this, things are a bit trickier there with the 2 dates etc. so decided to leave the summary for referrals alone for now.
* Updated relevant types to reflect that they are now generic forms.
* Added the `assessment` field as a preview field in the JM layout definition

Needs to be tested alongside https://github.com/techmatters/hrm/pull/173 to get the fix, but shouldn't be any worse than current versions against other versions of HRM

### Checklist
- [ X ] New tests added
- [ N/A ] Tested for chat contacts
- [ N/A ] Tested for call contacts

### Related Issues
CHI-1165

### Verification steps

* In a local version customise your NoteForm.json and add previewFields to a notes section of the `LayoutDefinitions.json`
* Check that the timeline, viewing, adding and editing notes works as expected - be aware that this will only work for new notes created after you changed the definition
* Repeat the above steps for referrals (changing `ReferralForm` this time).

### DIFF
--- a/hrm-form-definitions/form-definitions/jm-v1/LayoutDefinitions.json
+++ b/hrm-form-definitions/form-definitions/jm-v1/LayoutDefinitions.json
@@ -36,6 +36,9 @@
       },
       "documents": {
         "splitFormAt": 1
+      },
+      "notes": {
+        "previewFields": ["assessment"]
       }
     }
   }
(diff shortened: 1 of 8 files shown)


GPaoloni (approved):
Looks and works great! :rocket:

PR #721, opened 2022-04-26
CHI-1177 & CHI-1164 Fix Case Summary Edit & Refreshes Case List

stephenhand:

## Description

* Fixes CHI-1164 by refreshing the case list every time you close a case from it
* Also refreshes case search after closing a case
* Fixes CHI-1177 by removing some redundant redux state from Case.tsx 

### Checklist
- [ N/A ] New tests added
- [ N/A ] Tested for call contacts

### Related Issues
CHI-1177, CHI-1164

### Verification steps

See 'Steps To Reproduce' in associated tickets


plugin-hrm-form/src/components/case/caseDetails/CaseDetailsHeader.tsx
```diff
@@ -78,7 +78,6 @@ const CaseDetailsHeader: React.FC<OwnProps> = ({
                   onChange={handleClickChildIsAtRisk}
                   defaultChecked={Boolean(childIsAtRisk)}
                   disabled={!canEditChildAtRisk}
-                  autoFocus={canEditChildAtRisk}
```
murilovmachado, line 81: Here we want the autofocus: every time we open Case Screen (or any subscreens), we're making the first element on that screen focused by default. (It was a suggestion from the accessibility team)

murilovmachado (approved):
Everything looks good, just one minor thing on the autofocus (that's something we want)

plugin-hrm-form/src/components/case/caseDetails/CaseDetailsHeader.tsx
stephenhand: Putting it back doesn't seem to affect the fix, so done

PR #732, opened 2022-04-29
CHI 1121 Accessibility Enhancements for Case List

mmythily:
Primary reviewer:@GPaoloni  

## Description
- Removed 'expand' button and added a new styled component `CLCaseIDButton` based on button element for CaseId - keyboard buttons, tab and enter/space work
- Styling for CaseId column has been modified according to this [mock](https://marvelapp.com/prototype/6gcabah/screen/86290860)
    - Note that the sort arrow button  is not implemented. This requires more investigation
- Tested that table content is accessible using screen readers. 
- Added outline on case rows when hovered on mouse
- Adjusted the table column widths to accommodate the new changes. However, I did not change the order to avoid conflicts with upcoming changes to the Epic- see [mock](https://marvelapp.com/prototype/6gcabah/screen/86290860)

### Next Steps:
- Table headers are styled based on MUI components and require a larger refactor be accessible with a screen reader and keyboard. Refer to James Scholes's report and email on sorting. Namely, buttons need to be used for the sortable table headers and th elements should replace MUI's table components.
- Pagination is accessible by keyboard but there is no outline at focus on tab. I will address this as a different PR given there are some major styling refactors.
- Sort arrow button is not implemented as shown in mocks. This requires more investigation or just waiting until the Flex UI update. 

### Checklist
- [n/a] New tests added
- [x] Strings are localized

### Related Issues

### Verification steps
UI Changes to Case # column and other
- Use keyboard to navigate through case list page with tab and open a case with enter or space
- Use Screen Reader  to navigate through case list page and check that detailed information is also accessible
- Check with the updated [mocks](https://marvelapp.com/prototype/6gcabah/screen/86290860) to ensure implementation is identical.


plugin-hrm-form/src/components/caseList/CaseListTableHead.tsx
```diff
@@ -13,6 +13,7 @@ const CaseListTableHead = () => {
           column={ListCasesSortBy.ID}
           defaultSortDirection={ListCasesSortDirection.DESC}
           localizedText="CaseList-THCase"
+          aria-sort="descending"
```
nickhurlburt, line 16: This doesn't do anything -- there's no `aria-sort` prop on `CaseListTableHead`. The `defaultSortDirection` and `sortDirection` props will handle sort direction.

> Table headers are styled based on MUI components and require a larger refactor be accessible with a screen reader and keyboard

It may not require a larger refactor. According to [the MUI documentation](https://v3.mui.com/api/table-cell/), `TableCell` has a `sortDirection` prop that sets the `aria-sort` property. Possibly just passing that into `TableCell` could work. While we know MUI v3 doesn't have great accessibility support, it does have some, and we can at least give it a try.

At `CaseListTableHeadCell.tsx` Lines 77-79, I tried changing them to this:
```
    <CLTableCell style={{ width: width || '8%', cursor: cursor() }} scope="col" onClick={handleClick} sortDirection={sortDirection === ListCasesSortDirection.ASC ? 'asc' : 'desc'}>
      <CLTableHeaderFont style={{ borderBottom: borderBottom(), whiteSpace: 'nowrap' }}>
        <button><Template code={localizedText} /></button><span aria-hidden="true">{drawSort()}</span>
```

This results in rendered HTML like so:
image

It also makes the sortable headers reachable via tab. I believe this is more-or-less what we want, and what James had encouraged. There are two issues that I notice with this:
1. The 'button' isn't styled as we'd like
2. Because the `sortDirection` is a prop on the overall table, this results in every column having an `aria-sort` property. We'd need to edit the logic so that only the sorted column has an `aria-sort`.

But you could do a little work to fix that.

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -200,15 +200,17 @@
...
+  "CaseList-THCase": "Case",
+  "CaseList-THChildName": "Child",
   "CaseList-THSummary": "Summary",
   "CaseList-THCounselor": "Counsellor",
   "CaseList-THOpened": "Opened",
   "CaseList-THUpdated": "Updated",
   "CaseList-THFollowUp": "Follow Up",
   "CaseList-THCategory": "Category",
   "CaseList-ExpandButton": "Open case details",
+  "CaseList-StatusOpen": "Open",
```
nickhurlburt, line 212: We support other case statuses besides "Open" and "Closed" -- I think we want to get the status string directly from the case and display it, but @GPaoloni can confirm.

nickhurlburt (changes requested):
> Note that the sort arrow button is not implemented. This requires more investigation

Can we not just replace the use of [`ExpandMore`](https://mui.com/material-ui/material-icons/?query=expandmore&selected=ExpandMore) with [`ArrowDownward`](https://mui.com/material-ui/material-icons/?query=arrow&selected=ArrowDownward) at `CaseListTableHeadCell.tsx` Line 45? Is there something I'm missing?

> Added outline on case rows when hovered on mouse

Let's only keep this in if (1) a counselor can click on it to open the case view and (2) doing so does not harm accessibility otherwise

> Adjusted the table column widths to accommodate the new changes. However, I did not change the order to avoid conflicts with upcoming changes to the Epic- see [mock](https://marvelapp.com/prototype/6gcabah/screen/86290860)

Maybe there is some context I don't have, but why not change the order of the columns to match the mock? Is there another active story that it will conflict with?

Other issue:
When using a keyboard, after you select one of the links to see a case, focus is lost. That is, when the case detail is rendered there is no focused element. Focus should automatically be placed on the first focusable element in the case view. Talk with @murilovmachado on how to do this. Also, when returning from the case detail (after hitting 'Close'), focus should automatically be placed on the first focusable element in the case list.

mmythily:
This PR has outdated code and is well behind the master now. I have an updated PR here: https://github.com/techmatters/flex-plugins/pull/739

stephenhand:
@mmythily can we close this PR if it's never going to be merged (same for any others like that)?

PR #733, opened 2022-04-29
Pagination refactor to use nav and button & focus on tab key

mmythily:
Primary reviewer: @GPaoloni 

## Description
- Visual cue with a focus box has been added to ensure user can tab keyboard. Pagination component is updated to use semantic nav element to wrap the pagination UI and buttons instead of MUI components. 

### Checklist
- [n/a] New tests added
- [n/a] Strings are localized

### Related Issues

### Verification steps
- Use keyboard to tab and to navigate through pagination UI within the Case List page and Search Contacts/Cases page
- Use Screen Reader to navigate through pagination UI within the Case List page and Search Contacts/Cases page

### DIFF
--- a/plugin-hrm-form/src/components/Pagination.tsx
+++ b/plugin-hrm-form/src/components/Pagination.tsx
@@ -21,16 +21,19 @@ const renderPaginationButton = (page, handleChangePage) => n => {
   if (n === -1)
     return (
       <PaginationButton key={`ellipsis-${Math.random()}`}>
-        <ButtonText style={{ paddingTop: 10 }}>...</ButtonText>
+        <ButtonText>...</ButtonText>
       </PaginationButton>
     );
 
   return (
-    <ButtonBase key={`CaseList-pagination-${n}`} onClick={() => handleChangePage(n)}>
-      <PaginationButton highlight={page === n}>
-        <ButtonText highlight={page === n}>{n + 1}</ButtonText>
-      </PaginationButton>
-    </ButtonBase>
+    <PaginationButton
+      aria-label={`Page${n + 1}`}
+      highlight={page === n}
+      key={`CaseList-pagination-${n}`}
+      onClick={() => handleChangePage(n)}
+    >
+      <ButtonText highlight={page === n}>{n + 1}</ButtonText>
+    </PaginationButton>
   );
 };
 
@@ -43,16 +46,12 @@ type ChevronButtonProps = {
 const ChevronButton: React.FC<ChevronButtonProps> = ({ chevronDirection, onClick, templateCode }) => {
   const ChevronIcon = chevronDirection === 'left' ? ChevronLeft : ChevronRight;
   return (
-    <ButtonBase onClick={onClick}>
-      <PaginationChevron>
-        <HiddenText>
-          <Template code={templateCode} />
-        </HiddenText>
-        <ButtonText>
-          <ChevronIcon />
-        </ButtonText>
-      </PaginationChevron>
-    </ButtonBase>
+    <PaginationChevron onClick={onClick}>
+      <HiddenText>
+        <Template code={templateCode} />
+      </HiddenText>
+      <ChevronIcon style={{ margin: '-7px 0' }} />
+    </PaginationChevron>
   );
 };
 ChevronButton.displayName = 'ChevronButton';
@@ -76,17 +75,11 @@ const Pagination: React.FC<PaginationProps> = ({ page, pagesCount, handleChangeP
   };
 
   return (
-    <TableFooter data-testid="CaseList-TableFooter">
-      <PaginationRow transparent={transparent}>
-        <TableCell colSpan={8}>
-          <div style={{ display: 'flex', width: '100%', justifyContent: 'center' }}>
-            <ChevronButton chevronDirection="left" onClick={decreasePage} templateCode="CaseList-PrevPage" />
-            {getPaginationNumbers(page, pagesCount).map(renderButtons)}
-            <ChevronButton chevronDirection="right" onClick={increasePage} templateCode="CaseList-NextPage" />
-          </div>
-        </TableCell>
-      </PaginationRow>
-    </TableFooter>
+    <PaginationRow transparent={transparent} data-testid="CaseList-TableFooter">
+      <ChevronButton chevronDirection="left" onClick={decreasePage} templateCode="CaseList-PrevPage" />
+      {getPaginationNumbers(page, pagesCount).map(renderButtons)}
+      <ChevronButton chevronDirection="right" onClick={increasePage} templateCode="CaseList-NextPage" />
+    </PaginationRow>
   );
 };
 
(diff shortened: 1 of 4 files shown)


nickhurlburt:
@mmythily maybe this is specific to my browser, but it looks like there are some major UI issues to resolve?

image

plugin-hrm-form/src/components/Pagination.tsx
```diff
@@ -21,16 +21,19 @@ const renderPaginationButton = (page, handleChangePage) => n => {
...
     );
 
   return (
-    <ButtonBase key={`CaseList-pagination-${n}`} onClick={() => handleChangePage(n)}>
-      <PaginationButton highlight={page === n}>
-        <ButtonText highlight={page === n}>{n + 1}</ButtonText>
-      </PaginationButton>
-    </ButtonBase>
+    <PaginationButton
+      aria-label={`Page${n + 1}`}
```
nickhurlburt, line 30: Note there should probably be a space between Page and the number

plugin-hrm-form/src/components/Pagination.tsx
```diff
@@ -21,16 +21,19 @@ const renderPaginationButton = (page, handleChangePage) => n => {
   if (n === -1)
     return (
       <PaginationButton key={`ellipsis-${Math.random()}`}>
-        <ButtonText style={{ paddingTop: 10 }}>...</ButtonText>
+        <ButtonText>...</ButtonText>
```
nickhurlburt, line 24: Note that the "..." should not be focusable or reachable by tab, as is it not interactive
image

nickhurlburt: And actually, "previous page" shouldn't be active if we're on page 1, and "next page" shouldn't be active if we're on the final page.

plugin-hrm-form/src/styles/HrmStyles.tsx
```diff
@@ -525,11 +525,12 @@ type PaginationRowProps = {
   transparent?: boolean;
 };
 
-export const PaginationRow = styled(TableRow)<PaginationRowProps>`
+export const PaginationRow = styled('nav')<PaginationRowProps>`
```
nickhurlburt, line 528: We are probably supposed to identify what this nav is for screen readers, but I am not exactly sure how. About all I can come up with is to use (aria-labelledby)[https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#labeling_section_content] but that seems to require a header to be set.

I would say hunt around the internet for 15 minutes to see if we can label this somehow, and if you haven't found anything, then we'll follow up another day.

nickhurlburt (changes requested):
Looks good with a few changes to make non-interactive elements non-interactive.

plugin-hrm-form/src/components/Pagination.tsx
mmythily: Changes made as suggested

plugin-hrm-form/src/styles/HrmStyles.tsx
mmythily: From my research on this, it looks like semantic html elements like nav are already identified with a role. For instance, when using screen reader, at the first element focused within the navigation, the screen reader announces 'navigation'. For other non-semantic html elements such as div, role should be assigned to be identified by screen readers.

nickhurlburt: Yes, though when someone is using a screen reader, they won't know what type of navigation it is. We'd want some way to let the person know that this is a page selection navigation. I'm okay with leaving it for now and returning at a future point.

nickhurlburt (commented):
Good, now buttons are not actionable when they shouldn't be. Though there is a remaining accessibility bug: if you are on Page 2 and you select "Previous Page", or if you are on the penultimate page and you select "Next Page", then you lose focus. Focus should go to the first or last page number. As I believe you are already working on handling lost focus for the other PR, my guess is that it won't be a big lift to also do that for this one. If you are struggling with this, please reach out. I believe Murilo has some experience with this.

Also, I haven't really been looking at the code quality. @GPaoloni should still review this.

mmythily:
The new changes I made address the focus lost. They are persisting now.

plugin-hrm-form/src/styles/HrmStyles.tsx
```diff
@@ -525,11 +525,12 @@ type PaginationRowProps = {
...
+  display: flex;
+  justify-content: center;
   height: auto;
-  vertical-align: top;
   background-color: ${props => (props.transparent ? 'transparent' : props.theme.colors.base2)};
-  margin-top: -5;
+  margin: 3rem auto;
```
GPaoloni, line 533: I suggest using `px` as the unit for consistency with the rest of our styles, but nothing to worry about.

plugin-hrm-form/src/styles/caseList/index.ts
```diff
@@ -133,18 +133,23 @@ type PaginationButtonProps = {
   highlight?: Boolean;
 };
 
-export const PaginationButton = styled('div')<PaginationButtonProps>`
+export const PaginationButton = styled('button')<PaginationButtonProps>`
```
GPaoloni, line 136: Nice!

plugin-hrm-form/src/styles/caseList/index.ts
```diff
@@ -133,18 +133,23 @@ type PaginationButtonProps = {
...
   box-shadow: ${props => (props.highlight ? '0 1px 1px 0 rgba(0, 0, 0, 0.06)' : '0')};
   border-radius: 4px;
-  padding: 5px 10px;
-  margin: 5px 5px 0 5px;
+  padding: 6px 10px;
+  margin: 0 2px;
+  border: none;
+  &:focus {
+    outline: auto;
+    outline-color: #1976d2;
```
GPaoloni, line 145: I remember we agreed with Jana on having outlines black app-wide. Maybe they are different in the mocks, but for consistency I think we should have all of them black until we change all of them to something else.

GPaoloni (approved):
Looks good to me!

plugin-hrm-form/src/styles/caseList/index.ts
nickhurlburt: Ugh... looks like this is inconsistent:
image

image

@mmythily let's go with black for now, and we'll raise this with Jana later

nickhurlburt (approved):
With Gian's comments, looks good!

plugin-hrm-form/src/styles/caseList/index.ts
mmythily: Sounds good, Ill change to black for Case List components.

plugin-hrm-form/src/styles/HrmStyles.tsx
mmythily: That sounds good.

PR #738, opened 2022-05-04
Categories filter

murilovmachado:

## Description
This PR adds the categories filter to CaseList screen. (Front-end only)

image

image

image

### Checklist
- [ ] New tests added
- [x] Strings are localized

### Related Issues
Jira: https://bugs.benetech.org/browse/CHI-1119

### Verification steps
Categories filter should be all functional (front-end only). You can verify the payload sent will fill the categories values.


plugin-hrm-form/src/components/caseList/filters/Filters.tsx
```diff
@@ -59,6 +60,39 @@ const getInitialDateFilters = (): DateFilter[] => [
   },
 ];
 
+/**
```
stephenhand, line 63: I think the workaround is ok whilst we get guidance / a fix in react hook forms. My only question would be do we have to handle the name sanitising up in Filters.tsx? It would be nice if we could just black box this workaround away as an implementation detail of the CategoryFilter component if that were possible?

stephenhand (approved):
Ran it locally, looks good. Only one minor comment, otherwise code looks good too

plugin-hrm-form/src/components/caseList/filters/Filters.tsx
murilovmachado: Great suggestion, I've just done it! Thanks

PR #739, opened 2022-05-06
CHI 1121 Accessibility Enhancements for Case List 

mmythily:
Primary reviewer: @[GPaoloni](https://github.com/GPaoloni)

## Description
- Removed 'expand' button and added a new styled component CLCaseIDButton based on button element for CaseId - keyboard buttons, tab and enter/space work
- Styling for CaseId column has been modified according to this [mock](https://marvelapp.com/prototype/6gcabah/screen/86290860)
- Tested that table content is accessible using screen readers.
- Added outline on case rows when hovered on mouse and made clickable anywhere on row
- Rearranged and adjusted the table column widths to accommodate the new mocks. 
### Check Original [PR](https://github.com/techmatters/flex-plugins/pull/732) for these additional fixes:
- Case List Headers are accessible by keyboard and screen readers.
- Sort Arrow Button changed
- Added string parsing for case statuses
- Pending: focus on first case list element

### Checklist
- [n/a] New tests added
- [x] Strings are localized

### Related Issues

### Verification steps
- Use keyboard to navigate through case list page with tab and open a case with enter or space
- Use Screen Reader  to navigate through case list page and check that detailed information is also accessible
- Check with the updated [mocks](https://marvelapp.com/prototype/6gcabah/screen/86290860) to ensure implementation is identical.


plugin-hrm-form/src/styles/caseList/index.ts
```diff
@@ -43,6 +43,10 @@ export const CLTableRow = withStyles({
     borderColor: 'rgba(127, 134, 155, 0.07)',
     borderRadius: 4,
     boxShadow: '0 1px 1px 0 rgba(0, 0, 0, 0.06)',
+    '&:hover': {
+      outline: 'auto',
+      outlineColor: '#a0a8bd66',
```
GPaoloni, line 48: Same comment on the outline as prior PR.

plugin-hrm-form/src/styles/caseList/index.ts
```diff
@@ -104,12 +100,18 @@ export const CLTableHeaderFont = styled(FontOpenSans)`
...
+  background-color: transparent;
+`;
+CLHeaderSort.displayName = 'CLHeaderSort';
+
 type CLTableBodyFontProps = {
-  isOpenCase: boolean;
+  isOpenCase?: boolean;
 };
 
 export const CLTableBodyFont = styled(FontOpenSans)<CLTableBodyFontProps>`
```
GPaoloni, line 113: If this is not using `isOpenCase`, whoy don't just remove id? If the `CLTableBodyFontProps` ends up being empty you can just remove the type from here
`export const CLTableBodyFont = styled(FontOpenSans)`....`

plugin-hrm-form/src/styles/caseList/index.ts
```diff
@@ -118,17 +120,25 @@ export const CLTableBodyFont = styled(FontOpenSans)<CLTableBodyFontProps>`
 CLTableBodyFont.displayName = 'CLTableBodyFont';
 
 type CLCaseNumberContainerProps = {
-  isOpenCase: boolean;
+  isOpenCase?: boolean;
```
GPaoloni, line 123: Same as above.

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -200,15 +200,17 @@
   "CaseList-Cases": "Cases",
   "CaseList-Filters-CaseCount-Singular": "{{count}} case",
   "CaseList-Filters-CaseCount-Plural": "{{count}} cases",
-  "CaseList-THCase": "Case #",
-  "CaseList-THChildName": "Child's Name",
+  "CaseList-THCase": "Case",
```
GPaoloni, line 203: This is the string I referred as `"Case"` in a prior comment :P

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -200,15 +200,17 @@
...
+  "CaseList-THChildName": "Child",
   "CaseList-THSummary": "Summary",
   "CaseList-THCounselor": "Counsellor",
   "CaseList-THOpened": "Opened",
   "CaseList-THUpdated": "Updated",
   "CaseList-THFollowUp": "Follow Up",
   "CaseList-THCategory": "Category",
   "CaseList-ExpandButton": "Open case details",
+  "CaseList-StatusOpen": "Open",
+  "CaseList-StatusClosed": "Closed",
```
GPaoloni, line 213: Are this being used?

plugin-hrm-form/src/translations/pt-BR/flexUI.json
```diff
@@ -274,6 +274,8 @@
   "CaseList-THCategory": "Categoria",
   "NoCaseSummary": "Sem resumo do caso",
   "CaseList-ExpandButton": "Abrir detalhes do caso",
+  "CaseList-StatusOpen": "Aberto",
+  "CaseList-StatusClosed": "Fechado",
```
GPaoloni, line 278: Same question as above.

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -200,15 +200,17 @@
...
+  "CaseList-THCase": "Case",
+  "CaseList-THChildName": "Child",
   "CaseList-THSummary": "Summary",
   "CaseList-THCounselor": "Counsellor",
   "CaseList-THOpened": "Opened",
   "CaseList-THUpdated": "Updated",
   "CaseList-THFollowUp": "Follow Up",
   "CaseList-THCategory": "Category",
   "CaseList-ExpandButton": "Open case details",
+  "CaseList-StatusOpen": "Open",
```
nickhurlburt, line 212: We should be showing the actual string of the case status, not just open or closed. But arguably that's a feature change and not an accessibility thing, so I've added that to [CHI-1178](https://bugs.benetech.org/browse/CHI-1178) instead. So no need to address in this PR.

GPaoloni (approved):
Couple of comments but it looks & works really good, great one! The styles are on point and I'm loving the hover on the cases haha.

Also, thanks for moving more code to TS!

plugin-hrm-form/src/components/caseList/CaseListTableHeadCell.tsx
```diff
@@ -36,17 +36,18 @@ const CaseListTableHeadCell: React.FC<Props> = ({
...
-  const { featureFlags } = getConfig();
+  const { featureFlags, strings } = getConfig();
 
   const drawSort = () => {
     if (!featureFlags.enable_sort_cases) return null;
     if (!currentSort || !column || column !== currentSort.sortBy) return null;
 
     return (
-      <ExpandMore
+      <ArrowDownward
```
nickhurlburt, line 48: 👍

nickhurlburt (changes requested):
Several improvements but there are still some issues, especially around the table headers, which need to be resolved.

plugin-hrm-form/src/components/caseList/CaseListTableRow.tsx
```diff
@@ -41,30 +48,26 @@ const CaseListTableRow = ({ caseItem, counselorsHash, handleClickViewCase }) =>
   const isOpenCase = caseItem.status !== caseStatuses.closed;
 
   return (
-    <CLTableRow data-testid="CaseList-TableRow">
+    <CLTableRow data-testid="CaseList-TableRow" onClick={handleClickViewCase(caseItem)}>
       <CLNumberCell>
         <CLCaseNumberContainer isOpenCase={isOpenCase}>
-          <CLTableBodyFont isOpenCase={isOpenCase}>#{caseItem.id}</CLTableBodyFont>
+          <CLCaseIDButton aria-label={`Open Case ${caseItem.id}`} tabIndex={0} onClick={handleClickViewCase(caseItem)}>
```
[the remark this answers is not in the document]
mmythily: Ah, thanks for clarifying :)

plugin-hrm-form/src/styles/caseList/index.ts
mmythily: This particular one is for the the outline created when user hovers over the row with a mouse. In the mocks, it was this light grey....

nickhurlburt: Yes, this isn't a focus outline, so it's fine

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -126,7 +126,7 @@
   "Error-Form": "There is a problem with your submission.  Please check the form for errors.",
   "Error-CategoryRequired": "Required 1 category minimum, 3 categories maximum",
   "Error-Unexpected": "Unexpected error has occurred",
-  "Case-CaseNumber": "Case #",
+  "Case-CaseNumber": "Case",
```
nickhurlburt, line 129: I think this change has broken the case detail view:
image

That will need to be fixed

nickhurlburt (changes requested):

plugin-hrm-form/src/translations/en-US/flexUI.json
mmythily: I am using this like this:
` <Template code={`CaseList-Status${statusString}`} />`
I can remove it, but I am fairly certain this will work. For additional case statuses, the translations need to start with  **CaseList-Status**

mmythily: Also, I would have to remove Case Status unter the Case Id..
Case Statuses
.

nickhurlburt: I don't think that's necessary, though, because the form definitions already have the translated version. Checkout https://github.com/techmatters/flex-plugins/blob/master/hrm-form-definitions/form-definitions/br-v1/CaseStatus.json. We would just need to show the appropriate label.

While on the top, https://github.com/techmatters/flex-plugins/blob/master/hrm-form-definitions/form-definitions/za-v1/CaseStatus.json has an example of multiple case statuses.

nickhurlburt: I will note... there may be more complex reasons why this is hard in our existing system, and I may not fully understand those. But it seems redundant to me to need to maintain translations when we already have those strings in the form definitions.

mmythily:
I have worked on and resolved all of these points, except for these two, which need a closer look. 

https://github.com/techmatters/flex-plugins/pull/732#discussion_r862034714
I am not able to render a dynamic button that will focus and label the sort direction for screen readers for only the sortable headers. I end up transforming all  of the headers which is not desired. 

https://github.com/techmatters/flex-plugins/pull/739#discussion_r867226292
For this, I need some help figuring out the best practise for retrieving the form definition labels

PR #744, opened 2022-05-12
Changes on terraform scripts (default Flex Conversation Service) + Added UK Staging (III) config

GPaoloni:

## Description
This PR 
- Adds a small change on the import Twilio defaults script, to contemplate the possibility of default chat service used in Flex to be "Flex Conversation Service".
- Introduces a new variable to decide whether the helpline uses "Flex Conversation Service" or "Flex Chat Service" service for Flex. Defaults to "Flex Conversation Service".
- Introduces some terraform state for the test account I've created while working on this ticket.

Notes:
- RP Staging is kept in the original account. Changed the AWS Parameter Store to be the "Flex Conversation Service" for the CHAT_SERVICE variable, and triggered a Serverless re-deployment to update it.
- The Terraform changes where introduced in the "3rd account". The "2nd account" was manually edited to match the "Flex Chat Service" configuration but was not working. Today it worked seamlessly so it looks like the manual changes performed against it were successful (create "Flex Chat Service", roles, configurate Proxy, and a few more steps BEFORE running Terraform scripts). However, as I've said to @nickhurlburt, I consider going with "Flex Conversation Service" from now on to be a better, less error prone approach, as that is what we'll get from Twilio from now on. Happy to chat about the alternative tho.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1182)
- [N/A] New tests added
- [N/A] Strings are localized


twilio-iac/terraform-modules/flex/default/main.tf
```diff
@@ -8,7 +8,7 @@ terraform {
 }
 
 locals {
-  hrm_url = var.short_environment == "PROD" ? "https://hrm-production.tl.techmatters.org" : (var.short_environment == "STG" ? "https://hrm-test.tl.techmatters.org" : "https://hrm-development.tl.techmatters.org")
+  hrm_url = var.short_environment == "PROD" ? "https://hrm-production.tl.techmatters.org" : (var.short_environment == "STG" || var.short_environment == "STG-III"  ? "https://hrm-test.tl.techmatters.org" : "https://hrm-development.tl.techmatters.org")
```
GPaoloni, line 11: This is a dumb and ugly change, but I couldn't get around it using something like "contains substring". I'll revert before merging, but leaving to see if someone knows an alternative?

twilio-iac/terraform-modules/services/default/variables.tf
```diff
@@ -1,3 +1,9 @@
+variable "uses_conversations_service" {
+  default = true
+  type = bool
+  description = "Tells Terraform if the service being used is Flex Conversations Service"
+}
```
GPaoloni, line 5: Defaults to `true` as new accounts will always have this config, but leaving this to be variable dependent for backwards compatibility (Terraform tries to rename the "Flex Chat Service" otherwise).

twilio-iac/terraform-modules/flex/default/main.tf
stephenhand: We might have to move to passing the HRM url in as a variable anyway once we go multi-region - I think that would be preferable to a big nested set of conditions in a .tf file that try to work it out based on region / environment

scripts/src/terraformSupplemental/importDefaultTwilioResourcesToTerraform.ts
```diff
@@ -80,10 +80,19 @@ export async function importDefaultResources(
...
+    logWarning('Flex Chat Service not found to import');
+    logInfo('Trying to import Flex Conversation Service');
+    chatService = (await client.chat.services.list({ limit: 20 })).find(
+      (p) => p.friendlyName === 'Flex Conversation Service',
+    );
+  }
+
   if (chatService) {
     attemptTerraformImport(
       chatService.sid,
```
stephenhand, line 98: Might be nice to change the 'description' property below if we are using the conversations service. I know it's only for log output but it will help prevent confusion in the future if the name lines up

Also, as mentioned below - just want to check that the terraform resource type we are using here is correct?

twilio-iac/terraform-modules/services/default/main.tf
```diff
@@ -12,7 +12,7 @@ locals {
 }
 
 resource "twilio_chat_services_v2" "flex_chat_service" {
-  friendly_name = "Flex Chat Service"
+  friendly_name = var.uses_conversation_service == true ? "Flex Conversation Service" : "Flex Chat Service"
```
stephenhand, line 15: Should it still be a `twilio_chat_services_v2` resource type for conversations? Would it not be a  `twilio_conversations_services_v1`?

stephenhand (approved):
LGTM overall, just that query about the resource type really

PR #746, opened 2022-05-13
Chi 1167 redact counselor name on case pdf

acedeywin:
Primary reviewer: @stephenhand @nickhurlburt

### Checklist
- [ ] New tests added
- [ ] Strings are localized

### Related Issues
Fixes # N/A

### DIFF
--- a/plugin-hrm-form/src/components/case/casePrint/CasePrintDetails.tsx
+++ b/plugin-hrm-form/src/components/case/casePrint/CasePrintDetails.tsx
@@ -2,7 +2,7 @@
 /* eslint-disable react/prop-types */
 import React from 'react';
 import { Text, View, Image } from '@react-pdf/renderer';
-import { DefinitionVersionId } from 'hrm-form-definitions';
+import { DefinitionVersionId, DefinitionVersion } from 'hrm-form-definitions';
 
 import { getConfig } from '../../../HrmFormPlugin';
 import styles from './styles';
@@ -28,6 +28,7 @@ type OwnProps = {
   version: DefinitionVersionId;
   chkOnBlob?: string;
   chkOffBlob?: string;
+  definitionVersion: DefinitionVersion
 };
 
 type Props = OwnProps;
@@ -44,8 +45,11 @@ const CasePrintDetails: React.FC<Props> = ({
   version,
   chkOnBlob,
   chkOffBlob,
+  definitionVersion,
 }) => {
   const { strings } = getConfig();
+
+  const { hideCounselorDetails } = definitionVersion.layoutVersion.case;
 
   return (
     <View style={styles.caseDetailsContainer}>
@@ -75,6 +79,7 @@ const CasePrintDetails: React.FC<Props> = ({
           </View>
         </View>
       </View>
+      {hideCounselorDetails ? null : 
       <View style={styles.caseDetailsSubSection}>
         <View style={styles.caseCounsellorSection}>
           <View style={styles.flexColumn}>
@@ -91,7 +96,7 @@ const CasePrintDetails: React.FC<Props> = ({
         <View>
           <CasePrintCategories categories={categories} version={version} />
         </View>
-      </View>
+      </View>}
     </View>
   );
 };
--- a/plugin-hrm-form/src/components/case/casePrint/CasePrintDetails.tsx
+++ b/plugin-hrm-form/src/components/case/casePrint/CasePrintDetails.tsx
@@ -28,7 +28,7 @@ type OwnProps = {
   version: DefinitionVersionId;
   chkOnBlob?: string;
   chkOffBlob?: string;
-  definitionVersion: DefinitionVersion
+  definitionVersion: DefinitionVersion;
 };
 
 type Props = OwnProps;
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/HrmFormPlugin.tsx
```diff
@@ -24,12 +24,19 @@ export const getConfig = () => {
   const manager = Flex.Manager.getInstance();
 
   const hrmBaseUrl = `${manager.serviceConfiguration.attributes.hrm_base_url}/${manager.serviceConfiguration.attributes.hrm_api_version}/accounts/${manager.workerClient.accountSid}`;
-  const serverlessBaseUrl = manager.serviceConfiguration.attributes.serverless_base_url;
```
stephenhand, line 27: Is your linter set to max line width of 80? The project linter settings are 120 - can you try to get your IDE to use the project linter settings? Then it won't hassle you to do lots of reformats that aren't required

plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
```diff
@@ -62,6 +75,13 @@ const CasePrintView: React.FC<Props> = ({ onClickClose, caseDetails, definitionV
       },
     ];
 
+    console.log(
```
stephenhand, line 78: Remove before merging (I assume this was just for your own debugginr).

plugin-hrm-form/public/appConfig.example.js
```diff
@@ -1,20 +0,0 @@
-// your account sid
```
stephenhand, line 1: We should keep this! :-)

hrm-form-definitions/form-definitions/v1/LayoutDefinitions.json
```diff
@@ -37,6 +36,7 @@
     },
     "documents": {
       "splitFormAt": 1
-    }
+    },
+    "hideCounselorDetails": ["SafeSpot"]
```
nickhurlburt, line 40: Note that this file is actually the form definition for Childline Zambia, even though it's simply called `v1`.  (Zambia was our first live helpline.) SafeSpot's Aselo instance will never see this file. You will need to make a change in the `jm-v1` directory. 

Rather than list which helplines should be opted out, we would probably be better off if there's a redaction option for any field in the form definition, similar to how https://github.com/techmatters/flex-plugins/pull/709 is implemented. This would be a more forward-thinking solution that would allow more customization options for future helplines. Though I'll leave that up to a conversation between you and Steve.

nickhurlburt (commented):
Good to see the first PR of many to come!

For the `package-lock.json` files, we usually don't edit them unless we're making a `package.json` change. We can prevent edits to `package-lock.json` by running `npm ci` instead of `npm install`. This prevents us from having several different PRs around at the same time with slightly different dependencies.

hrm-form-definitions/form-definitions/v1/LayoutDefinitions.json
stephenhand: @nickhurlburt Unfortunately the counsellor information is hardcoded on the print view rather than coming from a form definition item, so it needs to be a separate flag

nickhurlburt: Ah, I see. So where should that configuration live? Service config? Flags?

acedeywin: @stephenhand Feedback has been implemented.

plugin-hrm-form/src/components/case/casePrint/CasePrintDetails.tsx
```diff
@@ -44,9 +45,21 @@ const CasePrintDetails: React.FC<Props> = ({
...
+  definitionVersion,
 }) => {
   const { strings } = getConfig();
 
+  const { helplines } = definitionVersion.helplineInformation;
+
+  const { hideCounselorDetails } = definitionVersion.layoutVersion.case;
+
+  const getHelplinesValues = helplines.map(
+    data =>
```
stephenhand, line 57: This looks like it doesn't take into account the case where 'hideCounselorDetails' is set to 'true'.  Just needs an ` || hideCounselorDetails === true` I think?

plugin-hrm-form/public/appConfig.example.js
stephenhand: @acedeywin Can we revert this file deletion - we keep this example to help people onboarding

stephenhand (commented):
Nearly there I think

Could you back out the remaining formatting changes in the PR? It makes it difficult to review the real changes with the noise of many reformats. Thanks.

plugin-hrm-form/src/components/case/casePrint/CasePrintDetails.tsx
acedeywin: It will be better to make `hideCounselorDetails` a boolean, where true will hide and vice versa. What do you think?

hrm-form-definitions/form-definitions/v1/LayoutDefinitions.json
stephenhand: @nickhurlburt - I think we're misunderstanding each other a little - the flag will still be in the form definitions and could be set by updating the appropriate file as usual, it just can't be scoped to a particular form item like in #709 - we need a standalone flag to toggle showing the counsellor details, and layoutDefinition.json seemed like the best place for that to go.

plugin-hrm-form/src/components/case/casePrint/CasePrintDetails.tsx
stephenhand: If you want to keep the ability to set redaction per helpline within an account, you can have the type be `boolean | string[] | undefined`

Or you can simplify things by having it be an optional boolean that sets the behaviour for all account helplines, since setting the option per account helpline was over and above the Acceptance Criteria on the card.

acedeywin:
@stephenhand I have done the formatting changes and updated the last feedback.

stephenhand:
So it looks like the PR build is failing because there are a couple of changes the linter requires in `plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx` after backing out the initial ones - VSCode should be able to apply these fixes automatically if it's using the eslint settings

As a final check, you can run `npm run lint` before you push, this runs the exact same check as the PR build so you know it will pass the PR check if that command reports no issues.

But the actual code changes all look good now, so good to merge once the build is green!

acedeywin:
@stephenhand feedback implemented and all jobs are running successfully. What's the next step?

nickhurlburt:
🎉  Looks great!

acedeywin:
@stephenhand @nickhurlburt Thank you.

PR #750, opened 2022-05-16
CHI-1161: Multi-AWS region terraform prep

stephenhand:
Primary reviewer: @nickhurlburt @GPaoloni 

## Description

* Allow HRM URL to be specified explicitly in TF configs. 
* Allow the AWS region for S3 buckets set in TF. 
* TerraformPOC example as a european deploy

### Checklist
- N/A New tests added
- N/A Strings are localized
- [-] Tested for chat contacts

### Related Issues
CHI-1161

### Verification steps

* Set up account with buckets & backend in a non standard region

### DIFF
--- a/twilio-iac/terraform-modules/aws/default/main.tf
+++ b/twilio-iac/terraform-modules/aws/default/main.tf
@@ -11,6 +11,11 @@ terraform {
   }
 }
 
+provider "aws" {
+  alias = "bucket"
+  region = var.bucket_region
+}
+
 
 locals {
   docs_s3_location = "tl-aselo-docs-${lower(var.short_helpline)}-${lower(var.environment)}"
@@ -19,6 +24,7 @@ locals {
 
 resource "aws_s3_bucket" "docs" {
   bucket = local.docs_s3_location
+  provider = aws.bucket
   cors_rule {
     allowed_headers = ["*"]
     allowed_methods = ["GET", "POST", "PUT"]
@@ -53,6 +59,7 @@ resource "aws_s3_bucket_public_access_block" "chat" {
   restrict_public_buckets = false
 }
 
+// Still put SSM parameters in default region, they don't store any helpline data & it keeps the workflow logic simpler
 resource "aws_ssm_parameter" "main_group" {
   for_each = {
     WORKSPACE_SID = jsonencode(["TWILIO", var.flex_task_assignment_workspace_sid, "Twilio account - Workspace SID"])
--- a/twilio-iac/terraform-modules/flex/default/main.tf
+++ b/twilio-iac/terraform-modules/flex/default/main.tf
@@ -8,7 +8,7 @@ terraform {
 }
 
 locals {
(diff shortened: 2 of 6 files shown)


twilio-iac/terraform-modules/aws/default/main.tf
```diff
@@ -19,6 +24,7 @@ locals {
 
 resource "aws_s3_bucket" "docs" {
   bucket = local.docs_s3_location
+  provider = aws.bucket
```
stephenhand, line 27: This does NOT migrate existing buckets automatically unfortunately. Manual steps outlined on CHI-1161 ticket

twilio-iac/terraform-modules/flex/default/main.tf
```diff
@@ -8,7 +8,7 @@ terraform {
 }
 
 locals {
-  hrm_url = var.short_environment == "PROD" ? "https://hrm-production.tl.techmatters.org" : (var.short_environment == "STG" ? "https://hrm-test.tl.techmatters.org" : "https://hrm-development.tl.techmatters.org")
+  hrm_url = var.hrm_url == "" ?  (var.short_environment == "PROD" ? "https://hrm-production.tl.techmatters.org" : (var.short_environment == "STG" ? "https://hrm-test.tl.techmatters.org" : "https://hrm-development.tl.techmatters.org")) : var.hrm_url
```
stephenhand, line 11: Going forward we probably want to give up on inferring the HRM URL completely and just pass it in all the time? Seems simpler

twilio-iac/terraform-modules/flex/default/variables.tf
```diff
@@ -2,17 +2,23 @@ variable "account_sid" {
   description = "Twilio Account SID"
   type        = string
 }
-
```
nickhurlburt, line 5: Was this whitespace change intended?

twilio-iac/terraform-modules/flex/default/variables.tf
```diff
@@ -2,17 +2,23 @@ variable "account_sid" {
...
+  type = string
+  default = ""
+}
+
 variable "short_environment" {
   description = "Short upper case environment identifier, typically 'PROD', 'STG' or 'DEV'"
   type        = string
 }
 
+
```
nickhurlburt, line 21: Was this whitespace change intended?

twilio-iac/terraform-modules/flex/default/main.tf
nickhurlburt: Probably, yes. Is there some way we could make it difficult to accidentally add a prod `hrm_url` to a non-prod instance?

nickhurlburt (approved):
Looks good, thanks!

twilio-iac/terraform-modules/flex/default/main.tf
stephenhand: Not sure, whatever we did might become harder to maintain as the number of instances increases - I think it would be hard to accidentally start actually using the wrong DB accidentally though, because you would also need the permissions configuration to be misaligned too (you would need the account permissions to be set up in the environment variables file in the S3 bucket for the wrong backend environment).

PR #755, opened 2022-05-19
CHI-1206: Fix loading up contacts into contact view

stephenhand:
Fix loading up contacts into contact view from a task form when the contact hasn't been saved.

Primary reviewer: @GPaoloni 

## Description

* Now loads contact from a task / form into the existing contacts store under a temporary ID before viewing it in the contact details view

### Checklist
- [ ] New tests added
- [ ] Strings are localized

### Related Issues
Fixes CHI-1206:

### Verification steps

1. Create an offline contact
2. Add it to a case
3. View the contact in the case timeline BEFORE saving

Also regression test all other scenarios where the contact details can be viewed

### DIFF
--- a/plugin-hrm-form/src/components/case/ViewContact.tsx
+++ b/plugin-hrm-form/src/components/case/ViewContact.tsx
@@ -11,22 +11,23 @@ import ContactDetails from '../contact/ContactDetails';
 import ActionHeader from './ActionHeader';
 import { CaseState } from '../../states/case/reducer';
 import type { CustomITask, StandaloneITask } from '../../types/types';
-import { loadRawContact, releaseContact } from '../../states/contacts/existingContacts';
+import { loadContact, loadRawContact, releaseContact } from '../../states/contacts/existingContacts';
 import { DetailsContext } from '../../states/contacts/contactDetails';
+import { taskFormToSearchContact } from '../../states/contacts/contactDetailsAdapter';
 
 const mapStateToProps = (state: RootState, ownProps: OwnProps) => {
   const form = state[namespace][contactFormsBase].tasks[ownProps.task.taskSid];
   const counselorsHash = state[namespace][configurationBase].counselors.hash;
   const caseState: CaseState = state[namespace][connectedCaseBase];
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/components/case/ViewContact.tsx
```diff
@@ -38,30 +39,43 @@ type OwnProps = {
...
       }
+      const temporaryId = `__unsavedFromCase:${connectedCase.id}`;
+      loadContactIntoState(taskFormToSearchContact(task, form, timeOfContact, counselor, temporaryId));
+      return () => releaseContactFromState(temporaryId);
     }
     return () => {
       /* no cleanup to do. */
     };
-  }, [tempInfo, counselorsHash, loadContactIntoState, releaseContactFromState]);
+  }, [
```
nickhurlburt, line 66: This has a "React Hook useEffect has a missing dependency: 'tempInfo'" warning... I don't know  `useEffect` well enough to know if that's a problem...

nickhurlburt (approved):
Looks good. I tested and can confirm that it fixes the issue (and I did a little more testing to see if it creates other issues, and I couldn't find any). I don't fully understand all that the code is done, but the general idea looks good.

plugin-hrm-form/src/components/case/ViewContact.tsx
stephenhand: It wouldn't matter practically because tempinfo is always the same on this screen, but I'll put it in to squash the warning

plugin-hrm-form/src/states/contacts/contactDetailsAdapter.ts
```diff
@@ -62,3 +65,34 @@ export const hrmServiceContactToSearchContact = (contact): SearchContact => {
     csamReports,
   };
 };
+
+export const taskFormToSearchContact = (task, form, date, counselor, temporaryId): SearchContact => {
```
GPaoloni, line 69: I'll be pleased the day we remove this representation, but great it's now contained in a helper.

GPaoloni (approved):
The code looks good to me, thanks for the fix!

PR #757, opened 2022-05-20
Chi 1193: Ukraine Staging Terraform

stephenhand:
Primary reviewer: @GPaoloni 

## Description

* Ukraine staging terraform config
* Minor tweaks to flex terraform config

### Related Issues
CHI-1193


twilio-iac/ukraine-staging/main.tf
```diff
@@ -0,0 +1,100 @@
...
+      source  = "twilio/twilio"
+      version = "0.11.1"
+    }
+  }
+
+  backend "s3" {
+    bucket         = "tl-terraform-state-twilio-ukr-staging"
+    key            = "twilio/terraform.tfstate"
+    dynamodb_table = "twilio-terraform-ukr-staging-locks"
+    region = "us-east-1"
```
GPaoloni, line 13: Is this set in `us`?

GPaoloni (approved):

twilio-iac/ukraine-staging/main.tf
stephenhand: Yes, I thought there was no harm in leaving the terraform state buckets all in the US - they contain no helpline data and aren't used in the running of the core system so their location has no performance implications. 

They could be moved to the same region as the helpline's backend in the future without difficulty if the need arises

PR #759, opened 2022-05-20
Fixed case updated time not showing when in same day as created at

GPaoloni:

## Description
This PR fixes a UI bug on Case screen, where "updated" date won't show if it was the same day as "opened".
Screenshot from 2022-05-20 14-42-38

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1201)
- [x] New tests added
- [N/A] Strings are localized

### Verification steps
- Create a new case.
- Update the case.
- Check that "Last Updated / Closed" is shown the current date (even if it's the same as "Opened").


stephenhand (approved):
LGTM

PR #760, opened 2022-05-20
Edit Contact permissions

mmythily:
Primary reviewer: @GPaoloni 

## Description
- Added a new edit contact permission action
- Added permission logic that will allow the user to edit contact based on two roles (isSupervisor and isOwner)
- Added the action for all helplines where only Supervisors can edit, except for IN and ZM, where owners can edit contact as well(as per @dee-luo's request)
- Added tests for testing permission logic for contacts

### Checklist
- [x] New tests added
- [n/a] Strings are localized

### Related Issues
Fixes [#1094](https://bugs.benetech.org/browse/CHI-1094)

### Verification steps
- In all helplines, only supervisor should be able to see the Edit button and edit.

### DIFF
--- a/plugin-hrm-form/src/___tests__/permissions/index.test.js
+++ b/plugin-hrm-form/src/___tests__/permissions/index.test.js
@@ -137,14 +137,10 @@ describe('Test different scenarios (random action) for Cases', () => {
 });
 
 describe('Test different scenarios (random action) for Contacts', () => {
-  const PermissionActionsValues = Object.values(PermissionActions);
-  const getRandomAction = () =>
-    PermissionActionsValues[Math.floor(Math.random() * 100) % PermissionActionsValues.length];
-
   each(
     [
       {
-        action: getRandomAction(),
+        action: 'editContact',
         conditionsSets: [['everyone']],
         workerSid: 'not owner',
         isSupervisor: false,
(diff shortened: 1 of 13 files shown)


mmythily:
It is the case that SearchContact is the current type definition for a Contact, and might need a refactor:
```
// Information about a single contact, as expected from search contacts endpoint (we might want to reuse this type in backend) - (is this a correct placement for this?)
export type SearchContact = {
  contactId: string;
  overview: {
    helpline: string;
    dateTime: string;
    name: string;
    customerNumber: string;
    callType: string;
    categories: {};
    counselor: string;
    notes: string;
    channel: string;
    conversationDuration: number;
    createdBy: string;
  };
  details: ContactRawJson;
  csamReports: CSAMReportEntry[];
};
```

GPaoloni:
@mythilytm Looks like we are sending the `twilioWorkerId` in the contact payload, but under the property `counselor` (see https://github.com/techmatters/hrm/blob/47e1936a2803b727b4dc2b813afbd2690e286585/hrm-service/src/contact/contact.ts#L149). Let's use that to decide who is the owner (and maybe leave a comment explaining what really is this field?).

GPaoloni (approved):
Looks good to me! I'd like to do some testing after the comment above is addressed tho.

dee-luo:
@GPaoloni @mmythily yes, isSupervisor only for all helplines, thanks! Humairaa and I will take care of helpline-specific changes at a later point.

GPaoloni (approved):
:rocket:

PR #762, opened 2022-05-23
Dependabot config to disable automatic PRs

johnhbenetech:

## Description
Example dependabot config that:
* limits to checking only 3 package directories
* reduces frequency to monthly checks
* ignores patch upgrades
* sets open PR limit to 0 essentially preventing PRs from getting opened until this is manually changed

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: "npm"
     directory: "/scripts/"
     schedule:
-      interval: "monthly"
+      interval: "weekly"
     open-pull-requests-limit: 0
     allow:
       - dependency-name: "*"
@@ -15,7 +15,7 @@ updates:
   - package-ecosystem: "npm"
     directory: "/plugin-hrm-form/"
     schedule:
-      interval: "monthly"
+      interval: "weekly"
     open-pull-requests-limit: 0
     allow:
       - dependency-name: "*"
@@ -27,7 +27,7 @@ updates:
   - package-ecosystem: "npm"
     directory: "/hrm-form-definitions/"
     schedule:
-      interval: "monthly"
+      interval: "weekly"
     open-pull-requests-limit: 0
     allow:
       - dependency-name: "*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
...
+updates:
+  - package-ecosystem: "npm"
+    directory: "/scripts/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 0
+    allow:
+      - dependency-name: "*"
+        dependency-type: "production"
+    ignore:
```
nickhurlburt, line 11: From [the docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore): "Security updates are unaffected by this setting." This is good, I would want patch updates for security to still be surfaced.

.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/scripts/"
+    schedule:
+      interval: "monthly"
```
nickhurlburt: We could do this weekly, I think.

nickhurlburt: (and for all the packages here)

.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/scripts/"
+    schedule:
+      interval: "weekly"
```
nickhurlburt, line 6: I changed this to weekly; for now, I intend to set a time to review these weekly, and eventually have a rotation on the team to do so.

nickhurlburt (approved):
Looks good to me, and I edited to check weekly. For now, I'm planning to review for critical dependencies every week.

nickhurlburt:
@johnhbenetech thank you, I'm going to go ahead and merge this. FYI @techmatters/aselo-reviewers this will eliminate Dependabot PRs for flex-plugins, though dependabot will still produce alerts in the 'Security' tab. For now, I will check these. If this goes well over the next week or two we can roll out to all repos.

PR #763, opened 2022-05-23
UI modification when No cases are available in Case List page

mmythily:

## Description
- In case of no cases / filtered down to zero cases to be shown in Case List, a table row with 'No Cases found.' is shown. As requested in https://marvelapp.com/prototype/df7eed1/screen/86858678 and https://marvelapp.com/prototype/df7eed1/screen/86858679
- Added a new string for No cases found as `CaseList-NoCases`
- Pagination component is hidden completely when no cases are shown

### Checklist
- [n/a] New tests added
- [x] Strings are localized

### Related Issues
Fixes # [1202](https://bugs.benetech.org/browse/CHI-1202) partially

### Verification steps
- Check the Case List page when filtered to zero cases in development
- Cross check string localization when using a different helpline

### DIFF
--- a/plugin-hrm-form/src/___tests__/components/caseList/CaseList.test.tsx
+++ b/plugin-hrm-form/src/___tests__/components/caseList/CaseList.test.tsx
@@ -112,8 +112,8 @@ test('Should dispatch fetchStarted and fetchSuccess actions if case lists return
   expect(screen.getByTestId('CaseList-Table')).toBeInTheDocument();
 
   expect(screen.getAllByTestId('CaseList-TableHead')).toHaveLength(1);
+  expect(screen.queryByTestId('CaseList-TableFooter')).not.toBeInTheDocument();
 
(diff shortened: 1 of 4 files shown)


mmythily:
In case of no cases found, I have also extended this to conditionally remove the Pagination component. https://marvelapp.com/prototype/df7eed1/screen/86858678
 As a result of this. a test is failing where the Footer is expected and not returned.
test('Should dispatch fetchStarted and fetchSuccess actions if case lists return', async () => {
...
  expect(screen.getAllByTestId('CaseList-TableFooter')).toHaveLength(1);
...
}
I am curious if the best practise is to keep this test/expectation and rework how Pagination is rendered, OR if I should update the test / add another one

mmythily:
I modified an existing test and added another to test when there are no cases to display

stephenhand (approved):
LGTM

PR #769, opened 2022-05-27
CHI-1200: all cases filter and sort

acedeywin:
Primary reviewer: @stephenhand @nickhurlburt 

## Note: 
Please, ignore the commit message. It was a very silly mistake. It's supposed to be CHI-1200: all cases filter and sort

## Description

- What this pull request does.
- Bugfix: 
- Custom date filters: Shifting Dates
- Custom date filters: Custom date display
- Fix search on a category filter 

### Checklist
- [N/A] Corresponding issue has been opened
- [N/A] New tests added
- [N/A] Strings are localized

### Related Issues
Fixes N/A

### Verification steps
- Note if migrations are required.
- For Custom date filters: Shifting Dates:
- Repro steps:
Select the Date Opened filter
Select Custom
Set End Date: 5/13/2022
Verify that does not Date shifts to 5/12/2022 automatically
This can be done for start date too

- For Custom date filters: Custom date display
- Replicate the above steps
- Select clear and select custom again, verify that the date fields populate for a user to fill out.

- For Fix search on a category filter 
- Search on a category filter
- Verify that the search start filtering after the user has typed three characters
Screenshot 2022-05-27 at 3 36 52 PM

Screenshot 2022-05-27 at 4 23 08 PM


plugin-hrm-form/src/components/caseList/filters/DateRangeFilter.tsx
```diff
@@ -80,10 +80,15 @@ const formToDateFilter = (
...
     return { option: values[selectedOptionField], exists: selected.exists };
   } else if (isFixedDateRange(selected)) {
+    const dateFrom = new Date(values.customDateRangeFrom);
+    const dateTo = new Date(values.customDateRangeTo);
+    const timeValue = 60 * 1000;
     return {
       option: values[selectedOptionField],
-      from: values.customDateRangeFrom ? new Date(values.customDateRangeFrom) : undefined,
-      to: values.customDateRangeTo ? endOfDay(new Date(values.customDateRangeTo)) : undefined,
+      from: values.customDateRangeFrom
```
stephenhand, line 88: This code works and fixes the issue - but I'd rather not resort to manually fixing up timezone offsets if it can possibly be avoided.

A little research shows that this bug stems from the fact that the behaviour of the built in JS `new Date(...)` constructor deviates from the ISO standard, insofar as it treats a time formatted 'YYYY-MM-DD' with no time info - (the format of the value we pass in from the form) as a UTC time, whereas the standard says it should be treated as a local time

On the other hand, the date-fns `parse` function implements the standard correctly and produces a local time. This fiddle demonstrates: https://jsfiddle.net/oeugx1qr/

If we were to simply replace the `new Date(...)` calls with date-fns `parse(...)` calls in the original code, it should fix it without us having to manually apply offsets.

This bug has a lesson for me to learn - never use the version of `new Date(...)` that takes a string, it has bugs!

stephenhand (commented):
Looks good, I've tested all the fixes locally and they look good. 

Would just like you to look at my comment on the timezone offset before you merge it

plugin-hrm-form/src/components/caseList/filters/DateRangeFilter.tsx
acedeywin: @stephenhand I saw the parse solution, but I decided to implement the current solution because it was well explained. I didn't see the shortcomings of the solution. I have gone ahead and implemented the parse solution. It still works fine as expected.

stephenhand: @acedeywin - As I said, your original code wasn't wrong, but in general, if a core API has an issue it's better practice to use a 3rd party library that already fixes an issue (especially if we already use the library elsewhere) rather than explicitly working around the issue for 2 reasons: 

1. The fix in the 3rd party library will be maintained, e.g. if the JS behaviour is ever changed to align with the ISO standard, date-fns parse will be updated to account for this, whereas we probably won't update our own fix until we start seeing a bug.

2. The best way to keep our code maintainable is to have as little of it as possible! Us reimplementing fixes already available elsewhere is just extra code for us to look after.

stephenhand (approved):
LGTM! Just need to fix the conflicts before merging - they look straightforward enough to resolve

acedeywin:
@stephenhand Merge conflict resolved and all tests are passing. PR is ready to be merged.

stephenhand:
If your PR has approvals, no unresolved issues and no conflicts, you can just go ahead and merge. Then deploy it and move your ticket into QA (let somebody on the product team know so they can test it).

PR #774, opened 2022-06-01
Terraform improvements + ECPAT Updates (Chatobts & Studio)

GPaoloni:

Wait for https://github.com/techmatters/flex-plugins/pull/767 to be merged before review.

NOTE: This PR might be much easier to review by commits. I know that it is a massive PR so please let me know if you want me to split it a bit more.

## Description
This PR 
- Updates terraform sumplemental (`twilioResources` script):
  - Adds `generate-chatbot-tf` command. This command, given a chatbot assistant sid (from a live Twilio account), generates the `.tf` file that matches the definition including assistant, tasks, task samples, field types, field values, stylesheet, default. Refer to the updated README for more information. NOTE: some things might be missing from an assistant definition (webhooks, querys), bear this in mind when generating from complex assistants.
  - Updates the `import-tf` to include tasks and task samples to be imported.
- Adds `twilio-iac/terraform-modules/chatbots/ecpat` folder, which contains the `.tf` files that matches the definitions of the 4 chatbots added to Ecpat staging account to be handled by terraform.
- Adds  `twilio-iac/terraform-modules/taskRouter/ecpat` folder that contains definitions to match the current state of the account. Relevant changes are:
  - Another task queue is introduced.
  - Master Workflow task routing is quite different that the default.
- Updates `twilio-iac/ecpat-staging/main.tf` file to match some of the updates in this account:
  - Added a new module `custom_chatbots` that contains the above mentioned chatbots.
  - Task Router module changed to be provided by the above mentioned one.
  - Studio flow updated.

~WIP:~
- [x] Factor out the Studio json definition to be reused.
- [x] Update other resources that were changed manually in the account (taskrouter).
  On this item, changing to a new `/taskRouter/ecpat` module was the easiest solution for me, as the task router config for this account includes a new task queue that is then used as part of the Master Workflow. If there's a way to rewrite some parts of `default` and `ecpat` configs for this module happy to take revisit, but I couldn't see it in this iteration.
- [x] Check why all of the assistants tasks are showing additions in the `actions` (all of them). Might need some help here (@stephenhand :sweat_smile:).
UPDATE (using twilio provider `v0.17.0`): on this issue, I've tried creating a new task within one of the chatbots from Twilio console, then imported it using plain `twilio import` command (no scripts involved). The behavior is the same: `terraform plan` shows that the `action` of the task resource will be added, even if the definition is the exact same as it is in Twilio console. I've also tried changing the `actions_url` property of the resource and it shows the difference as expected (not adding but modifying).
I've `terraform apply`ed on the task resource and the task actions remained unmodified, so it seems safe to do this to get the terraform state in sync with the actual Twilio state, but I don't want to go with this until we feel safe.
Does this seems like something worth opening an issue in the Twilio provider?
- [x] Rollback the backend (now pointing to local, correct is S3), and actually update the remote tf state (and maybe migrate my local state to avoid re-importing all of the resources).

NOTE: I have updated the remote terraform state for the account to include the imports (chatbots, task queue, etc), but have not `apply`ed anything against it yet. You put together a .tfvars file for this account (or ask me for it) and then
`➜ terraform plan -var-file ecpat-staging.private.tfvars` to see all the changes this will perform.
What to expect:
- "Changes" on every `modules.custom_chatbots.twilio_autopilot_assistants_tasks_v1` resource (it detectes that there were no `actions` for every task so it "creates them"). This is the issue that I've already discussed with you @stephenhand. I think is still safe to apply, but will open an issue to the provider.
- "Changes" on every `module.custom_chatbots.twilio_autopilot_assistants_v1` resource (it detects that the bots don't have `defaults` nor `stylesheet` configs). Same as above.
- A change on `module.survey.twilio_taskrouter_workspaces_workflows_v1.survey_workflow`: expression for the workflow filter changes since the helpline name changed.
- Another provider ugliness on `module.taskRouter.twilio_taskrouter_workspaces_workflows_v1.master_workflow` (json evaluates to the same but some spacing have changed, dunno where tho :) )
Screenshot from 2022-06-06 17-03-47

### Checklist
- [ ] New tests added
- [N/A] Strings are localized

### Verification steps
- Create `ecpat-staging.private.tfvars` file and fill it with proper values.
- Run `terraform plan -var-file ecpat-staging.private.tfvars`.
- Check that the changes are the expected ones, and help me be sure that we indeed want to apply this changes and have no fear of blowing things xD.
Screenshot from 2022-06-06 17-34-12


twilio-iac/kidshelpphone-staging/main.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     twilio = {
       source  = "twilio/twilio"
-      version = "0.11.1"
+      version = "0.17.0"
```
nickhurlburt, line 5: Should/can we refactor this out into a single variable across all helplines?

scripts/src/terraformSupplemental/generateChatbotResource.ts
```diff
@@ -0,0 +1,336 @@
...
+    .flat(Number.MAX_VALUE)
+    // Filter potential empty strings
+    .filter((s) => s)
+    // Converts into a single string with an empty line between each resource
+    .join('\n\n');
+
+  // Replace all the occurrences of serverless url for the terraform representation grabed from vars
+  const output = !serverlessUrl
+    ? fileString
+    : fileString.replace(serverlessUrl, '$'.concat('{var.serverless_url}'));
```
stephenhand, line 328: Why not just use a single quoted literal? i.e.  `'${var.serverless_url}'` - just to prevent potential confusion?

scripts/src/terraformSupplemental/generateChatbotResource.ts
```diff
@@ -0,0 +1,336 @@
...
+    `${referenceName}_${fieldType.uniqueName}`,
+    fieldTypeProperties,
+  );
+
+  const fieldTypeValuesResources = generateFieldTypeValuesTF(fieldType, referenceName);
+
+  return [fieldTypeResource, fieldTypeValuesResources];
+};
+
+const generateResourceTF =
```
stephenhand, line 283: I know you were talking about using DI, but I think what you've done here is fine. You are dealing with a fixed subset of resources and you need to support all of them, so the simplest and clearest code is to hardcode the relationships like you have done here

stephenhand (approved):
LGTM!

scripts/src/terraformSupplemental/generateChatbotResource.ts
GPaoloni: A lint rule complains because thinks that the `${}` should be in a templated string, but we actually want that to be passed to terraform in that form, so I thought this to be the clearest way to indicate that without silencing linter. Happy to change if makes more sense in other way tho.

PR #786, opened 2022-06-15
E2E test integration in Github CI 

mmythily:
I am hoping to help with direction on the which workflow to implement e2e test.

- I am curious if the first step here is to use a separate github workflow to test (e2e_workflow_yml) or to keep using plugin-hrm-ci workflow because that is where it will live( was my original assumption). Murilo mentioned it might be helpful to have this run only once the PR is merged to master, but he wanted to check with you and rest of the team. 

- What will also help - if I make a new e2e_workflow.yml, can I trigger on:[push] as well? Is this a bad practice to have 2 workflows on push?

(Note that I am running the e2e tests before the jest tests to save time only)

### DIFF
--- a/.github/workflows/plugin-hrm-form-ci.yml
+++ b/.github/workflows/plugin-hrm-form-ci.yml
@@ -41,11 +41,6 @@ jobs:
     - name: Run Lint
       run: npm run lint --if-present
       working-directory: ./plugin-hrm-form
-    - name: Run Tests
-      run: TWILIO_ACCOUNT_SID=${{secrets.E2E_DEV_ACCOUNT_SID}} TWILIO_AUTH_TOKEN=${{secrets.E2E_DEV_AUTH_TOKEN}} npm run test:jest
-      working-directory: ./plugin-hrm-form
-      env:
-        CI: true
     - name: Install e2e-test Packages
       run: npm ci
       working-directory: ./e2e-tests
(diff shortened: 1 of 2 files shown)


mmythily:
For my own notes... Murilo suggested this if I were to test with a separate workflow.
```
name: E2E Tests

on:
  push:
    branches:
    - master
   ......
   - name: Create appConfig.js
  run: cp ./public/appConfig.example.js ./public/appConfig.js
  working-directory: ./plugin-hrm-form
  shell: bash
- name: Replace Twilio account SID in appConfig.js
  uses: falnyr/replace-env-vars-action@master
  env:
    TWILIO_ACCOUNT_SID: ${{ secrets.E2E_ACCOUNT_SID}}
  with:
    filename: ./plugin-hrm-form/public/appConfig.js
## All the commands to build aselo
## Then the command to run playwrite tests
[...]
```

stephenhand:
It should definitely be a separate workflow, it's not bad practice at all to have more than one workflow on the same action. If you set it up as a second workflow, it will appear as a separate item on the 'checks' list at the bottom. 

This is what we want, because for the time being we don't want E2E test fails to block merges, we just want visibility of them until we are confident they are stable enough to block.

Even then, I think @murilovmachado is right, the E2E test should only block on merges to master, or perhaps even deploys - but that's out of scope for this story - just running as a separate push triggered workflow is all we need

stephenhand:
Also, we should set the concurrency for the E2E so it only runs one at a time. I think it would be tough to get these tests running reliably in parallel: https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/

stephenhand:
You were asking about the `appConfig.js` earlier? There's an example of writing one in a workflow in the main action for the deploy workflows: https://github.com/techmatters/flex-plugins/blob/5b8c9db28a85657a0954e39df34e84504bd5bc83/.github/actions/main-action/action.yml#L61

stephenhand:
Also, we should set the concurrency of the E2E workflow so it only ever runs one set of E2E tests at a time: https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/

And you should remove the `headless: false` line from the checked in `playwright.config.js` file - we want the browser to run headlessly by default, especially in CI

PR #788, opened 2022-06-16
Validate form definitions

GPaoloni:

## Description
This PR adds `validate.ts` file which contains code to test the validity of a form based on the specifications provided (thanks @stephenhand for sharing the code you authored for this).
Also adds a test file that will iterate over all the definitions forms and apply the validation functions.

As for now this is testing:
- JSONs are valid.
- `HouseholdForm`, `IncidentForm`, `NoteForm`, `PerpetratorForm`, `ReferralForm`, `DocumentForm`, `CallerInformationTab`, `CaseInformationTab`, `ChildInformationTab`, `callTypeButtons`, and `IssueCategorizationTab` (for all helplines) are all valid (required fields are present and `validator` functions are not throwing assertions).

Notes:
- `validator` functions are being unused for now.
- `br` definitions are failing the checks because they do not contain `firstName` and `lastName` fields for `HouseholdForm` and `PerpetratorForm`, but they are required according to https://benetech.app.box.com/notes/875135859260?s=vfukg4wxf5g9sj3hugt5qw70ijl0eezf. Should we allow them to not-be-required? Or fix the `br` definitions?
EDIT: As discussed with @dee-luo, we'll relax the requirements on `HouseholdForm` and `PerpetratorForm`.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1004)
- [x] New tests added
- [N/A] Strings are localized

### DIFF
--- a/hrm-form-definitions/package.json
+++ b/hrm-form-definitions/package.json
@@ -14,7 +14,9 @@
     "@babel/preset-env": "^7.16.5",
     "@babel/preset-typescript": "^7.16.5",
     "@types/jest": "^27.0.3",
+    "@types/lodash": "^4.14.182",
     "@types/node": "^17.0.8",
+    "@types/react": "^17.0.38",
     "@typescript-eslint/eslint-plugin": "^4.28.3",
     "@typescript-eslint/parser": "^4.28.3",
     "babel-core": "^6.26.3",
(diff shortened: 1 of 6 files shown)


hrm-form-definitions/package.json
```diff
@@ -14,7 +14,9 @@
     "@babel/preset-env": "^7.16.5",
     "@babel/preset-typescript": "^7.16.5",
     "@types/jest": "^27.0.3",
+    "@types/lodash": "^4.14.182",
     "@types/node": "^17.0.8",
+    "@types/react": "^17.0.38",
```
nickhurlburt, line 19: Does it matter if this is not the same version of react that our frontend is using?

stephenhand (approved):
LGTM, great to get this done!

hrm-form-definitions/package.json
stephenhand: Probably not, since it's just a dev dependency - but why is it needed? I couldn't see any extra imports in the PR that would need it? (same Q for the react-hook-form dep below)

GPaoloni: I think since this are just the types it won't really affect, but definitively something to check just in case.

PR #790, opened 2022-06-17
Chi1125 e2e workflow

stephenhand:
Primary reviewer:

### Checklist
- [ ] New tests added
- [ ] Strings are localized

### Related Issues
Fixes #....


stephenhand:
@mmythily - You nearly had the workflow working, here were the final tweaks I needed to do:

* The `npm run start` command to run the flex server needs to be run with an ampersand at the end, i.e. `npm run start &` - This is because `npm run start` runs in the terminal foreground, and doesn't return control to the user to run further commands, this causes the whole e2e workflow to stop dead at that command. The ampersand forces the command to run in the background so the workflow can move on to subsequent steps
* Add a `private/secret.js` to the source before building - the values weren't important but the file needed to be there or the module system barfs trying to run the plugin
* The plugin path in the appConfig.js needed adjusting slightly - the one we copied from the deploy job was only suitable for when the flex plugin is deployed to a Twilio account, not when we run it in the dev server like we do for the E2E tests

.github/workflows/plugin-hrm-form-e2e-test.yml
```diff
@@ -0,0 +1,87 @@
+name: End to End Testing CI
+
+# Action 
+on: [push]
+concurrency: e2e
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
```
murilovmachado, line 6: This workflow will be run on every push for every branch. Is that intended? Or de we want to limit it a little bit more (maybe run against push to master)?

e2e-tests/playwright.config.ts
```diff
@@ -8,6 +8,8 @@ const config: PlaywrightTestConfig = {
     storageState: 'temp/state.json',
     baseURL: process.env.PLAYWRIGHT_BASEURL ?? 'http://localhost:3000',
     permissions: ['microphone'],
+    screenshot: 'only-on-failure',
+    video: 'retain-on-failure',
```
murilovmachado, line 12: Nice that playwright has these configs!

murilovmachado (approved):
Looks good to me!

.github/workflows/plugin-hrm-form-e2e-test.yml
stephenhand: I thought we could try it this way for starters - if it causes issues we can always change the trigger.

It's nice to see the E2E result on the PR and we aren't saying every PR needs to pass E2E right now, until we have confidence in them, it's just advisory.

Just out of interest, if we run the tests only on merges to master, will it actually back out the merge if it fails? Or would we have to implement that ourselves?

stephenhand: @murilovmachado - I changed it to be a `pull_request` trigger, so you at least need to open a PR for the branch before it starts running E2E tests against it. We could possibly further refine it to only run on PRs targeting master, but I don't think that would cut down the number of runs very much anyway

murilovmachado: > Just out of interest, if we run the tests only on merges to master, will it actually back out the merge if it fails? Or would we have to implement that ourselves?

I think it doesn't back out the merge. Also, I'm not sure if we want to roll back immediately after E2E tests fail. I'd rather not to rollback on failing for now. What do you think?

PR #792, opened 2022-06-21
Update Flex UI Version Automatically on Deploy

murilovmachado:
Primary reviewer: @GPaoloni 

## Description
This PR adds one more step to the deployment flow:
- It now sets **Flex UI Version** on the deployed environment to be the same as in the `package-lock.json` file

### Verification steps
If needed, you can manually try this on development:
1. Manually downgrade UI Version in here: https://flex.twilio.com/admin/versioning/
2. Trigger deploy from this branch
3. Refresh the Flex Page and see the UI Version has changed


murilovmachado:
> I just wonder if this would not be simpler if it were implemented as one of our NPM scripts rather than a github action? Was there a particular reason that approach wouldn't work?

Yeah, I think moving it to be an NPM script will be easier to maintain. I'll make this change and test it before merging it.

plugin-hrm-form/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "plugin-hrm-form",
   "version": "0.0.0",
   "private": true,
+  "type": "commonjs",
```
GPaoloni, line 5: Not super aware of the implications of this but can't it affect other parts of the project?

plugin-hrm-form/src/scripts/updateFlexUiVersion.ts
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable camelcase */
+import { setFailed, info, error } from '@actions/core';
```
GPaoloni, line 2: Does this still sets the action to "failed" if it fails? That's great if so!

PR #804, opened 2022-06-29
CHI1180 Edit contact modal

mythilytm:
Primary reviewer: @GPaoloni 

## Description
- Changed the Edit Contact Sections to imitate a modal-like container. With a editingContact state, components are hidden based on when user edits a contact. 
- Added Edit <Section Name> and close button. As such, added these localized strings -  "Contact-EditCaller",
  "Contact-EditChild", "Contact-EditCategories",  "Contact-EditSummary"
- Removed duplicate headers in Contact view with in Case List page

### Checklist
- [n/a] New tests added
- [x] Strings are localized

### Related Issues
Fixes # [1180](https://bugs.benetech.org/browse/CHI-1180)

### Verification steps
- Check these four ways to get to a Edit Contact Section
1. From Contacts Queue view, search for a contact and attempt to edit a contact section
2. From Contacts Queue view, search for a case and attempt to edit a contact section
3. From Case List view, open a case, and the contact, and attempt to edit a contact section
4. From Search Contacts/Case, attempt to edit a contact section

### DIFF
--- a/plugin-hrm-form/src/components/contact/EditContactSection.tsx
+++ b/plugin-hrm-form/src/components/contact/EditContactSection.tsx
@@ -14,6 +14,7 @@ import { ContactDetailsRoute, DetailsContext, navigateContactDetails } from '../
 import { ContactDetailsSectionFormApi, IssueCategorizationSectionFormApi } from './contactDetailsSectionFormApi';
 import { refreshRawContact } from '../../states/contacts/existingContacts';
 import CloseCaseDialog from '../case/CloseCaseDialog';
+import * as t from '../../states/contacts/actions';
 
 type OwnProps = {
   context: DetailsContext;
@@ -33,6 +34,8 @@ const EditContactSection: React.FC<Props> = ({
   navigateForContext,
   refreshContact,
   contactDetailsSectionForm,
+  setEditContactPageOpen,
+  setEditContactPageClosed,
   children,
 }) => {
   const methods = useForm({
@@ -55,6 +58,10 @@ const EditContactSection: React.FC<Props> = ({
      * of adding any dependency inside the array
      */
     setInitialFormValues(methods.getValues());
+    setEditContactPageOpen();
+    return () => {
+      setEditContactPageClosed();
+    };
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
@@ -134,6 +141,8 @@ const EditContactSection: React.FC<Props> = ({
 const mapDispatchToProps = {
   navigateForContext: navigateContactDetails,
   refreshContact: refreshRawContact,
+  setEditContactPageOpen: t.setEditContactPageOpen,
+  setEditContactPageClosed: t.setEditContactPageClosed,
 };
 
 const mapStateToProps = (state: RootState, ownProps: OwnProps) => ({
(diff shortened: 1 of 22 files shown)


plugin-hrm-form/src/components/tabbedForms/TabbedForms.tsx
```diff
@@ -303,8 +312,9 @@ TabbedForms.displayName = 'TabbedForms';
 const mapStateToProps = (state: RootState, ownProps: OwnProps) => {
   const routing = state[namespace][routingBase].tasks[ownProps.task.taskSid];
   const contactForm = state[namespace][contactFormsBase].tasks[ownProps.task.taskSid];
+  const editContactFormOpen = state[namespace][contactFormsBase].editingContact;
```
nickhurlburt, line 315: @mythilytm I would imagine that the problem we talked about -- where the editing modal disappears if you leave the task and then return -- is due to this not being scoped to the task. That is, it does not match the two lines above it, which use a map keyed by `taskSid`.

I am not fully certain of that, but it'd be one place to check out.

mythilytm: I added a jira ticket to check on resolving this missign feature - [Persist changes to contacts on edits](https://bugs.benetech.org/browse/CHI-1302)

plugin-hrm-form/src/components/contact/EditContactSection.tsx
```diff
@@ -55,6 +63,10 @@ const EditContactSection: React.FC<Props> = ({
      * of adding any dependency inside the array
      */
     setInitialFormValues(methods.getValues());
+    setEditContactPageOpen();
+    return () => {
+      setEditContactPageClosed();
+    };
```
GPaoloni, line 69: One thing I've spotted is that if you are editing a contact tab, then go to another view and back to the contact, it closes the edit contact tab view. That behavior plus this "cleanup" (`setEditContactPageClosed` when this component unmounts) is likely preventing a different bug, where the "is editing contact" flag will stay true across multiple tasks because of it not being per-task-scoped. I haven't really digged into that scenario tho, so I might be wrong.

GPaoloni (approved):
Looks great! Small comments but the behavior is the expected to me.

plugin-hrm-form/src/styles/case/index.tsx
```diff
@@ -376,3 +376,13 @@ export const CloseDialogText = styled('p')`
 `;
 
 CloseDialogText.displayName = 'CloseDialogText';
+
+export const EditContactContainer = styled('div')`
```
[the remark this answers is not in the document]
mythilytm: Agreed on needing a reorganization. I am not sure where this should ideally live.

plugin-hrm-form/src/components/contact/EditContactSection.tsx
mmythily: setEditContactPageClosed() is required for cleanup because the hidden adjacent components need to be visible again. True that editing is not currently not scoped per task. I think it will be resolved in https://bugs.benetech.org/browse/CHI-1302

plugin-hrm-form/src/components/search/index.tsx
```diff
@@ -182,6 +189,7 @@ const Search: React.FC<Props> = props => {
   const { currentPage, currentContact, searchContactsResults, searchCasesResults, form, routing } = props;
 
   return (
+    // TODO: Needs converting to a div and the className={editContactFormOpen ? 'editingContact' : ''} adding, but that messes up the CSS
```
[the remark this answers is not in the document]
mythilytm: I ended up adding the class in a different component. Thanks for the catch on this comment

PR #807, opened 2022-06-29
Contactless E2E smoke test (with Case)

GPaoloni:

Note: this PR needs https://github.com/techmatters/hrm/pull/208 to be merged and deployed to work as intended (code can reviewed tho).
Update: above PR has been deployed and the tests are good now :heavy_check_mark: 

## Description
This PR
- Centralizes all the env vars in a single location where I'm adding dotenv configuration to consume from `.env` file (this is an easier workflow for me). This is all captured in the first commit so it will be super easy to rollback, let me know if I should.
- Adds `case.ts` helper code to interact with CaseHome screen.
- Adds `tests/offlineContact.spec.ts` test suit that
  - Creates an offline contact and add it to a new case.
  - Creates one note and one household entry in this new case.
  - Checks that the case is good in hrm after.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1129)
- [x] New tests added
- [n/a] Strings are localized


e2e-tests/tests/offlineContact.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+            'Content-Type': 'application/json',
+            Authorization: `Bearer ${token}`,
+          },
+        };
+
+        const response = await window.fetch(url, options);
+        return response.json();
+      },
+      [caseId],
+    );
```
GPaoloni, line 136: Here I'm getting away from adding `node-fetch` as a dep by using the browsers fetch. Happy to change if there's some concern or if it's hard to read.

e2e-tests/case.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  const { getNewCaseId } = selectors;
+
+  return {
+    getNewCaseId,
+    addCaseSection,
+    saveCaseAndEnd,
+  };
+};
```
mythilytm, line 61: Nice!!

e2e-tests/case.ts
```diff
@@ -0,0 +1,61 @@
...
+  sectionTypeId: 'note' | 'referral' | 'household' | 'perpetrator' | 'incident' | 'document';
+  items: T;
+};
+
+export const caseHome = (page: Page) => {
+  const caseHomeArea = page.locator('div.Twilio-CRMContainer');
+  const selectors = {
+    addSectionButton: (sectionTypeId: string) =>
+      caseHomeArea.locator(
+        `//button[@data-testid='Case-${
```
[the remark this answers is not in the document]
GPaoloni: I don't know I saw the other selectors written like this and just followed along xD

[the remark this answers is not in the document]
stephenhand: Playwright supports a lot of different styles of selectors for finding elements - `//` denotes an XPath style selector: https://playwright.dev/docs/selectors#xpath-selectors - there are probably other selector styles you can use to find the same element but I know XPath pretty well (because I'm old enough to have been writing code when XML was cool :-P )

e2e-tests/global-setup.ts
```diff
@@ -3,6 +3,20 @@ import { FullConfig } from '@playwright/test';
 import { differenceInMilliseconds } from 'date-fns';
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { oktaSsoLoginViaApi, oktaSsoLoginViaGui } from './okta/sso-login';
+import { config as dotenvConfig } from 'dotenv';
+dotenvConfig();
+
+export const environmentVariables = {
```
stephenhand, line 9: Could we move this out to a separate file? Fells a bit weird to be importing `global-setup.ts` into test code

e2e-tests/tests/offlineContact.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    console.log('Open a new offline contact');
+    const addOfflineContactButton = pluginPage.locator(
+      `//button[@data-fs-id='Task-AddOfflineContact-Button']`,
+    );
+    await addOfflineContactButton.waitFor({ state: 'visible' });
+    await expect(addOfflineContactButton).toContainText('Offline Contact Record');
+    await addOfflineContactButton.click();
+
+    // Wait for 3 seconds in case the counselors dropdown is not yet populated
+    await pluginPage.waitForTimeout(3000);
```
stephenhand, line 31: Can we try to wait for that the dropdown content to populate using a `waitFor()` rather than just waiting 3 seconds? We should use fixed delays as a last resort because E2E tests are slow anyway and a fixed wait here and a fixed wait there starts to add up.

Sometimes they are unavoidable but we should try our best :-)

e2e-tests/tests/offlineContact.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    });
+
+    await casePage.saveCaseAndEnd();
+
+    // Check if the case got properly saved in HRM
+    const resultCase = await pluginPage.evaluate(
+      async ([caseIdArg]) => {
+        const manager = (window as any).Twilio.Flex.Manager.getInstance();
+        const token = manager.user.token;
+        const hrmBaseUrl = `${manager.serviceConfiguration.attributes.hrm_base_url}/${manager.serviceConfiguration.attributes.hrm_api_version}/accounts/${manager.workerClient.accountSid}`;
```
stephenhand, line 121: Ha, nice! I would have tried to intercept HTTP calls and scraped the token from the headers, but this is much nicer :-)

stephenhand (approved):
Only minor points & qs - great stuff!

e2e-tests/tests/offlineContact.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+    ]);
+    console.log('Plugin page visited.');
+  });
+
+  test.afterAll(async () => {
+    await pluginPage?.close();
+  });
+
+  test('Offline Contact', async () => {
+    console.log('Open a new offline contact');
```
stephenhand, line 26: @GPaoloni - sorry I know I approved this, only spotted this once I was revieing Mythily's PR

When you get chance could you possibly refactor some of these contactless task Playwright interactions being done directly in the tests out to component object methods, like you did the case & form interactions?

I know they are fairly small but I elaborate on the reasoning behind factoring direct Playwright calls out of test methods here: https://github.com/techmatters/flex-plugins/pull/811#pullrequestreview-1028355770

PR #809, opened 2022-06-29
Added enabled flag to flex flows set to true

GPaoloni:
## Description
This PR adds `enabled = true` to the two workflows created on account setup.

### Verification steps
Check that an up-to-date account has no changes when `terraform plan`ing with this changes (Flex Flows should remain `enabled`).

Screenshot from 2022-06-29 16-47-47


murilovmachado (approved):
Looks good! :rocket:

mythilytm (approved):

PR #810, opened 2022-06-30
[Bug fix] Search date range not filtering accurately for Cases

GPaoloni:
Primary reviewer: @stephenhand might be overwhelmed by PRs? @murilovmachado

## Description
This PR fixes the bug described in the ticket linked below, where "the Search date range filter is not pulling the correct cases in the search result based on the filter applied". This is fixed in the "search thunk" action, by capturing the date ranges specified and bundling them in the search payload as they are expected by the search endpoint. I choose to do so in the thunk because this function is used in `src/components/PreviousContactsBanner.tsx` and `src/components/search/index.tsx`, but happy to capture this behavior at a component level if that feels more correct.

This PR supersedes https://github.com/techmatters/hrm/pull/207.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1203)
- [x] New tests added
- [n/a] Strings are localized

### Verification steps
- Start development server.
- Perform searches using date ranges and confirm that the results are as expected.
- Optional: confirm that "previous contacts search" still works as expected (previous contacts banner).

### DIFF
--- a/plugin-hrm-form/src/___tests__/states/search/actions.test.ts
+++ b/plugin-hrm-form/src/___tests__/states/search/actions.test.ts
@@ -131,6 +131,52 @@ describe('test action creators', () => {
     expect(dispatch).toBeCalledWith({ type: t.SEARCH_CASES_SUCCESS, taskId, searchResult });
   });
 
+  test('searchCases bundles dateFrom and dateTo under filters object if provided', async () => {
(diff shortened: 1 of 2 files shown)


stephenhand (approved):
LGTM and looks to work on my local tests

PR #811, opened 2022-07-01
CHI1130 E2E Case list smoke test

mythilytm:

Note: Add case section test is present, but due to a backend permission issue, I have commented this out to unblock this PR

## Description
- Adds `caseList.ts` utility module that interacts with the Case List page (including opening case and case section views)
- Adds `tests/caselist.spec.ts` test suite:
    - Filter by Status, Counselor, and Categories
    - Open the first case in the results
    - Open and close the print view
    - Add / Update Summary
    - Add a Case Section(had to be commented out temporarily)
    - Close the case

### Checklist
- [x] New tests added
- [n/a] Strings are localized

### Related Issues
Fixes # [1130](https://bugs.benetech.org/browse/CHI-1130)

### Verification steps
- Run `*env variables* npx playwright test tests/caselist.spec.ts --headed` in e2e-tests/


mythilytm:
The failing e2e test is possibly due to database permission for case related editing.  I was able to make changes to case with changes to permissions in flex-plugins, but PUT requests have been getting 401 errors.

stephenhand (commented):
Basic flow is good, great work adding more data-testids than trying to use more brittle selectors. Just would like to see the refactoring I suggested

Basically, very little of the actual Playwright API calls should be in the test itself, the test body should mostly be a series of calls to methods on these page & component objects, rather than direct Playwright calls (beforeXXX and afterXXX methods are a bit different, no big deal if they are doing lots of Playwright calls)

The idea is that a test method should almost read like an English description of the test logic, i.e. 'do this, then do that, check this, then do that, etc.) such that even a non coder could get a good idea of what the basic flow is without having to figure out what selectors are doing etc.

As with almost any 'rule' in this job, it's not absolute, there's always exceptions, but it's a good thing to aim for. Also, remember, you can put helper functions in the test files themselves, if you want to try and make a test body read more clearly but the helper logic is only relevant to the current test suite.

Also I spotted that @GPaoloni has a bit of direct playwright logic in his test methods so I'll bug him about that

Sorry to both of you - I didn't properly go through the specifics of the test writing style I was going for so you couldn't have known - I should have talked through all this up front instead of asking you to change stuff when you are at the stage of raising PRs :-(

I think it should mostly just be lifting & shifting code though, the basic logic you already have is good

e2e-tests/caseList.ts
```diff
@@ -0,0 +1,67 @@
...
+    caseCloseButton: caseListPage.locator(`//button[@data-testid='CaseHome-CloseButton']`),
+    updateCaseButton: caseListPage.locator(`//button[@data-testid='CaseHome-Update-Button']`),
+    //Case Section view
+    saveCaseItemButton: caseListPage.locator(
+      `//button[@data-testid='Case-AddEditItemScreen-SaveItem']`,
+    ),
+  };
+
+  //Filter
+  async function filterCases(filter: Filter, option: string) {
```
stephenhand, line 44: You might want to limit this method to only setting text based filters - perhaps taking a TextFilters union type that excludes the date based ones?

Sorry, this is probably a bit nitpicky for a WIP

e2e-tests/tests/webchat.spec.ts
```diff
@@ -114,5 +114,6 @@ test.describe.serial('Web chat caller', () => {
     ]);
     console.log('Saving form');
     await form.save();
+    await statusIndicator(pluginPage).setStatus(WorkerStatus.OFFLINE);
```
mythilytm, line 117: Subsequent tests and local development tends to lag when the worker status is left online after chat module is finished. I think this will help with the flow better. Curious about your thoughts.

e2e-tests/caseList.ts
mythilytm: This made it easier!

e2e-tests/.gitignore
```diff
@@ -1,3 +1,4 @@
 # File created during tests
 temp/
-.env
\ No newline at end of file
+.env
+test-results/
```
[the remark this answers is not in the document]
stephenhand: No it should only prevent the folder being checked in, it shouldn't affect it being written as part of a workflow

e2e-tests/tests/webchat.spec.ts
stephenhand: There's certainly no harm in leaving the worker in a consistent state.

Maybe it's worth doing this as an `after` step, or in a `finally` block to make sure it runs whether the test passes or not?

e2e-tests/tests/caselist.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    await page.openFirstCaseButton();
+
+    await page.viewClosePrintView();
+
+    await page.updateCaseSummary();
+
+    // commenting out the following add case section to bypass the test due to backend permission rule
+    /*
+    await page.addCaseSection({
```
stephenhand, line 40: Backend permissions have been aligned now, can you uncomment this and push? I'd like to see the test output

mmythily: Uncommenting and pushing. Still unable to save on my local.

nickhurlburt:
@mythilytm @stephenhand who has the next step on this? Just checking through the list of open PRs and checking for anything that might have been forgotten about

mythilytm:
@stephenhand After changes to the backend permissions set to open(everyone can write), this helped the case for aselo-alerts to edit and save again. We discussed this should be reverted back. Meanwhile, I am hoping you can review the tests and utility methods I wrote.

PR #829, opened 2022-07-07
CHI-1302: Edit contact state in Redux

stephenhand:
Primary reviewer: @mmythily 

Fixes the issue where the state of a contact halfway through being edited is lost when switching tasks

## Description

* Adds a 'draftContact' to the existing contacts redux state, which is a partial SearchContact holding the content being edited
* The existence of this draft and and which part of the contact is in the draft now determine which route should be shown, rather than a route.
* Reference counts for existing contacts are replaced with named references, this way edited states are held in redux rather than being released when the user switches tasks

### Checklist
- [X] New tests added
- N/A Strings are localized

### Related Issues
Fixes CHI-1302

### Verification steps

* Open a contact from the search within a contactless task, edit each section, saving normally (regression)
* Repeat the above but switch to another task partway through making the edits on the form and switch back - the form should be in exactly the state you left it
* Repeat the above but for a contact you are viewing from a case you searched for from a contactless task
* Repeat the above but for a contact you are viewing from a top level search
* Repeat the above but for a contact you are viewing from within a case you opened in a top level search
* Repeat the above but for a contact you are viewing from within a case opened from the case list

### DIFF
--- a/plugin-hrm-form/src/___tests__/components/case/ViewContact.test.tsx
+++ b/plugin-hrm-form/src/___tests__/components/case/ViewContact.test.tsx
@@ -154,8 +154,9 @@ describe('View Contact', () => {
           tasks: {},
           existingContacts: {
             'TEST ID': {
-              contact,
-              refCount: 1,
+              savedContact: contact,
+              references: new Set(['task-id']),
+              categories: { gridView: false, expanded: {} },
             },
           },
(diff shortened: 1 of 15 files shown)


mythilytm (approved):
The feature works just as desired. And the code quality is just next level.

PR #837, opened 2022-07-13
CHI-1322: Changes to a contact are now reflected on the version attached to a case

stephenhand:
Primary reviewer: @GPaoloni

A better for would be for the Case to save it's connected contacts to the existingContacts as soon as it loads and use that everywhere in the case, but that would be a bigger change with much higher potential for knock on bugs, so didn't seem appropriate when we were trying to get a new prod version out.

## Description

* The ViewContact.tsx page now sends an 'Update Case Contact' action when the viewed action changes
* This action replaces the contact in the parent case's `connectedContacts` collection with the updated version
* This required yet another 'Contact' -> 'Same Contact in a Slightly Different Format' conversion function. :sadpanda

### Checklist
- [X] New tests added
- N/A Strings are localized

### Related Issues
Fixes #....

### DIFF
--- a/plugin-hrm-form/src/___tests__/states/case/actions.test.ts
+++ b/plugin-hrm-form/src/___tests__/states/case/actions.test.ts
@@ -1,16 +1,21 @@
-import { Case } from '../../../types/types';
+import { Case, SearchContact } from '../../../types/types';
 import {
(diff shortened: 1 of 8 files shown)


stephenhand:
PR should have been raised against deploy branch, not master

PR #839, opened 2022-07-14
CHI-1106: Text transcript contact details

stephenhand:
Primary reviewer: @GPaoloni 

## Description

* Adds a link to transcripts on contact details where available
* Add reservation to HRM contact data so transcripts / recordings can be retrieved

### Checklist
- N/A New tests added
- [X] Strings are localized

### Related Issues
CHI-1106

### Verification steps

* Create a webchat contact with some chat content
* View the contact
* Open the transcript and ensure it's contents are what you expect


plugin-hrm-form/src/translations/es-ES/flexUI.json
```diff
@@ -419,6 +419,8 @@
     "ContactDetails-GeneralDetails-DateTime": "Fecha y Hora",
     "ContactDetails-GeneralDetails-KeepConfidential": "¿Mantener la confidencialidad?",
     "ContactDetails-GeneralDetails-OKToCall": "¿Puede llamar/SMS para el trabajador social?",
+    "ContactDetails-LoadRecording-Button": "Grabación abierta",
```
janorivera, line 422: "Abrir Grabación" will be better in this case :)

stephenhand: Updated, thanks!

plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
```diff
@@ -251,6 +289,9 @@ const mapStateToProps = (state: RootState, ownProps: OwnProps) => ({
   savedContact: state[namespace][contactFormsBase].existingContacts[ownProps.contactId]?.savedContact,
   draftContact: state[namespace][contactFormsBase].existingContacts[ownProps.contactId]?.draftContact,
   detailsExpanded: state[namespace][contactFormsBase].contactDetails[ownProps.context].detailsExpanded,
+  canViewTranscript: (state.flex.worker.attributes.roles as string[]).some(
+    role => role.toLowerCase().startsWith('wfo') && role !== 'wfo.',
+  ),
```
murilovmachado, line 294: By looking at [Twilio Documentation](https://www.twilio.com/docs/flex/admin-guide/setup/sso-configuration/insights-user-roles#understanding-flex-insights-roles), the role `wfo.quality_process_manager` doesn't have permission to `Play Calls`, which I assume it also cannot view transcripts (it can only create questionnaires).

Also, wouldn't it be more secure if we whitelist the roles that can view transcripts?

murilovmachado (approved):
Code looks good to me.

I thought this view transcript/call was just a POC, and the real feature would come after we implement all transcripts storage and API ourselves. I couldn't see a corresponding JIRA card either. So my question is: should we be merging this or just leave it as a hanging PR (POC)?

plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
stephenhand: Updated - this isn't a security feature, it's merely a quality of life feature to not show the button in cases where the transcript would not be available anyway. The security is all handled by Flex Insights.

stephenhand:
> Code looks good to me.
> 
> I thought this view transcript/call was just a POC, and the real feature would come after we implement all transcripts storage and API ourselves. I couldn't see a corresponding JIRA card either. So my question is: should we be merging this or just leave it as a hanging PR (POC)?

This was initially supposed to be a 'v1' of exposing transcripts and voice recordings, which would be around for at least few months before we implemented external transcript storage. Unfortunately it spent so long with Twilio that it might not be in place long before we implement the external transcript solution.

AFAIK I still think we want to implement this as a v1, so we have something in place in the event that external transcripts taker longer than planned or get deprioritised for some reason.

@dee-luo can confirm or correct the above

Either way, it's behind feature flags, so we can merge it whether we plan to enable it before we implement external storage or not.

PR #842, opened 2022-07-19
CHI-1155: Inconsistent Saving Behavior on Case

acedeywin:
Primary reviewer: @stephenhand @murilovmachado @GPaoloni 

## Description

- Feature Modification: Modify Editing Case Summary and fix inconsistent saving behavior

### Checklist
- [x] New tests added
- [x] Strings are localized

### Related Issues
Fixes CHI-1155

### Verification steps

- Click on a case list added by you or add offline contact and save it to case.
- Click on the Case Overview Edit button
- Update Case Summary, Follow up Date, Case Status, or Child is at risk option
- Save or Cancel

Screenshot 2022-08-02 at 6 29 17 PM
Screenshot 2022-08-02 at 6 29 26 PM


acedeywin:
@stephenhand @murilovmachado @GPaoloni @mmythily this PR is ready for review.

acedeywin:
@stephenhand @murilovmachado @GPaoloni @mythilytm The necessary tests have been added and this PR is ready for review.

stephenhand:
Not sure what the mock ups show (the link on the ticket took me to mocks for another feature), but it seems to me like the 'Update' button on the Case landing page will always be disabled and should be removed now?

hrm-form-definitions/form-definitions/br-v1/caseForms/CaseSummaryForm.json
```diff
@@ -0,0 +1,32 @@
+[
```
stephenhand, line 1: It seems pointless to have this defined as json. Having a json definition gives the impression that this is customisable, when in fact it isn't.

`AddEditCaseSummary.tsx` expects 4 fields with specific names, these are the only 4 fields that get populated from the original case data, and they are the only 4 that get saved back. If anyone were to add any other fields to this definition, they would not work.

Even if the UI could handle a dynamic form, we don't have a place on the backend data to save the additional fields - the case summary data structure is fixed

So having a definition for every helpline that must, in practice be exactly the same seems pointless and will only lead to confusion for the product team in the future.

We should remove these forms and do one of 2 things, either hardcode the form items directly in the AddEditCaseSummary.tsx form, or have a single file that defines the case summary form layout for all helplines. I think the first option seems more straightforward, but the second probably needs less change from what you currently have.

plugin-hrm-form/src/states/case/types.ts
```diff
@@ -171,6 +171,7 @@ export type CaseDetails = {
   referrals: t.ReferralEntry[];
   notes: NoteActivity[];
   documents: t.DocumentEntry[];
+  caseSummaries?: t.CaseSummaryEntry[];
```
stephenhand, line 174: This should not be an array or optional, cases will always have 1 case summary,
```suggestion
  caseSummary: t.CaseSummaryEntry;
```

plugin-hrm-form/src/types/types.ts
```diff
@@ -59,11 +63,13 @@ export type CaseInfo = {
   referrals?: ReferralEntry[];
   incidents?: IncidentEntry[];
   documents?: DocumentEntry[];
+  caseSummaries?: CaseSummaryEntry[];
```
stephenhand, line 66: Same as above, this should not be an array or optional, cases will always have 1 case summary,
```suggestion
  caseSummary: t.CaseSummaryEntry;
```

acedeywin:
@stephenhand I also thought of removing the 'Update' button but wasn't sure. So I left it for review to decide. I think it should be removed. It no longer has any use.

hrm-form-definitions/form-definitions/br-v1/caseForms/CaseSummaryForm.json
acedeywin: @stephenhand  I totally agree. I have removed the CaseSummary JSON files and have hard-coded the JSON within the EditCaseSummary file. It is cleaner and simple this way.

plugin-hrm-form/src/states/case/types.ts
acedeywin: @stephenhand I removed it. Turns out that the declaration is totally useless here!

plugin-hrm-form/src/types/types.ts
acedeywin: Done!

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+  const methods = useForm(reactHookFormOptions);
+  const [openDialog, setOpenDialog] = React.useState(false);
+
+  const { getValues } = methods;
+
+  const [l, r] = React.useMemo(() => {
+    const createUpdatedTemporaryFormContent = (payload: CaseItemPayload): EditTemporaryCaseInfo => {
+      const isEdited = !isEqual(initialForm, payload);
+      if (isEditTemporaryCaseInfo(temporaryCaseInfo)) {
```
stephenhand, line 147: Since EditTemporaryCaseInfo is the only valid option for this form, can we not simply validate that it is an EditTemporaryCaseInfo and throw an error if not, similar to how we do in ViewCaseItem?

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -0,0 +1,314 @@
...
+  return null;
+};
+
+export type EditCaseSummaryProps = {
+  task: CustomITask | StandaloneITask;
+  counselor: string;
+  definitionVersion: DefinitionVersion;
+  exitItem: () => void;
+  routing: AppRoutesWithCaseAndAction;
+  itemType: string;
```
stephenhand, line 67: There is no need to pass in an 'itemType' when there is only ever one, we can hardcode it in the component, same with several of these other injected parameters

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -0,0 +1,314 @@
...
+       * This way changes are reflected when you go back to the view after an edit
+       */
+      temporaryCaseInfo.info.updatedAt = now;
+      temporaryCaseInfo.info.updatedBy = workerSid;
+      info.summary = temporaryCaseInfo.info.form.caseSummary as string;
+      info.followUpDate = temporaryCaseInfo.info.form.date as string;
+      info.childIsAtRisk = temporaryCaseInfo.info.form.inImminentPhysicalDanger as boolean;
+      status = temporaryCaseInfo.info.form.caseStatus as string;
+
+      newInfo = applyTemporaryInfoToCase(
```
stephenhand, line 198: I don't think this logic needs to be injected for this case, there is only one save operation it does, it can be inline here

acedeywin: Done!

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
acedeywin: I tried and it didn't come out as expected. The code keeps breaking once I hit the `Save` button.

acedeywin: I tried it and the outcome wasn't buggy. The page breaks whenever I hit the `Save` button.

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
acedeywin: itemType can be hard-coded (done that) but the rest can't. They are either props from the parent component or props from state management.

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
acedeywin: Done.

acedeywin: @stephenhand  All feedback has been implemented.

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -0,0 +1,300 @@
...
+        isEdited,
+      };
+    };
+
+    const updateCallBack = () => {
+      const formValues = getValues();
+      updateTempInfo(createUpdatedTemporaryFormContent(formValues), task.taskSid);
+    };
+    const generatedForm = createFormFromDefinition(formDefinition)([])(initialForm, firstElementRef)(updateCallBack);
+    if (layout.splitFormAt) return splitAt(layout.splitFormAt)(disperseInputs(7)(generatedForm));
```
stephenhand, line 157: Can we simplify this and remove the 'layout' variable completely? 'splitFormAt' will always be '3'.

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -0,0 +1,300 @@
...
+  const { temporaryCaseInfo } = connectedCaseState;
+
+  if (!isEditTemporaryCaseInfo(temporaryCaseInfo)) {
+    return null;
+  }
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const firstElementRef = useFocus();
+
+  const layout = {
```
stephenhand, line 88: This is redundant now (see below).

acedeywin: Done!

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
acedeywin: Done!

stephenhand (approved):

stephenhand (commented):
I think one of the E2E tests edits a case summary, and it will still be expecting the old UI, so that would need updating to use the modified UI to get it green again

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -0,0 +1,281 @@
...
+      ],
+    },
+    {
+      name: 'date',
+      type: 'date-input',
+      label: 'Follow Up Date',
+    },
+    {
+      name: 'inImminentPhysicalDanger',
+      label: 'Child is in imminent physical danger',
```
nickhurlburt, line 105: `label: 'inImminentPhysicalDanger'` as a translation key?

nickhurlburt: or `label: 'Case-ChildIsAtRisk'`, see https://github.com/techmatters/flex-plugins/blob/master/plugin-hrm-form/src/translations/en-US/flexUI.json#L198

acedeywin: @nickhurlburt @stephenhand  This has been implemented and a PR raised for it https://github.com/techmatters/flex-plugins/pull/920

PR #847, opened 2022-07-22
Cache updated config when reducers run (for flex 2.0 upgrade).

stephenhand:
Primary reviewer: @GPaoloni 

## Description

This workaround subscribes to redux reducer cycles and updates a cached configuration object to a version generated using the same code `getConfig` currently uses to generate that object on demand. `getConfig()` now just returns that cached object.

I decided to do the workaround for the time being, because whilst this is technically 'pre-work', I'm concerned that the changes around flex redux between the 2 versions would mean there might be differences in the 'post getConfig()' API, and these differences would still need to be maintained for the lifetime of the fork, whereas this solution should work 'as is' for both versions

### Checklist
- N/A New tests added
- N/A Strings are localized

### Related Issues
Fixes CHI-1283

### Verification steps

General regression of Flex UI features

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.tsx
+++ b/plugin-hrm-form/src/HrmFormPlugin.tsx
@@ -68,7 +68,16 @@ const readConfig = () => {
   };
 };
 
-let cachedConfig = readConfig();
+let cachedConfig;
+
+try {
+  cachedConfig = readConfig();
+} catch (err) {
+  console.log(
+    'Failed to read config on page load, leaving undefined for now (it will be populated when the flex reducer runs)',
+    err,
+  );
+}
 
 export const getConfig = () => cachedConfig;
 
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/HrmFormPlugin.tsx
```diff
@@ -276,5 +289,16 @@ export default class HrmFormPlugin extends FlexPlugin {
     }
 
     manager.store.addReducer(namespace, reducers);
+    /*
+     * Direct use of 'subscribe' is generally discouraged.
+     * This is a workaround until we deprecate 'getConfig' in it's current form after we migrate to Flex 2.0
+     */
+    manager.store.subscribe(() => {
```
GPaoloni, line 296: Thanks for the comment, I'm fine with breaking the rules in this case, the benefits outweigh the risks I think haha. Do you have an approach in mind to replace this in the future already?

GPaoloni (approved):
Looks great to me!

plugin-hrm-form/src/HrmFormPlugin.tsx
stephenhand: I was thinking we remove the parameterless `getConfig` and replace it with one where passing in some state is required? Then components can map the state required and pass it in, and reducers just use the state as is. Any other consumers of getConfig would need the state passing in from whereverr they are called. When we introduce Sagas or Thunks we will use their idiomatic techniques for passing in the state. 

This would require a fair bit of refactoring so I didn't want to rush it in before Flex 2.0 - this way we could even do it incrementally if need be, although it would be nice to get it done and this workaround removed pretty quickly after moving to Flex 2.0 in master

I'm still a little unclear on exactly how we get the service configuration from the redux state, that's something I was also putting off until after we merge flex 2.0 to master :-P

PR #849, opened 2022-07-25
Chi 1316 childline zimbabwe staging

acedeywin:
Primary reviewer: @GPaoloni 

## Description

- New feature: Childline Zimbabwe Staging Setup

### Checklist
- [N/A] New tests added
- [N/A] Strings are localized

### Related Issues
Fixes CHI-1316


.github/workflows/zimbabwe_staging.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+      # Call main-action to compile and deploy
+      - name: Executing main-action
+        uses: ./.github/actions/main-action
+        with:
+          account-sid: ${{ secrets.Zw-v1_STG_ACCOUNT_SID }}
+          auth-token: ${{ secrets.Zw-v1_STG_AUTH_TOKEN }}
```
GPaoloni, line 33: This two are not really following the naming we use for this, as we don't need "versions", just the country code. Could you change that?
Example
Screenshot from 2022-07-25 17-24-11

twilio-iac/terraform-poc-account/main.tf
```diff
@@ -107,7 +107,6 @@ module aws_monitoring {
 }
 
 module github {
-
```
GPaoloni, line 110: Thanks, this one bugged me and forgot to change haha

.github/workflows/zimbabwe_staging.yml
GPaoloni: This secrets are created by the `github` terraform module, and after checking they are there, with the proper name, so you'll only need to change the name here in this workflow and reference the right one.
Screenshot from 2022-07-25 17-32-21

GPaoloni (approved):
Just one small comment, but other than that looks good to me! Approving as I don't think this needs a re-review after you do the change :)

twilio-iac/terraform-poc-account/main.tf
acedeywin: @GPaoloni That was a clumsy mistake. Fixed.

PR #861, opened 2022-07-29
CHI-1106: Fixed zero artefact at the bottom of contact view.

stephenhand:
Primary reviewer: @mythilytm 

## Description

* Ensured 'Open Recording / Transcript' toggle is coerced to boolean so extra zero never appears
* Added 'Invalid Row'…fallback on case list to make it more resilient to bad case data.
* Added defensive code to ensure cases with missing / invalid definition versions use the configured definition version as a fallback

### Checklist
- N/A New tests added
- N/A Strings are localized

### Related Issues
Fixes CHI-1106

### Verification steps

* Open offline contact details with voice type channel, ensure there is no zero at the bottom
* Repeat for an offline text type channel
* Ensure that where transcripts & recordings ARE available, they still display correctly


mythilytm (approved):
I checked between this and the previous branch for transcripts/recordings. Looks good!

PR #863, opened 2022-07-29
Added delete events to the taskrouter callback

GPaoloni:

## Description
Enables task deletion events to raise the taskrouter callback. Will impact on new helplines accordingly to the changes in https://github.com/techmatters/serverless/pull/278. 
For already existing accounts this change must be done manually.

I'm adding the ecpat config but haven applied the changes, is just to reflect the state once we do so (anyways I think there are still lots of things not in sync between the account and tf state).


stephenhand (approved):
LGTM

PR #864, opened 2022-08-09
chi-1336: fix child and caller info to be copied appropriately

acedeywin:

## Description

- Bug fix: Fix child and caller info to be copied appropriately

### Checklist
- [N/A] New tests added
- [x] Strings are localized

### Related Issues
Fixes CHI-1336

### Verification steps

1. Agent Desktop > Create offline contact > Search for Contacts and Cases
2. Search for a Counsellor
3. Select to View a contact (contact should have both child and caller)
4. Select Copy next to the Child section
5. Select Copy next to the Caller section
6. Say 'Yes' to the pop-up dialogue box
7. Verify that both the Caller and Child's tabs are populated appropriately 

Screenshot 2022-08-09 at 9 45 36 AM


murilovmachado (changes requested):
@acedeywin I've run it locally and everything works as expected.

But I think we should make the implementation different. I'll talk to you so we can jump on a call so we can discuss the details together. The main point is that I think we don't need this new action `CHECK_BUTTON_DATA` and the `buttonData` state.

PR #865, opened 2022-08-09
Added a script to safely patch service configurations with updated feature flags.

stephenhand:
Primary reviewer: @GPaoloni 

## Description

Adds a command to the `twilioResources` set of scripts that allows the caller to specify a set of feature flags without affecting any other feature flags or service configuration settings. 

### Checklist
- [X] New tests added
- N/A Strings are localized

### Verification steps

Use the script to update Aselo Development

### DIFF
--- a/scripts/src/terraformSupplemental/updateFlexServiceConfiguration.ts
+++ b/scripts/src/terraformSupplemental/updateFlexServiceConfiguration.ts
@@ -3,7 +3,7 @@ import { logDebug, logSuccess } from '../helpers/log';
 
 const TWILIO_FLEX_CONFIGURATION_ENDPOINT = 'https://flex-api.twilio.com/v1/Configuration';
 
-export default async function updateFlexServiceConfiguration(payload: unknown) {
+export async function updateFlexServiceConfiguration(payload: unknown) {
   logDebug('Patching flex configuration with:', payload);
   const response = await axios.post(TWILIO_FLEX_CONFIGURATION_ENDPOINT, payload, {
     auth: {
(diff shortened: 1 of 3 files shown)


scripts/src/terraformSupplemental/updateFlexServiceConfiguration.ts
```diff
@@ -22,3 +22,30 @@ export default async function updateFlexServiceConfiguration(payload: unknown) {
...
+      `Error response from Flex Service Configuration GET, HTTP Status ${response.statusText} (${
+        response.status
+      }). Response: ${response.data === 'object' ? JSON.stringify(response.data) : response.data}`,
+    );
+  }
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  const { account_sid, ui_attributes, attributes } = response.data;
+  await updateFlexServiceConfiguration({
+    account_sid,
+    ui_attributes,
```
murilovmachado, line 48: I think we don't need `ui_attributes` in here, right?

murilovmachado (approved):
Look good! Just one question on `ui_attributes`.

scripts/src/terraformSupplemental/updateFlexServiceConfiguration.ts
stephenhand: I was never too sure about that? I always post it back whenever I manually patch a config in Postman, but if you're sure it's redundant we can remove it.

Sorry, was a bit trigger happy with hitting 'Merge' before I saw this comment. Can address it next time I'm in this code though

PR #868, opened 2022-08-10
Updating dependencies and unit tests modifications for Flex 2.0 migration

mythilytm:

## Description
- Adds identity-obj-proxy dependency and update jest config to map and read css files
- Updates TabPressWrapper to remove use of deprecated 'RootRef' MUI component which is causing other tests to fail
- Removes zombie code / components and its tests and styling references, namely: Menu and MenuItem
- Updates Search (ContactDetails and Search) related unit tests to ensure they pass and use react-testing-library

### Checklist
- [x] New tests added
- [n/a] Strings are localized

### Related Issues
Fixes # [CHI-1286](https://bugs.benetech.org/browse/CHI-1286)
Bug Ticket pending for failing TabPressWrapper test [CHI-1364](https://bugs.benetech.org/browse/CHI-1364)

### Verification steps
- In `plugin-hrm-form`, install dependencies with `UNBUNDLED_REACT=true npm ci --legacy-peer-deps` 
- Run `npm run test:jest`


stephenhand (approved):
LGTM - if there's more tests to fix they can be done in follow up PRs, or added to this one - your call

mythilytm:
@stephenhand I am requesting review again. 
I thing you might speculate to my choice to remove Menu and MenuItem. I realize after the fact that we should have had a discussion on this. If this is not ideal, I am happy to revert the code to include these files. 

After a little bit of digging past code, I realized Menu was used at Case Save from main page
Menu and MenuItem

stephenhand:
> 

If you're absolutely sure these components are no longer used, absolutely, delete them & their tests. We can always get them back from git should we need them again

stephenhand (approved):

PR #869, opened 2022-08-11
CHI-1234: Copy Household

stephenhand:

## Description

* Adds a new 'copy-to' type of form element, with a 'savable' flag set to false to indicate it shouldn't be persisted to the form.
* Refactors section specific transformation / update / find operations into a separate API interface
* Added logic to copy a case item to another section when a 'copy-to' checkbox is selected

### Checklist
- [X] New tests added
- N/A Strings are localized

### Related Issues
Fixes CHI-1234

### Verification steps

In Aselo Development

1. Create a household
2. Populate all the fields
3. Tick the 'Also a perpetrator' box
4. Check a perpetrator is created with all the same details

Repeat the above steps WITHOUT checking the box, ensure nothing is created
Repeat the above steps but create a perpetrator and copy it to a household

We can also test edge cases not present in the baseline demo forms by either running flex locally and editing the demo definition, or by deploying it to staging environments where different perp / household member defintions exist


plugin-hrm-form/package-lock.json
```diff
@@ -3879,7 +3879,7 @@
     "@types/json5": {
       "version": "0.0.29",
       "resolved": "https://registry.npmjs.org/@types/json5/-/json5-0.0.29.tgz",
-      "integrity": "sha1-7ihweulOEdK4J7y+UnC86n8+ce4=",
+      "integrity": "sha512-dRLjCWHYg4oaA77cxO64oO+7JwCwnIzkZPdrrC71jQmQtlhM556pwKo5bUzqvZndkVbeFLIIi+9TC40JNF5hNQ==",
```
murilovmachado, line 3882: Why `package-lock.json` changed without changing `package-json`? This is because there were changes to `hrm-form-definitions` module?

hrm-form-definitions/form-definitions/demo-v1/caseForms/HouseholdForm.json
```diff
@@ -161,5 +161,12 @@
     "name": "ethnicity",
     "label": "Ethnicity",
     "type": "input"
+  },
+  {
+    "name": "copyToPerpetrator",
+    "label": "Also a perpetrator",
+    "type": "copy-to",
+    "target": "perpetrators",
+    "saved": false
```
murilovmachado, line 170: I suppose this tags the property to not be saved in the DB, right?

murilovmachado (approved):
I've run it locally and it worked correctly. Code is much cleaner now, thanks for putting extra effort into that.

plugin-hrm-form/package-lock.json
stephenhand: No, it was accidental (I use `npm install` rather than `npm ci` locally because it's faster., I just don't check in the updated `package-lock.json`, but sometimes I forget 😅 ), will back it out

PR #876, opened 2022-08-16
CO Post training updates

janorivera:
Primary Reviewer
@Humairaa127 

This is the latest round of updates based on feedback and meetings with the Te Guío team.


Humairaa127:
Did they also want Email and Other as offline contact channels?
Should the case manager be Juan Felipe or Angelica? As she holds the Te Guio email address and will lead the helpline.
The document form description is in English? Should it be Spanish?

PR #883, opened 2022-08-22
 CHI1364 Tab Press Component tests and behavior fix in Flex 2.0

mythilytm:

## Description
- Modifies the tests for TabPressWrapper HOC to use UserEvent API to mock for user actions instead of react-test-renderer. 
- Adds'@testing-library/jest-dom/extend-expect' import and a small change in config file to help jest read certain failing tests when using toHaveFocus function

### Checklist
- [x] New tests added (modified)
- [n/a] Strings are localized

### Related Issues
Fixes # [CHI-1364](https://bugs.benetech.org/browse/CHI-1364)

### DIFF
--- a/plugin-hrm-form/src/___tests__/TabPressWrapper.test.js
+++ b/plugin-hrm-form/src/___tests__/TabPressWrapper.test.js
@@ -1,5 +1,8 @@
 import React from 'react';
 import renderer from 'react-test-renderer';
+import { getByTestId, render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom/extend-expect';
 
 import './mockStyled';
 import TabPressWrapper from '../components/TabPressWrapper';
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/___tests__/TabPressWrapper.test.js
```diff
@@ -12,173 +15,211 @@ const getLastElement = component => component.getInstance().lastElementRef.curre
...
+  const firstElement = getByTestId(container, 'tabIndex2');
+  userEvent.tab();
+  expect(firstElement).toHaveFocus();
+
+  const secondElement = getByTestId(container, 'tabIndex2Again');
+  userEvent.tab();
+  expect(secondElement).toHaveFocus();
+
+  const lastElement = getByTestId(container, 'tabIndex7Again');
+  userEvent.tab();
```
stephenhand, line 215: Perhaps a bit easier to tell how many tabs if this was a for loop :-P

stephenhand (approved):
So no fixes were required in the code after all?

Also - not for this PR but might be good in a follow up - you could try to do the refactoring you were talking about - move the component to be a functional one, convert the component and test code to be typescript.

If you were refactoring it would also be nice to update these test names to describe what is actually being tested. Good test names should describe the initial scenario, what action is taken, and what the expected result. A name like '<TabPressWrapper> children with only one tabIndex' only has the 1st of these - as such people need to read the implementation code to know what it is actually trying to test

But, as I said, for another PR

plugin-hrm-form/src/___tests__/TabPressWrapper.test.js
mythilytm: > So no fixes were required in the code after all?

I made changes to the earlier PR removing use of Rootref. Currently it is working but sometimes does not work much like the random rendering of styling based on different components.

mythilytm:
> So no fixes were required in the code after all?

I made changes to the earlier PR removing use of Rootref. Currently it is working but sometimes does not work much like the random rendering of styling based on different components.

plugin-hrm-form/src/___tests__/TabPressWrapper.test.js
mythilytm: I modified to use count of focusable elements and test for the last element to use in for loop using the count

plugin-hrm-form/src/___tests__/TabPressWrapper.test.js
```diff
@@ -11,174 +14,214 @@ const getLastElement = component => component.getInstance().lastElementRef.curre
...
+  expect(secondElement).toHaveFocus();
+
+  const lastElement = getByTestId(container, 'tabIndex3');
+  userEvent.tab();
+  expect(lastElement).toHaveFocus();
 });
 
-test('<TabPressWrapper> children with tabIndexes: 1 and 3', () => {
-  const component = renderer.create(
+test('<TabPressWrapper> children with tabIndexes: 1 and 3. Tabbing through each element will cycle focus based on tabIndex', () => {
```
GPaoloni, line 105: I might be missing something but seems to me like this one is not really testing the "cycling" (one more tab goes to 1st index instead than losing focus)? Worth adding if not, this is a great one to have!

PR #911, opened 2022-08-30
CHI-1375: Fix case item view not updating

stephenhand:
This PR changes ViewCaseInfo.tsx to read the information it wants to display directly from the tasks `connectedCase` rather than passing a TemporaryCaseInfo.

At the moment it creates a TCI if the user clicks 'Edit' to pass to `AddEditCaseItem.tsx`

Question: It seems like `AddEditCaseItem.tsx` doesn't need to be passed a TCI info either really, it could just create one when initially opened and remove it when closed, would this work and be cleaner? I could do this as extra work in this PR or as a follow up if we think it's worth doing

Primary reviewer:

@GPaoloni 

## Description
* Refactor ViewCaseItem.tsx to look up data using ID rather than have it passed as TemporaryCaseInfo. 
* Other minor fixes / refactors

### Checklist
- [X] New tests added (existing ones updated)
- N/A Strings are localized

### Related Issues
Fixes CHI-1375

### Verification steps

See ticket, but also test switching tasks with edited forms to check dialog behaviour works as expected

### DIFF
--- a/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
+++ b/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
@@ -161,7 +161,7 @@ const AddEditCaseItem: React.FC<Props> = ({
 
     return splitInHalf(disperseInputs(7)(generatedForm));
   }, [
-    createUpdatedTemporaryFormContent,
+    temporaryCaseInfo,
     formDefinition,
     initialForm,
     firstElementRef,
--- a/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
+++ b/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
@@ -65,6 +65,28 @@ const getTemporaryFormContent = (temporaryCaseInfo: TemporaryCaseInfo): CaseItem
   return null;
 };
 
+const createUpdatedTemporaryFormContent = (
+  payload: CaseItemPayload,
+  initialForm: CaseItemPayload,
+  temporaryCaseInfo: TemporaryCaseInfo,
+): AddTemporaryCaseInfo | EditTemporaryCaseInfo => {
+  const isEdited = !isEqual(initialForm, payload);
+  if (isEditTemporaryCaseInfo(temporaryCaseInfo)) {
+    return {
+      ...temporaryCaseInfo,
+      info: { ...temporaryCaseInfo.info, form: payload },
+      isEdited: isEdited || temporaryCaseInfo.isEdited,
+    };
+  } else if (isAddTemporaryCaseInfo(temporaryCaseInfo)) {
+    return {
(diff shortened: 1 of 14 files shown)


plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
```diff
@@ -271,13 +272,6 @@ const AddEditCaseItem: React.FC<Props> = ({
             updatingCounsellor={updatingCounsellorName}
             updated={updated}
           />
-          <CloseCaseDialog
-            data-testid="CloseCase-Dialog"
-            openDialog={openDialog}
-            setDialog={() => setOpenDialog(false)}
-            handleDontSaveClose={close}
-            handleSaveUpdate={methods.handleSubmit(saveAndLeave, onError)}
-          />
```
murilovmachado, line 280: Were there two repeated `<CloseCaseDialog />` at this screen?

plugin-hrm-form/src/components/case/CaseHome.tsx
```diff
@@ -95,17 +91,13 @@ const CaseHome: React.FC<Props> = ({
...
   type CaseItemInfo<T extends TemporaryCaseInfo> = T['info']; // A bit redundant but looks cleaner than the anonymous subtype reference syntax
 
-  const onCaseItemActionClick = <T extends ViewTemporaryCaseInfo>(
-    targetSubroute: CaseSectionSubroute,
-    targetAction: T['action'],
-  ): ((entry: CaseItemInfo<T>) => void) => (entry: CaseItemInfo<T>) => {
-    updateTempInfo({ screen: targetSubroute, info: entry, action: targetAction }, task.taskSid);
-    changeRoute({ route, subroute: targetSubroute, action: targetAction } as AppRoutes, task.taskSid);
+  const onViewCaseItemClick = (targetSubroute: CaseSectionSubroute) => (id: string) => {
+    changeRoute({ route, subroute: targetSubroute, action: CaseItemAction.View, id } as AppRoutes, task.taskSid);
```
murilovmachado, line 100: Nice!

plugin-hrm-form/src/components/case/ViewCaseItem.tsx
```diff
@@ -43,29 +43,21 @@ const ViewCaseItem: React.FC<Props> = ({
   task,
   routing,
   counselorsHash,
-  temporaryCaseInfo,
```
murilovmachado, line 46: Great that we're not using `temporaryCaseInfo` here anymore

murilovmachado (approved):
Great PR, I just love it when we can reduce state! So good to see we are not using `temporaryCaseInfo` at `ViewCaseItem`.

Also, answering your question: yes, it would be very good if you can make `AddEditCaseItem` to not depend on `temporaryCaseInfo` as well.

PR #912, opened 2022-08-31
Alejandro terraform project 1

janorivera:
Primary reviewer:
@stephenhand 

## Description
Adding new channel modules  and updating Te Guío main.tf to match staging environment.

- [x] Added web channel
- [x] Added Messenger (Facebook) channel
- [x] Added Twitter channel
- [x] Added Instagram channel
- [x] Added WhatsApp channel
- [x] Added Line channel
- [x] Added flow templates 
- [x] Modified Te Guío main.tf
- [x] Tested provisioning via Terraform
 
If someone can take a look at the channel-attributes for each channel just to be sure that I have those ones correct. And also to check the filter for the Master Workflow since I modified it a little bit to include twilio_numbers.


stephenhand (approved):
This looks great!

twilio-iac/teguio-staging/variables.tf
```diff
@@ -77,4 +77,16 @@ variable "feature_flags" {
 
 variable "messaging_flow_contact_identity" {
   default = "+17752526377"
-}
\ No newline at end of file
+}
+
+variable "twitter_channel" {
```
stephenhand, line 82: I'm wondering why we need these variables, would we not simply remove the module from the TF config if we didn't want the channel any more? What's the scenario where we want to toggle their creation with a variable?

janorivera: @stephenhand

stephenhand: Sorry, I read this and forgot to reply.

I see what you are saying, although it bothers me a bit that we have a toggle that would only be used in initial rollout and then essentially be redundant. It also concerns me a little that somebody could enable / disable these channels without raising a PR. One of the benefits of adopting a configuration as code approach is that we get a 'paper trail' in the repo as to when things were changed. With them exposed as variables that can be toggled on & off via the CLI we won't get this

A compromise might be to have them as variables in the modules, but hardcode their values in the `main.tf` file? Then they can be configured in advance, and enabled with a very small change, but a PR still needs to be raised & reviewed?

(I appreciate right now people can just apply whatever changes they like from their local machine with no 'paper trail', but we hope to change this process in the future :-))

twilio-iac/terraform-modules/channels/instagram/channel-attributes.tftpl
```diff
@@ -0,0 +1,9 @@
+{
```
stephenhand, line 1: I wonder if (in the future, not this PR) there would be value in trying to make these channel modules a single generic module where the channel attributes and flow are generated externally and passed in? It would DRY things up considerably.

Or are there too many other special little tweaks required outside of these 2 variables required in individual channels to make this worth doing? I can see some create an SSM parameter and some don't - this could easily made optional in a generic module, but if there's lots of these little 'gotchas' it might be best to keep them separate as they are here.

stephenhand (approved):
LGTM - couple of comments but no blockers

twilio-iac/terraform-modules/channels/instagram/channel-attributes.tftpl
janorivera: yes! that would be a good improvement and I also thought of that while I was doing this, but I agree that it will make more sense to make it on a second iteration.

janorivera:
@stephenhand ok, I made a change, what do you think?

stephenhand (approved):
LGTM, great work @janorivera!

GPaoloni (commented):
This is awesome, great work @janorivera!! :fire: :rocket:

PR #919, opened 2022-09-05
CHI-1375: View Contact from case refactor

stephenhand:
Only review AFTER #911 is merged.

## Description

This refactor builds on PR #911 by removing the need for a `temporaryCaseInfo` object when viewing a contact.

* The cases existing contact, or the unsaved new contact, are now saved into redux under the `existingContacts` map as soon as they are loaded into `Case.tsx`
* The `existingContacts` reducers & actions have been updated to support loading & releaseing multiple contacts to support the above.
* `ViewContact.tsx` has now been simplified to only be passed an ID to load the contact from redux with
* Removed unused associated TemporaryCaseInfo types and updated the view contact route type to include an ID.
* Removed some dead code from `Case.tsx & `CaseHome.tsx` associated with previous case editing workflow

TODO: Fix broken tests, add new tests, general tidying

### Checklist
- [ ] New tests added
- N/A Strings are localized

### Related Issues
Related to CHI-1375

### Verification steps

Regression test cases functionality, especially switching tasks, case list / search / new case scenarios and viewing / updating contacts from a case.


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -103,46 +93,77 @@ const Case: React.FC<Props> = ({
...
+      return sortActivities(timelineActivities);
+    },
+    // savedContacts is not present but savedContactsJson is
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [newContact, savedContactsJson, task, form, props.connectedCaseId, props.connectedCaseState?.connectedCase],
+  );
 
-    getTimeline();
-  }, [task, form, props.connectedCaseId, props.connectedCaseState?.connectedCase, setLoading]);
+  useEffect(() => {
```
stephenhand, line 146: Would be good to handle this sort of thing in thunks or sagas in the future, populate the existing contacts cache on the service level, not up in the component

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -103,46 +93,77 @@ const Case: React.FC<Props> = ({
...
-        if (isStandaloneITask(task)) return;
-        const date = getDateFromNotSavedContact(task, form);
+  const [loadedContactIds, setLoadedContactIds] = useState([]);
+  const { connectedCase, prevStatus } = props?.connectedCaseState ?? {};
+  // This is to provide a stable dep for the useEffect that generates the timeline
+  const savedContactsJson = JSON.stringify(savedContacts);
+
+  const timeline: Activity[] = useMemo(
+    () => {
+      /**
```
stephenhand, line 112: Should we perhaps push 'activities' down to the redux state? Then they could be recalculated in reducers rather than up in components and wouldn't need to be passed around in CaseDetails prop drilling

plugin-hrm-form/src/states/contacts/existingContacts.ts
```diff
@@ -33,83 +33,102 @@ export const LOAD_CONTACT_ACTION = 'LOAD_CONTACT_ACTION';
...
+        },
+      };
+      return [
+        c.contactId,
+        {
+          ...currentContact,
+          savedContact: action.replaceExisting || !current.references.size ? c : state[c.contactId].savedContact,
+          references: action.reference ? current.references.add(action.reference) : current.references,
+        },
+      ];
```
murilovmachado, line 87: Just a naming question: what's `Kvps`? I've found it hard to guess even after reading the code.

plugin-hrm-form/src/states/routing/types.ts
```diff
@@ -19,87 +19,88 @@ export const NewCaseSectionSubroutes = {
...
 
 // Routes that may lead to Case screen (maybe we need an improvement here)
 export type AppRoutesWithCase =
   // TODO: enum the possible subroutes on each route
-  | AppRoutesWithCaseAndAction
-  | AppRoutesWithCaseActionAndId
+  | AddCaseSectionRoute
+  | EditCaseSectionRoute
+  | ViewCaseSectionRoute
+  | CaseViewContactRoute
```
murilovmachado, line 94: These new routes have much better naming, thanks.

murilovmachado (approved):
Looks good!

plugin-hrm-form/src/components/case/Case.tsx
murilovmachado: It seems like a good idea. Just make sure the reducer code doesn't get too complex, I tend to think they should be simple functions (but I may be wrong, not sure from where I got this feeling)

stephenhand: So as of 2020 redux are recommending putting as much logic as possible in reducers: https://redux.js.org/style-guide/#put-as-much-logic-as-possible-in-reducers. But obviously we want to compose complex logic together from simple functions as we would any other code

plugin-hrm-form/src/states/contacts/existingContacts.ts
stephenhand: Key-Value-Pairs - will rename

PR #923, opened 2022-09-06
Chi 1258 workflow to deploy flex helpline stg env

acedeywin:

## Description

- New feature: Workflow that can deploy Flex to all helpline staging environments

### Checklist
- [N/A] New tests added
- [x] Strings are localized

### Related Issues
Fixes CHI-1258


.github/workflows/deploy-helplines-parallel-worflows.yml
```diff
@@ -0,0 +1,211 @@
+# This is a basic workflow to deploy Flex to all helpline staging environments
+
+name: Deploy Flex to all Helpline Staging Environments
+
+on:
+  workflow_dispatch:
+    inputs:
+      confirm:
```
GPaoloni, line 8: It would be great to "stop the workflow" if the user does not types in the confirmation message, to prevent accidental runs, but maybe I'm being paranoid :P

.github/workflows/deploy-helplines-parallel-worflows.yml
```diff
@@ -0,0 +1,211 @@
...
+        description: This is dangerous! Are you sure you want to trigger this workflow? Type "deploy helplines workflow" to continue.
```
stephenhand, line 9: Do we really need this?

.github/workflows/deploy-helplines-parallel-worflows.yml
```diff
@@ -0,0 +1,211 @@
...
+        required: true
+      changelog:
+        description: Release Changelog - Add a comment about the changes to be included in this release.
+        required: true
+
+jobs:
+  deploy-canada_staging:
```
stephenhand, line 16: Could we use a `matrix` here to DRY up this code? We could make it static, but then we would have to update this workflow every time we add a new staging helpline.

Could we generate the matrix in a preceding step by reading all of the workflow files that end with `_staging.yml` using the technique described here: https://tomasvotruba.com/blog/2020/11/16/how-to-make-dynamic-matrix-in-github-actions/ - that way we wouldn't have an extra step to update this workflow as part of the new helpline setup?

It depends on whether you can use dynamic values in the `uses` property as to whether this is feasible - although for flex even if it isn't, you could still use a static matrix and the `custom_helpline.yml` workflow (but the same approach wouldn't work for serverless)

stephenhand: Hrmm, looks like expressions can't be used in workflows after all, that's disappointing - it still might be nice to DRY this up using a matrix of parameters to pass to the `custom_helpline.yml` rather than using the individual helpline deploy .yml files - this would also mean you only have to add the `workflow_call` event to a single helpline.

The matrix would have to be specified statically though, because we don't have a way to figure out the short codes we need dynamically

GPaoloni: For the "dynamic approach", we could trigger the workflows using `curl`/github cli + a bash script. It will be uglier from a UI perspective tho, since this approach allows us to "re-reun failed jobs" and we have more visual insights on the status of the whole deployment. But yeah, maybe the benefit of having a workflow that does not requires maintenance outweigh the cons.

stephenhand: I think just DRYing up the workflow based approach would best for now - I think the build information we would lose wouldn't be worth saving a step in the helpline setup process

acedeywin: @stephenhand @GPaoloni It worked perfectly with the custom helpline technique.

Screenshot 2022-09-09 at 11 26 00 PM

.github/workflows/deploy-helplines-parallel-worflows.yml
```diff
@@ -0,0 +1,43 @@
...
+      environment_code: ${{ matrix.environment_code }}
+      changelog: ${{ inputs.changelog }}
+
+  deploy-uk_staging_III: 
+    uses: techmatters/flex-plugins/.github/workflows/uk_staging_III.yml@CHI-1258_workflow_to_deploy_flex_helpline_stg_env
+    secrets: inherit
+    with:
+      changelog: ${{ inputs.changelog }}
+
+  do-something-after:
```
stephenhand, line 35: Can  we change this for something more useful, perhaps a Slack message saying 'All Staging helplines successfully deployed'. Then we don't need to go through al the individual success messages to be sure they all deployed.

.github/workflows/deploy-helplines-parallel-worflows.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  deploy-helplines:
+    strategy:
+      fail-fast: false
+      matrix:
+        short_helplines: ['CA', 'CL', 'PH', 'ET', 'JM', 'MW', 'IN', 'RO', 'BR', 'ZA', 'CO', 'TH', 'UK', 'UKR', 'ZM', 'ZW']
+        environment_code: [STG]
+    uses: techmatters/flex-plugins/.github/workflows/custom_helpline.yml@CHI-1258_workflow_to_deploy_flex_helpline_stg_env
+    secrets: inherit
```
GPaoloni, line 23: I'm not a fan of this, because it breaks the "less privilege rule", but maybe there's no other way around this being a matrix job. @acedeywin can you confirm if that's the case?

.github/workflows/deploy-helplines-parallel-worflows.yml
```diff
@@ -0,0 +1,43 @@
...
+      changelog:
+        description: Release Changelog - Add a comment about the changes to be included in this release.
+        required: true
+
+jobs:
+  deploy-helplines:
+    strategy:
+      fail-fast: false
+      matrix:
+        short_helplines: ['CA', 'CL', 'PH', 'ET', 'JM', 'MW', 'IN', 'RO', 'BR', 'ZA', 'CO', 'TH', 'UK', 'UKR', 'ZM', 'ZW']
```
nickhurlburt, line 20: If we go with this approach, please add a step to our helpline setup documentation to make sure we add it here.

.github/workflows/deploy-helplines-parallel-worflows.yml
acedeywin: @GPaoloni this is the best I can come up with in order to keep the code as DRY as possible.

.github/workflows/deploy-helplines-parallel-worflows.yml
acedeywin: @nickhurlburt Sure. I had that in mind. I will do that as soon as this PR gets approved.

.github/workflows/deploy-helplines-parallel-worflows.yml
stephenhand: @GPaoloni these are not privileges being assigned to a user, just being variables being made available to a script - and we own the calling script and the called script anyway, they are in the same repo so if you can change one you can change the other. So I think in this case not using 'inherit' is just extra boilerplate, there is no added security.

.github/workflows/deploy-helplines-parallel-worflows.yml
acedeywin: @stephenhand Done!

.github/workflows/deploy-helplines-parallel-worflows.yml
```diff
@@ -0,0 +1,69 @@
...
+      changelog: ${{ inputs.changelog }}
+
+  deploy-uk_staging_III: 
+    uses: techmatters/flex-plugins/.github/workflows/uk_staging_III.yml@CHI-1258_workflow_to_deploy_flex_helpline_stg_env
+    secrets: inherit
+    with:
+      changelog: ${{ inputs.changelog }}
+
+  # Send Slack notifying success
+  send-slack-message:
```
stephenhand, line 36: Could we not use this action: https://github.com/techmatters/flex-plugins/blob/master/.github/actions/notify-deploys/action.yml instead of duplicating the steps?

acedeywin: @stephenhand For some reason, I am not able to use it. The workflow can't read/find the action.yml file within the notify-deploys folder.

stephenhand: I see from this diff https://github.com/techmatters/flex-plugins/commit/29f5c2a38d9cc71574851360c42de3184bce3af7 you were trying to use an absolute path. Could you try using a relative path like the deploy workflows use to call the main action? i.e. `./.github/actions/notify-deploys`

stephenhand (approved):
I replied about the notify deploy action - I think we should be able to get this working. If not, merge it as is and I'll see if I can refactor it to use the action in another PR

.github/workflows/deploy-helplines-parallel-worflows.yml
acedeywin: @stephenhand  I also tried it. got the same result as the absolute path.

stephenhand: OK, I already approved it. Go ahead and merge as is

PR #926, opened 2022-09-08
Hungary customization updates

dee-luo:
-Flex UI Translations


dee-luo:
@janorivera on the changes you made - the leading 0s actually just need to be saved on the backend, they don't need to appear in the frontend to the user in the form (so in the value, not the label)! And then for select type fields that are **required**, the blank "" label value should also be "". I'm not totally sure how the required field logic works on the engineering side, but my understanding is that having the "" value will throw the required field error.

PR #928, opened 2022-09-13
Alejandro co terraform updates 2

janorivera:
Primary reviewer:
@stephenhand 
## Description
Changes for CO configuration and v2.0 of channel modules. 

- Channel modules: Just 2 modules needed now, one for custom channels and another for twilio channels.  Channels are activated via the local variables.
- Changes to the CO file to match current structure.
- Added CO chatbots
- Added a translations directory to allow localization of messages (idea that needs approval). 

### Checklist
- [x] Channel modules v2.0
- [x] CO main.tf configurations 
- [x] Added CO chabots
- [x] Localization
- [x] Successful apply
- [x] Test Twilio account


twilio-iac/terraform-modules/taskRouter/default/main.tf
```diff
@@ -9,9 +9,9 @@ terraform {
 
 locals {
   helplines = var.helplines == null ? [var.helpline] : var.helplines
-  helplines_filter = "helpline IN [${join(", ", formatlist("'%s'", local.helplines))}]"
+  helplines_filter = var.helplines == null ? "1==1" : "helpline IN [${join(", ", formatlist("'%s'", local.helplines))}]"
```
stephenhand, line 12: Nice

stephenhand (approved):
This is a great refactor Alejandro!

janorivera:
thanks for the approval @stephenhand ! I did this based on your advice so, much obliged!

twilio-iac/terraform-modules/channels/custom-channel/main.tf
```diff
@@ -13,30 +13,30 @@ locals {
...
-  resource "twilio_flex_flex_flows_v1" "twitter_flow" {
+  resource "twilio_flex_flex_flows_v1" "channel_flow" {
     channel_type  = "custom"
     chat_service_sid = var.flex_chat_service_sid
-    friendly_name = "Flex Twitter Channel Flow"
+    friendly_name = "Flex ${title(var.channel_name)} Channel Flow"
     integration_type = "studio"
-    contact_identity     = "twitter"
-    integration_flow_sid = twilio_studio_flows_v2.twitter_messaging_flow.sid
+    contact_identity     = "${var.channel_name}"
```
GPaoloni, line 32: Questions here @janorivera:
- Why is this different than in the "twilio channels" where we specify the address as part of the identity?
- This won't scale with multiple instagram accounts in one account, for example. Above could be the solution for that?

I may be talking nonsense tho :P

twilio-iac/terraform-modules/studioFlow/default/main.tf
```diff
@@ -199,7 +199,7 @@ locals {
         "type": "send-to-flex",
         "name": "webAttributes",
         "properties": {
-          "attributes": "{\"ip\":\"{{trigger.message.ChannelAttributes.pre_engagement_data.ip}}\",\"name\": \"{{trigger.message.ChannelAttributes.from}}\", \"channelType\": \"{{trigger.message.ChannelAttributes.channel_type}}\", \"channelSid\": \"{{trigger.message.ChannelSid}}\", \"helpline\": \"{{trigger.message.ChannelAttributes.pre_engagement_data.helpline}}\", \"ignoreAgent\":\"\", \"transferTargetType\":\"\",\n\"memory\": {{widgets.ChatBot.memory | to_json}}}",
+          "attributes": "{\"ip\":\"{{trigger.message.ChannelAttributes.pre_engagement_data.ip}}\",\"name\": \"{{trigger.message.ChannelAttributes.from}}\", \"channelType\": \"{{trigger.message.ChannelAttributes.channel_type}}\", \"channelSid\": \"{{trigger.message.ChannelSid}}\", \"helpline\": \"{{trigger.message.ChannelAttributes.pre_engagement_data.helpline}}\", \"ignoreAgent\":\"\", \"transferTargetType\":\"\",\"taskLanguage\": \"es-CO\",\n\"memory\": {{widgets.ChatBot.memory | to_json}}}",
```
GPaoloni, line 202: This might be addressed later on, but I don't think we want to default to `es-CO` language from here (that is the default shared across all accounts).

GPaoloni:
@janorivera Very cool refactor indeed, thanks!! :rocket:

twilio-iac/terraform-modules/channels/custom-channel/main.tf
janorivera: Good questions!
1.- For Twilio channels the API actually rejects if I don't create the channel with a specific contact_identity, it doesn't allow to put only the name of the channel.
2.- Yeah, this whole logic has been done thinking that we won't need more than one instace of each social media channel per account. So if the need arises then I'll need to refactor this code to support that, when that time comes I'll address this contact_identity attribute as well as you are suggesting.  The first attempt of this was to replicate what we are already doing from our how to manuals :).

twilio-iac/terraform-modules/studioFlow/default/main.tf
janorivera: @GPaoloni good catch! I'll make a correction :)

twilio-iac/terraform-modules/channels/custom-channel/main.tf
GPaoloni: Excellent! Thanks for the reply. Sure I don't expect any changes until they are needed, I just wanted to confirm with you hehe.

PR #931, opened 2022-09-19
CHI-1400 & CHI-1417

stephenhand:
Primary reviewer: @acedeywin 

## Description

* Styling on case list background is fixed (random white rectangles are gone)
* Custom Case status transitions are presented as options to the counsellor on the Edit Case Summary form, respecting the available transitions specified in the definition.
* Custom Case statuses are displayed in the case details section as expected
* Converted CaseDetails.jsx to TypeScript

### Checklist
- [N/A] New tests added
- [N/A] Strings are localized

### Related Issues
Fixes CHI-1400, CHI-1417

### Verification steps

See tickets

### DIFF
--- a/plugin-hrm-form/src/components/case/EditCaseSummary.tsx
+++ b/plugin-hrm-form/src/components/case/EditCaseSummary.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable react/jsx-max-depth */
 /* eslint-disable react/prop-types */
 import { v4 as uuidV4 } from 'uuid';
-import React from 'react';
+import React, { useMemo } from 'react';
 import { Template } from '@twilio/flex-ui';
 import { connect } from 'react-redux';
 import { FieldValues, FormProvider, SubmitErrorHandler, useForm } from 'react-hook-form';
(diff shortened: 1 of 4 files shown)


plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -65,49 +65,59 @@ const EditCaseSummary: React.FC<Props> = ({
...
+          name: 'caseSummary',
+          label: 'SectionName-CaseSummary',
+          type: 'textarea',
+        },
+      ];
+    } catch (e) {
+      console.error('Failed to render edit case summary form', e);
+      return [];
+    }
+  }, [connectedCaseState.prevStatus, definitionVersion]);
```
GPaoloni, line 116: `useMemo`  + almost constant dependencies :rocket:

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -142,7 +152,11 @@ const EditCaseSummary: React.FC<Props> = ({
     };
     const generatedForm = createFormFromDefinition(formDefinition)([])(initialForm, firstElementRef)(updateCallBack);
     return splitAt(3)(disperseInputs(7)(generatedForm));
-  }, [formDefinition, initialForm, firstElementRef, temporaryCaseInfo, getValues, updateTempInfo, task.taskSid]);
+  }, [formDefinition, initialForm, firstElementRef, editTemporaryCaseInfo, getValues, updateTempInfo, task.taskSid]);
```
GPaoloni, line 155: In the same line of above comment, I _believe_ `editTemporaryCaseInfo` changes pretty often, kinda invalidating this `useMemo` instance :p (same occurs in others case components, where this object slipped in).

GPaoloni (approved):
Looks great! Thanks for moving more code to TS :raised_hands: 

One thing I've noticed is that dev account have in it's config `permissionConfig: 'dev'`, and from `src/permissions/dev.json` I see that `"caseStatusTransition": [ ]`, which means no one should be allowed to do this op. The backend is kicking me with 401 when I attempt to transition the status as expected. Should we "disable" the drop-down when there are no permissions? Is this what you were commenting yesterday?

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
stephenhand: > Looks great! Thanks for moving more code to TS 🙌
> 
> One thing I've noticed is that dev account have in it's config `permissionConfig: 'dev'`, and from `src/permissions/dev.json` I see that `"caseStatusTransition": [ ]`, which means no one should be allowed to do this op. The backend is kicking me with 401 when I attempt to transition the status as expected. Should we "disable" the drop-down when there are no permissions? Is this what you were commenting yesterday?

Yes, there's another ticket open for that: https://bugs.benetech.org/browse/CHI-1427 - we decided to just scope this one to the custom case statuses and fix permissions in a different ticket

stephenhand:
> Looks great! Thanks for moving more code to TS 🙌
> 
> One thing I've noticed is that dev account have in it's config `permissionConfig: 'dev'`, and from `src/permissions/dev.json` I see that `"caseStatusTransition": [ ]`, which means no one should be allowed to do this op. The backend is kicking me with 401 when I attempt to transition the status as expected. Should we "disable" the drop-down when there are no permissions? Is this what you were commenting yesterday?

Yes, there's another ticket open for that: https://bugs.benetech.org/browse/CHI-1427 - we decided to just scope this one to the custom case statuses and fix permissions in a different ticket

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
stephenhand: `editTemporaryCaseInfo` is just `temporaryCaseInfo` casted to the `EditTemporaryCaseInfo` type, so it's just the same as it was before - but I take your overall point

stephenhand:
> 

One point of order - I think this should probably return a 403 rather than a 401. A 401 is technically a redirect that brings somebody to a page where they can authenticate to access the thing they were trying to get at originally. If the request is already authenticated, but the user isn't permitted to perform the operation / access the resource, I think it's a 403.

Although I also can't really think of a practical reason why it would make a difference :-P

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
GPaoloni: Yes, `temporaryCaseInfo` is causing the same in other case components. Talked about this with someone else in the team a while ago (can't remember with who tho haha).

GPaoloni: BTW I don't have a good idea on how to stop this other than wrapping the temporary object in a ref (like suggested [here](https://stackoverflow.com/a/63905181)). Just raising this conversation as I believe we either want to preserve the memo at low re-renders or just get rid of it :P

PR #933, opened 2022-09-22
Setting taskLanguage programatically

janorivera:
FYI @GPaoloni , What do you think about this? This will set the task language correctly for accounts with 1 language.
I aware though that this does not support multiple languages so my plan will be to work on that on a second round of changes.

### DIFF
--- a/twilio-iac/teguio-staging/main.tf
+++ b/twilio-iac/teguio-staging/main.tf
@@ -69,7 +69,7 @@ module twilioChannel {
       serverless_service_sid = var.serverless_service_sid
       master_workflow_sid = module.taskRouter.master_workflow_sid
       chat_task_channel_sid = module.taskRouter.chat_task_channel_sid
-      channel_attributes = var.custom_channel_attributes != "" ? var.custom_channel_attributes : file("../terraform-modules/channels/twilio-channel/channel-attributes/${each.key}-attributes.tftpl")
+      channel_attributes = var.custom_channel_attributes != "" ? var.custom_channel_attributes : templatefile("../terraform-modules/channels/twilio-channel/channel-attributes/${each.key}-attributes.tftpl",{language=var.language})
       flow_description = "${title(each.key)} Messaging Flow"
       pre_survey_bot_sid = module.custom_chatbots.pre_survey_bot_es_sid
       target_task_name = var.target_task_name
--- a/twilio-iac/teguio-staging/main.tf
+++ b/twilio-iac/teguio-staging/main.tf
@@ -69,7 +69,7 @@ module twilioChannel {
       serverless_service_sid = var.serverless_service_sid
       master_workflow_sid = module.taskRouter.master_workflow_sid
       chat_task_channel_sid = module.taskRouter.chat_task_channel_sid
-      channel_attributes = var.custom_channel_attributes != "" ? var.custom_channel_attributes : templatefile("../terraform-modules/channels/twilio-channel/channel-attributes/${each.key}-attributes.tftpl",{language=var.language})
+      channel_attributes = templatefile("../terraform-modules/channels/twilio-channel/channel-attributes/${each.key}-attributes.tftpl",{language=var.language})
       flow_description = "${title(each.key)} Messaging Flow"
       pre_survey_bot_sid = module.custom_chatbots.pre_survey_bot_es_sid
       target_task_name = var.target_task_name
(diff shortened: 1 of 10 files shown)


twilio-iac/teguio-staging/main.tf
```diff
@@ -69,7 +69,7 @@ module twilioChannel {
       serverless_service_sid = var.serverless_service_sid
       master_workflow_sid = module.taskRouter.master_workflow_sid
       chat_task_channel_sid = module.taskRouter.chat_task_channel_sid
-      channel_attributes = var.custom_channel_attributes != "" ? var.custom_channel_attributes : file("../terraform-modules/channels/twilio-channel/channel-attributes/${each.key}-attributes.tftpl")
+      channel_attributes = var.custom_channel_attributes != "" ? var.custom_channel_attributes : templatefile("../terraform-modules/channels/twilio-channel/channel-attributes/${each.key}-attributes.tftpl",{language=var.language})
```
GPaoloni, line 72: Since this is an implementation of an account (provides actual values to the more abstract `modules`), I think you could just say `channel attributes = var.custom_channel_attributes` if you know there are some, or the template file there aren't.

twilio-iac/terraform-modules/channels/twilio-channel/main.tf
```diff
@@ -14,7 +14,7 @@ locals {
       master_workflow_sid = var.master_workflow_sid
       chat_task_channel_sid = var.chat_task_channel_sid
       pre_survey_bot_sid = var.pre_survey_bot_sid
-      channel_attributes = var.custom_channel_attributes != "" ? var.custom_channel_attributes : file("${path.module}/channel-attributes/${var.channel_name}-attributes.tftpl")
+      channel_attributes = var.custom_channel_attributes != "" ? var.custom_channel_attributes : templatefile("${path.module}/channel-attributes/${var.channel_name}-attributes.tftpl",{language=var.language})
```
GPaoloni, line 17: What if `var.language` is empty here? It will cascade to   `"taskLanguage":"",`? What about removing the `taskLanguage` if there is not `var.language`? I'm even sure if that's possible in a template file tho.

GPaoloni (approved):
Looks good, just throwing some thoughts so we can dive deeper into this, thinking in future re usability!

Also Iyou are adding `"taskLanguage":"${language}",` to fb and web attributes only. Won't it make sense to throw in all of the templates, just to keep the rest updated?

twilio-iac/teguio-staging/main.tf
janorivera: aah I thought that taskLanguage was only used for the post survey messages, so I added it only on those channels that I thought it will be needed (TwilioChannels with Autopilot), is taskLanguage used somewhere else? anyway I can add it in the other channels, np. 

You are right about the channel attributes = var.custom_channel_attributes, I'll make the change, thanks for that!

twilio-iac/terraform-modules/channels/twilio-channel/main.tf
janorivera: I'll need to test that! but yeah I don't think that I can add that or remove it with a condition,  I'll need to use another template and that would be too much... hopefully it will default to "taskLanguage":"" and that should default to en-US right? @GPaoloni

GPaoloni: _I believe_ yes, that will be the case (as the task language is calculated [like this](https://github.com/techmatters/flex-plugins/blob/9d51f9f8e509745b817a464246b24001e6c5e9dd/plugin-hrm-form/src/utils/setUpActions.js#L139)). BTW just realized it might not work as expected with `taskLanguage`, but with `language` task attribute instead o.O

Let me know if this is a bug, we can rename the property in the code if that messes with too much things.

janorivera:
@GPaoloni  OK I made some changes, what do you think?

twilio-iac/terraform-modules/channels/custom-channel/variables.tf
```diff
@@ -1,3 +1,9 @@
+variable "language" {
+  type = string
+  default = ""
+  description = "Override the default language by setting this"
+}
```
GPaoloni, line 5: I would be great to test if this being empty results in `""` (empty string) or `' "" '` (a string that contains double quotes).

GPaoloni (approved):
LGTM!

twilio-iac/terraform-modules/channels/custom-channel/variables.tf
janorivera: ok I'll test this out, although I can't do it with the account that I'm currently working (CO) because this one is already using the language for the messages, but I'll test it out with TH or ZM 
So I'll merge it for now

janorivera: @GPaoloni  ok it gets passed as this (so I think is ok! right?) :
{
  "name": "{{trigger.message.ChannelAttributes.from}}",
  "channelType": "{{trigger.message.ChannelAttributes.channel_type}}",
  "channelSid": "{{trigger.message.ChannelSid}}",
  "twilioNumber": "{{trigger.message.ChannelAttributes.twilioNumber}}",
  "ignoreAgent": "",
  "language":"",
  "transferTargetType": "",
  "memory": {{widgets.ChatBot.memory | to_json}}
}

PR #937, opened 2022-10-03
End Chat button will wrapup task and trigger post survey

mythilytm:
Primary reviewer: @GPaoloni 

## Description
- Refactored post survey to be triggered on task wrapup(End Chat button) as opposed to on task complete

### Checklist
- [n/a] New tests added
- [n/a] Strings are localized

### Related Issues
Fixes #[1409](https://bugs.benetech.org/browse/CHI-1409)

### Verification steps
- The Serverless code required for this change to work has been deployed. If not, check into this branch (https://github.com/techmatters/serverless/pull/319) and deploy
- After a webchat based chat has been initiated, in Flex, 'END CHAT' for that task/conversation. This will trigger the post survey for the user in webchat

### DIFF
--- a/plugin-hrm-form/src/utils/setUpActions.js
+++ b/plugin-hrm-form/src/utils/setUpActions.js
@@ -368,7 +368,7 @@ const triggerPostSurvey = async (setupObject, payload) => {
  * @param {ReturnType<typeof getConfig> & { translateUI: (language: string) => Promise<void>; getMessage: (messageKey: string) => (language: string) => Promise<string>; }} setupObject
  * @returns {import('@twilio/flex-ui').ActionFunction}
  */
-export const afterCompleteTask = setupObject => async payload => {
+export const afterWrapupAction = setupObject => async payload => {
   const { featureFlags } = setupObject;
 
   if (featureFlags.enable_post_survey) {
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/utils/setUpActions.js
```diff
@@ -368,7 +368,7 @@ const triggerPostSurvey = async (setupObject, payload) => {
  * @param {ReturnType<typeof getConfig> & { translateUI: (language: string) => Promise<void>; getMessage: (messageKey: string) => (language: string) => Promise<string>; }} setupObject
  * @returns {import('@twilio/flex-ui').ActionFunction}
  */
-export const afterCompleteTask = setupObject => async payload => {
+export const afterWrapupTask = setupObject => async payload => {
```
GPaoloni, line 371: The fact that you are replacing the whole logic of `afterCompleteTask` and moving it to `afterWrapupTask` is introducing a bug here! `removeContactForm` is being dragged to this event and as a consequence, if a worker hits "end chat" before submitting the contact, all info is gone.

I think what you want to do is keep the logic unrelated to post survey in `afterCompleteTask` as it is, and only move to `afterWrapupTask` what is relevant in this case.

mythilytm: Thanks for the clarification

GPaoloni (approved):
Looks good and works as expected!

PR #938, opened 2022-10-05
CHI-1427: Rework EditCaseSummary to use caseWorkingCopy redux state

stephenhand:
Primary Reviewer @murilovmachado @acedeywin 

## Description

* Migrates the EditCaseSummary.tsx form to use the caseWorkingCopy redux state rather than TemporaryCaseInfo
* Remove TemporaryCaseInfo type, actions & redux state
* Remove several other redundant actions / reducers
* Remove some redundant component level code

### Checklist
- [X] New tests added
- N/A Strings are localized

### Related Issues
Fixes #....

### Verification steps

Regression test case adding & editing, including switching between multiple cases in various states


acedeywin (approved):
LGTM.

plugin-hrm-form/src/components/case/CaseSummary.tsx
```diff
@@ -42,7 +36,6 @@ const CaseSummary: React.FC<Props> = ({ task, connectedCaseState, updateCaseInfo
         // Add Case summary doesn't show up as default value
         placeholder={readonly ? strings.NoCaseSummary : strings['Case-AddCaseSummaryHere']}
         value={summary}
-        onChange={e => handleOnChange(e.target.value)}
```
murilovmachado, line 45: We're removing `onChange` here because editing now is done at a separate component: `EditCaseSummary.tsx`, right?

murilovmachado: If the answer is true, I believe we do not need `readonly` props in this component, because it should always be `true`, I guess.

murilovmachado (approved):
Looks good to me! Just one question on `CaseSummary.tsx` file.

plugin-hrm-form/src/components/case/CaseSummary.tsx
stephenhand: You are correct on both points, I removed the readonly prop

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -69,32 +73,32 @@ const EditCaseSummary: React.FC<Props> = ({
...
         },
         {
-          name: 'inImminentPhysicalDanger',
+          name: 'childIsAtRisk',
           label: 'Case-ChildIsAtRisk',
           type: 'checkbox',
         },
         {
-          name: 'caseSummary',
+          name: 'summary',
```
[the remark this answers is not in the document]
stephenhand: The new names actually reflect the props in the DB JSON object, so we can just use a spread operator to overwrite the 3 props in the `info` blob, and add `status` to the top level separately.

Previously the names in the form didn't align with what was in the DB JSON, but they were explicitly mapped

PR #940, opened 2022-10-06
CHI-1427: Case summary permissions

stephenhand:

## Description

* Now the 'Edit' button for the case summary appears if the user is permitted to edit anything on that form
* The form elements on the 'Edit Case Summary' form are disabled when a user does not have permission to use them
* Added an optional function to the form generation function to set whether controls are enabled or not - this seemed more appropriate than adding something to the definition itself, as it is a transient property that would be set be the code generating the form, rather than being specified declaratively in the definition file
* Rejigged the Case reducer so it takes the global HRM State as well as the case state - this means it can access the form definiton for the case. It means I had to not use 'combineReducers' for that particular reducer.
* Added a precalculated 'availableStatusTransitions' array to the case redux state.
* The Case UI uses the 'availableStatusTransitions' to determine if there are any status transitions the user can make, and will render the button if there are any
* The Edit Case Summary UI will use the 'availableStatusTransitions' to determine what status options are available

### Checklist
- [X] New tests added
- N/A Strings are localized

### Related Issues
Fixes CHI-1427

### Verification steps

See CHI-1427 repro steps


plugin-hrm-form/src/states/index.ts
```diff
@@ -27,21 +28,25 @@ const reducers = {
...
 export type RootState = FlexState & { [namespace]: HrmState };
+const combinedReducers: (state: HrmState, action) => HrmState = combineReducers(reducers);
+
+// Combine the reducers
+const reducer = (state: HrmState, action): HrmState => {
+  return {
+    ...combinedReducers(state, action),
+    [connectedCaseBase]: ConnectedCaseReducer(state, (state ?? {})[connectedCaseBase], action),
+  };
+};
```
murilovmachado, line 50: I had a hard time until I could understand this piece. Can you add some comments on why you're doing it this way?

murilovmachado (approved):
Looks good to me!

PR #967, opened 2022-10-12
Fixed hungarian messages not working as expected

GPaoloni:
## Description
This PR:
- Fixes a subtle bug where default messages were returned in cases where it shouldn't.
- Fixes the `hu-HU` messages not working as intended cause a bad reference.

### Verification steps
- Start development server.
- Submit a new task via webchat (or any other chat) but do not accept it inmediately.
- Edit the task attributes from Twilio console, adding `"language": "hu-HU",`.
- Accept the task and confirm that the translated messages are sent.


stephenhand (approved):
LGTM

PR #971, opened 2022-10-13
Added helplineLanguage option to be set via tf

GPaoloni:
## Description
This PR adds the option to set `helplineLanguage` service config attribute via terraform (under `flex` module).
I'm leaving the default value as `""` (empty string) since I believe that won't cause any issues in our UI code (if the helpline language is falsy, we don't use it).

@janorivera In order to set this in PL, you should pass down to the flex module the variable `helpline_language`.

Question here @stephenhand @janorivera:
I see that there are two different modules addressing the same, `terraform-modules/flex/default` and `/terraform-modules/flex/service-configuration`, the first being used almost everywhere, except in `teguio_staging`, `teguio_production` (>.<), `thailan_staging` and `zimbabwe_staging`. Is this correct, or we did something wrong along the way?
I see that the `default` one configures Flex flows (`messaging_flow` & `webchat_flow`) out of the box, so maybe the second is the most recent one, accordnig to the recent changes on "per channel boxed config"?

### Verification steps
- Provide `helpline_language` to the `module flex` from the `main.tf` for a given account.
- `terraform init`
- `terraform plan -var-file=....`

### DIFF
--- a/twilio-iac/terraform-modules/flex/default/main.tf
+++ b/twilio-iac/terraform-modules/flex/default/main.tf
@@ -65,7 +65,8 @@ locals {
     "pdfImagesSource": "https://tl-public-chat.s3.amazonaws.com",
     "logo_url": "https://aselo-logo.s3.amazonaws.com/145+transparent+background+no+TM.png",
     "multipleOfficeSupport": var.multi_office_support,
-    "serverless_base_url": var.serverless_url
+    "serverless_base_url": var.serverless_url,
+    "helplineLanguage": var.helpline_language
   }})
 }
 
--- a/twilio-iac/terraform-modules/flex/default/variables.tf
+++ b/twilio-iac/terraform-modules/flex/default/variables.tf
@@ -35,6 +35,12 @@ variable "definition_version" {
   type        = string
 }
 
+variable "helpline_language" {
+  description = "Keyword that determines the language to be used as default across the helpline"
+  type        = string
+  default = ""
+}
+
 variable "feature_flags" {
   description = "Map of feature flag settings. All values should be boolean"
   type = map(bool)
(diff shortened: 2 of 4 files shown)


stephenhand:
I'm happy to move service configuration out into a separate module if it's easier to work with. I presume the helplines using this one aren't using the flex/default module, and setting up the stuff it sets up elsewhere?

If that's the case, perhaps every helpline should use the service-configuration module, and delete the null resource from the flex/default module. Then every helpline would pull in the service configuration module and only pull in the default one if re4quired?

We would have to ensure we add the correct 'moved' directives too for migration purposes (although maybe not, the worst that would happen is the service config would get reset to the TF version in the next `terraform apply`)

stephenhand (approved):

janorivera:
ok there is another "language" variable that is used for localization of messages in studio flow. I can consolidate both into one variable. 
@GPaoloni To answer your question, yes, as  you said, the flex/default was the module used before the channel implementation approach, I left it there to have some backward compatibility. I created the service-configuration under the same module to maintain the location but yes I agree that this resource doesn't make too much sense to keep it under "flex/" if it doesn't have anything to do with it? I thought that the logic behind it was that the service-configuration changes how our "flex" ui behaves?  anyway @stephenhand I'm happy to change it too to a more appropriate module, any recomendations for a name?

janorivera (approved):

stephenhand:
@janorivera - I'm happy to leave it under Flex, I agree the service configuration is flex related stuff. 

I wasn't talking about moving it, but if we migrate old accounts from using the service config resource in flex/default to using the one in flex/service-configuration, we can use a 'moved' directive to tell TF that this is actually the same resource in a different place, we aren't deleting from flex/default and creating something unrelated in flex/service-configuration, we want it to be moved, with its current state intact: https://www.terraform.io/language/modules/develop/refactoring

So should we be migrating all the accounts away from flex/default to channel modules?

PR #972, opened 2022-10-14
CHI-1425: Minor terraform updates

stephenhand:
Primary reviewer: @janorivera 

## Description
* Added triggers to API key provisioners so they will rerun when the SSM parameter name would change. 
* Moved variables that shouldn't be changed at execution time to locals in Kek Vonal
* Added github module to Kek Vonal

### Checklist
- N/A New tests added
- N/A Strings are localized

### Related Issues
CHI-1425

### Verification steps

None required, TF changes already applied

### DIFF
--- a/twilio-iac/kek-vonal-staging/main.tf
+++ b/twilio-iac/kek-vonal-staging/main.tf
@@ -16,8 +16,36 @@ terraform {
 }
 
 locals {
+
+  helpline  = "Kék Vonal"
+  short_helpline = "HU"
+  operating_info_key = "hu"
+  environment = "Staging"
+  short_environment = "STG"
+  definition_version = "hu-v1"
+  permission_config = "hu"
+  multi_office  = false
+
+  feature_flags = {
+      "enable_fullstory_monitoring": false,
+      "enable_upload_documents": true,
+      "enable_post_survey": false,
+      "enable_case_management": true,
+      "enable_offline_contact": true,
+      "enable_filter_cases": true,
+      "enable_sort_cases": true,
+      "enable_transfers": true,
+      "enable_manual_pulling": true,
+      "enable_csam_report": true,
+      "enable_canned_responses": true,
+      "enable_dual_write": false,
+      "enable_save_insights": true,
+      "enable_previous_contacts": true
+    }
+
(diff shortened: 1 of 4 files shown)


twilio-iac/kek-vonal-staging/main.tf
```diff
@@ -16,8 +16,36 @@ terraform {
...
+      "enable_filter_cases": true,
+      "enable_sort_cases": true,
+      "enable_transfers": true,
+      "enable_manual_pulling": true,
+      "enable_csam_report": true,
+      "enable_canned_responses": true,
+      "enable_dual_write": false,
+      "enable_save_insights": true,
+      "enable_previous_contacts": true
+    }
```
GPaoloni, line 44: We used to declare this in `variables.tf` file. I think doing so here it's even cleaner, maybe we should follow this path going forward :P

twilio-iac/terraform-modules/hrmServiceIntegration/default/main.tf
```diff
@@ -3,6 +3,12 @@ locals {
 }
 
 resource "null_resource" "hrm_static_api_key" {
+  triggers = {
+    helpline = var.helpline
+    short_helpline = var.short_helpline
+    environment = var.environment
+    short_environment = var.short_environment
+  }
```
GPaoloni, line 11: Nice, just learned this :rocket:

GPaoloni (approved):

twilio-iac/kek-vonal-staging/main.tf
stephenhand: Yes, as I said to @janorivera - these variables really should require a PR for someone to change, rather than being able to override them when we apply, so yes, would like to move those vars to locals for all accounts in time

twilio-iac/terraform-modules/hrmServiceIntegration/default/main.tf
stephenhand: @janorivera - if any of those 'trigger' values changes from what it was when Terraform was last applied, it will rerun the provisioner. So in this case it will automatically recreate the API Key and SSM parameter if the helpline or environment info changes

PR #978, opened 2022-10-18
Move Transfers to Serverless

murilovmachado:
Primary reviewer: @GPaoloni 

## Description
The goal of this PR is to move transfer code into serverless. In order words, this is mostly removing code here. It also fixes some transfer scenarios that are slightly different now. There's still some code left here that could be moved to Serverless in future iterations.

Serverless PR: https://github.com/techmatters/serverless/pull/330

### Checklist
- [x] New tests added
- [x] Strings are localized

### Related Issues
https://bugs.benetech.org/browse/CHI-1350

### Verification steps
After Serverless is deployed, transfers scenarios should behave just fine.


stephenhand:
@murilovmachado - how do we need to manage upgrades? I presume serverless needs to be upgraded before deploying this version.

Have we tested the current flex UI with the updated serverless? Does trying to do the work on the frontend AND serverless cause any potential issues?

murilovmachado:
> Have we tested the current flex UI with the updated serverless? Does trying to do the work on the frontend AND serverless cause any potential issues?

We better deploy serverless and flex-plugins at the same time. If we deploy only serverless, we'll get some issues related to some controls being visible when they shouldn't. (Maybe bigger issues on transfers' edge cases either)

plugin-hrm-form/src/utils/transfer.js
```diff
@@ -220,38 +174,5 @@ export const closeCallOriginal = async task => {
 export const closeCallSelf = async task => {
   await setTransferRejected(task);
   await returnTaskControl(task);
-  await Actions.invokeAction('HangupCall', { sid: task.sid });
+  await Actions.invokeAction('CompleteTask', { sid: task.sid });
```
GPaoloni, line 177: This is because the task isn't closed by TaskRouter callback listeners for voice contacts, right?

GPaoloni (approved):
Looks good to me.

plugin-hrm-form/src/utils/transfer.js
murilovmachado: Right!

PR #979, opened 2022-10-18
[WIP] UI functionality for external transcript viewing

GPaoloni:
Primary reviewer:

## Description
This PR adds logic to download the transcript from the S3 location if possible.
Notes:
- The view of the transcript have zero styling. I have tried several times to re-use Twilio stuff, but they rely on the task/channel. If the goal for us is to "import the transcript from S3", as I have understood, we may need to add our own styles.
- Currently, when we export the transcripts , we include the `from` attribute from the messages, which represent the `identity` of the sender within the Twilio Chat Service (`➜ twilio api:chat:v2:services:users:fetch --service-sid=<service sid> --sid=<identity> -o=json`). To retrieve it's friendly name, we have to do some manipulation, either in Flex when loading the transcript, or in the process of exporting the transcript. I think the second is better as it will save some API calls from the client (and sound more like ETL?).

WIP:
- Play around with Twilio stuff to see if I can reuse the webchat styles?
- Tests.

### Checklist
- [ ] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1342)
- [ ] New tests added
- [ ] Strings are localized

### Verification steps
- Un-comment line 296 of `src/components/contact/ContactDetailsHome.tsx` to point to a fixed existing transcript object.
- Comment lines 40 to 54 of `src/components/contact/TranscriptSection.tsx` so the transcript is shown as if it were the corresponding one for this contact.
- `npm run dev`.
- Search and open the transcript section of any contact.

### DIFF
--- a/plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
+++ b/plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
@@ -25,6 +25,7 @@ import { DetailsContext, toggleDetailSectionExpanded } from '../../states/contac
 import { getPermissionsForContact, PermissionActions } from '../../permissions';
 import { createDraft, ContactDetailsRoute } from '../../states/contacts/existingContacts';
 import { getConfig } from '../../HrmFormPlugin';
+import TranscriptSection from './TranscriptSection';
 
 // TODO: complete this type
 type OwnProps = {
@@ -40,6 +41,7 @@ type Props = OwnProps & ReturnType<typeof mapStateToProps> & typeof mapDispatchT
 /* eslint-disable complexity */
 // eslint-disable-next-line sonarjs/cognitive-complexity
 const ContactDetailsHome: React.FC<Props> = function ({
+  contactId,
   context,
   detailsExpanded,
   showActionIcons = false,
@@ -295,9 +297,11 @@ const ContactDetailsHome: React.FC<Props> = function ({
           buttonDataTestid="ContactDetails-Section-Transcript"
           showEditButton={false}
         >
-          <SectionActionButton type="button" onClick={() => console.log('>>>> Pressed')}>
-            <Template code="ContactDetails-LoadTranscript-Button" />
-          </SectionActionButton>
+          {/* <TranscriptSection contactId={contactId} transcriptUrl={savedContact.details.mediaUrls[0].url} /> */}
+          <TranscriptSection
+            contactId={contactId}
+            transcriptUrl="https://tl-aselo-docs-as-development.s3.amazonaws.com/transcripts/2022/08/09/20220809131844-WT1d7624f7c9d015a52d825e3be75ea105.json"
+          />
         </ContactDetailsSection>
       )}
     </DetailsContainer>
--- a/plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
+++ b/plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
@@ -5,6 +5,7 @@ import { Actions, Insights, Template } from '@twilio/flex-ui';
 import { connect } from 'react-redux';
 import { callTypes } from 'hrm-form-definitions';
 
+import { ContactMediaType } from '../../types/types';
 import {
   DetailsContainer,
   NameText,
@@ -14,7 +15,7 @@ import {
 } from '../../styles/search';
 import ContactDetailsSection from './ContactDetailsSection';
 import SectionEntry from '../SectionEntry';
-import { channelTypes } from '../../states/DomainConstants';
+import { channelTypes, isChatChannel, isVoiceChannel } from '../../states/DomainConstants';
 import { isNonDataCallType } from '../../states/ValidationRules';
 import { formatCategories, formatDuration, formatName, mapChannelForInsights } from '../../utils';
 import { ContactDetailsSections, ContactDetailsSectionsType } from '../common/ContactDetails';
@@ -108,6 +109,7 @@ const ContactDetailsHome: React.FC<Props> = function ({
     CHILD_INFORMATION,
     ISSUE_CATEGORIZATION,
     CONTACT_SUMMARY,
+    TRANSCRIPT,
   } = ContactDetailsSections;
   const addedBy = counselorsHash[createdBy];
   const counselorName = counselorsHash[counselor];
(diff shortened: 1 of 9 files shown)


stephenhand:
We should definitely be trying to put as much of the relevant transcript metadata as is feasible into the transcript document. We may want to consider if we want the lambda to look up the friendly name, or if we want it to be saved along with the contact, but only if we already have it in Flex at that point, definitely don't want to be doing extra lookup calls in the front end. 

If it is available we could save it as part of the `conversationMedia` item potentially, then pass it along with the message if the additional parameters blob.

GPaoloni:
> If it is available we could save it as part of the conversationMedia item potentially, then pass it along with the message if the additional parameters blob.

I'm not sure it is to be honest. Looks like Flex has several helpers to fetch information about the different members of channel, but in last instance those are wrapped API calls to the Chat -> Users APIs.

plugin-hrm-form/src/services/ServerlessService.ts
```diff
@@ -166,6 +166,16 @@ export const getFileDownloadUrl = async (fileNameAtAws: string, fileName: string
   return response;
 };
 
+/**
+ * Gets a file download url from S3, using the object url as constructed by AWS
+ */
+export const getFileDownloadUrlFromUrl = async (objectUrl: string, fileName: string = undefined) => {
```
stephenhand, line 172: Sorry for the dumb Q, but how is this different to the one we use to download case docs?

stephenhand (approved):
Just the query about the show / hide logic, LGTM otherwise

plugin-hrm-form/src/services/ServerlessService.ts
GPaoloni: While it's not done in the serverless side, I thought that leaving the door open for fetching from a different bucket wont harm, just in case we want S3 objects that are not under docs bucket (we have that info in the url). Also thought that manipulating the URL (what we have in the media urls) within the UI was not as clean as wrapping that logic in a service. They could be unified tho.

plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
```diff
@@ -127,13 +131,14 @@ const ContactDetailsHome: React.FC<Props> = function ({
...
+  const recordingAvailable = Boolean(
+    featureFlags.enable_voice_recordings &&
+      isVoiceChannel(channel) &&
       canViewTranscript &&
-      savedContact.details.conversationMedia?.length &&
-      typeof savedContact.overview.conversationDuration === 'number',
+      savedContact.details.conversationMedia?.length,
+    // && typeof savedContact.overview.conversationDuration === 'number',
   );
+  const showTranscriptSection = featureFlags.enable_transcripts && isChatChannel(channel);
```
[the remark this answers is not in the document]
GPaoloni: How can we differentiate those two? Also aren't we planning to back fill the "legacy" ones?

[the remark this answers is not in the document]
stephenhand: The 'pending' transcript could simply not have a URL, or we can have a state field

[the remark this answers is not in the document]
stephenhand: I don't know if we are planning to backfill transcripts from the beginning of time, but even if we do, we can't guarantee a transcript for every contact is available in Twilio

Also, as discussed on the estimation, we are planning to implement transcript viewing permission purely on the backend, by simply not returning a conversationMedia item for the transcript to the client if the counsellor doesn't have transcript viewing permission.

plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
```diff
@@ -269,17 +274,30 @@ const ContactDetailsHome: React.FC<Props> = function ({
...
+          handleExpandClick={() => toggleSection(TRANSCRIPT)}
+          buttonDataTestid="ContactDetails-Section-Transcript"
+          showEditButton={false}
+        >
+          <TranscriptSection
+            contactId={contactId}
+            canViewTranscript={canViewTranscript}
+            transcriptAvailable={savedContact.details.mediaUrls?.some(m => m.type === ContactMediaType.TRANSCRIPT)}
+            // transcriptUrl="https://tl-aselo-docs-as-development.s3.amazonaws.com/transcripts/2022/08/09/20220809131844-WT1d7624f7c9d015a52d825e3be75ea105.json"
+            transcriptUrl={savedContact.details.mediaUrls ? savedContact.details.mediaUrls[0].url : undefined}
```
stephenhand, line 297: We should be checking the type of the conversationMedia item, since this collection will eventually have recordings in as well. Not strictly required whilst we only have one type of media, but no harm in having it in

stephenhand (commented):
Sorry, also noticed this PR rips out support for viewing the existing Twilio stored transcripts. We need to be able to support both. If the item in `conversationMedia` has a `reservationSid` rather than a `url` it should load the overlay rather than the new transcript view (if it has both, prefer the externally stored transcript).

We probably want separate feature flags for Twilio and external transcript support as well

This will make migration much easier to manage, and perhaps some helplines will prefer to kee pusing the Twilio transcripts going forward (if not we can rip out support once everybody is migrated).

stephenhand (commented):
.

GPaoloni:
@stephenhand The "verification step" section in the PR description is updated accordingly to this changes.

plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
```diff
@@ -138,7 +139,17 @@ const ContactDetailsHome: React.FC<Props> = function ({
       savedContact.details.conversationMedia?.length,
     // && typeof savedContact.overview.conversationDuration === 'number',
   );
-  const showTranscriptSection = featureFlags.enable_transcripts && isChatChannel(channel);
+
+  const twilioStoredTranscript =
+    featureFlags.enable_twilio_transcripts && savedContact.details.conversationMedia?.find(isTwilioStoredMedia);
+  const externalStoredTranscript =
+    featureFlags.enable_extrernal_transcripts && savedContact.details.conversationMedia?.find(isS3StoredTranscript);
```
stephenhand, line 146: typo `enable_external_transcripts`

plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
```diff
@@ -138,7 +139,17 @@ const ContactDetailsHome: React.FC<Props> = function ({
...
+  const showTranscriptSection = Boolean(
+    canViewTranscript &&
```
stephenhand, line 148: I think `canViewTranscript` is only relevant to the Twilio transcripts, it checks the Twilio worker's roles to see if the worker would have permission to load up the transcript via the API before showing a button that doesn't work

So you probably want:

```
      isChatChannel(channel) &&
      savedContact.details.conversationMedia?.length &&
      ((canViewTranscript && twilioStoredTranscript) || externalStoredTranscript),
```

Perhaps want to rename that variable to 'canViewTwilioTranscript' to prevent similar confusion in the future

nickhurlburt:
> We should definitely be trying to put as much of the relevant transcript metadata as is feasible into the transcript document as opposed to forcing lookups on the frontend - especially as we don't know how long some of this data will be kept around.

Yes. A key thing to remember is that by exporting transcripts from Twilio, we are working toward a day when we can delete data out of Twilio -- _especially_ any PII or sensitive data (which includes transcripts). This is especially important for helplines that do not want PII stored in US-based systems. This will only be some helplines. So we want to think of our transcript features as being able to work even if all caller PII has been removed from Twilio.

PR #982, opened 2022-10-19
[Bug Fix]: Start New Chat appears after post survey initialization

GPaoloni:
Primary reviewer: @stephenhand @mmythily 

## Description
This PR addressed two bugs:
1) Calls to `maybeExcludeDeactivateChatChannel` were moved to it's own definition in https://github.com/techmatters/flex-plugins/pull/947, causing the [ChatOrchestrator](https://assets.flex.twilio.com/docs/releases/flex-ui/1.23.3/ChatOrchestrator.html) to behave as default (channels are deactivated on wrapup). This caused the first bug where every chat contact was "closed" before the post survey had a chance to execute.
2) Counselors where still able to read the message after the task was wrapped, exposing the child's responses to the post survey. This is prevented by removing some listeners from the Flex UI on wrapup. Note: I would have preferred removing the counselor from the channel instead, so that could be done from taskrouter callback and not the UI, but doing so causes Flex to entirely unmount the chat from the UI, making imposible for the counselor to see the conversation when it should be available. Same occurs if we try to orchestrate `LeaveChatChannel` on wrapup, as can be seen in [732e6ec](https://github.com/techmatters/flex-plugins/commit/732e6eccd988d5b87894d6c6b595044155ae2a12).

Also played around with some TS.
Test are WIP, but this can be reviewed :)

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1468)
- [WIP] New tests added
- [n/a] Strings are localized

### Verification steps
- Start development server.
- Send a new webchat contact.
- Accept it in your local instance of Flex.
- Move the task to wrapup (either with end chat button in Flex or with the new end chat button in webchat if available).
- Confirm that:
  - The channel is still open, and the webchat can still interact with the post survey.
  - The counselor does not receives any further interactions.

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.tsx
+++ b/plugin-hrm-form/src/HrmFormPlugin.tsx
@@ -68,7 +68,7 @@ const readConfig = () => {
   };
 };
 
-let cachedConfig;
+let cachedConfig: ReturnType<typeof readConfig>;
 
 try {
   cachedConfig = readConfig();
--- a/plugin-hrm-form/src/___tests__/utils/setUpActions.test.ts
+++ b/plugin-hrm-form/src/___tests__/utils/setUpActions.test.ts
@@ -1,7 +1,10 @@
-import { ITask } from '@twilio/flex-ui';
+/* eslint-disable camelcase */
+import { ITask, StateHelper, TaskHelper, ChatOrchestrator } from '@twilio/flex-ui';
 
-import { afterCompleteTask } from '../../utils/setUpActions';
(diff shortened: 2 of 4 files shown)


stephenhand (approved):
Sorry - my bad on the first one

LGTM, haven't tested it though. Can do tomorrow

mythilytm (approved):
Looks great! 🚀 
Got a chance to test it in different scenarios. Webchat and Flex work well together is displaying the post survey and 'Start New Chat' accordingly.

PR #1001, opened 2022-10-27
Mask PII identifiers for all channel type contacts / tasks 

mythilytm:
Primary reviewer: @GPaoloni or @stephenhand 

## Description
- In this PR, the twilio strings and aselo components with PII are masked for all channel types. All Aselo components have been masked. 
- Two twilio related components need work. 
    - TaskListItem.secondLine (Hopefully, there is a better way than using handlebars)
    - Remove IP address in the first message from child (only webchat)
- `maskIdentifiers` flag implemented in this PR will be replaced in the next PR #1017 . 

### Checklist
- [x] New tests added - Tests updated
- [x] Strings are localized

### Related Issues
Fixes # 1471

### Verification steps
- Change `maskIdentifiers` in all relevant files to `true` and Check for PII where they usually appear in
  - Task List
  - Task panel when a call comes in
  - Task panel during an active call (except the message bubble... still working on this)
  - Info panels for calls
  - Previous calls banner
  - Contact Details 
  - Agents page


plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -370,5 +370,10 @@
   "BottomBar-DontSave": "No, Don't Save",
   "BottomBar-Save": "Yes, Save",
 
-  "CaseSummary-None": "- No case summary -"
+  "CaseSummary-None": "- No case summary -",
+
+
+  "MaskIdentifiers": "XXXXXX",
+  "TaskInfoPanelContent": "<h1>TASK CONTEXT</h1><h2>Task type</h2><p>{{task.channelType}}</p><h2>Task created on</h2><p>{{task.dateCreated}}</p><h2>Task priority</h2><p>{{task.priority}}</p><h2>Task queue</h2><p>{{task.queueName}}</p><h2>Task Sid</h2><p>{{task.taskSid}}</p><h2>Reservation Sid</h2><p>{{task.sid}}</p><hr /><h1>CUSTOMER CONTEXT</h1><h2>Customer name / phone number</h2><p>AdminXXXXXXXXXXX</p><h2>Country</h2><p>{{task.attributes.caller_country}}</p><hr /><h1>ADDONS</h1><p>No add-ons enabled. To expand your experience, visit</p><a href=\"https://www.twilio.com/marketplaceadd-ons\" target=\"blank\">Twilio Marketplace</a>",
```
mythilytm, line 377: I hardcoded these but I am exploring how to change this dynamically

plugin-hrm-form/src/utils/setUpComponents.js
```diff
@@ -436,16 +438,31 @@ export const setupTwitterChatChannel = () => {
...
   TwitterChatChannel.templates.TaskListItem.firstLine = 'TaskHeaderLineTwitter';
   TwitterChatChannel.templates.TaskCard.firstLine = 'TaskHeaderLineTwitter';
-  TwitterChatChannel.templates.TaskCanvasHeader.title = 'TaskHeaderLineTwitter';
+  // TwitterChatChannel.templates.TaskCanvasHeader.title = 'TaskHeaderLineTwitter';
   TwitterChatChannel.templates.Supervisor.TaskCanvasHeader.title = 'TaskHeaderLineTwitter';
   TwitterChatChannel.templates.Supervisor.TaskOverviewCanvas.title = 'TaskHeaderLineTwitter';
 
+  if (maskIdentifiers) {
+    // Task List
+    TwitterChatChannel.templates.TaskListItem.firstLine = 'MaskIdentifiers';
```
stephenhand, line 456: Can we write a function for this, rather than repeating the logic per channel?

plugin-hrm-form/src/components/search/SearchForm/index.jsx
```diff
@@ -237,7 +238,13 @@ class SearchForm extends Component {
                     </Box>
                     <span>
                       <Template code="PreviousContacts-OnlyShowRecordsFrom" /> <Template code={source} />{' '}
-                      <Bold>{checkBoxName}</Bold>
+                      {maskIdentifiers ? (
```
stephenhand, line 241: Probably not worth it for 2, but if this switch is repeated elsewhere, a small component for identifiers might be worth doing

plugin-hrm-form/src/components/PreviousContactsBanner.tsx
```diff
@@ -46,6 +46,8 @@ const PreviousContactsBanner: React.FC<Props> = ({
   changeRoute,
   editContactFormOpen,
 }) => {
+  const maskIdentifiers = true;
```
stephenhand, line 49: Needs to be hardcoded false if you want to merge this

stephenhand:
Looks like the E2E test failed because the chat task didn't come through to Flex. Not sure if it's related to this PR or some issue with Twilio, probably worth trying to rerun them

plugin-hrm-form/src/HrmFormPlugin.tsx
```diff
@@ -42,7 +42,7 @@ const readConfig = () => {
   const featureFlags = manager.serviceConfiguration.attributes.feature_flags || {};
   const contactsWaitingChannels = manager.serviceConfiguration.attributes.contacts_waiting_channels || null;
   const { strings } = (manager as unknown) as { strings: { [key: string]: string } };
-
+  console.log('>strings', strings);
```
stephenhand, line 45: Probably need to remove these debug logs to merge?

mythilytm: ✅

plugin-hrm-form/src/components/PreviousContactsBanner.tsx
mythilytm: ✅

plugin-hrm-form/src/utils/setUpComponents.js
mythilytm: ✅

plugin-hrm-form/src/translations/en-US/flexUI.json
stephenhand: Could we move the HTML content into code somewhere and have separate keys for all the bits of translatable text in it?

HTML in translations is a bit yucky. The odd <span> tag we might get away with, but let's try to avoid it where possible

plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
```diff
@@ -176,10 +180,17 @@ const ContactDetailsHome: React.FC<Props> = function ({
...
           value={formattedChannel}
         />
-        <SectionEntry
-          description={<Template code="ContactDetails-GeneralDetails-PhoneNumber" />}
-          value={isPhoneContact ? customerNumber : ''}
-        />
+        {maskIdentifiers ? (
+          <SectionEntry
+            description={<Template code="ContactDetails-GeneralDetails-PhoneNumber" />}
+            value={strings.MaskIdentifiers}
```
GPaoloni, line 186: If we can pass a React component here, passing down `<Template code="MaskIdentifiers" />` will save the need for calling `getConfig` in this component. Worth trying I think.

plugin-hrm-form/src/components/search/ContactPreview/ChildNameAndDate.tsx
```diff
@@ -4,6 +4,7 @@ import { format } from 'date-fns';
 import { Template } from '@twilio/flex-ui';
 import { CallTypes } from 'hrm-form-definitions';
 
+import { getConfig } from '../../../HrmFormPlugin';
```
GPaoloni, line 7: Probably unused?

plugin-hrm-form/src/utils/setUpComponents.js
```diff
@@ -436,16 +436,19 @@ export const setupTwitterChatChannel = () => {
...
-  TwitterChatChannel.templates.IncomingTaskCanvas.firstLine = 'TaskHeaderLineTwitter';
+  /*
+   * modify TwitterChatChannel here
+   * TwitterChatChannel.templates.IncomingTaskCanvas.firstLine = 'TaskHeaderLineTwitter';
+   */
   TwitterChatChannel.templates.CallCanvas.firstLine = 'TaskHeaderLineTwitter';
   TwitterChatChannel.templates.TaskListItem.firstLine = 'TaskHeaderLineTwitter';
   TwitterChatChannel.templates.TaskCard.firstLine = 'TaskHeaderLineTwitter';
-  TwitterChatChannel.templates.TaskCanvasHeader.title = 'TaskHeaderLineTwitter';
+  // TwitterChatChannel.templates.TaskCanvasHeader.title = 'TaskHeaderLineTwitter';
```
GPaoloni, line 446: Can this be removed?

plugin-hrm-form/src/translations/en-US/flexUI.json
mythilytm: I think I can do something like this: https://www.twilio.com/docs/flex/developer/ui/localization-and-templating#templating-support-with-mustache-style-syntax
I am curious where would be a good place for this to live

GPaoloni (commented):
Looks and works mostly good, just some small comments and this
- 2nd line of task items still shows the identifiers
  Screenshot from 2022-11-04 14-15-08
- Same for message bubbles (but I see you are working on that)
  Screenshot from 2022-11-04 14-16-12

plugin-hrm-form/src/HrmFormPlugin.tsx
```diff
@@ -183,6 +185,15 @@ const setUpComponents = (setupObject: SetupObject) => {
...
 
   if (featureFlags.enable_canned_responses) Components.setupCannedResponses();
+
+  if (maskIdentifiers) {
+    const { strings } = getConfig();
+    strings.TaskInfoPanelContent = strings.TaskInfoPanelContentMasked;
+    Flex.MessagingCanvas.defaultProps.memberDisplayOptions = {
+      theirDefaultName: 'XXXXX',
+      theirFriendlyNameOverride: true,
+    };
```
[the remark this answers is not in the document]
mythilytm: I have to move the helpers for masking to Hrmformplugin.tsx because `setUpComponents` does not allow running `getConfig` as required by permissions configuration. This is the only snag I have left for implementing permissions

plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
mythilytm: I looked at different ways of implementing what you mentioned, but I am not able to pass a react component into `value`. I get the following error:
`Type 'Element' is not assignable to type 'string | number | boolean'.ts(2322)
SectionEntry.tsx(14, 3): The expected type comes from property 'value' which is declared here on type 'IntrinsicAttributes & Props & { children?: ReactNode; }'`

plugin-hrm-form/src/components/search/ContactPreview/ChildNameAndDate.tsx
mythilytm: ✅

plugin-hrm-form/src/translations/en-US/flexUI.json
nickhurlburt: I suspect this has been rewritten in Flex 2, so whatever we do now may need to be re-done as soon as Flex 2 becomes our primary branch. We should still make sure that the goal of this story (masking any PII) is achieved in this PR, but I would argue that we don't need anything especially well-factored knowing we'll have to update it very soon anyway.

plugin-hrm-form/src/HrmFormPlugin.tsx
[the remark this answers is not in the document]
GPaoloni: Why not? Also you could just use `setupObject`, which is basically the same object. 
I wonder, how will this work once this is moved to be a permission instead than a flag? :thinking:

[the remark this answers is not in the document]
mythilytm: Call for `getConfig()` interrupts `setUpComponents` and the application will nto load complemently with a lot of twilio components missing.

plugin-hrm-form/src/translations/en-US/flexUI.json
mythilytm: Sounds good. I will leave this for now. Should I add a ticket to update `TaskInfoPanel` across all translations once Flex 2.0 is merged?

nickhurlburt: Sure, yes. I think we can just call it a task vs a story; should be a quick thing.

stephenhand (approved):
Looks like everything is disabled by default so LGTM

PR #1003, opened 2022-10-27
CHI-1285: Fix welcome message

stephenhand:
Primary reviewer: @GPaoloni 

## Description

Fixed a couple of issues sending a welcome message, found because it broke the E2E tests

* Removed style from tabs that the new version of material was choking
* Migrated the Welcome Message listener & sender code to the Conversations API

### Checklist
- [ ] New tests added
- N/A Strings are localized

TODO: Some unit test coverage for AfterAcceptTask handler

### Related Issues
CHI-1285

### Verification steps

* Start a webchat conversation
* Check that the Welcome Message is sent to the caller when the counsellor accepts the chat.

### DIFF
--- a/plugin-hrm-form/src/styles/HrmStyles.tsx
+++ b/plugin-hrm-form/src/styles/HrmStyles.tsx
@@ -444,11 +444,6 @@ export const StyledSearchTab = withStyles({
   selected: {
     backgroundColor: '#ffffff',
   },
-  labelIcon: {
-    '& > span > span': {
-      padding: 0,
-    },
-  },
 })(StyledTab);
 StyledSearchTab.displayName = 'StyledSearchTab';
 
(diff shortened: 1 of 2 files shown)


GPaoloni (approved):
Code looks good!

>Migrated the Welcome Message listener & sender code to the Conversations API

I thought this changes would not be required for now.. Maybe the internals are the same (Programmable Chat) but is accessed via different API in the code?

stephenhand:
> Code looks good!
> 
> > Migrated the Welcome Message listener & sender code to the Conversations API
> 
> I thought this changes would not be required for now.. Maybe the internals are the same (Programmable Chat) but is accessed via different API in the code?

Their docs said the APIs should be backwards compatible, but in reality, the listeners in old API didn't seem to be firing at the expected points, whereas the ones in the Conversations API were working

PR #1009, opened 2022-11-01
CHI-1416 Handle temporary and permanent transcript errors

robert-bo-davis:
Primary reviewer: @GPaoloni 

## Description
This handles temporary and permanent transcript related errors separately. This has a sibling PR in techmatters/serverless#349 to simplify local dev setup.

### Checklist

### Related Issues
CHI-1416

### Verification steps
Missing aws keys in serverless .env: permanent
Bad aws key in serverless .env: permanent
Bad filename in db: permanent

switch out endpoint in serverless:/functions/getFileDownloadUrl.ts s3 client to induce service failure: temporary

```js
const s3Client = new AWS.S3({
    endpoint: 'http://google.com',
});
```


stephenhand (approved):

plugin-hrm-form/src/components/contact/TranscriptSection.tsx
```diff
@@ -35,13 +46,83 @@ const TranscriptSection: React.FC<Props> = ({
   myIdentity,
   transcript,
   loadTranscript,
+
+  // eslint-disable-next-line sonarjs/cognitive-complexity
```
robert-bo-davis, line 50: I got down to 16, but couldn't get below 15 without a lot more refactoring. 😦

plugin-hrm-form/src/components/contact/TranscriptSection.tsx
```diff
@@ -24,9 +25,19 @@ type OwnProps = {
   externalStoredTranscript?: S3StoredTranscript;
   loadConversationIntoOverlay: () => Promise<void>;
 };
+
 // eslint-disable-next-line no-use-before-define
 type Props = OwnProps & ReturnType<typeof mapStateToProps> & typeof mapDispatchToProps;
 
+class TranscriptFetchResponseError extends Error {
```
stephenhand, line 32: Added this comment when I originally approved but I mustn't have saved it 🤦 

Just asking if there's a particular advantage to having a class extending Error in here? Whilst using classes are fine our usual 'house style' is to prefer factories & type aliases - so in this case we would just add props to an Error instance and alias a union type if we needed to.

We can't filter catch logic by type in TS / JS (grr), but are you thinking we could use `instanceof` checks here to split error handling logic by error type, rather than having to write type guards like we would if we just had a type alias?

As I said, totally fine with the code how it is, just wanted to understand your thinking.

robert-bo-davis: Could you send a link to an example of the factory & type alias pattern? I'm not familiar with it.

Extending classes for exception handling is just what I'm used to. Yes, it allows for `instanceof` case/if splitting for exception handling. It also allows for simplified testing like [this](https://github.com/techmatters/hrm/pull/257/files#diff-fa5fcd4223299b8d44f882e99541fa020f3a06d44321ecb46fa93129871d4307R26).

stephenhand: Fair enough, tbf exceptions look to be one of the places in TS where inheriting classes is the better option, because building an error in a factory would put extra crap on the call stack too

robert-bo-davis: While looking for an example, I found [this](https://github.com/techmatters/hrm/blob/master/hrm-service/src/contact/contact-routes-v0.ts#L90)... which... yeah.

Like a lot of other languages, in my experience, extending exception classes and using `instanceof` is just kind of the way you do it even though js/ts won't filter catch blocks because the alternatives are all pretty icky.

robert-bo-davis: @stephenhand I've never seen that approach. Interesting! I'm also have only really ever used TS in a couple of personal projects so I have a *lot* to learn.

If you could point me in the direction of some examples of the "factories & type aliases" pattern in our code, I'd definitely like to understand the house style a bit more. I'm not sure I understand what you mean by factory in a context where their isn't any class inheritance.

robert-bo-davis: Ahhh... Is this [explanation of FP factory functions](https://chrisman.github.io/1.html) more in line with what you mean by "factory"?

stephenhand: Yes that's the kind of pattern I'm talking about, basically sidestep the JS prototypal inheritance hairball for the most part - in our code the 'Page Objects' in the E2E tests show the approach in our code pretty well, e.g.: https://github.com/techmatters/flex-plugins/blob/master/e2e-tests/webchat.ts 

We rarely have cause to use fancier patterns like mixins, simple construction time composition suits our purposes most of the time.

Also, we are judicious about when we chose to define object methods as opposed to using separate functions. Obvs the more OO structure encapsulates better, but can cause unexpected behaviours when copying & object content around using spread operators & such (which, you've no doubt already noticed, we use a LOT, another reason we don't use a lot of class structures).

But as I say, exception handling is one instance where JS prototypal inheritance wins, it removes a lot of boilerplate for no real drawbacks vs other approaches

robert-bo-davis: Thanks for the writeup and example! 

Of course, now that I've spent a bit of time researching to try to understand the benefits/risks, I discovered a potential TS transpilation issue. Possibly what @murilovmachado was [running into here](https://github.com/techmatters/infrastructure-config/blob/master/lambdas/src/lineSetup/invalidInputPayloadException.ts#L12).

It looks like there are some issues around extending builtins and [TS transpilation](https://github.com/microsoft/TypeScript/wiki/FAQ#why-doesnt-extending-built-ins-like-error-array-and-map-work) that mean we may want to avoid using inheritance for errors as well, at least in front end code, depending on the browsers we are targeting. At minimum we would need to use the prototype hackery in the constructor that they suggest in that FAQ.

plugin-hrm-form/src/services/ServerlessService.ts
```diff
@@ -169,7 +169,16 @@ export const getFileDownloadUrl = async (fileNameAtAws: string, fileName: string
  * Gets a file download url from S3, using the object url as constructed by AWS
  */
 export const getFileDownloadUrlFromUrl = async (objectUrl: string, fileName: string = undefined) => {
-  const [bucketName, fileNameAtAws] = objectUrl.replace('https://', '').split('.s3.amazonaws.com/');
+  let bucketName, fileNameAtAws;
+  [bucketName, fileNameAtAws] = objectUrl.replace('https://', '').split('.s3.amazonaws.com/');
```
GPaoloni, line 173:
```suggestion
  let [bucketName, fileNameAtAws] = objectUrl.replace('https://', '').split('.s3.amazonaws.com/');
```

GPaoloni (approved):
Code looks good!
I'm trying to run the local env to see the different errors in action.

PR #1010, opened 2022-11-02
feature: CSAM: Self-Generated Report UI - initial commit

acedeywin:

## Description

- New feature: CSAM: Self-Generated Report UI 
### Checklist
- [N/A] New tests added
- [N/A] Strings are localized

### Related Issues
Fixes CHI-1406

### Verification steps

- Click on the "Offline Contact Record" button
- Select any of the categories
- Click on the "External Reports" dropdown button 

Note: This is the mock implementation, so there is not much going on here. There are lots of commented codes that will be used in the integration task. You will have to ignore the mess. No test has been implemented yet.

Screenshot 2022-11-02 at 2 50 45 PM


stephenhand:
This looks like it could use some DRYing up. Since a large amount of the UI is the same between the two types of CSAM reports, we should think about how we can extend & reuse more of the existing CSAM components for the new type of report rather than having a completely separate UI with lots of duplication

Similarly, the redux code could be consolidated into a single CSAM state. We are never going to be filling in both different types of CSAM report at the same time for the same task, so we don't need 2 separate, but very similar sets of states / reducers / actions defined - instead we should have a single state that can represent a partially complete counsellor reported or self-reported CSAM.

stephenhand (changes requested):
This looks like it could use some DRYing up. 

Since a large amount of the UI is the same between the two types of CSAM reports, we should think about how we can extend & reuse more of the existing CSAM components for the new type of report rather than having a completely separate UI with lots of duplication

Similarly, the redux code could be consolidated into a single CSAM state. We are never going to be filling in both different types of CSAM report at the same time for the same task, so we don't need 2 separate, but very similar sets of states / reducers / actions defined - instead we should have a single state that can represent a partially complete counsellor reported or self-reported CSAM.

acedeywin:
@stephenhand I have refactored the code to be as DRY as possible.

plugin-hrm-form/src/components/tabbedForms/CSAMReportDropdown.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const CSAMReportDropdown: React.FC<Props> = ({ handleCSAMType, dropdown, setCSAMType }) => {
+  return (
+    <StyledCSAMReportDropdown style={{ display: dropdown ? 'block' : 'none' }}>
+      <StyledCSAMReportHeader>
+        <Template code="TabbedForms-CSAMReportButton" />
+      </StyledCSAMReportHeader>
+      <StyledCSAMReportDropdownList
+        onClick={() => {
+          setCSAMType(true);
```
stephenhand, line 23: I don't think there should have to be 2 separate steps to set the CSAM type and route. We should be able to route to the page to set the CSAM report with the information about which of type it is in a single operation. The injected handler should take a parameter for the type and build the route accordingly

Also, we shouldn't use a boolean for CSAM type. Booleans should only be used in cases where 'true' and 'false' actually make sense, not just as a substitute for 2 arbitrary values.  Something informative like an enum or a union of string values should be used here instead.

plugin-hrm-form/src/states/csam-report/reducer.ts
```diff
@@ -13,6 +13,7 @@ type CSAMReportState = {
   tasks: {
     [taskId: string]: TaskEntry;
   };
+  createLinkForChild: boolean;
```
stephenhand, line 16: A global boolean like this would cause issues if anyone were ever to be filing CSAM reports on 2 tasks simultaneously. As commented elsewhere, it shouldn't be needed though, we can handle the CSAM type in the routing logic

plugin-hrm-form/src/states/csam-report/types.ts
```diff
@@ -2,6 +2,7 @@
 export const UPDATE_FORM = 'csam-report/UPDATE_FORM';
 export const UPDATE_STATUS = 'csam-report/UPDATE_STATUS';
 export const CLEAR_CSAM_REPORT = 'csam-report/CLEAR_CSAM_REPORT';
+export const SET_CSAM_TYPE = 'SET_CSAM_TYPE';
 
 export type CSAMReportForm = {
```
stephenhand, line 7: We should have separate types for self-reported CSAM report and counsellor reported CSAM. We can have a union type for use in the functions but we should be using the type system to prevent people filling in properties for both types of report on the same instance

stephenhand (commented):
Few comments. 

Also we need a feature flag to be able to disable the new UI - it should toggle the rendering the new dropdown or the old button. This way we can roll out versions with this code but without the backend.

plugin-hrm-form/src/components/CSAMReport/CSAMReportFormDefinition.ts
```diff
@@ -65,6 +67,21 @@ export const definitionObject: CSAMFormDefinitionObject = {
     required: { value: true, message: 'RequiredFieldError' },
     maxLength: { value: 100, message: '100 characters max.' },
   },
+  childAge: {
```
stephenhand, line 70: I think the 2 types of CSAM reports are better with separate form definitions

acedeywin: I have refactored this.

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
```diff
@@ -88,6 +99,32 @@ export const CSAMReportScreen: React.FC<Props> = ({
     return Object.entries(definitionObject).reduce(reducerFunc, null);
   }, [firstElementRef, initialForm, methods, taskSid, updateFormAction]);
 
+  const childCSAMformElements = React.useMemo(() => {
```
stephenhand, line 102: Can we not have 1 form definition member, and we chose the definition object based on csamType?

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
```diff
@@ -52,6 +61,8 @@ export const CSAMReportScreen: React.FC<Props> = ({
   csamReportState,
   routing,
   counselorsHash,
+  csamType,
```
stephenhand, line 64: We still want csamType to be derived from the route rather than setting separately in redux

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
acedeywin: @stephenhand There's no way this can be achieved since we have separate form definition objects, and we still need to reference each form element within the CSAMReportFormScreen.tsx file. As I pointed out yesterday, we would also need different form element definitions since we have separate form definition objects for the csam types.

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
acedeywin: @stephenhand I am not sure I understand what you mean. Both `csamTypes` use the same route, which is `csam-report`.  Since the csamTypes can either be `self-report` or `counsellor-report`, the default type specified in the reducer is `counsellor-report` since that's the initial csam type that existed prior.

stephenhand: Both csam types use the same route now. But they don't have to. There's no reason why there couldn't be a `self-report-csam-report` and a `counsellor-csam-report` subroute. seems like this would eliminate the need for a global csamType in redux, which as was noted before, could cause issues.

You could still pass the csamType down to the CSAMReportDefinition component, just infer it from the route

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
stephenhand: We can use a union type here as well, pass one property that could be either type of definition down to CSAMReportFormScreen, then use a type guard / discriminator to determine which of the 2 types it is internally within CSAMReportFormScreen, which is where you need to know which of the 2 types it is

stephenhand: Or, more simply, we could still have the 2 separate properties in CSAMReportFormScreen, but here we have a single function that produces both definitions, in an object or an array. Just means you don't duplicate the whole function

plugin-hrm-form/src/components/case/ActionHeader.tsx
```diff
@@ -52,7 +54,18 @@ const ActionHeader: React.FC<Props> = ({
             />
           </CaseActionDetailFont>
         )}
-        {!added && (
+        {addedForCSAMCLC && (
```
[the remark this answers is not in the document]
acedeywin: @stephenhand This has been fixed.

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
acedeywin: @stephenhand It's done.

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
```diff
@@ -64,28 +73,31 @@ export const CSAMReportScreen: React.FC<Props> = ({
...
       const generatedInput = getInputType([], onUpdateInput)(e);
-      const initialValue = initialForm[e.name] === undefined ? initialValues[e.name] : initialForm[e.name];
+      const csamInitialValues = { ...initialValues, ...childInitialValues };
+      const initialValue = initialForm[e.name] === undefined ? csamInitialValues[e.name] : initialForm[e.name];
 
       return index === 0 ? generatedInput(initialValue, firstElementRef) : generatedInput(initialValue);
     };
 
     // Function used to generate the inputs with a reduce
+    const csamDefinitionObject = { ...definitionObject, ...childDefinitionObject };
```
stephenhand, line 90: Sorruy, this isn't what I was suggesting - apologies if you got that impression. My 2 suggestions were to either:

- use a union type to return an instance that could be the child definition OR the counsellor definition
- use a single function to return BOTH definitions, but both as separate objects, either as array items or object properties. i.e. `return [childReportDefinition, counsellorReportDefinition]` or `return {childReportDefinition, counsellorReportDefinition}`. You could then destructure them using `const [childReportDefinition, counsellorReportDefinition] = React.useMemo(() => { ...`

What you have here merges the two objects together into a combined blob.

stephenhand (commented):
The routing changes look good- it's just the form definition management and the CLC naming that need looking at now

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
acedeywin: @stephenhand option one always returns an empty object `{}`. For option 2, I am not sure how to go about that, but I implemented something that may be close.

stephenhand (approved):

stephenhand (commented):
Code looks ok to me now, I will test it in the morning

stephenhand:
Also, remember the story is not complete until this is merged up into Flex 2 - QA will be done on Flex 2

plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx
stephenhand: Option one wouldn't do that unless there was an issue with the implementation - but what you have is a version of option 2 which looks ok.

janorivera:
hey @acedeywin!  just wanted to say thanks for taking the time to update all the translation files!

PR #1011, opened 2022-11-02
UI styling for external transcript viewing

GPaoloni:
Primary reviewer: @mythilytm or @murilovmachado ?

## Description
This PR adds styles to kinda-match the Twilio ones when viewing external transcripts.

NOTE: There's one subtle detail that I have not addressed in this PR (kinda realized this just now), that is the missing "horizontal ruler with date" displayed in the Twilio chat. I'll add this one in a subsequent PR as that needs some extra handling of the messages list, but I'd like to get some early feedback on this styles if possible.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1343)
- [ ] New tests added
- [x] Strings are localized

### Verification steps
- `npm run dev`.
- Search for the contacts that have first name "Transcript".
- Open the one that contains an external transcript, and load.
- Compare the UI styles with the ones provided in the mocks.
- Play with the different places where we can search contacts.

### DIFF
--- a/plugin-hrm-form/src/components/contact/MessageItem.tsx
+++ b/plugin-hrm-form/src/components/contact/MessageItem.tsx
@@ -32,7 +32,9 @@ const MessageItem: React.FC<Props> = ({ isCounsellor, isGroupedWithPrevious, mes
       <MessageBubbleContainer isCounsellor={isCounsellor}>
         <MessageBubleInnerContainer>
           <MessageBubbleHeader>
-            <MessageBubbleNameText isCounsellor={isCounsellor}>{message.from}</MessageBubbleNameText>
+            <MessageBubbleNameText isCounsellor={isCounsellor}>
+              {message.friendlyName || message.from}
+            </MessageBubbleNameText>
             <MessageBubbleDateText isCounsellor={isCounsellor}>
               {format(new Date(message.dateCreated), 'hh:mm a')}
             </MessageBubbleDateText>
--- a/plugin-hrm-form/src/components/contact/MessageItem.tsx
+++ b/plugin-hrm-form/src/components/contact/MessageItem.tsx
@@ -16,31 +16,31 @@ import {
 import { TranscriptMessage } from '../../states/contacts/existingContacts';
 
 type Props = {
-  isCounsellor: boolean;
+  isCounselor: boolean;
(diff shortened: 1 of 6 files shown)


plugin-hrm-form/src/components/contact/TranscriptSection.tsx
```diff
@@ -44,21 +36,26 @@ const TranscriptSection: React.FC<Props> = ({
 
   // Preferred case, external transcript is already in local state
   if (transcript) {
+    const messagesToRender = transcript.messages.reduce<
+      (TranscriptMessage & { isCounsellor: boolean; isGroupedWithPrevious: boolean })[]
+    >((accum, m, index) => {
+      const isCounsellor = m.from.startsWith('gian');
+      // const isCounsellor = m.from === myIdentity;
```
GPaoloni, line 43: Since the only transcript visible at this moment ishas an outdated shape, this will force the counselor messages to be displayed accordingly. Before merging, I will cleanup this to be
`const isCounsellor = m.isCounsellor`
and the flag should be attached to each message when the transcript is exported (Lambda function).

plugin-hrm-form/src/components/contact/MessageItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <MessageItemContainer isCounsellor={isCounsellor} isGroupedWithPrevious={isGroupedWithPrevious}>
+      {!isCounsellor && (
+        <AvatarContainer isGroupedWithPrevious={isGroupedWithPrevious}>
+          {!isGroupedWithPrevious && <Icon icon="DefaultAvatar" />}
+        </AvatarContainer>
+      )}
+      <MessageBubbleContainer isCounsellor={isCounsellor}>
+        <MessageBubleInnerContainer>
+          <MessageBubbleHeader>
+            <MessageBubbleNameText isCounsellor={isCounsellor}>{message.from}</MessageBubbleNameText>
```
GPaoloni, line 35: Same as prior message, here I'm displaying the `m.from` which is the `identity` property of the Chat Channel User. When the transcript is exported, each message should have the friendly name (if any) attached to it, and that's what we will wnt to display here.

murilovmachado (approved):
The code and visuals look good to me!

PR #1019, opened 2022-11-08
CHI-970: Deduplicate flex deploy logic, improve deploy messages, bump Twilio CLI versions

stephenhand:
Review after #1018 

Primary reviewer: @GPaoloni 

## Description

* Route all deploys via the Custom Deploy workflow, so individual deploys are consistent with mass deploys
* Add the branch / tag to deployment messaging
* Update Twilio CLI versions in the deploy action
* Split

### Checklist
- N/A New tests added
- N/A Strings are localized

### Related Issues
Blocks CHI-970 (workflow deploys aren't working for flex 2.0)

### Verification steps

Use the individual & mass deploy workflows, then review the output to ensure they work as expected


.github/actions/main-action/action.yml
```diff
@@ -88,18 +90,23 @@ runs:
...
+      # Release flex plugin
+      - name: Run deploy command
         env:
-         TWILIO_ACCOUNT_SID: ${{ inputs.account-sid }}
-         TWILIO_AUTH_TOKEN: ${{ inputs.auth-token }}
-         UNBUNDLED_REACT: true
-        run: npm install twilio-cli -g && twilio plugins:install @twilio-labs/plugin-flex@4.3.4 && twilio flex:plugins:deploy --changelog "${{ github.event.inputs.changelog }} (commit ${{ github.sha }})"
+          TWILIO_ACCOUNT_SID: ${{ inputs.account-sid }}
+          TWILIO_AUTH_TOKEN: ${{ inputs.auth-token }}
+          UNBUNDLED_REACT: 'true'
```
GPaoloni, line 108: Is this still required for Flex 2? Or is it just for compatibility?

hrm-form-definitions/src/formDefinition/loadDefinition.ts
```diff
@@ -32,6 +36,28 @@ export enum DefinitionVersionId {
...
+      isDependentSelectDefinitionWithReferenceOptions(json) ||
+      isSelectDefinitionWithReferenceOptions(json)
+    ) {
+      const { optionsReferenceKey, ...restOfJson } = json;
+      const options = referenceData[optionsReferenceKey];
+      if (!options) {
+        throw new Error(
+          `Reference key '${optionsReferenceKey}' for form item '${json.name}' not found.`,
+        );
+      }
```
GPaoloni, line 54: Would it make sense to do some simple validation just to make sure this matches with type expected by `option`s (a record)? That could catch errors when updating the def itself, since the plugin would break as soon as it tries to load the definition.

GPaoloni: You already did in `hrm-form-definitions/src/specification/validate.ts` :D

.github/actions/main-action/action.yml
stephenhand: I'm not sure - I could try removing it? I don't use it locally on Flex 2 so it should be ok - I'll give it a go

hrm-form-definitions/src/formDefinition/loadDefinition.ts
stephenhand: Yes the test should do the more in-depth validation. I just thought a key name mismatch might be a common enough error scenario it would be helpful to call it out at runtime

PR #1021, opened 2022-11-08
CHI-1491: Fix canned responses

stephenhand:

## Description
* Fix canned responses by using conversation SID in the UI

### Checklist
- N/A New tests added
- N/A Strings are localized

### Related Issues
CHI-1419

### Verification steps

See ticket

### DIFF
--- a/plugin-hrm-form/src/components/CannedResponses.tsx
+++ b/plugin-hrm-form/src/components/CannedResponses.tsx
@@ -2,26 +2,23 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
 import { Actions, withTheme } from '@twilio/flex-ui';
+import { MessageInputChildrenProps } from '@twilio/flex-ui-core/src/components/channel/MessageInput/MessageInputImpl';
 
 import { getConfig } from '../HrmFormPlugin';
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/components/CannedResponses.tsx
```diff
@@ -2,26 +2,23 @@
...
 // eslint-disable-next-line no-use-before-define
-type Props = OwnProps;
+type Props = MessageInputChildrenProps;
 
 // eslint-disable-next-line react/display-name
 const CannedResponses: React.FC<Props> = props => {
   const cannedResponses = useSelector(selectCannedResponses);
   const { strings } = getConfig();
-  const { channelSid } = props;
+  const { conversationSid } = props;
```
GPaoloni, line 18: Does this means that `channelSid` is not available anymore? If so, we should also address this in the places where we use it (i.e. saving the contact).

GPaoloni (approved):
Works great!

plugin-hrm-form/src/components/CannedResponses.tsx
stephenhand: We should perhaps consider recording the conversationSid though

PR #1035, opened 2022-11-14
Cross platform tweaks for local E2E test runs

stephenhand:

## Description

The call to 'source' in the local E2E test run jobs wasn't cross platform, and I believe it was only required because variables were not loading correctly from the `.env` file in the node process via `dotenv`

This was due to an issue where a bug in `dotenv`s bash parsing was incorrectly interpreting the hash at the start of a value as a comment. I have worked around this by quoting the value in the source SSM parameter, so now the values are loaded by that package correctly and call to `source` can be removed 

* Removed 'source' call in local NPM jobs now it should be redundant.
* Limited local runs to 1 worker. 
* Added feature flag check on PR template

### Verification steps

Run `npm run test:local` from the e2e-test folder

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,6 +10,7 @@ Primary reviewer:
 ### Checklist
 - [ ] Corresponding issue has been opened
 - [ ] New tests added
+- [ ] Feature flags added
 - [ ] Strings are localized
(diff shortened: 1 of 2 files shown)


robert-bo-davis (approved):
Looks great to me! Good catch. I was able to test on MacOS and it still works there.

PR #1063, opened 2022-11-24
CSAM: Update Contact Details To Show Self Reported CSAM

GPaoloni:
Primary reviewer: @stephenhand or @acedeywin

This PR depends on https://github.com/techmatters/hrm/pull/268.

## Description
Adds logic to support issuing `self-generated` CSAM reports, store them in HRM and display in Contact Summary (tabbed forms) and Contact Details.

Screenshot from 2022-11-25 14-46-36
Screenshot from 2022-11-25 14-45-57

NOTE: 
- Instead than removing "oprhan CSAM records" in case of errors from external API, we could be add some retry logic, if considered worth it.
- This PR does not address the "report status screen", since that will depend on the external API rather than HRM.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1454)
- [WIP] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Start local hrm sitting in https://github.com/techmatters/hrm/pull/268, expose the endpoing (using `ngrok`?).
- `npm run dev`, and point to the local server (in Chrome console, `Twilio.Flex.Manager.getInstance().serviceConfiguration.attributes.hrm_base_url = <the url of step 1>`).
- Save some CSAM reports of both kinds (self generated and counsellor generated).
- Confirm that reports are properly stored, and linked to the contact once this is saved.
- Force an error in line 137 of `plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx` and confirm that the orphan report is removed from the DB if it can't be aknowledged.


plugin-hrm-form/src/components/SectionEntry.tsx
```diff
@@ -11,14 +11,34 @@ import { getConfig } from '../HrmFormPlugin';
 
 type Props = {
   description: React.ReactNode | string;
-  value: string | number | boolean;
+  value: React.ReactNode | string | number | boolean;
```
stephenhand, line 14: I wonder if SectionEntry is due a refactor? 

We could make it a compound component where its children are the value, then move the value's text formatting logic into the `SectionValueText` component (would need converting to a 'real' component, maybe rename it to something indicating it's a form value being displayed), so SectionEntry doesn't need to take a form definition or a layout if the value it's rendering doesn't need it. We could also move the `SectionValueText` / `DownloadFile` switch logic into the calling component.

It could also be changed from taking a description string / element to just taking a description key, with an optional parameter to set some additional styling.

Whilst we're at it, if these 'Section' components aren't used outside of the ContactDetails page we could move them into the components/contact directory

plugin-hrm-form/src/services/CSAMReportService.ts
```diff
@@ -1,23 +1,44 @@
...
+): Promise<CSAMReportEntry> => {
   const options = {
     method: 'POST',
     body: JSON.stringify(body),
   };
 
   return fetchHrmApi('/csamReports', options);
 };
+
+export const aknowledgeCSAMReport = async (reportId: CSAMReportEntry['id']): Promise<CSAMReportEntry> => {
```
stephenhand, line 28: This endpoint is a little unusual. PATCH normally indicates 'I'm going to supply the parts of the resource that I want to update in the payload' - so this case it would be something like `PATCH /csamReports/${reportId} BODY: { acknowledged: true }`.

If you are going for more of a 'command' type endpoint like the one you have here, I would normally make it a POST

This is very picky REST anorak stuff though

stephenhand (approved):
Couple of comments, no blockers though

plugin-hrm-form/src/services/CSAMReportService.ts
stephenhand: > Couple of comments, no blockers though

Well, the misspelling might be a blocker :-P

stephenhand:
Extra changes LGTM

PR #1066, opened 2022-11-28
CHI-1526: Optimised flex log proxying to only proxy errors. 

stephenhand:

## Description

The rollbar proxing change is to address performance issues observed since the migration to Flex 2, where large amounts of browser console logging lead to high resource usage & slow updates due to work being done when proxying console.log calls to the Rollbar SDK

* Updated Flex console.log proxying to only proxy 'console.error' to Rollbar, rather than all console mewthods, since the Rollbar client is configured to filter out all log events below 'error' level befopre uploading anyway
* Updated application logic directly driven from NODE_ENV to work off separate environment variables
* Updated webpack to use dotenv-webpack to combine .env files and CLI env vars into browser's 'process.env' 
* Added .env files with defaults based on NODE_ENV
* Converted setUpMonitoring to TS

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-1526

### Verification steps

* Add a 'console.error' log line to your local version
* Set 'ENABLE_MONITORING=true' for your command line environment
* Run the plugin locally
* Check the error is reported in Rollbar
* Run the plugin again without the 'ENABLE_MONITORING' environment variable dset
* Check the error is NOT reported


murilovmachado (approved):
Looks good!

plugin-hrm-form/webpack.config.js
```diff
@@ -1,17 +1,20 @@
...
    */
 
-  config.plugins.push(
-    new webpack.EnvironmentPlugin([
-      'REACT_HRM_BASE_URL',
-      'REACT_SERVERLESS_BASE_URL',
-    ]),
-  );
-
+  const envPath = isDev ? '.env.dev' : '.env.prod';
```
robert-bo-davis, line 9: Awesome!

Kindof a nit and probably, but I would generally try to match `.env.{env}` naming to NODE_ENV for more future flexibility ex: `.env.${process.env.NODE_ENV}`. 

I have used these fairly lightweight libraries to handle this kind of thing in the past: [dotenv-flow](https://www.npmjs.com/package/dotenv-flow) and [dotenv-flow-webpack](https://www.npmjs.com/package/dotenv-flow-webpack). They add some nice functionality to dotenv like automagically loading `.env.${process.env.NODE_ENV}` files and ordered loading of multiple .env files. So if you have a .env in your project root and a `.env.${process.env.NODE_ENV}`, the .env is loaded first and treated as the defaults and then any overrides in `.env.${process.env.NODE_ENV}` are applied. Comes in handy in my experience as things get more complex.

robert-bo-davis (approved):

plugin-hrm-form/webpack.config.js
robert-bo-davis: That example is a little confusing. `.env.${process.env.NODE_ENV}` would be `.env.development`, `.env.test`, `.env.production`, etc. Sorry still on first cup of coffee.

stephenhand: Love `dotenv-flow-webpack`, less code, more functionality, more cascadeyness. Migrated the PR to use it.

robert-bo-davis:
Migration to dotenv-flow LGTM!

PR #1067, opened 2022-11-28
Alejandro mt updating tf scripts

janorivera:
Primary reviewer:
@Humairaa127 

## Description

Small modifications to add channel modules.

### Checklist
- [x]  Corresponding issue has been opened
- [x]  Plan reviewed
- [x]  Changes applied

### Verification steps

Webchat studio flow and flex channels should have been created.


twilio-iac/mt-kellimni-staging/main.tf
```diff
@@ -85,14 +87,33 @@ module "taskRouter" {
...
-  source = "../terraform-modules/studioFlow/default"
+module twilioChannel {
+  for_each = local.twilio_channels
+  source = "../terraform-modules/channels/twilio-channel"
+  channel_contact_identity = each.value.contact_identity
+  channel_type = each.value.channel_type
+  pre_survey_bot_sid = module.chatbots.pre_survey_bot_sid
+  target_task_name = local.target_task_name
+  channel_name = "${each.key}"
+  janitor_enabled = !local.enable_post_survey
```
GPaoloni, line 98: Would it makes sense to leverage this one to be calculated within the module, or could be a case where this two are both `true` or both `false`?

janorivera: I am not so sure!, I will need to pass the enable_post_survey as a variable in the module and that seems a bit "out of place" or misleading, also this will require to modify the module and existing accounts that are already using this module, I don't think that the improvement is impactful enough to be worth the effort. Maybe an improvement for the future?

PR #1071, opened 2022-11-30
Refactor SectionEntry

GPaoloni:

## Description
This PR refactors SectionEntry component, as suggested in https://github.com/techmatters/flex-plugins/pull/1063#discussion_r1033368722. 
- Moves `Section` and `SectionEntry` components into `/common/forms/`. It makes sense for me to live there since this is only used to "consume forms", i.e. show in the UI data collected by different forms. Happy to reorganize if makes more sense somewhere else.
- Small change to `presentValue` function: now it takes in functions that help render in the proper way. This avoids us carrying around `strings` object where is not necessary, while keeping the `CasePrint` working as expected (because of the underlying library used, we can't pass in React components here, only plain strings).

Further suggestions on how to organize this code are welcome :)

>We could also move the SectionValueText / DownloadFile switch logic into the calling component.

I'm not addressing this here because `file-upload` input type could be, theoretically, used in any customizable form in our app. Leveraging that responsibility to "the caller" would be messier in my opinion. On the other hand, I have not seen the form upload input being used anywhere but in the "Upload Document" part of cases. Is it worth keeping this as a potential customizable input?

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1540)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Start development server.
- Confirm that ContactDetails screen looks the same as it did before.
- Confirm that CasePrint works in the same manner as it did before.
- General regression test.


plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
```diff
@@ -300,11 +294,14 @@ const ContactDetailsHome: React.FC<Props> = function ({
                     <Template code="Category" /> {index + 1}
```
stephenhand, line 294: Wonder if an optional 'descriptionStyle' prop addresses this case better than allowing 'description' to have a node? Then we could rename 'description' to 'descriptionKey' to make it super clear a key is expected?

stephenhand (approved):
LGTM, minor Qs

PR #1091, opened 2022-12-08
CHI-1560: Validate all form inputs in tests

stephenhand:
Primary reviewer: @GPaoloni 

## Description

* Fixed form validation tests so basic validations are run on ALL form inputs, not just ones that are defined in the base specification
* Added explicit check that the input type is supported to form definition validation tests
* Converted form definition input 'type' from a union string literal to a string enum, so the values can be iterated in tests
* Split each form definition being validated into a separate test, so failure reports are more granular & useful
* Fixed issue with zw IncidentForm.json

### Checklist
- [X] Corresponding issue has been opened
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized
- [X] Tested for chat contacts
- N/A Tested for call contacts

### Related Issues
Fixes CHI-1560

### Verification steps

Automated PR checks validate these changes

PR #1097, opened 2022-12-09
Remove button workaround

stephenhand:
Review after #1091

Primary reviewer: @GPaoloni 

## Description

Remove workaround for issue introduced in v2.0.0-beta.3 that seems to be fixed in v2.0.0 GA


murilovmachado (approved):
LGTM

PR #1105, opened 2022-12-09
[Bug Fix] "Next" button for data form is pushed off of the screen

GPaoloni:
Primary reviewer: @stephenhand @mythilytm @murilovmachado 
(Mentioning multiple ppl cause maybe you have a smarter idea about fixing this one).

## Description
This PR fixes the bug mentioned in the Jira card, where the "previous contacts banner" will push down the content of the tabbed forms, causing the "Next" button to hidden under a scroll.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1552)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- `npm run dev`.
- Start a live contact form a number/ip that already exists in the DB.
- Confirm that the button bar is not being pushed below the limits of the screen.

### DIFF
--- a/plugin-hrm-form/src/components/TaskView.tsx
+++ b/plugin-hrm-form/src/components/TaskView.tsx
@@ -17,6 +17,7 @@ import PreviousContactsBanner from './PreviousContactsBanner';
 import { Flex } from '../styles/HrmStyles';
 import { isStandaloneITask } from './case/Case';
 import { getHelplineToSave } from '../services/HelplineService';
+import { YellowBannerHeight } from '../styles/previousContactsBanner';
 
 type OwnProps = {
   task: CustomITask;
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/components/TaskView.tsx
```diff
@@ -84,12 +85,22 @@ const TaskView: React.FC<Props> = props => {
...
       {isFormLocked && <FormNotEditable />}
-      <HrmForm task={task} featureFlags={featureFlags} />
+      <Flex
+        flexDirection="column"
+        style={{
+          // This fixes a UI bug where the PreviousContactsBanner pushes the container down
+          height: enablePreviousContacts ? `calc(100% - ${YellowBannerHeight}` : '100%',
+          width: '100%',
+        }}
+      >
```
murilovmachado, line 101: I believe it can work like this:
```
<Flex
        flexDirection="column"
        style={{
          // This fixes a UI bug where the PreviousContactsBanner pushes the container down
          height: '100%',
          width: '100%',
          overflow: 'auto',
        }}
      >
```

murilovmachado (approved):
Visually it works perfectly. But I think we can achieve the same result without calculating the height (that forces us to keep track of the yellow banner height).

I've posted down below how to do this. Let me know if that works, if not, I'm fine with merging this PR as it is.

plugin-hrm-form/src/components/TaskView.tsx
GPaoloni: It works! I have tried so many different things, can't believe this one was not in the list haha :facepalm: 
Thank you!!

PR #1110, opened 2022-12-12
adding terraform lock files

janorivera:
Primary reviewer:
@robert-bo-davis 

Adding terraform lock files.


robert-bo-davis (approved):
LGTM. Thanks!

PR #1115, opened 2022-12-13
CHI-1584-warm_transfers_fixes

stephenhand:
Primary reviewer: @GPaoloni 

## Description

* Fix warm transfer to show Accept / Reject buttons in handoff. 
* Fix Accept / Reject buttons to get task correctly

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-1584

### Verification steps

* Regression test warm call transfers
* Regression test other transfers?

### DIFF
--- a/plugin-hrm-form/src/utils/transfer.ts
+++ b/plugin-hrm-form/src/utils/transfer.ts
@@ -36,7 +36,7 @@ export const shouldShowTransferButton = (task: ITask) =>
   (!isTransferring(task) || hasTaskControl(task));
 
 export const shouldShowTransferControls = (task: ITask) =>
-  !isOriginalReservation(task) && isTransferring(task) && hasTaskControl(task) && TaskHelper.isTaskAccepted(task);
+  !isOriginalReservation(task) && isTransferring(task) && !hasTaskControl(task) && TaskHelper.isTaskAccepted(task);
 
 /**
  * Indicates if the current counselor has sole control over the task. Used to know if counselor should send form to hrm backend and prevent the form from being edited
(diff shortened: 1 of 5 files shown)


plugin-hrm-form/src/utils/transfer.ts
```diff
@@ -36,7 +36,7 @@ export const shouldShowTransferButton = (task: ITask) =>
   (!isTransferring(task) || hasTaskControl(task));
 
 export const shouldShowTransferControls = (task: ITask) =>
-  !isOriginalReservation(task) && isTransferring(task) && hasTaskControl(task) && TaskHelper.isTaskAccepted(task);
+  !isOriginalReservation(task) && isTransferring(task) && !hasTaskControl(task) && TaskHelper.isTaskAccepted(task);
```
stephenhand, line 39: Really not 100% sure about this change, but the 'sidWithTaskControl' doesn't get set until the receiver accepts the transfer, which is when we want the buttons to be dismissed, so I'm thinking we actually want the buttons showing _before_ we take control of the task?

Please validate!

(it appears to work though...)

stephenhand:
Can't really wait to merge this - please review retrospectively though

murilovmachado:
The buttons are showing up alright to me, approved ✅ . 

One thing that's funny is that in the UI it seems the original counselors are not being removed when warm transferring (even though they are). We end up with things like these:

image

This picture shows 5 people in a call, where there are actually only 2. Looking at the task attributes on Twilio, things look right. But there may be other things not 100% right on the Twilio side. But this is not affecting the overall behavior, just this misleading UI, that seems to indicate there're other counselors (even repeated ones). I'd say it's not related to this PR. Just registering so we can tackle this later on.

PR #1122, opened 2022-12-15
Create FeatureFlags type with descriptions

murilovmachado:
Primary reviewer: @stephenhand 

## Description
Creates the type FeatureFlags, with a short description of each feature flag.

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-1559

PR #1114, opened 2022-12-13
Move task wrapup and post survey trigger logic from Flex to Serverless

GPaoloni:
This PR is related to https://github.com/techmatters/serverless/pull/373

Primary reviewer: @stephenhand 

## Description
This PR:
- Wraps post survey triggering under a temporary feature flag `post_survey_serverless_handled`. If it's enabled, it means that the triggering logic will be run from serverless (see related PR). If not, we'll continue to support the existing behavior.
- Moves the logic to unsubscribe counselor from "post survey - child interactions" from [ActionsManager framework](https://assets.flex.twilio.com/docs/releases/flex-ui/2.0.0/ui-actions/ActionsManager/) to the newly introduced [FlexEvents APIs](https://assets.flex.twilio.com/docs/releases/flex-ui/2.0.0/ui-actions/FlexEvent/), which are event listeners for TarkRouter events.
- Commit [592febe](https://github.com/techmatters/flex-plugins/pull/1114/commits/592febe909e75a41e29c0ab3970d8597e5b8a235) pins Node engine to `^14.0` in `package.json`. This is the same one specified in `.nvmrc`, but grants some benefits for those using a custom zsh plugin, which will turn red if you are not in the correct version. This comes handy to me, as I introduced some wrong builds because I didn't realize I was installing some dev dependency using Node 16, which caused `package-lock.json` to be recalculated, causing [build errors](https://github.com/techmatters/flex-plugins/actions/runs/3698859431/jobs/6265605295). I can rollback this change if not desired.
  Screenshot from 2022-12-14 23-48-50
  Screenshot from 2022-12-14 23-49-02

### Checklist
- [x] [Corresponding issue has been opened
](https://tech-matters.atlassian.net/browse/CHI-1479)
- [x] New tests added
- [x] Feature flags added
- [n/a] Strings are localized
- [x] Tested for chat contacts
- [n/a] Tested for call contacts

### Verification steps
- Make sure that https://github.com/techmatters/serverless/pull/373 is deployed.
- Update feature flags to contain `post_survey_serverless_handled = false`. Can be done from `flex-plugins/scripts` folder like
  `❯ npm run twilioResources patch-feature-flags -- -f post_survey_serverless_handled:false`.
- `npm run dev`.
- Confirm that everything works as expected:
  - Post survey is triggered when the counselor ends the chat.
  - The counselor stops receiving messages when the task is in wrapup mode.
[...]

PR #1123, opened 2022-12-15
Add post-survey-serverless feature flag to types

nickhurlburt:
Primary reviewer: @GPaoloni 

## Description
It looks like the combination of #1122 and #1114 has resulted in a build failure, because the flag `post_survey_serverless_handled` introduced in #1114 was not added to the types in #1122. This should resolve that issue. I am not sure exactly how to test this, but I suspect that if it builds, it's okay.


mythilytm (approved):
Works great. The breaking change works locally as well now

PR #1133, opened 2022-12-20
CHI-1275: Search components minor rename & reorg

stephenhand:
Primary reviewer: @GPaoloni 

## Description

* Renames the contact preview header for consistency with the case equivalent
* Moves some shared preview components to top level
* Renames Case / Contact preview components to be module entry points

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-1275

### Verification steps

* Automated tests should be sufficient


plugin-hrm-form/src/components/search/CasePreview/index.tsx
```diff
@@ -28,7 +28,7 @@ const connector = connect(mapStateToProps);
 
 type Props = OwnProps & ConnectedProps<typeof connector>;
 
-const CasePreview: React.FC<Props> = ({ currentCase, onClickViewCase, counselorsHash, definitionVersions }) => {
+const Index: React.FC<Props> = ({ currentCase, onClickViewCase, counselorsHash, definitionVersions }) => {
```
GPaoloni, line 31: Why is this change? Couldn't it preserve the name even if moved to `index.tsx`?

GPaoloni (approved):
Looks good (and cleaner)!

plugin-hrm-form/src/components/search/CasePreview/index.tsx
stephenhand: No, that was just Webstorm getting a little overly enthusiastic with auto-renaming - reverted

PR #1138, opened 2022-12-21
Add ability to save Contact ID to Insights

GPaoloni:

This PR is related to https://github.com/techmatters/serverless/pull/381.

WIP:
- [x] CI
- [ ] Add tests

## Description
This PR adds support for storing contact id as part of the task attributes, to make it available in Insights.(*)
It also makes available any other properties from the contact as saved in the DB.
This can be referenced in the "one to many" insights configs like
```
{
    "attributeName": "conversation_attribute_10",
    "insightsObject": "conversations",
    "paths": [
      "savedContact.id"
    ]
  }
```
Note that `savedContact.id` is the path to access the saved contact id, since `applyCustomUpdate` (`plugin-hrm-form/src/services/InsightsService.ts`) is now updated like
`const dataSource = { taskAttributes, contactForm, caseForm, savedContact };`.

It also updates the demo form definitions to store the contact id under `conversations_attribute_10`.

(*) To achieve this, we need to have a final "set task attributes" before completing it, but after the contact was saved in HRM. 
On the other hand, since the creation of a contact in HRM depends on us providing a unique task sid, we still have a dependency on a valid Twilio task being available at submission time.
Above constrains implies that:
- For live contacts, the only thing we need to do is "change the order" of the operations. The task was available the whole time, so we just save the contact, update the task attributes, and proceed to complete the task.
- For offline contacts, this required a bit more of adjustments. Previously, we assumed we had everything needed for the "final task attributes" even before saving the contact, so the task could just receive the attributes, be assigned and completed all within the same operation. Now we need to:
  - Create the task, and assign to the corresponding worker.
  - Submit the form in HRM.
  - Generate the final task attributes, update it and then complete it.
  Since some things can go wrong in this process, we also need to account for a cleanup state to prevent the user getting stuck with "stale tasks" that are created but never completed/canceled. For this purpose, in the related PR, we are also adding a short life span for the offline-contact-tasks.

This changes make the offline contact assignment process a bit more time consuming since an extra call to a Serverless function is needed.
I think this implementation suffices for the time given, but is also a good moment to think if our design is correct.
Some questions to think about:
- Do we really want to make the contacts to dependent on a unique task sid being provided?
- Could we use/generate a unique identifier in some other way? 
- Could we inverse the order, and make tasks dependent of the contact id? If so, offline contacts assignment would be back to a single call to Serverless since we could create-assign-complete the task all after the contact is already there.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1498)
- [WIP] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- `npm ci` & `npm start`.
- Make sure that you are using https://github.com/techmatters/serverless/pull/381
  - If you are using remote, confirm that's deployed.
  - If you are using local server, you can `npm start` serverless repo, expose the port via `ngrok http <PORT>` and then point to it in Flex like `Twilio.Flex.Manager.getInstance().serviceConfiguration.attributes.serverless_base_url = 'https://the-url-you-got-from-ngrok'`. Remember to provide `.env` file with at least the following vars (look at the values in Twilio Console -> Serverless -> .env): `TWILIO_WORKSPACE_SID`, `TWILIO_CHAT_TRANSFER_WORKFLOW_SID`, `ACCOUNT_SID`, `AUTH_TOKEN`.
- Submit  contacts (live and offline).
  - Everything should work as expected and with no errors. Contact should be submitted in DB.
  - In Twilio Console -> Taskrouter -> Tasks, find the corresponding task to the contact you just saved and make sure that the id of the submitted contact is properly stored under `conversations.conversations_attribute_10`.

### DIFF
--- a/plugin-hrm-form/src/services/InsightsService.ts
+++ b/plugin-hrm-form/src/services/InsightsService.ts
@@ -1,5 +1,5 @@
 /* eslint-disable camelcase */
-import { get, cloneDeep } from 'lodash';
+import { get, cloneDeep, merge } from 'lodash';
 import {
   DefinitionVersion,
   callTypes,
@@ -16,7 +16,7 @@ import { isNonDataCallType } from '../states/ValidationRules';
 import { mapChannelForInsights } from '../utils/mappers';
 import { getDateTime } from '../utils/helpers';
 import { TaskEntry } from '../states/contacts/reducer';
-import { Case, CustomITask } from '../types/types';
+import { Case, CustomITask, Contact } from '../types/types';
 import { formatCategories } from '../utils/formatters';
 import { getDefinitionVersions } from '../HrmFormPlugin';
 import { shouldSendInsightsData } from '../utils/setUpActions';
@@ -68,6 +68,7 @@ type InsightsUpdateFunction = (
   attributes: TaskAttributes,
   contactForm: TaskEntry,
   caseForm: Case,
+  savedContact: Contact,
 ) => InsightsAttributes;
 
 const sanitizeInsightsValue = (value: string | boolean) => {
@@ -318,10 +319,10 @@ export const processHelplineConfig = (
 };
 
 const applyCustomUpdate = (customUpdate: OneToManyConfigSpec): InsightsUpdateFunction => {
-  return (taskAttributes, contactForm, caseForm) => {
+  return (taskAttributes, contactForm, caseForm, savedContact) => {
     if (isNonDataCallType(contactForm.callType)) return {};
 
-    const dataSource = { taskAttributes, contactForm, caseForm };
+    const dataSource = { taskAttributes, contactForm, caseForm, savedContact };
     // concatenate the values, taken from dataSource using paths (e.g. 'contactForm.childInformation.province')
     const value = customUpdate.paths.map(path => sanitizeInsightsValue(get(dataSource, path, ''))).join(delimiter);
 
@@ -353,6 +354,7 @@ export const mergeAttributes = (
 ): TaskAttributes => {
   return {
     ...previousAttributes,
+    ...newAttributes,
     conversations: {
       ...previousAttributes.conversations,
       ...newAttributes.conversations,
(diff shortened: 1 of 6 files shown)


stephenhand:
I wouldn't lose too much sleep over the extra call. I would put this on the 'stuff that gets fixed / easier once we start creating contacts in HRM straight away' - if we're saving the contact along the way, when the time comes to put it into insights (assuming we want to do it at the same point), we will already have an ID.

At that point we can just use the contactID for idempotence and we can just record the taskID along with the other data (I think? not 100%)

plugin-hrm-form/src/services/formSubmissionHelpers.ts
```diff
@@ -37,19 +36,35 @@ export const completeContactlessTask = async (task: CustomITask) => {
...
+      const finalAttributes = buildInsightsData(inBehalfTask, contactForm, caseForm, savedContact);
+      await assignOfflineContactResolve({
+        action: 'complete',
+        taskSid: inBehalfTask.sid,
+        finalTaskAttributes: finalAttributes,
+      });
+      return savedContact;
+    } catch (err) {
+      // If something went wrong remove the task for this offline contact
+      await assignOfflineContactResolve({ action: 'remove', taskSid: inBehalfTask.sid });
```
stephenhand, line 58: We might want to consider not awaiting this - if something goes wrong with assignOfflineContactResolve, this is the error that gets throw, rather than the original, which is probably more useful?

plugin-hrm-form/src/types/types.ts
```diff
@@ -125,6 +125,26 @@ export type ContactRawJson = {
   conversationMedia: ConversationMedia[];
 };
 
+export type Contact = {
```
stephenhand, line 128: We will have to merge this with the one in #1136 - shouldn't be too hard. They have different names, but I think we've implemented their use in different places, so the merge shouldn't be too horrible

stephenhand (approved):
Not had chance to test yet (is there a serverless component or does it just use existing fns? But code looks good

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -380,16 +382,16 @@ const getInsightsUpdateFunctionsForConfig = (
  * Note: config parameter tells where to go to get helpline-specific tests.  It should
  * eventually match up with getConfig().  Also useful for testing.
  */
-export const buildInsightsData = (task: CustomITask, contactForm: TaskEntry, caseForm: Case) => {
-  const previousAttributes = task.attributes;
+export const buildInsightsData = (task: CustomITask, contactForm: TaskEntry, caseForm: Case, savedContact: Contact) => {
```
stephenhand, line 385: @GPaoloni  - sorry, didn't spot this before it was merged, but can we not do away with `contactForm` here now? Since I'm assuming it's basically the same data as `savedContact` without the ID? 

Not only does it cut down on some duplication / complexity, but relying on the form TaskEntry type less gives us less work to do in the future when we want to be always running off a redux layer containing saved contacts

PR #1143, opened 2022-12-23
CHI-1616 tf secret management

robert-bo-davis:
## Description

This adds a [secret manager system](https://github.com/techmatters/flex-plugins/pull/1143/files#diff-d46864bd173499ed0b0f99a8681222526112a6661c707e6c80b2260c57b6c0c4R1) that is meant to be run the first time a helpline is configured, but that can be run anytime to manage a set of secrets that are stored in SSM. These SSM secrets replace the `.private.tfvars` that must exist on each user's local machine.

These secrets include the twilio tokens for the relevant account and those stored tokens can be used whenever we need twilio access. This eliminates the need for a user to keep them in their environment variables.

It also adds a single `make setup-new-environment` command that reduces the number of individual commands a user has to run to setup a new environment.

This *does* require a move to the make/docker/tf setup for all users. It still needs to be tested on windows.

The parts related to serverless_url can be ignored as they are removed in #1144.

~#1144, #1145, and #1151 should be merged before this PR~

~The `-d` flag in `twilio-iac/scripts/twilioResourceImportAccountDefaults.sh` should be removed before this is merged.~

### Checklist
- [ ] Corresponding issue has been opened

### Related Issues
Fixes CHI-1616

### Verification steps

Run `make manage-ssm-secrets` from a helpline dir. Add the appropriate SSM params. Then run `make plan`.

~This also adds a `make setup-new-environment` command that reduces the steps required to spin up a new env. It is configured for a dry run currently, so it is safe to run that command.~

PR #1144, opened 2022-12-23
CHI-1616 automate serverless url data

robert-bo-davis:
## Description

This adds an external data source via python script that is able to gather data about serverless environments so that we don't have to swap out a manually managed `serverless_url` variable during account setup.

It also refactors internal naming of twilio credentials to be a bit more explicit and handles some changes that needed to be made after merging down from master.

This should be merged before #1143 and after #1145 and #1151

### Checklist

### Related Issues
Fixes CHI-1616


twilio-iac/makefiles/scripts.make
```diff
@@ -1,13 +1,10 @@
-setup-new-environment: manage-ssm-secrets init twilio-resources-import-account-defaults apply noop-serverless-url manage-ssm-secrets apply
+setup-new-environment: manage-ssm-secrets init twilio-resources-import-account-defaults apply
 
 manage-ssm-secrets:
 	docker run -it --rm $(DEFAULT_ARGS) $(DOCKER_IMAGE):$(TF_VER) ../scripts/secretManager/manageSecrets.py $(MY_ENV)
 
 twilio-resources-import-account-defaults:
 	docker run -it --rm $(DEFAULT_ARGS) $(DOCKER_IMAGE):$(TF_VER) ../scripts/twilioResourceImportAccountDefaults.sh $(MY_ENV)
```
GPaoloni, line 7: Is it possible that this was the step that failed with MT prod? I mean, is it possible that this command being run in this way didn't succeeded as expected when running from the local terminal, or that the failure input was missed?

twilio-iac/makefiles/scripts.make
```diff
@@ -1,13 +1,10 @@
...
 
-get-serverless-url:
-	docker run -it --rm $(DEFAULT_ARGS) $(DOCKER_IMAGE):$(TF_VER) ../scripts/getServerlessUrl.py $(MY_ENV)
-
 noop-serverless-url:
 	docker run -it --rm $(DEFAULT_ARGS) $(DOCKER_IMAGE):$(TF_VER) ../scripts/noop/getServerlessUrl.sh $(MY_ENV)
```
GPaoloni, line 10: Is this noop command used?

twilio-iac/makefiles/scripts.make
```diff
@@ -1,13 +1,10 @@
-setup-new-environment: manage-ssm-secrets init twilio-resources-import-account-defaults apply noop-serverless-url manage-ssm-secrets apply
+setup-new-environment: manage-ssm-secrets init twilio-resources-import-account-defaults apply
```
GPaoloni, line 1: `make setup-new-environment` would run all this commands subsequently? Or how is this one supposed to be used?

twilio-iac/terraform-modules/serverless/default/outputs.tf
```diff
@@ -8,4 +8,9 @@ output "serverless_service_sid" {
 output "serverless_environment_production_sid" {
   description = "Twilio SID of the 'production' environment under above service"
   value = twilio_serverless_services_environments_v1.production.sid
+}
+
+output "serverless_environment_production_url" {
+  description = "Twilio URL of the 'production' environment under above service"
+  value = "https://${data.external.service_environment_production.result.domain_name}"
 }
```
GPaoloni, line 16: :sunglasses: :rocket:

GPaoloni (approved):
Looks REALLY good to me, thanks for removing the probably-most-annoying step for new accounts.
Just leaving a couple of non-blocking questions.
A few more: 
- Do we need to take any actions before running this in an existing account, or the new `data "external" "service_environment_production"` will be computed as expected when we execute the `plan`/`apply`?
- This changes mean that we don't need to provide mocked serverless url in the first apply, nor update to the real one after it, right? If so, could you update the readme accordingly?

twilio-iac/makefiles/scripts.make
robert-bo-davis: This code wasn't in place yet when we set up MT prod. I think I missed this step when setting up MT prod. Sorry about that!

twilio-iac/makefiles/scripts.make
robert-bo-davis: Nope! Good catch. I removed it.

twilio-iac/makefiles/scripts.make
robert-bo-davis: Correct. Make will run "batch" targets like that sequentially. Which allows us to do fancy thing like this.

robert-bo-davis:
>Do we need to take any actions before running this in an existing account, or the new data "external" "service_environment_production" will be computed as expected when we execute the plan/apply?

The external will "just work" without any need for changes. ;)

twilio-iac/terraform-modules/serverless/default/getServerlessServiceEnvironmentData.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
```
robert-bo-davis, line 1: @stephenhand I want to make sure you are aware of the external data source pattern implemented in this and the next 2 files in case it could be useful as a work around for the limitations of the twilio provider in the future.

robert-bo-davis: The limitation I mean is the fact that the twilio provider only ever returns the SID of any resources it provisions and doesn't provide any data sources. This is a fairly simple pattern for working around that and implementing our own data sources inside of terraform.

PR #1145, opened 2022-12-26
CHI-1617 Github action for terraform plan and role based terraform

robert-bo-davis:
Primary reviewer:

## Description
This sets terraform up for role based privilege escalation and adds a github workflow to run terraform plan.

This depends on techmatters/infrastructure-config#105

This should be merged after #1151 and #1153
This should be merged before #1144

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #CHI-1617

### Verification steps

See test run here: https://github.com/techmatters/flex-plugins/actions/runs/3795028591 The `plan.log` artifact is the human readable diff. Secrets should be properly marked as sensitive since they are pulled from SSM and TF will prevent their accidental display.


.github/workflows/terraform_plan.yml
```diff
@@ -0,0 +1,45 @@
...
+        working-directory: ./twilio-iac/${{ inputs.helpline }}
+        run: |
+          echo "Running terraform init for ${{ inputs.helpline }}"
+          terraform init > /dev/null
+          echo "Running terraform refresh for ${{ inputs.helpline }}"
+          terraform refresh -input=false > /dev/null
+          echo "Running terraform plan for ${{ inputs.helpline }}"
+          terraform plan -input=false -refresh=false -out=/tmp/tfplan > /dev/null
+          echo "Running terraform show for ${{ inputs.helpline }}"
+          terraform show -no-color /tmp/tfplan > /tmp/plan.log
```
robert-bo-davis, line 40: Secrets will be marked sensitive by TF as long as they are being pulled from SSM, so shouldn't show up in any plans. If we wanted to be doubly careful, we could encrypt plan.log and require the end user to decrypt it.

.github/workflows/terraform_plan.yml
```diff
@@ -0,0 +1,45 @@
...
+        description: Directory name of helpline to run plan against
+        required: true
+        type: string
+env:
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  AWS_DEFAULT_REGION: us-east-1
+
+jobs:
+  deploy:
```
GPaoloni, line 19: Maybe not the appropriate job name? :P

.github/workflows/terraform_plan.yml
```diff
@@ -0,0 +1,45 @@
...
+name: Terraform Plan
+
+# Controls when the action will run.
+on:
+  workflow_dispatch:
+    inputs:
+      helpline:
+        description: Directory name of helpline to run plan against
+        required: true
+        type: string
```
GPaoloni, line 12: Not necessarily now, but it might be good for the solutions engs if we provide a dropdown with the options for this matter (this will add more burden to "adding new helpline" too, as this file must be updated).

scripts/src/terraformSupplemental/setEnvFromSsm.ts
```diff
@@ -1,9 +1,28 @@
-import { SSM } from 'aws-sdk';
+import { SSM, STS } from 'aws-sdk';
 
 export const setEnvFromSsm = async (accountDirectory: String): Promise<void> => {
-    const ssm = new SSM();
+    const sts = new STS();
+
+    const timestamp = (new Date()).getTime();
+    const params = {
+      RoleArn: 'arn:aws:iam::712893914485:role/tf-twilio-iac-ssm-ro',
```
GPaoloni, line 8: What does the `-ro` suffix means here?

twilio-iac/aarambh-production/main.tf
```diff
@@ -11,6 +11,14 @@ terraform {
...
     dynamodb_table = "terraform-locks"
     encrypt        = true
+    role_arn       = "arn:aws:iam::712893914485:role/tf-twilio-iac-production"
+  }
+}
+
+provider "aws" {
+  assume_role {
+    role_arn     = "arn:aws:iam::712893914485:role/tf-twilio-iac-${lower(var.environment)}"
+    session_name = "tf-${basename(abspath(path.module))}"
```
GPaoloni, line 21: Does it makes sense for to have `tf-twilio-iac-${lower(var.environment)}` or `tf-twilio-iac-production` in both places here? Same questions applies for all the analogous changes.

twilio-iac/scripts/secretManager/secretManager/ssm_client.py
```diff
@@ -1,5 +1,29 @@
...
+        return client
+
+    ts = time.time()
+
+    stsClient = boto3.client("sts")
+
+    response = stsClient.assume_role(
+        RoleArn="arn:aws:iam::712893914485:role/tf-twilio-iac-ssm-admin",
+        RoleSessionName="secret-manager" + str(ts),
+    )
```
GPaoloni, line 20: This role is only for SSM setup?

GPaoloni (approved):
Looks good to me!

Silly question: In the description it says that
>and adds a github workflow to run terraform apply

Is just for `plan`ning tho, right?

scripts/src/terraformSupplemental/setEnvFromSsm.ts
robert-bo-davis: In the assume role based IAM privilege escalation pattern a running process should always be run with only the privileges it needs. In this case the `tf-twilio-iac-ssm-ro` role is a read only role for `tf-twilio-iac-ssm` as opposed to the `tf-twilio-iac-ssm-admin` role that has read/write permissions.

twilio-iac/aarambh-production/main.tf
robert-bo-davis: Do you mean matching the session name to the role name?

The role needs to be environment specific. I *believe* that you can't have duplicate session names. So if we used `tf-twilio-iac-production` for the session name then our session names will act as a secondary lock preventing folks from running two staging or production terraform plan/apply steps at the same time even though dyanmodb would allow it.

Using the module path means the session matches the locking behavior of dynamodb locks.

robert-bo-davis:
>Is just for planning tho, right?

oops. yes. Fixed

scripts/src/terraformSupplemental/setEnvFromSsm.ts
GPaoloni: Awesome thanks. I don't think this is a blocker in any way, but I got confused as `-ro` is actually a helpline code (Romania :P), so maybe `-read` and `read-write` could confusion-proof this two? Or adding comments, which is definitively a smaller change (I think the first will impact infra tf too).

.github/workflows/terraform_plan.yml
robert-bo-davis: Good catch! fixed

twilio-iac/aarambh-production/main.tf
GPaoloni: No sorry, I meant the way `role_arn` in composed (fixed in the first occurrence, dynamic in the second).

.github/workflows/terraform_plan.yml
robert-bo-davis: Once [this is done](https://github.com/community/community/discussions/12029) (or once we have a slack bot that can handle some of this stuff) We'll be able to dynamically produce a list of options.

For now, I'd suggest we optimize for ease of dynamic automation over ease of end user use. Reducing the steps and "automatability" is more future proof than slight simplification of end user experience in the short term for internal tools.

twilio-iac/scripts/secretManager/secretManager/ssm_client.py
robert-bo-davis: Yes. This role has read/write access to the ssm params. See [this answer ](https://github.com/techmatters/flex-plugins/pull/1145#discussion_r1060893299)for more info about the pattern. I'm also happy to discuss further.

twilio-iac/aarambh-production/main.tf
robert-bo-davis: Ohh... This is a limitation of terraform. state config happens before anything else and can't be dynamic at all. 😦 

Before long we'll probably end up with terragrunt which will let us dynamically generate the state configuration. But for now we are stuck with the two different ways of setting the same thing.

robert-bo-davis: Terragrunt depends on [CHI-1569](https://tech-matters.atlassian.net/browse/CHI-1569). It is part of my internal roadmap to get us to maximally automated new helpline configuration, but I'm trying not to get too far ahead of myself yet. ;)

scripts/src/terraformSupplemental/setEnvFromSsm.ts
robert-bo-davis: Ahh! Yes. `ro` is a pattern I've used for a long time, but more explicit is more better. Fixed in both places!

robert-bo-davis: I used `tf-twilio-iac-ssm-read-only` and `tf-twilio-iac-ssm-admin` because `tf-twilio-iac-ssm-read-write` would be a bit of a misnomer in this case, I think.. The `tf-twilio-iac-ssm-admin` role can create, delete, and otherwise manage the ssm params related to `tf-twilio-iac` if there was a role that could only read and modify existing ssm params, I think the `tf-twilio-iac-ssm-read-write` would make sense, but we're getting pretty deep into a semantics hole at that point.

stephenhand (approved):

PR #1147, opened 2022-12-27
Create a Prod Release workflow for Flex plugins

GPaoloni:
Primary reviewer:

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### DIFF
--- a/.github/workflows/plugin-hrm-form-production-release.yml
+++ b/.github/workflows/plugin-hrm-form-production-release.yml
@@ -1,0 +1,49 @@
+# Workflow to create a new production release from a given qa tag
+
+name: Create a QA candidate release
+
+# Controls when the action will run.
+on:
+  workflow_dispatch:
+    inputs:
+      tag-prefix:
+        description: Release tag prefix (generally the version, e.g. v1.2.3).
+        required: true
+      title:
+        description: Release title - The title that will be given to this pre-release.
+        required: true
+
+jobs:
+  generate-release:
+    needs: run-e2e-tests
+
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      - name: GitHub Tag Name example
+       run: |
+         echo "Tag name from github.ref_name: ${{  github.ref_name }}"
+
+      # - name: Create pre release
+      #   uses: ./.github/actions/generate-production-release
+      #   with:
+      #     tag-prefix: ${{ inputs.tag-prefix }}
+      #     tag-suffix: 'qa'
+      #     title: ${{ inputs.title }}
+      #     repository: ${{ github.repository }}
+      #     repo_token: ${{ secrets.GITHUB_TOKEN }}
+      #   id: create_pre_release
+
+      # Send Slack notifying success
+      # - name: Slack aselo-deploys channel
+      #   uses: ./.github/actions/notify-deploys
+      #   with:
+      #     aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      #     aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      #     aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+      #     slack-message: "`[Flex]` Release from ${{ github.ref_type }} `${{ github.ref_name }}` requested by `${{ github.triggering_actor }}` completed with SHA ${{ github.sha }}. Release tag is ${{ steps.create_pre_release.outputs.generated-pre-release-tag }} :rocket:."
+      #   env:
+      #     SLACK_BOT_TOKEN: ${{ env.GITHUB_ACTIONS_SLACK_BOT_TOKEN }}


.github/actions/generate-production-release/action.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Create a release
+      uses: ncipollo/release-action@v1.12.0
+      with:
+        commit: ${{ github.sha }}
+        tag: ${{ steps.get_release_tag.outputs.generated-tag }}
+        name: ${{ inputs.title }}
+        bodyFile: "release-notes/${{ steps.get_release_tag.outputs.generated-tag }}.md"
+        prerelease: false
+        # allowUpdates: true ## do we want to allow "updates" to a release?
+      id: create_release
```
GPaoloni, line 48: This allows only one tag to exist per production version. I assumed this is what we want here, to treat the "production releases" as immutable (once we cut a prod release, further prod releases will have a greater sem-ver tag). Let me know if this is not desired.

.github/actions/generate-production-release/get-release-tag.ts
```diff
@@ -0,0 +1,29 @@
+import { setOutput, setFailed } from '@actions/core';
+
+async function generateReleaseTag() {
+  const qaTag: string = process.env.REF;
+
+  // Test that the tag matches the format we use for QA tags
+  if (!qaTag.includes('-qa')) {
+    throw new Error(`Provided REF env var does not matches the expected QA tag. Please provide a QA tag generated by "Create a QA candidate release" workflow`);
+  }
```
GPaoloni, line 9: This is assuming we'll only want to create prod-ready releases from already tested `-qa` tags. While you can manually create fake QA releases that didn't went thru the correct process, assuming this is only used with good intentions I think this is a good safety check to have. This line will abort if we ever call it with a branch (unless the branch for some reason contains `-qa` xD). Let me know if you think this is good or a better check can be used.

.github/actions/generate-production-release/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "generate-production-release",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "get-release-tag": "npm install && npx ts-node --esm ./get-release-tag.ts"
```
robert-bo-davis, line 8: Are we gaining enough from typescript here to make it worth incurring the additional overhead and extending the execution time by installing TS dependencies and transpiling on every execution?

.github/actions/generate-production-release/action.yml
```diff
@@ -0,0 +1,48 @@
...
+    description: "The generated release tag"
+    value: ${{ steps.get_release_tag.outputs.generated-tag }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '14.x'
```
robert-bo-davis, line 18: Node 14 is in maintenance mode and EOLs in April. Does it make sense to start defaulting to 16.x? Or better yet 18.x for new development?

.github/actions/generate-production-release/get-release-tag.ts
robert-bo-davis: I would generally shy away from protecting folks from themselves by adding training wheels where they aren't absolutely required in a release process. In my experience, on a trusted team, explicitly defined process with very few hurdles for going around the process if it is required tends to lead to fewer headaches in the long run. It allows for more flexibility around experimentation with process changes in the the future and doesn't get in the way when the process needs to be bypassed during an atypical release cycle or an emergency.

.github/actions/generate-production-release/package.json
GPaoloni: As there are other packages that we need to install, adding TS to the install step is not a really relevant extra burden. The compile step is almost instant too, as it's a single file. I'm not firmly against removing it, but TS it's my default choice, not only for the type safety, because the code inference and suggestions (I really love it while writing code hehe). [A run of this workflow is 25 seconds](https://github.com/techmatters/flex-plugins/actions/runs/3790543286) right now :P. Happy to change to plain JS if it's not something we want to deal with anyways!

.github/actions/generate-production-release/action.yml
stephenhand: This is correct, once a release is cut, further ones would be patch increments

.github/actions/generate-production-release/get-release-tag.ts
stephenhand: I would say it might be overkill to require a 'qa' tag

But perhaps it would be good to enforce running this against a 'tag' type ref rather than a 'branch' type ref

Then at least it can be run against an `EMERGENCY_HOTFIX_I_DID_AT_4AM` tag or w/e, and we may experiment with having different types of pre-release in the future (e.g. 'qa' to denote a build with a story ready to test vs 'rc' to indicate a production release candidate)

I would prefer it if we at least guarded against cutting releases directly off branches though. If it's so much of an emergency we can't take the time to create a tag, we probably should just lean into the fact that we're 'off piste' and deploy a 'non production' build to release, rather than trying to give it a veneer of respectability :-)

.github/actions/generate-production-release/get-release-tag.ts
```diff
@@ -0,0 +1,29 @@
+import { setOutput, setFailed } from '@actions/core';
+
+async function generateReleaseTag() {
```
stephenhand, line 3: Does this need to be `async`?

.github/actions/generate-production-release/get-release-tag.ts
stephenhand: In this specific case I really have no preference - it's a trivial amount of code but the overhead to transpile it seems trivial too

stephenhand (commented):
Overall looks great, I like the workflow

On the specific case of the file vs workflow parameter - I would prefer to go with a workflow parameter, for the following reason:

We really want the production builds to be cut from commits that have previously been tagged by the QA release workflow, because we know those build, and have passed the E2E tests

However, I can see the following thing happening:

We cut a QA release, goes to staging, no problems are found so it's good to release.
Great! Let's run our nice new prod release workflow against that tag.
Oh nuts, it failed because we have no release notes, no problem, lets create some
Ok, got that release notes PR merged, let's go!
Oh wait, now the commit I want to release with the notes doesn't have a QA tag, I need to run ANOTHER QA release
Finally I can run the prod release...

The above can be mitigated by process, ensuring release notes are created before cutting a 'release candidate' QA release. But that might introduce a problem where out of date release notes were added a while back are present and don't have the correct info on them, so we have to go through the above rigmarole anyway, to get a 'QA release' out with correct, up to date notes.

Because Github releases can have their descriptions updated retroactively, I would suggest we have a description workflow parameter with a default value of 'TODO'. Then for trivial releases (patch releases or whatever) we can replace that with an inline comment. Otherwise we can leave it as 'TODO' then somebody can go into the Github GUI and add the notes after the release it created. 

The other beneifit I see is that we only keep the notes in one place - on GH, so don't need to worry about keeping the GH version and the repo version in sync, if we need to tweak  them after the fact for instance.

Does this sound reasonable? @nickhurlburt @robert-bo-davis

stephenhand (approved):
Nice that we don't need any custom actions at all now

PR #1160, opened 2023-01-05
Add Support to view more languages in pdf, including Thai

mythilytm:

## Description
- Replace 'Open Sans' with Roboto for better character font support and use NotoSansThaiLooped for Thai characters for viewing pdf documents (for `react-pdf/renderer` library)
- The reasoning for using NotoSansThaiLooped was because Noto supports languages and fonts used worldwide: 
See https://fonts.google.com/noto/specimen/Noto+Sans and https://notofonts.github.io/
 * In the future, when more languages are added, adding a compatible fontFamily might be necessary. Check https://notofonts.github.io/

### Checklist
- [n/a] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Related Issues
Fixes # [1618](https://tech-matters.atlassian.net/browse/CHI-1618)

### Verification steps
- In `HrmFormPlugin.tsx`, change line 71 to 
`definitionVersion: DefinitionVersionId.huV1,` to check hungarian (Ukrainian character are supported as well now)
`definitionVersion: DefinitionVersionId.thV1,` to check thai character support
- Insert some phrases of those languages in `components/case/casePrint/CasePrintView.tsx` or its child components or hardcode values that will appear in the PDF format
Screenshot 2023-01-05 at 6 53 58 PM


plugin-hrm-form/src/components/case/casePrint/CasePrintView.tsx
```diff
@@ -31,11 +31,11 @@ type OwnProps = {
 type Props = OwnProps;
 
 const CasePrintView: React.FC<Props> = ({ onClickClose, caseDetails, definitionVersion, counselorsHash }) => {
-  const { pdfImagesSource, strings } = getConfig();
```
mythilytm, line 34: I needed to use `definitionVersion` property of config object which was competing with `definitionVersion` in line 33. Not sure if there is a better refactor, but for now I had to remove this destructuring assignment.

stephenhand: You can use aliasing in destructuring, e.g.

```
const { pdfImagesSource, strings, definitionVersion: myAliasForDefinitionVersion } = getConfig();
```

I always find the syntax a bit odd, to me it would make more sense if the alias and the property were on opposite sides of the colon - but it's handy for cases like this.

This is also why the eslint `no-shadow` rule is actually pretty handy...

plugin-hrm-form/src/components/case/casePrint/styles/index.ts
```diff
@@ -30,6 +46,36 @@ Font.registerEmojiSource({
...
+/**
+ * Roboto font works for all languages/fonts supported thus far.
+ * However, Thai characters are not readable, hence substituting to NotoSansThaiLooped is required.
+ * https://fonts.google.com/noto/specimen/Noto+Sans
+ * In the future, when more languages are added, adding a compatible fontFamily might be necessary
+ *
+ * @param definitionVersion: DefinitionVersionId
+ */
+export const changeRegisteredFonts = (definitionVersion: DefinitionVersionId) => {
+  if (definitionVersion === DefinitionVersionId.thV1) {
```
stephenhand, line 63: We still need a better way of differentiating between which font is required than hardcoded checks for the form definition set being used

A boolean flag in the form definition named something like `thaiCharactedPrintSupport` (or 'South East Asian character support if we would use it potentially for more than Thai?), assumed to be false if not present would suffice

stephenhand (commented):
LGTM except for the comment on how we determine if the alternative font is required, I'd like to see that addressed before we merge

plugin-hrm-form/src/components/case/casePrint/styles/index.ts
```diff
@@ -30,6 +46,33 @@ Font.registerEmojiSource({
...
+const { caseSummaryText } = summaryStyles;
+
+/**
+ * 'Roboto' font family works for all languages/fonts supported thus far.
+ * However, Thai characters are not readable with Roboto, hence substituting to NotoSansThaiLooped is required.
+ * https://fonts.google.com/noto/specimen/Noto+Sans
+ * In the future, when more languages are added, adding a compatible fontFamily might be necessary
+ *
+ */
+export const setThaiFontFamily = () => {
```
stephenhand, line 61: I think `useThaiFontFamily` is most descriptive, but this is a better than previously

PR #1166, opened 2023-01-09
Refactor prepopulateForm method to ensure preEngagamentData populates in the absence of survey

mythilytm:

## Description
- This PR improves on the `prepopulateForm` method to ensure `preEngagamentData` populates in the absence of `survey`. 
- Also, fields with mixed-checkbox as type are parsed to read the value for the checkbox in flex. 

### Checklist
- [n/a] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Related Issues
Fixes # [1071](https://tech-matters.atlassian.net/browse/CHI-1071)

### Verification steps
- Check with webchat and flex configurations for Dev, CL and CA

### DIFF
--- a/plugin-hrm-form/src/utils/prepopulateForm.ts
+++ b/plugin-hrm-form/src/utils/prepopulateForm.ts
@@ -111,9 +111,9 @@ export const getValuesFromPreEngagementData = (
   tabFormDefinition.forEach((field: FormItemDefinition) => {
     if (prepopulateKeys.indexOf(field.name) > -1) {
       if (field.type === 'mixed-checkbox') {
-        if (preEngagementData[field.name] === 'yes' || preEngagementData[field.name] === 'Yes') {
+        if (preEngagementData[field.name]?.toLowerCase() === 'yes') {
           values[field.name] = true;
-        } else if (preEngagementData[field.name] === 'no' || preEngagementData[field.name] === 'No') {
+        } else if (preEngagementData[field.name]?.toLowerCase() === 'no') {
           values[field.name] = false;
         }
         return;
@@ -127,7 +127,7 @@ export const getValuesFromPreEngagementData = (
 // eslint-disable-next-line sonarjs/cognitive-complexity
 export const prepopulateForm = (task: ITask) => {
   const { memory, preEngagementData } = task.attributes;
-
+  if (memory === null && preEngagementData === null) return;
   const { currentDefinitionVersion } = getDefinitionVersions();
   const { tabbedForms, prepopulateKeys } = currentDefinitionVersion;
   const { ChildInformationTab, CallerInformationTab, CaseInformationTab } = tabbedForms;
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -109,6 +110,14 @@ export const getValuesFromPreEngagementData = (
   const values = {};
   tabFormDefinition.forEach((field: FormItemDefinition) => {
     if (prepopulateKeys.indexOf(field.name) > -1) {
+      if (field.type === 'mixed-checkbox') {
+        if (preEngagementData[field.name] === 'yes' || preEngagementData[field.name] === 'Yes') {
```
stephenhand, line 114: You can use `toLowerCase()` in situations like this to do a case insensitive comparison, rather than comparing multiple strings with the same letters in different cases, e.g.

```
if (preEngagementData[field.name]?.toLowerCase() === 'yes') {
```

plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -118,29 +127,62 @@ export const getValuesFromPreEngagementData = (
...
+    Manager.getInstance().store.dispatch(
+      RoutingActions.changeRoute(
+        { route: 'tabbed-forms', subroute: 'childInformation', autoFocus: true },
+        task.taskSid,
+      ),
+    );
+    return;
+  }
+
+  const { answers } = task.attributes.memory.twilio.collected_data.collect_survey;
```
stephenhand, line 157: You could still get here if both memory and preengagment are null, and it would crash. Is this a problem, or is that case eliminated as a possibility before calling the function?

plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -118,29 +127,62 @@ export const getValuesFromPreEngagementData = (
...
 
-  // If this task came from the pre-survey
-  if (memory || preEngagementData) {
-    const { answers } = task.attributes.memory.twilio.collected_data.collect_survey;
+  // If can't know if call is child or caller, do nothing here
+  if (!answers.about_self || !['Yes', 'No'].includes(answers.about_self.answer)) return;
```
stephenhand, line 160: If we have no survey at all, we treat preengagement as a child, whereas if we have a survey but no question about filling the survey for yourself, it doesn't do anything. This seems a little inconsistent, should we not also assume the caller is the child in this case as well?

stephenhand (approved):
Few non-blocking comments. It seems like this logic could be expressed more neatly if you use the 'null object pattern' https://en.wikipedia.org/wiki/Null_object_pattern to cut down on the amount of conditional checks, but I haven't been through it to prove it so I'll trust you that this is the best approach

plugin-hrm-form/src/utils/prepopulateForm.ts
mythilytm: Fixed this to account for no about_self property to default to child

plugin-hrm-form/src/utils/prepopulateForm.ts
mythilytm: Added a check early in the function in this case

plugin-hrm-form/src/utils/prepopulateForm.ts
mythilytm: 👍

PR #1168, opened 2023-01-09
Childline Thailand TH TF Production Account setup

janorivera:
Primary reviewer:
@GPaoloni 

## Description

Ready for apply.
Production basic implementation for TH. 
This implementation doesn't have the channels yet (that will come later) and it doesn't implement any chatbots.
Tested on Staging.

### Checklist
- [x] Updates made to staging 
- [x] Confirmed updates on staging
- [x] Tested on staging

### Related Issues
Fixes #....


Humairaa127 (commented):
The age limit is 18, so the list should be 19+

twilio-iac/thailand-production/main.tf
```diff
@@ -0,0 +1,193 @@
...
+    "enable_voice_recordings": false,
+    "enable_twilio_transcripts": true,
+    "enable_external_transcripts": false,
+    "post_survey_serverless_handled": true,
+    "enable_csam_clc_report": false
+  }
+
+  secrets = jsondecode(data.aws_ssm_parameter.secrets.value)
+  //Channels [Facebook | Line | Instagram | Twitter]
+  twilio_channels = {"facebook" = {"contact_identity" = "messenger:59591583805", "channel_type" ="facebook"}}
```
GPaoloni, line 69: This is repeated in `twilio_numbers`.  think that's fine but maybe we could DRY in the future and have a single place to edit this identifier?

GPaoloni (approved):
Looks good!

twilio-iac/thailand-production/main.tf
janorivera: yes definitely, we will need to pull from twilio_channels and custom_channels, but it should be easy to do as an improvement in the future

PR #1171, opened 2023-01-10
Linea Libre CL Production Account setup

janorivera:
Primary reviewer:
@GPaoloni 
## Description

Linea libre account setup
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


GPaoloni (commented):
I see that all of the CL staging modules are being re-defined here as `linealibre-cl-staging`, but they already exist under `cl-linealibre-staging`. I think if this is applied as it is, it may cause weird things? @robert-bo-davis Can this lead to broken state, or just duplicate modules that both mess with the same key `twilio/cl/terraform.tfstate` of `tl-terraform-state-staging`?

nickhurlburt:
@janorivera is this something you need to discuss with Bo, or what's the next step?

robert-bo-davis:
>I see that all of the CL staging modules are being re-defined here as linealibre-cl-staging, but they already exist under cl-linealibre-staging. I think if this is applied as it is, it may cause weird things? @robert-bo-davis Can this lead to broken state, or just duplicate modules that both mess with the same key twilio/cl/terraform.tfstate of tl-terraform-state-staging?

Off the top of my head, I believe that there will be an issue, but it is't insurmountable. The terraform secrets SSM won't be found and will need to be reconfigured with the same values as the already applied terraform module by running `make manage-ssm-secrets` from the new directory before initing or applying it. The old secret (`/terraform/twilio-iac/cl-linealibre-staging/secrets.json`) will need to be deleted along with the old directory once the plan/apply for this dir has been validated and applied.

I *think* those will be the only changes needed.

GPaoloni:
But is there any reason to move away the already existing `cl-linealibre-staging`?

janorivera:
@GPaoloni  yes! I'm trying to bring consistency to our naming conventions. Right now most of our names are close to  "Helpline Name " "Code" "Environment", I think the more close we move to a standard way of naming the better. Open to ideas on what the definitive naming convention should be though.

janorivera:
I think that depending on the "type of field", things could be a little different, but maintaining the order of fields could be consistent across the board? 
Like:
Datadog apps, twilio account names, etc could be : Helpine Name (Code) Environment
repos and others?  could be: helplinename-code-environment 
and so on...
I'm happy to whatever we decide but I do think that is important to have it.

janorivera:
Btw @GPaoloni  I was working on this before I saw your comment, so is tested and ready to go, I did what Bo suggested.  Happy to change it back to what it was before if needed.

janorivera:
excellent! Ok I'll stop trying to change other directories, anyway, this one is already done and applied according to the steps you mentioned. The only missing step is to delete the AWS terraform secrets.  Do you have permission to do that?

robert-bo-davis:
>The only missing step is to delete the AWS terraform secrets. Do you have permission to do that?
Done!

GPaoloni (approved):
@janorivera The changes are fine! I just wanted to ensure tf will remain in a consistent state. LGTM!

robert-bo-davis:
Oops, I missed something. @janorivera could you please also make sure the [helpline terraform spreadsheet](https://app.box.com/file/1109527438079) is up to date with the path change (and any other changes you have made)?

PR #1173, opened 2023-01-11
CHI-1484: Missing translation

stephenhand:
Quick extra translation, if our resident native speakers could haver a quick look :-)


murilovmachado (approved):
Looks good to me! and the pt-BR translations are correct 👍

PR #1176, opened 2023-01-12
Form generators refactor - First steps and factored out FormInput

GPaoloni:
Primary reviewer:

## Description
This PR is a proposal to bring some order to our form generators module. It also includes some changes in the way they are generated.
- Removes `createFormFromDefinition` function. This is instead replaced by a custom hook `useCreateFormFromDefinition` that will receive mostly the same parameters and
  - Abstracts away the repeated logic in every place where a form was built in the previous way (that's why I think a custom hook is a good fit).
  - ~Will take care of memoizing the computed form resulting from the given form definition. This was previously a burden in the caller components, and in most cases after some time we ended up ruining the memoization anyways. 
    To see the benefits of this change you can sit in commit https://github.com/techmatters/flex-plugins/pull/1176/commits/b816e89b86cff9a61bde6ccf4695d7aceced4f5b, which is the current state of our forms plus a debug log. Interact with the tabbed forms while keeping an eye to the Chrome console and you'll see that the form is being re-built in every re-render of the parent component. Then sit in commit https://github.com/techmatters/flex-plugins/pull/1176/commits/328286f58174376fee9e71797cd3da1433a94070 that already handles the creation of the form via this hook, and check that it is only created once, and then kept in sync via rerenders (but the form as such does not needs to be entirely rebuilt from the definition).~ 
    Removed after @stephenhand gathered [empirical evidence that this memo is not really necessary](https://github.com/techmatters/flex-plugins/pull/1176#issuecomment-1382104859) :P
- Introduces a file `inputGenerator.tsx`, with `createInput` function which is intended to replace the existing `getInputType` one. It is intended to be simpler as the logic for each component should be factored out from it, leaving this function as the link between a `FormItemDefinition` and it's corresponding input.
- Adds a new folder under components, `forms/` where I think we could "factor out" our custom inputs (right now they all live in the 1k lines file `src/components/common/forms/formGenerators.tsx.tsx` :sweat_smile:). Inside I've exemplified with top level files `types.ts` and `styles.tsx`, where the code shared among all the component types will live. There's also a `FormInput/` module that contains `FormInput.tsx` and `styles.tsx`, which make the base input that corresponds to `FormInputType.Input`. This is the easier one since it's the base style, but you get the idea. I'm also bundling the unit tests for this module within itself. Happy to move this to the usual `__tests__` folder if preferred by the team, I've just been thinking this way can be easier to work with.

Why I think this refactor is a good idea (this are mostly my opinions):
- The code for the form generators module is more organized and easier to maintain.
- The styles for the above will live in a more specialized place, which will help reducing the amount of code we have in the root `src/styles/HrmStyles.tsx`.
- The code that calls the form generators is cleaner too, it just needs to call single custom hook and will receive the generated form.
- Writing tests for the UI, the `createItem` and the hook as separate things will be easier to handle than trying to write tests for the big monster we currently have :laughing:.

### Checklist
- [x] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Related Issues
Fixes #....

### Verification steps
Manual regression test to verify the forms generated (tabbed forms, case forms..) are still working as intended, no change in the behavior is expected.
  - Forms are generated and works as expected, preserving the same UI. The required and error states should be the same (asterisk and error message appears upon validation, respectively).
  - The `updateCallback` is invoked as expected, properly updating the Redux state (if pertinent).


stephenhand (commented):
This is an awesome refactor.

Much happier that this is refactored internally in Flex rather than going into the form defs module, didn't really want a clatter of React deps going in there.

Although it might make sense to put them in a 3rd module at some point, but only when something else needs to consume them.

I would move the 'forms' directory out of the 'common' directory as part of this refactor. I would probably move everything else out of there too - there doesn't really seem to be any distinction between stuff in components/common and any of the other components...

stephenhand:
FYI the E2E test failures seem to be related to your changes, rather than just flake

plugin-hrm-form/src/components/forms/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './components';
+
+export { default as useCreateFormFromDefinition } from './hooks/useCreateFormFromDefinition';
```
GPaoloni, line 3: Does it makes sense for this module to live in `/components/forms` or would you prefer this to go one level up, outside of `components` folder? I think we could treat `forms` as it's own module right?

stephenhand (approved):
Looks great! 🚀

PR #1180, opened 2023-01-16
CHI-1644: Refactor fetch wrappers & config. Add resource service stubs

stephenhand:
## Description

* Adds a base fetch wrapper for all the service specific ones to wrap
* Added some tests for the low level fetch wrapper APIs
* Tidied up error handling
* Moved config out of HrmFormPlugin.tsx. (we should now test it!)
* Moved shared state client related code out of HrmFormPlugin.tsx

### Checklist
- [X] New tests added
- [X] Feature flags added
- N/A Strings are localized

### Related Issues
Preparing for CHI-1644

### Verification steps

* Automsated tests should cover it


plugin-hrm-form/src/services/ResourceService.ts
```diff
@@ -0,0 +1,19 @@
...
+// eslint-disable-next-line import/no-unused-modules
+export const referrableResourcesEnabled = () => Boolean(getReferrableResourceConfig().resourceBaseUrl);
+// eslint-disable-next-line import/no-unused-modules
+export const getResource = async (resourceID: string): Promise<ReferrableResource> => {
+  // return fetchResourceApi(`resources/`);
+  return {
+    id: resourceID,
+    name: 'Example Client Side Stubbed Resuource',
+  };
+};
```
murilovmachado, line 19: I know this is just a stub and everything will be replaces, but there's 2 points here:
- Shouldn't it return a promise? like `Promise.resolve({ id, name })`;
- typo at name string, it's written `Resuource` instead of `Resource`

plugin-hrm-form/src/services/fetchApi.ts
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Low level fetch wrapper to provide some sensible defaults & rudimentary error handling
+ * You would normally wrap this rather than calling it directly, see fetchProtectedApi & fetchHrmApi
+ * @param baseUrl
+ * @param endpointPath
+ * @param token
+ * @param options
+ */
+export const fetchApi = async (baseUrl: URL, endpointPath: string, options: RequestInit): Promise<any> => {
```
murilovmachado, line 28: Question on `baseUrl` type: Wouldn't it be of type `string` instead of `URL`?
~~Looking at `readConfig()` it seems it's a `string`.~~

UPDATE: I've seen where you're converting it to URL.

plugin-hrm-form/src/services/fetchHrmApi.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Factored out function that handles api calls hosted in HRM backend.
+ * Will throw Error if server responses with http error code.
+ * @param endPoint endpoint to fetch from (withouth the host part of url, e.g. "/cases/contacts").
+ * @param {Partial<RequestInit>} options Same options object that will be passed to the fetch function (here you can include the BODY of the request)
+ * @returns {Promise<any>} the api response (if not error)
+ */
+const fetchHrmApi = (endPoint: string, options: Partial<RequestInit> = {}): Promise<any> => {
+  const { hrmBaseUrl, token } = getConfig();
+
+  return fetchApi(new URL(hrmBaseUrl), endPoint, {
```
murilovmachado, line 14: Oh, I see here (and in the other fetch-wrapper files) you're creating the `URL` object given a `string`. Is there a reason to not just use the `string` as it is? Ex: would we get any descriptive error in case we pass a malformed URL string?

murilovmachado (approved):
LGTM! Thanks for the refactoring

plugin-hrm-form/src/services/ResourceService.ts
stephenhand: Keeping it as a promise to mimic the real API, which will be async

plugin-hrm-form/src/services/fetchHrmApi.ts
stephenhand: Exactly. We get some free validation which will give us a nicer error to debug if it's fed garbage

Also to avoid the Primitive Obsession antipattern: https://medium.com/the-sixt-india-blog/primitive-obsession-code-smell-that-hurt-people-the-most-5cbdd70496e9

We should always prefer types that more specifically represent the purpose of the variable over strings, numbers & bools. Especially when they are already built into the runtime like `URL`

PR #1189, opened 2023-01-19
Audio Notifications for Reserved tasks and new messages

mythilytm:

## Description
- This PR adds audio notifications - a repeating alert when a task gets reserved, and a single beep(bell) when the counsellor receives a  new message. 
- Introduces `notifyReservedTask` that will trigger an alert based on workerClient's `reservationCreated` listener and `notifyNewMessage` that will trigger an alert based on conversationsClient's `messageAdded` listener.
- In a separate PR, Bo [added](https://github.com/techmatters/infrastructure-config/pull/112) an s3 bucket for each environement with Cloudfront for static assets. 
- To use this, new service configuration attribute has been added called `assets_bucket_url`. For all helplines, configurations will need to be updated with `assets_bucket_url` to ensure audio notifications work. 

### Checklist
- [n/a] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized
- [x] Added a new property in Service Configuration

### Related Issues
Fixes #[1619](https://tech-matters.atlassian.net/browse/CHI-1619)

### Verification steps
- Start a conversation in webchat or other chat channels. 
- Ensure Flex is active in browser and audio is turned on. 
- For checking reserved task notification, change to 'Available' as worker and ensure no other tasks are active. A continuous beep should persist. To stop the alert, change status of the worker or task
- For checking new message notification, (this is tricky - use a phone or other device to ensure flex is active) send new messages from the chat channel. A single bell sound should happen.


plugin-hrm-form/src/utils/audioNotifications.ts
```diff
@@ -0,0 +1,44 @@
+import * as Flex from '@twilio/flex-ui';
+
+/**
+ * An audio alert when a task is reserved to the counsellor. Stops when accepted or other event that changes the worker or task status
+ */
+export const notifyReservedTask = reservation => {
+  let media;
+  const manager = Flex.Manager.getInstance();
+  const assetsBucketUrl = manager.serviceConfiguration.attributes.assets_bucket_url;
```
stephenhand, line 9: Could we use `getConfig()`rather than going direct to the Flex Manager here?

plugin-hrm-form/src/utils/setUpComponents.tsx
```diff
@@ -346,3 +349,13 @@ export const setupCannedResponses = () => {
   Flex.MessageInput.Content.add(<CannedResponses key="canned-responses" />);
   Flex.MessageInputV2.Content.add(<CannedResponses key="canned-responses" />);
 };
+
+/**
+ * An audio alert when a counsellor receives a new message
+ */
+export const setupNotifications = () => {
```
stephenhand, line 356: I would put this in the `audioNotifications.ts` file, we should be trying to break this file up, not add more to it

plugin-hrm-form/src/utils/setUpComponents.tsx
```diff
@@ -78,6 +79,8 @@ const setUpRerenderOnReservation = () => {
   const manager = Flex.Manager.getInstance();
 
   manager.workerClient.on('reservationCreated', reservation => {
+    notifyReservedTask(reservation);
```
stephenhand, line 82: We should set this up in a separate `manager.workerClient.on('reservationCreated', ...)` call and call it from `setUpNotifications`, keeping all the audio notifications set up in the same function

stephenhand (commented):
Code logic looks good, just some notes on how it's organised

plugin-hrm-form/src/utils/setUpComponents.tsx
```diff
@@ -346,3 +349,13 @@ export const setupCannedResponses = () => {
   Flex.MessageInput.Content.add(<CannedResponses key="canned-responses" />);
   Flex.MessageInputV2.Content.add(<CannedResponses key="canned-responses" />);
 };
+
+/**
+ * An audio alert when a counsellor receives a new message
+ */
+export const setupNotifications = () => {
+  const manager = Flex.Manager.getInstance();
+  manager.conversationsClient.on('messageAdded', messageInstance => {
```
GPaoloni, line 358: I think there a few things here:
- As discussed via Slack, a check to filter out counselor messages will be nice (`if (messageInstance.author !== manager.user.identity)`).
- This subscribes to event `messageAdded` on every conversation. You want only the ones regarding this particular counselor. You could set a taskrouter event listener that triggers on task `accepted`. Using `TaskHelper.isChatBasedTask` and `StateHelper.getConversationStateForTask` you can get the specific conversation, and then attach the listener on `messageAdded`. To see this idea in action you can check on `src/utils/setUpTaskRouterListeners.ts`. If you go this way, I agree with Steve's comment of moving this code, but feels a better fit to the file I'm referring here.

plugin-hrm-form/src/utils/audioNotifications.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const notificationTone = 'ringtone';
+  const notificationUrl = `${assetsBucketUrl}/notifications/${notificationTone}.mp3`;
+
+  if (document.visibilityState === 'visible') {
```
GPaoloni, line 14: This check causes the alerts to only sound if we are in Flex tab. Is this intended?

GPaoloni (commented):
Code looks really clean and logic is in place! Just a few more little things on top of Steve's ones.

plugin-hrm-form/src/utils/setUpComponents.tsx
GPaoloni: Sorry, I just realized, if you use afterAcceptTask action, setUpActions would be the correct place.

plugin-hrm-form/src/utils/setUpComponents.tsx
mythilytm: Moved everything relevant to audio alerts out of setupComponents

plugin-hrm-form/src/utils/audioNotifications.ts
```diff
@@ -0,0 +1,44 @@
...
+    media = Flex.AudioPlayerManager.play({
+      url: notificationUrl,
+      repeatable: true,
+    });
+  }
+  const taskStatuses = ['accepted', 'canceled', 'rejected', 'rescinded', 'timeout'];
+  taskStatuses.forEach(status => {
+    reservation.once(status, () => {
+      Flex.AudioPlayerManager.stop(media);
+    });
```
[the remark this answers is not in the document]
mythilytm: Todo

plugin-hrm-form/src/utils/setUpComponents.tsx
mythilytm: Thanks for the direction!

plugin-hrm-form/src/utils/audioNotifications.ts
mythilytm: According to the requirement, yes. https://tech-matters.atlassian.net/browse/CHI-1619?focusedCommentId=12277

plugin-hrm-form/src/utils/setUpActions.ts
```diff
@@ -185,6 +186,10 @@ export const afterAcceptTask = (featureFlags: FeatureFlags, setupObject: SetupOb
 ) => {
   const { task } = payload;
 
+  if (TaskHelper.isChatBasedTask(task)) {
```
stephenhand, line 189: I think we should set up separate handlers for each individual action we want to take on a particular event, rather than having one big handler that does everything.

At the moment we get strange behaviour, like the welcome message not sending because prepopulating forms had an error. Separating the handlers would make things more robust and break up the code more, make it easier to follow.

Doesn't need to be for this PR though

plugin-hrm-form/src/utils/audioNotifications.ts
```diff
@@ -0,0 +1,81 @@
+import { StateHelper, Manager, AudioPlayerManager } from '@twilio/flex-ui';
```
stephenhand, line 1: This should have some tests. I'm ok to merge this PR so it can be deployed for QA without the tests, but I'd want to see a PR adding test coverage raised & merged before the associated task is moved to 'Done' if we do it that way.

Or adding tests as part of this PR works too

stephenhand (approved):
Nothing blocking this PR, but we should have tests before the story is 'done'.

plugin-hrm-form/src/utils/audioNotifications.ts
mythilytm: I'll add tests in a separate PR for this module.

PR #1193, opened 2023-01-21
chore: transpile TS on init not every execution

robert-bo-davis:
## Description

TS transpile is *very* slow inside of virtualized docker (on macos or windows). The Alpine Linux based terraform container runs node 18.x. Node/npm version mismatches between npm install and running scripts from the scripts folder cause issues. This moves `npm install` and `tsx` of the scripts into the init phase so that they are run inside of the container with the correct node versions.


stephenhand (approved):
LGTM

twilio-iac/scripts/installScripts.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cd ../../scripts/
+
+npm i
```
GPaoloni, line 5: Isn't `npm ci` preferred for this cases? No really something needed, just asking.

GPaoloni (approved):
Great! From the user's perspective nothing changes, we run the same commands, right?

robert-bo-davis:
> Great! From the user's perspective nothing changes, we run the same commands, right?

Yup!

PR #1194, opened 2023-01-23
CHI-1653: fix greeting

stephenhand:
## Description

* Fixes CHI-1653
* Makes the prepopulateForm function handle no preengagement data

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-1653

### Verification steps

See ticket - ensure greeting appears for channels with no preengagement

### DIFF
--- a/plugin-hrm-form/src/utils/prepopulateForm.ts
+++ b/plugin-hrm-form/src/utils/prepopulateForm.ts
@@ -127,22 +127,21 @@ export const getValuesFromPreEngagementData = (
 // eslint-disable-next-line sonarjs/cognitive-complexity
 export const prepopulateForm = (task: ITask) => {
   const { memory, preEngagementData } = task.attributes;
-  if (memory === null && preEngagementData === null) return;
+  if (!memory && !preEngagementData) return;
 
   const { currentDefinitionVersion } = getDefinitionVersions();
   const { tabbedForms, prepopulateKeys } = currentDefinitionVersion;
   const { ChildInformationTab, CallerInformationTab, CaseInformationTab } = tabbedForms;
   const { preEngagement, survey } = prepopulateKeys;
 
-  // PreEngagementData Values
-  const childInfoValues = getValuesFromPreEngagementData(
-    preEngagementData,
-    ChildInformationTab,
-    preEngagement.ChildInformationTab,
-  );
-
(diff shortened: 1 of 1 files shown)


stephenhand:
@mmythily - Looking more at `prepopulateForm.ts` are we sure that blank preengagement values aren't going to wipe values from the survey? This line: https://github.com/techmatters/flex-plugins/blob/ec60c25f5d55679cb29651823de3dac8b74404fe/plugin-hrm-form/src/utils/prepopulateForm.ts#L121 looks like it could wipe survey values with empty spaces if nothing is set for the fieled pre-engagement, rather than leaving it as it is

murilovmachado (approved):
Using `!memory` is fine in this case, because it cannot be `0` or `""`. But, in general, I'd use one of the below, in Typescript:
- `memory === null || memory === undefined`
- `memory == null`

But for this PR `!memory` is totally fine.

plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -127,22 +127,21 @@ export const getValuesFromPreEngagementData = (
 // eslint-disable-next-line sonarjs/cognitive-complexity
 export const prepopulateForm = (task: ITask) => {
   const { memory, preEngagementData } = task.attributes;
-  if (memory === null && preEngagementData === null) return;
+  if (!memory && !preEngagementData) return;
```
murilovmachado, line 130: (Forgot to mark in the code what I was referring to)

stephenhand: Linter doesn't like `== null` and I find the explicit double check very verbose

> Using `!memory` is fine in this case, because it cannot be `0` or `""`. But, in general, I'd use one of the below, in Typescript:
> 
> * `memory === null || memory === undefined`
> * `memory == null`
> 
> But for this PR `!memory` is totally fine.

Linter doesn't like `== null` and I find the explicit double check very verbose. `(memory ?? null) === null` is an ok compromise. 

But as you say, any falsey value would be invalid here, so if anything `!memory` is better, covers more incorrect cases. I suppose the best check would be `typeof memory === 'object'`, but I think this is grand :-P

PR #1198, opened 2023-01-25
Th config updates post visit

Humairaa127:

## Description
- Translation corrections
- Dropdown value adjustments in case and contact forms

### Checklist
- [ ] Strings are localized


plugin-hrm-form/src/translations/th-TH/flexUI.json
```diff
@@ -1,585 +1,585 @@
 {
```
stephenhand, line 1: Can we revert the indentation change? It just makes doing diffs over time way harder if there's a commit where every line has changed with an extra tab

stephenhand (commented):
Just one minor comment then LGTM

stephenhand (approved):
LGTM!

PR #1203, opened 2023-01-26
Update serverless deploy workflow template

stephenhand:
## Description

Updates the serverless workflow template to align with how the current serverless workflows are now configured. Changed terminology from 'release' to 'deploy'

### DIFF
--- a/scripts/templates/serverless-workflow-template
+++ b/scripts/templates/serverless-workflow-template
@@ -6,12 +6,9 @@ on:
   workflow_dispatch
 jobs:
   deploy:
-    runs-on: ubuntu-latest
-
-    steps:
-      uses: ./.github/workflows/custom_helpline.yml
-      secrets: inherit
-      with:
-        helpline_code: "<SHORT_HELPLINE>"
-        environment_code: "<SHORT_ENVIRONMENT>"
-        changelog: "Deploying individual account"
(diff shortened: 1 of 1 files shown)


scripts/templates/serverless-workflow-template
```diff
@@ -1,135 +1,14 @@
...
-# Controls when the action will run.
 on:
-  workflow_dispatch:
-  
-  workflow_call:
-    inputs:
-      changelog:
-        description: Release Changelog - Add a comment about the changes to be included in this release.
-        required: true
-        type: string
```
GPaoloni, line 14: This workflows won't be callable anymore? I thought it was needed for "deploy all" ([workflow_call](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call)), but maybe it isn't? I don't see `workflow_call` present in the other ones either.

GPaoloni (approved):

scripts/templates/serverless-workflow-template
stephenhand: Deploy All now calls `custom_helpline.yml` directly so the helpline flows don't need to be callable, they are literally just a UI convenience now

murilovmachado (approved):
LGMT

PR #1207, opened 2023-01-27
Style the 'single resource view'

GPaoloni:
## Description
This PR adds [the styles](https://marvelapp.com/prototype/ieh6cb7/screen/88434650) to the "view single resource" screen.

Screenshot from 2023-01-27 17-46-51

NOTES: 
- ~The last commit of this PR breaks the resources view, since the data does not matches the shape expected.~  The last commit does not shows any attributes since the returned example resource does not contains any. You must hardcode a mocked response (like in https://github.com/techmatters/flex-plugins/pull/1207/commits/07cccd6dbb76ffbbaaf8deca680749cb5b529975).
- This PR just styles according to the mocks, assuming the resources as described in that particular scenario, for KHP. I assume we'll have some sort of layout specification per account, that tells this view what it should display for each resource. (Already talked a bit about this with Steve, but is something still unclear to me).

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1419)
- [n/a] New tests added
- [n/a] Feature flags added
- [x] Strings are localized

### Verification steps
- `❯ git checkout 07cccd6 -- src/services/ResourceService.ts` so the API returns a mocked resource.
- Confirm that the view resource screen matches the styles.


plugin-hrm-form/src/services/ResourceService.ts
```diff
@@ -1,10 +1,15 @@
 import fetchResourceApi from './fetchResourceApi';
 import { getReferrableResourceConfig } from '../hrmConfig';
 
+export type ReffarebleResourceAttributeValue = string | string[] | { id: string; value: string; color?: string }[];
```
stephenhand, line 4: * ReferrableResourceAttributeValue

plugin-hrm-form/src/___tests__/StandaloneSearch.test.jsx
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
```
stephenhand, line 1: Convert to .tsx whilst you're in here? :-P

stephenhand (approved):
LGTM, couple of small nits but nothing blocking

GPaoloni:
>Nice refactor creating useExpandableOnOverflow. Would it make more sense as a compound component than a hook maybe? Not even sure if that would work.

I tried to implement as a component, but since the two places where we use it differ in very specific css/the way the UI is composed, I ended up having a HoC that will wrap the entire `children` and fed it with the expandable/isOverflowing logic as props. Having a hook does the same but within the same component, rather than wrapping it with the HoC. I can go with the HoC if that's preferred, for me hooks are the default :P

PR #1213, opened 2023-01-31
[PoC] Demonstrate redux-promise-middleware

GPaoloni:
Note: since this is a PoC, I'm not fixing the CI unless we decide is worth the effort :stuck_out_tongue_closed_eyes: 

## Description
This PR is (kinda) part of investigating different options for async behavior in our Redux actions.
After doing some reading, it looks like we can't use Redux middleware yet (Sagas and Thunks are still out :disappointed:)
- [Twilio documentation](https://www.twilio.com/docs/flex/developer/ui/redux#:~:text=You%20might%20be%20wondering%20why,this%20is%20not%20yet%20supported)
- [Github issue referencing this exact same thing](https://github.com/twilio/flex-plugin-builder/issues/171)

However, we can use [redux-promise-middleware](https://github.com/pburtchaell/redux-promise-middleware) ([as it is bundled with Flex](https://www.twilio.com/docs/flex/developer/ui/redux#write-asynchronous-actions)).
The idea of this middleware is that we provide a single action creator of the shape of 
```
const asyncAction = () => ({
  type: 'ACTION_TYPE',
  payload: new Promise(...),    // or asyncFunction()
})
```
and the middleware will capture the action, and since it's `payload` is a `Promise`, it will decompose it into three different actions: 
- `ACTION_TYPE_PENDING`: dispatched when `asyncAction` is called.
- `ACTION_TYPE_FULFILLED`: dispatched when the promise resolves successfully. The `payload` of this second action will be the value to which the promise resolved.
- `ACTION_TYPE_REJECTED`: dispatched when the promise is rejected. The `payload` of this second action will be the value to which the promise was rejected.

The pros:
- We don't need to manually write actions creators to express the different actions involved in an async action.
- Easy to use.

The cons:
- Does not really adds much value, since we still need to express the reducer logic and the action types. The second can be somewhat mitigated if we use [createAsyncAction](https://github.com/pburtchaell/redux-promise-middleware/tree/main/examples/using-promise-actions#using-promise-middleware-with-redux-promise-middleware-actions) helper.
- Forces us to separate the complex logic that involves multiple dispatches (e.g. fetch missing definition versions after loading the case list). Since this is the main use case to advocate for thunks/sagas, I'm not fully convinced of refactoring our current code like in this PoC, but happy to change my mind if the team thinks it's a good idea.

This PR have two different commits to review:
- 56badcb0fdf3b5ac860c7c61f8aff6499601c9ce, where I'm demonstrating how we could use this middleware to express our async actions.
- 19b3e489761b956db8a357605f237640ca19ed8f, where I've introduced a custom hook to handle the "fetch missing definition versions" logic. The idea of this is to show how we could share this behavior between our components, if we migrate to redux pomise and don't want to repeat ourselves.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/jira/software/c/projects/CHI/boards/7?modal=detail&selectedIssue=CHI-1237)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Checking out to both commits should work as usual, specially around the case list view. Note that I've introduced a case with version `co-v1`, to include different definition versions.
- Changing `fetchCaseList` (`src/states/caseList/listContent.ts`) to return `payload: Promise.reject(new Error('Booom!')),` should trigger the error page for this view.

### DIFF
--- a/plugin-hrm-form/src/components/caseList/CaseList.tsx
+++ b/plugin-hrm-form/src/components/caseList/CaseList.tsx
@@ -8,16 +8,14 @@ import Case from '../case';
 import { StandaloneITask, ListCasesQueryParams, ListCasesFilters, ListCasesSort } from '../../types/types';
 import CaseListTable from './CaseListTable';
 import { CaseListContainer, CenteredContainer, SomethingWentWrongText } from '../../styles/caseList';
-import { listCases } from '../../services/CaseService';
 import { CaseLayout } from '../../styles/case';
 import * as CaseActions from '../../states/case/actions';
-import * as ConfigActions from '../../states/configuration/actions';
 import { StandaloneSearchContainer } from '../../styles/search';
-import { getCasesMissingVersions } from '../../utils/definitionVersions';
 import { caseListBase, namespace, RootState } from '../../states';
 import { undoCaseListSettingsUpdate } from '../../states/caseList/reducer';
 import { dateFilterPayloadFromFilters } from './filters/dateFilters';
 import * as ListContent from '../../states/caseList/listContent';
+import { useLoadDefinitionVersions } from '../../hooks/useLoadDefinitionVersions';
 
 export const CASES_PER_PAGE = 10;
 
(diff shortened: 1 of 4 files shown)


stephenhand:
Sorry Gian, I didn't realise you'd done this when I did my resources placeholder search PR (#1214)

PR #1214, opened 2023-02-01
CHI-1464: Placeholder search

stephenhand:
## Description

* Resources search form
* Resources search results
* Redux layer implemented with `react-promise-middleware` and `react-middleware-promise-acton`

### Checklist
- [x] New tests added
- [x] Feature flags added
- [ ] Strings are localized

### Related Issues
CHI-1646

### Verification steps

See ticket


plugin-hrm-form/src/components/resources/search/SearchResourcesResults.tsx
```diff
@@ -0,0 +1,146 @@
...
+            <p>{error.message}</p>
+          </>
+        )}
+        {currentPageResults.map(result => (
+          <ResourcePreview
+            key={result.id}
+            resourceResult={result}
+            onClickViewResource={() => viewResource(result.id)}
+          />
+        ))}
```
murilovmachado, line 127: Can we make it an HTML `list` of `ResourcesPreview`?
Using semantic HTML makes it much easier for accessibility.

plugin-hrm-form/src/states/resources/index.ts
```diff
@@ -65,6 +67,16 @@ export const viewResourceAction = (id: string): ViewResourceAction => ({
...
 
+const OPEN_SEARCH = 'resource-action/open-search';
+
+type OpenSearchAction = {
+  type: typeof OPEN_SEARCH;
+};
+
+export const openSearchAction = (): OpenSearchAction => ({
+  type: OPEN_SEARCH,
+});
```
murilovmachado, line 78: I think `navigateToSearchAction` and `NAVIGATE_TO_SEARCH` would be more clear. The first time I read `openSearch` it was not super clear to me what it was actually doing. But this is just my opinion and also very minor, feel free to keep or change these names.

plugin-hrm-form/src/translations/en-US/flexUI.json
```diff
@@ -405,5 +405,9 @@
...
 
-  "Resources-LoadResourceError": "Something went wrong trying to load this resource."
+  "Resources-LoadResourceError": "Something went wrong trying to load this resource.",
```
murilovmachado, line 408: Should it be `Resources-ResourceSearchError`?

murilovmachado (approved):
Looks good to me. But I couldn't really test everything, I don't know what search term would actually return results.

It seems it's missing the label `Resources-ResourceSearchError`:
image

stephenhand:
I think the version of HRM that is required for the search to work was undeployed from dev, I will address your other comments. Hopefully the HRM PR will be merged soon so there will be no risk of the endpoint being undeployed

plugin-hrm-form/src/components/resources/search/SearchResourcesResults.tsx
stephenhand: We don't do this with the other search results, they are just divs. How exactly would it be structured? A `ul` tag enclosing the whole list and a `li` tag around each item - just enclosing the current structure?

murilovmachado: Yeah, that would be enough.

About the old search results screen, that's a tech debt.

stephenhand:
@murilovmachado - your comments should be addressed now, and a compatible version of HRM should be on dev

murilovmachado:
@stephenhand Thanks, both code and behavior look good to me!

GPaoloni (approved):
Looks great!

PR #1217, opened 2023-02-02
home page shortcut link to counselor toolkit

acedeywin:

## Description

- New feature: Home page shortcut link to counselor toolkit

### Checklist
- [N/A] New tests added
- [x] Feature flags added
- [x] Strings are localized

### Related Issues
Fixes CHI-1547

### Verification steps

- Verify that there is an information icon on the side nav
- Click on the information Icon on the side nav
Screenshot 2023-02-02 at 11 20 05 AM


plugin-hrm-form/src/components/toolkits/setUpCounselorToolkits.tsx
```diff
@@ -0,0 +1,31 @@
...
+import React from 'react';
+
+import CounselorToolkitSideLink from './CounselorToolkitSideLink';
+import { getConfig } from '../../HrmFormPlugin';
+
+// eslint-disable-next-line import/no-unused-modules
+export const setUpCounselorToolkits = () => {
+  const { featureFlags } = getConfig();
+  // eslint-disable-next-line no-unused-expressions
+  featureFlags.enable_counselor_toolkits &&
```
stephenhand, line 27: We should also check if a kmsUrl is configured on the default helpline before showing this icon, in addition to the feature flag

plugin-hrm-form/src/components/toolkits/setUpCounselorToolkits.tsx
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+
+import * as Flex from '@twilio/flex-ui';
+import React from 'react';
+
+import CounselorToolkitSideLink from './CounselorToolkitSideLink';
+import { getConfig } from '../../HrmFormPlugin';
```
stephenhand, line 21: `getConfig` from `HrmFormPlugin` is deprecated, use `getHrmConfig` from `hrmConfig` instead

plugin-hrm-form/src/styles/global-overrides.css
```diff
@@ -43,6 +43,10 @@ div.Twilio-MainHeader button.MuiButton-root {
     background-color: rgb(18, 28, 45);
     border: none;
 }
+div.css-1nnjdk7:hover {
```
stephenhand, line 46: I'm not sure what effect this is implementing but we can't override the autogenerated CSS class names like this - as well as being very obfuscated and hard to maintain, these generated names could change at any time, andf the effect would be lost

We need to try to achieve this with CSS that doesn't reference one of these class names, or if this is not achievable we need to go back to Jana & Dee

stephenhand (changes requested):
Couple of changes, the only tricky one might be the CSS override

plugin-hrm-form/src/components/toolkits/setUpCounselorToolkits.tsx
acedeywin: @stephenhand `featureFlags` does not exist in `getHrmConfig` but in `getConfig` of `hrmConfig`.

stephenhand: `getAseloFeatureFlags()` can be used to get the feature flages from `hrmConfig`

acedeywin: @stephenhand Done!

plugin-hrm-form/src/components/toolkits/setUpCounselorToolkits.tsx
stephenhand: @acedeywin - this one is still outstanding

stephenhand (approved):

stephenhand (changes requested):
Oops, forgot there was 1 change still outstanding

plugin-hrm-form/src/components/toolkits/setUpCounselorToolkits.tsx
acedeywin: @stephenhand, I have implemented the changes. I did that in the `plugin-hrm-form/src/components/toolkits/CounselorToolkitSideLink.tsx` file.

stephenhand (approved):

PR #1221, opened 2023-02-03
Audio notifications test Coverage

mythilytm:
Primary reviewer:

## Description
- Includes additional error handling for audio notifications
- Added test suite for notifyNewMessage and notifyReservedTask
-
### Checklist
- [x] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Related Issues
Fixes #....


plugin-hrm-form/src/___tests__/utils/audioNotifications.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        url: 'http://assets.fake.com/notifications/ringtone.mp3',
+        repeatable: true,
+      },
+      expect.any(Function),
+    );
+    const taskStatuses = ['accepted', 'canceled', 'rejected', 'rescinded', 'timeout'];
+    taskStatuses.forEach(status => {
+      expect(reservation.on).toHaveBeenCalledWith(status, AudioPlayerManager.stop);
+    });
+  });
```
mythilytm, line 148: I am struggling to have this test case defined. Especially at line 145, I am hoping for some help pass this test.

mythilytm:
Currently, I am testing the following scenarios:
Notifications for new messages
- audio notification should play when author of a message is the customer (or Bot or System) and the document is visible
- audio notification should not play when author of a message is the customer (or Bot or System) but the document is not visible
- audio notification should not play when author of a message is the counselor even when document is visible

Notifications for reserved tasks
- audio notification should play when a reservation is created with an online agent
- audio notification should not play when document is not visible but task is reserved
- audio notification should stop when the reservation changes its state to one of the specified statuses
- audio notification should stop when the timeout ends(after 15000ms)
 
 I am happy to get suggestions to add meaningful tests.

plugin-hrm-form/src/___tests__/utils/audioNotifications.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  Manager: {
+    getInstance: () => mockFlexManager,
+  },
+  AudioPlayerManager: {
+    play: jest.fn(),
+    stop: jest.fn(),
+  },
+}));
+
+describe('notifyNewMessage', () => {
```
stephenhand, line 47: As we previously discussed, you shouldn't be making private methods public to test them - we should test the public methods

In the case of event subscribers, a good technique is to 'capture' the handler functions by mocking the subscribe method - so for example:

```
describe('subscribeReservedTaskAlert handler', () => {
  let handlerToTest
  beforeEach(() => {
    // Assumes manager.workerClient.on has been mocked as a jest.fn()
    manager.workerClient.on.mockimplementation((event, handler) => handlerToTest);
    // Now the notifyReservedTask handler will be assigned to 'handlerToTest' by the above mock implementation, ready to use in your tests
    subscribeReservedTaskAlert();
  );

  test('Does something', () => {
    handlerToTest(...);
    expect(...)
    ...
  });
})

```

plugin-hrm-form/src/___tests__/utils/audioNotifications.test.ts
stephenhand: This is because you do not call `reservation.on` with `AudioPlayerMessage.stop` - you call it with an anonymous function that wraps `AudioPlayerMessage.stop`

There are only 2 ways around this, either you make the wrapped function referenecable from the test, which is icky because you have to either make it a public property of the module, exposing private implementation details, or passing it in, which is polluting the public API with useless parameters just for tests

OR

You have to use the same 'capture' technique outlined in the comment above - i.e. your expectation here only checks that any function was provided, but you provide `on` with a `mockImplementation` that assigns the handler to a variable. Then you can call the handler separately in your test and verify that it does in fact call `AudioPlayerMessage.stop`.

The latter is more hoops to jump through, but it doesn't require you to make any compromises with your production code

stephenhand:
Actually - looking into it, you don't have to explicitly capture function arguments using mockImplementation and an external variable if you don't want to

Jest keeps a record of every time a mock function was called and the arguments it was called with on the mocks property, so you can just grab the passed in handlers from there after the subscribe method is called, without the need for any prior setup

https://stackoverflow.com/questions/41939511/how-can-i-get-the-arguments-called-in-jest-mock-function

stephenhand:
Also this file and it's associated tests shouldn't be in `utils`, we should be emptying this folder, not piling more stuff in there. It should either be in a meaningful folder with related functionality or on the top level if no such folder already exists or could be created

PR #1222, opened 2023-02-04
Configuration API tidy up, local feature flag overrides

stephenhand:
## Description

The main goal of this is to prevent weird issues cropping up in tests caused by pulling in the module that defines the root component (which naturally has a huge clatter of it's own deps) just to use config. As well as preventing issues it might speed the unit tests up a little, because many tests will be pulling in a much smaller dependency tree.

In addition I added a way to locally override feature flags and split the CI into separate linting & testing jobs

* Removes everything exported from `HrmFormPlugin.tsx` except the component itself
* Converts SearchForm.jsx from a JS class component to a TS functional component
* Splits CI into test & lint workflows

Substantial code changes:

[.github/workflows/plugin-hrm-form-ci.yml](https://github.com/techmatters/flex-plugins/pull/1222/files#diff-8fadd5be0e45e81bdc8dc1054cd57c3be00216ccb037137922eb3960e7873bd1)
[plugin-hrm-form/src/HrmFormPlugin.tsx](https://github.com/techmatters/flex-plugins/pull/1222/files#diff-682c3153310c88c02dbc47bb1da8e7f4cabb7eb1c50223b37adf0cd114018939)
[plugin-hrm-form/src/components/search/SearchForm/index.tsx](https://github.com/techmatters/flex-plugins/pull/1222/files#diff-df20b8bb2a1673466682ac2bb05228240fe0ff46e77de7798c5e679bf6a34345)
[plugin-hrm-form/src/hrmConfig.ts
](https://github.com/techmatters/flex-plugins/pull/1222/files#diff-f0628c848601806b1655b83645667f8886f2a606e44250e440a3b949113cb705)

### Checklist
- N/A Corresponding issue has been opened
- [X] New tests added
- [X] Feature flags added
- [X] Strings are localized

### Related Issues
None

### Verification steps

Automated tests should cover it


plugin-hrm-form/src/components/search/SearchForm/index.tsx
```diff
@@ -0,0 +1,268 @@
+/**
```
stephenhand, line 1: I changed this too much & git deleted / recreated 😢 I can go back and recommit the changes in a way that gets git to move it instead if preferred, but I think the diff would be a complete sh*tshow anyway, given how much is changed

plugin-hrm-form/src/components/decorateTab.jsx
```diff
@@ -19,7 +19,7 @@ import ErrorIcon from '@material-ui/icons/Error';
 import { Template } from '@twilio/flex-ui';
 
 import { StyledTab } from '../styles/HrmStyles';
-import { formIsValid } from '../states/ValidationRules';
+import { formIsValid } from '../states/validationRules';
```
GPaoloni, line 22: I don't think this file is used at all. Maybe we can get rid of it?

plugin-hrm-form/src/components/search/SearchForm/index.tsx
GPaoloni: >Converts SearchForm.jsx from a JS class component to a TS functional component

:raised_hands:

plugin-hrm-form/src/hrmConfig.ts
```diff
@@ -48,8 +51,15 @@ const readConfig = () => {
...
-  const featureFlags: FeatureFlags = manager.serviceConfiguration.attributes.feature_flags || {};
+  const featureFlagsFromEnvEntries = Object.entries(process.env)
+    .filter(([varName]) => varName.startsWith(featureFlagEnvVarPrefix))
+    .map(([name, value]) => [
+      name.substring(featureFlagEnvVarPrefix.length),
+      (value ?? 'false').toLowerCase() === 'true',
+    ]);
+  const featureFlagsFromEnv = Object.fromEntries(featureFlagsFromEnvEntries);
+  const featureFlagsFromServiceConfig: FeatureFlags = manager.serviceConfiguration.attributes.feature_flags || {};
+  const featureFlags = { ...featureFlagsFromServiceConfig, ...featureFlagsFromEnv };
```
GPaoloni, line 62: :monocle_face: Allows to configure feature flags from env vars?

GPaoloni (approved):
Looks very good to me, thanks for the cleanup!

The only suggestion I have is renaming `getResourceStrings`, as resource now have a different meaning in our system, it's kinda confusing. Just `getStrings` should suffice?

PR #1223, opened 2023-02-05
CHI-1569 terraform refactor

robert-bo-davis:
Primary reviewer:

## Description

Initial POC work on terraform refactor using terragrunt and dynamic configuration. After some thought and experimentation, moving forward with terragrunt will make the refactor MUCH less painful.

This has been split into a multiple PR tree to (try to) simplify review. They are:

#1312
#1313
#1314
#1315
#1316 
#1317
#1318

## Backwards Compatibility

At this stage backwards compatibility has been prioritized over fully implemented best practices. The separation of concerns between helpline configuration and running stages/modules is still a bit blurry in this iteration. This is a start, but once migration is complete, and as we migrate away from Autopilot more refactoring should done to get closer to the goal of stages/modules being fully configuration driven. 

## What is missing/next?

- The TS scripts in `./scripts` haven't been updated to support terragrunt in this PR
- Splitting flex/serviceConfiguration module into its own system instead of using null_resource
- Because of missing script updates, this isn't ready for full apply of the config stage yet, only plan
- Channel configuration is not nearly as configuration driven as it should be so that we can maintain backwards compatibility and to simplify migration. It should be refactored once helplines are migrated.

## Terragrunt Basics

Terragrunt works by generating files using `generate` blocks for things like [the backend configuration](https://github.com/techmatters/flex-plugins/blob/da6fd860f0c38e2343a7485a54f71007221de566/twilio-iac/stages/terragrunt.root.hcl#L44) and [providers](https://github.com/techmatters/flex-plugins/blob/da6fd860f0c38e2343a7485a54f71007221de566/twilio-iac/stages/terragrunt.root.hcl#L50). 

It uses an `inputs` [block](https://github.com/techmatters/flex-plugins/blob/da6fd860f0c38e2343a7485a54f71007221de566/twilio-iac/stages/provision/terragrunt.hcl#L23) to define variables to be passed to a module defined in the `terraform` block. In the example we are just passing through] the terragrunt `locals.config` as variables for the [terraform apply of the specified stage module](https://github.com/techmatters/flex-plugins/blob/da6fd860f0c38e2343a7485a54f71007221de566/twilio-iac/stages/provision/terragrunt.hcl#L41).

See the [terragrunt overview doc](https://github.com/techmatters/flex-plugins/blob/da6fd860f0c38e2343a7485a54f71007221de566/twilio-iac/docs/terragrunt.md) for more details.

## Directory Structure in the future

The basic layout when all is done in the twilio-iac directory will be a more consolidated tree:

### `helplines`

This contains the configurations for each helpline and environment. These configuration `.hcl` files are organized for cascading overrides. The structure of this is not set in stone and can be easily expanded on in the future as the need arises, but in this example it goes like so:

`helplines/defaults.hcl` should contain global defaults. 
`helplines/{short_helpline}/common.hcl` contains defaults for the helpline.
`helplines/{short_helpline}/{environment.hcl` contains configuration overrides for each environment.
`helplines/{short_helpline}/files/*.tf contains helpline specific files to be copied directly into a stage
`helplines/{short_helpline}/templates/*.tftpl contains helpline specific templates to be used by terraform an/or the terragrunt configuration system

### `makefiles`

The makefile system lives here.

### `stages`

This directory contains the main `.hcl` file used to tell terragrunt what to do. It also contains Makefiles because it is where a user (or github action) will go to run `make apply`. 

All of the make commands take two arguments in this example:
`HL` - the short_helpline code to tell terragrunt and all other systems which helpline is being acted on.
`HL_ENV` - an environment. This could default to `staging` so it only has to be provided if you want to act on some other environment.

So a full apply make command for mt staging is `make apply HL=mt HL_ENV=staging`.

Terragrunt also uses directory traversal to provide some magic plan/apply steps that should be used sparingly, but can help with verifying applies of all stages. From the `stages` root instead of a specific stage, you can run `make plan-all HL=mt HL_ENV=staging`. This will show the plan output for an apply of all stages. There are quite a few caveats we can discuss in detail later.

The layout of the terragrunt config will likely evolve over various iterations, but this is probably good enough to solve today's problems effectively.

### `modules`

This is where the terraform modules live, I'll likely rename from `terraform-modules` to just `modules` in a later PR since that is a more common naming convention.

### `translations`

Same structure as before migration

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


robert-bo-davis:
This has been split into a multiple PR tree to (try to) simplify review. They are:

#1312
#1313
#1314
#1315
#1316 
#1317
#1318

PR #1224, opened 2023-02-06
Chi 1661 caselist date filter error

acedeywin:

## Description

Bug fix - fix case list date filter error. While investigating the fix to this issue, I found that we can only use onClick to perform event actions rather than onChange while filtering form date.
### Checklist
- [N/A] New tests added
- [N/A] Feature flags added
- [N/A] Strings are localized

### Related Issues
Fixes CHI-1661

### DIFF
--- a/plugin-hrm-form/src/components/caseList/filters/DateRangeFilter.tsx
+++ b/plugin-hrm-form/src/components/caseList/filters/DateRangeFilter.tsx
@@ -5,6 +5,7 @@ import { Template } from '@twilio/flex-ui';
 import { endOfDay, format, parse } from 'date-fns';
 import ArrowDropUp from '@material-ui/icons/ArrowDropUp';
 import ArrowDropDown from '@material-ui/icons/ArrowDropDown';
+import _ from 'lodash';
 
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/components/caseList/filters/DateRangeFilter.tsx
```diff
@@ -74,7 +75,8 @@ const formToDateFilter = (
   filterOptions: DateFilterOption[],
   values: ReactHookFormValues,
 ): DateFilterValue | undefined => {
-  const [, selected] = filterOptions.find(([opt]) => opt === values[selectedOptionField]);
+  const { ...selected } = Object(filterOptions.find(([opt]) => opt === values[selectedOptionField]))[1];
```
stephenhand, line 78: This works but is a little messy and hard to follow, all we really need to do to handle the undefined cases is add some defensive checks:

````ts
const [, selected] = filterOptions?.find(([opt]) => opt === values[selectedOptionField]) ?? []
````

acedeywin: @stephenhand, you are looking at the wrong PR. [Here](https://github.com/techmatters/flex-plugins/pull/1228) is the updated PR. Your suggested implementation works perfectly.

PR #1225, opened 2023-02-06
Refactor reserved task notification logic

mythilytm:
Primary reviewer: @GPaoloni 

## Description
- Checking reservation status on an interval to ensure clean up if notification clean up is missed. 
- ~~Change setTimeout to 2 minutes(from 15 seconds)~~  Not anymore. Removing this

### Checklist
- [n/a] Corresponding issue has been opened
- [x] New tests added ( in a different PR)
- [n/a] Feature flags added
- [n/a] Strings are localized

### Related Issues
Fixes [remove 15 second timer ](https://tech-matters.slack.com/archives/CV4QM23UJ/p1675437711969629?thread_ts=1675198539.557809&cid=CV4QM23UJ)

### Verification steps
- Check for behavior around stopping media on a reserved task.

### DIFF
--- a/plugin-hrm-form/src/utils/audioNotifications.ts
+++ b/plugin-hrm-form/src/utils/audioNotifications.ts
@@ -104,9 +104,7 @@ const notifyReservedTask = reservation => {
       reservation.on(status, stopAudio);
     });
 
-    setTimeout(() => {
-      stopAudio();
-    }, 120000);
+    setTimeout(stopAudio, 120000);
 
     const checkForReservedTask = () => {
       if (reservation.task.status === 'reserved') {
--- a/plugin-hrm-form/src/utils/audioNotifications.ts
+++ b/plugin-hrm-form/src/utils/audioNotifications.ts
@@ -76,6 +76,12 @@ const notifyNewMessage = messageInstance => {
   }
 };
 
+/** notifyReservedTask plays ringtone when an agent has a reserved task in pending state.
+ *  The notification stops when the reservation is not longer in pending. 
+ *  There is a check, checkForPendingReservation, in case, notification does not stop as expected.
+ * 
+ * @param reservation 
+ */
 const notifyReservedTask = reservation => {
   try {
     const { assetsBucketUrl } = getHrmConfig();
@@ -97,24 +103,16 @@ const notifyReservedTask = reservation => {
       );
     }
 
-    const stopAudio = () => AudioPlayerManager.stop(media);
+    const stopAudio = () => media && AudioPlayerManager.stop(media);
 
     const taskStatuses = ['accepted', 'canceled', 'rejected', 'rescinded', 'timeout'];
     taskStatuses.forEach(status => {
       reservation.on(status, stopAudio);
     });
 
-    setTimeout(stopAudio, 120000);
-
-    const checkForReservedTask = () => {
-      if (reservation.task.status === 'reserved') {
-        setTimeout(checkForReservedTask, 5000);
-      } else {
-        stopAudio();
-      }
-    };
-
-    checkForReservedTask();
+    const checkForPendingReservation = () =>
+      reservation.status === 'pending' ? setTimeout(checkForPendingReservation, 5000) : stopAudio;
+    checkForPendingReservation();
   } catch (error) {
     console.error('Error in notifyReservedTask:', error);
   }
--- a/plugin-hrm-form/src/utils/audioNotifications.ts
+++ b/plugin-hrm-form/src/utils/audioNotifications.ts
@@ -76,11 +76,12 @@ const notifyNewMessage = messageInstance => {
   }
 };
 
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/utils/audioNotifications.ts
```diff
@@ -14,81 +14,112 @@
...
+
+    const checkForReservedTask = () => {
+      if (reservation.task.status === 'reserved') {
+        setTimeout(checkForReservedTask, 5000);
+      } else {
+        stopAudio();
+      }
+    };
+
+    checkForReservedTask();
```
mythilytm, line 119: Changes to cleanup audio notifications logic here...

plugin-hrm-form/src/utils/audioNotifications.ts
```diff
@@ -14,81 +14,112 @@
...
+          repeatable: true,
+        },
+        (error: AudioPlayerError) => {
+          console.log('AudioPlayerError:', error);
+        },
+      );
+    }
 
-  let media;
+    const stopAudio = () => AudioPlayerManager.stop(media);
```
GPaoloni, line 100: Is there any risk of raising errors if we call `AudioPlayerManager.stop` whit something nullish? If so a check like `const stopAudio = () => media && AudioPlayerManager.stop(media);` may be worth?

plugin-hrm-form/src/utils/audioNotifications.ts
GPaoloni, line 117: I'm not against all of this to make sure the alert goes off, but may be an overkill. I think just relying on `checkForReservedTask` is enough. Worth asking to someone else in the team to get extra eyes on this!

GPaoloni (approved):
Looks great and works good in my end!

plugin-hrm-form/src/utils/audioNotifications.ts
mythilytm: Makes sense!

plugin-hrm-form/src/utils/audioNotifications.ts
mythilytm: @stephenhand Curious about your input.

plugin-hrm-form/src/utils/audioNotifications.ts
```diff
@@ -14,81 +14,110 @@
...
+
+    setTimeout(stopAudio, 120000);
+
+    const checkForReservedTask = () => {
+      if (reservation.task.status === 'reserved') {
+        setTimeout(checkForReservedTask, 5000);
+      } else {
+        stopAudio();
+      }
+    };
```
GPaoloni, line 115: A note on this that I've realized after approving the PR: here you are checking the `reservation.task.status`, which is not the same as the `reservation.status`. I think you want the later here, You can see more of the reservation status [here](https://www.twilio.com/docs/taskrouter/api/reservations#:~:text=Possible%20values%20for%20ReservationStatus%3A), but TL;DR, you can keep waiting for as long as `reservation.status === 'pending'`.

The reason is that the `reservation.task` represents the [task router task resource](https://www.twilio.com/docs/taskrouter/api/task), which will be shared among all the workers until one picks it up (accepts it). The reservation resource on the other hand is created each time the task is reserved to a worker. If it accepts it, that's the last reservation for this task, but if not, more reservations will be created until it gets accepted/canceled.

plugin-hrm-form/src/utils/audioNotifications.ts
GPaoloni: @mythilytm as a note related to [this comment](https://github.com/techmatters/flex-plugins/pull/1225#discussion_r1097909619), note that what you are calling `taskStatuses` are actually reservation statuses. Can you rename the variable to make it clear?

plugin-hrm-form/src/utils/audioNotifications.ts
```diff
@@ -97,24 +103,16 @@ const notifyReservedTask = reservation => {
...
-      if (reservation.task.status === 'reserved') {
-        setTimeout(checkForReservedTask, 5000);
-      } else {
-        stopAudio();
-      }
-    };
-
-    checkForReservedTask();
+    const checkForPendingReservation = () =>
+      reservation.status === 'pending' ? setTimeout(checkForPendingReservation, 5000) : stopAudio;
```
GPaoloni, line 115: I think you want to call `stopAudio` here

```suggestion
      reservation.status === 'pending' ? setTimeout(checkForPendingReservation, 5000) : stopAudio();
```

plugin-hrm-form/src/utils/audioNotifications.ts
mythilytm: Thanks @GPaoloni.  Changed to `checkForPendingReservation`

GPaoloni (approved):
LGTM

PR #1231, opened 2023-02-07
Removed chain of dead code

GPaoloni:
## Description
This PR removes a chain of code that was actually not used elsewhere than in their own definitions and tests. I pulled from this thread while reviewing https://github.com/techmatters/flex-plugins/pull/1222 :P 

### Verification steps
- Tests are not broken.
- Manual review that I'm not removing stuff that's actually used.


stephenhand (approved):
I always approve of deleting code! LGTM once the conflicts are sorted

PR #1234, opened 2023-02-08
Fetch definitions from S3 instead of bundling

murilovmachado:
## Description
Make Aselo fetch form definitions from S3 instead of bundling them.

### Related Issues
Fixes [CHI-1663](https://tech-matters.atlassian.net/browse/CHI-1663)

### Verification steps
Build `plugin-hrm-form` and check that the form definition JSONs are not bundled.
Check on the network tab that all form definitions are fetched individually.


GPaoloni (approved):
Looks and works good to me!

plugin-hrm-form/src/definitionVersions.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (definitionVersionId === 'demo-v1') return 'as';
+  if (definitionVersionId === 'v1') return 'zm';
+
+  return definitionVersionId.substring(0, 2);
+};
+
+const getVersionFromDefinitionVersionId = (definitionVersionId: string) => {
+  if (definitionVersionId === 'v1') return 'v1';
+
+  return definitionVersionId.substring(definitionVersionId.length - 2);
```
GPaoloni, line 45: Comment got lost in the review. It's fine to have this just let's remember the codes can't be longer 2 characters neither the versions (`ukr-v1` or `as-v11` won't work).

plugin-hrm-form/src/definitionVersions.ts
```diff
@@ -0,0 +1,54 @@
+/**
```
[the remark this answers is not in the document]
robert-bo-davis: Hopefully there is a relatively quick followup iteration to get rid of all of this hackery before it bites us. 😜

plugin-hrm-form/src/definitionVersions.ts
robert-bo-davis: Hopefully there is a relatively quick followup iteration to get rid of all of this hackery before it bites us. 😜

robert-bo-davis (approved):
tested locally a bit and it LGTM! I'm interested to see the performance difference. We should definitely do a few before and after throttled loads via chrome dev tools so we can compare. That doesn't work in local dev.

stephenhand (commented):
There's a couple of places in the code where we call `getDefinition(...)` directly - they are all wrapped in a 'only if the definition isn't already in Redux' checks, but since some of them are in repeated components like case list rows, there's a possibility it could kick off a crapload of requests if lots of rows with an unloaded definition are rendered at the same time.

Since we aren't using form definitions versioning right now it's probably only a theoretical concern so doesn't need any action in the PR. That said, I think sooner rather than later, perhaps as part of the planned tidy up we should refactor definitions to be loaded up in the Redux layer using redux-middleware-promise and only ever get definitions from redux in the UI layer. 

This will help us avoid any performance 'landmines' in the future where we inadvertently write UI code that spams requests for definitions

plugin-hrm-form/src/definitionVersions.ts
murilovmachado: yeah, there are some hacky pieces in this file, but the idea is that we get rid of this whole file soon. If you look at the comments in the file, it's intended to be temporary.

plugin-hrm-form/src/definitionVersions.ts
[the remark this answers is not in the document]
murilovmachado: This file hopefully will be fully removed from the project soon. It's there just to deal with temporary workarounds.

PR #1242, opened 2023-02-10
CHI-1480: Address some race conditions loading components

stephenhand:
## Description

* Update any code that uses `currentVersionDefinition` to expect it to be potentially undefined
* Update some components that were calling `getDefinitionVersion` to map the configuration redux state to properties instead

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-1480

### Verification steps

* E2E tests should pass
* General regression

### DIFF
--- a/plugin-hrm-form/src/components/case/Case.tsx
+++ b/plugin-hrm-form/src/components/case/Case.tsx
@@ -115,14 +115,17 @@ const Case: React.FC<Props> = ({
 
   const timeline: Activity[] = useMemo(
     () => {
+      const { connectedCaseId, definitionVersions, connectedCaseState } = props;
       /**
        * Gets the activities timeline from current caseId
        * If the case is just being created, adds the case's description as a new activity
        */
-      if (!props.connectedCaseId) return [];
+      if (!connectedCaseId) return [];
+
+      const { connectedCase } = connectedCaseState;
 
       const timelineActivities = [
-        ...getActivitiesFromCase(props.connectedCaseState.connectedCase),
+        ...getActivitiesFromCase(connectedCase, definitionVersions[connectedCase.info.definitionVersion]),
         ...getActivitiesFromContacts(savedContacts ?? []),
       ];
 
@@ -215,7 +218,7 @@ const Case: React.FC<Props> = ({
       return firstConnectedContact.rawJson.caseInformation.categories;
     }
     if (form?.categories && form?.helpline) {
-      return transformCategories(form.helpline, form.categories);
+      return transformCategories(form.helpline, form.categories, definitionVersion);
     }
     return null;
   };
@@ -307,7 +310,6 @@ const Case: React.FC<Props> = ({
     summary,
     childIsAtRisk,
     office,
-    version,
     contact: firstConnectedContact,
     contacts: savedContacts ?? [],
   };
--- a/plugin-hrm-form/src/components/case/caseActivities.ts
+++ b/plugin-hrm-form/src/components/case/caseActivities.ts
@@ -14,10 +14,11 @@
  * along with this program.  If not, see https://www.gnu.org/licenses/.
  */
 
+import { DefinitionVersion } from 'hrm-form-definitions';
+
 import { Activity, ConnectedCaseActivity, NoteActivity } from '../../states/case/types';
 import { Case, NoteEntry, ReferralEntry } from '../../types/types';
 import { channelTypes } from '../../states/DomainConstants';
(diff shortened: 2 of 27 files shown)


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -208,14 +214,16 @@ const Case: React.FC<Props> = ({
 
   const handleCloseSection = () => changeRoute(closeSubSectionRoute(), task.taskSid);
 
-  if (!props.connectedCaseState) return null;
+  const definitionVersion = props.definitionVersions[version] ?? props.currentDefinitionVersion;
```
GPaoloni, line 217: I'm sure if it's correct to default to the `currentDefinitionVersion` if the `definitionVersions[version]` is not in state :thinking:

plugin-hrm-form/src/components/case/caseActivities.ts
```diff
@@ -117,9 +118,7 @@ const connectedContactActivities = (caseContacts): ConnectedCaseActivity[] =>
     })
     .filter(cca => cca);
 
-export const getActivitiesFromCase = (sourceCase: Case): Activity[] => {
-  const { definitionVersions } = getDefinitionVersions();
-  const formDefs = definitionVersions[sourceCase.info.definitionVersion];
+export const getActivitiesFromCase = (sourceCase: Case, formDefs: DefinitionVersion): Activity[] => {
```
GPaoloni, line 121: :raised_hands: I dream about the day only components consume this configurations from state/hooks and all the function get their dependencies as parameters haha

plugin-hrm-form/src/hrmConfig.ts
```diff
@@ -125,12 +125,10 @@ export const getTemplateStrings = () => cachedConfig.strings;
 export const getAseloFeatureFlags = (): FeatureFlags => cachedConfig.featureFlags;
 
 /**
- * Helper to expose the forms definitions without the need of calling Manager
+ * DO NOT USE IN REACT COMPONENTS! Using this in react components can lead to race conditions on loading where the component loads before the definition is ready
+ * Map the configuration redux state to the React component properties instead. This way if the component loads before the definitions are ready, it will reload once they are
+ * Ideally this should only be used in code that is invoked independently of React components, like Flex action event handlers
```
GPaoloni, line 130: Not for this PR for sure, but I was thinking that a good refactor to in the short term would be to use a custom hook `useDefinitionVersion(versionId: string)`, that internally:
- uses `useSelector` to fetch the correct definition and if it's not present
- runs an effect to fetch and load the state.

When/if we move to something like sagas that behavior can still be wrapped under the same custom hook, so we'll make our lives easier for improving this behavior. Anyways just a thought, I'll play with the idea if I get the time.

plugin-hrm-form/src/states/case/types.ts
```diff
@@ -97,7 +97,6 @@ export type CaseDetails = {
   summary: string;
   childIsAtRisk: boolean;
   office?: HelplineEntry;
-  version?: DefinitionVersionId;
```
GPaoloni, line 100: I think we still want to save which version was used to save this case?

GPaoloni (approved):
Looks good to me! Just leaving questions and thought as this issue with the definitions keeps biting us from time to time. Nothing blocking tho.

plugin-hrm-form/src/states/case/types.ts
stephenhand: We didn't seem to be using it anywhere? This is the `CaseDetails` type that primarily gets used for priniting, not the main `Case` type that gets used in most places

I'd like to get rid of that type one day - but not today :-P

plugin-hrm-form/src/components/case/Case.tsx
stephenhand: Yes I think you're right, just being overly defensive

plugin-hrm-form/src/hrmConfig.ts
stephenhand: I'm still not crazy about custom hooks as a paradigm - they feel a bit JFM to me now, but I'm sure you'll bring me round!

If we double down on redux the way I'd like though, I think most components will be mapping state to props anyway, so we won't need specialist hooks for specific state

I think we can move the whole 'fetch if not present' logic down into the redux layer even without thunks & sagas - if we can we just need standard state & dispatch mappings up in the component layer

PR #1249, opened 2023-02-13
Map hrm-test to staging, not development. Switch off E2E video

stephenhand:
## Description
* Map hrm-test to staging, not development. 
* Switch off E2E video

### DIFF
--- a/plugin-hrm-form/src/definitionVersions.ts
+++ b/plugin-hrm-form/src/definitionVersions.ts
@@ -32,7 +32,7 @@ const getEnvironmentFromHrmBaseUrl = (manager: Flex.Manager) => {
 
   /*
    * hrm-test is an alias of hrm-staging that we should deprecate & remove, but some accounts are still configured to point at it
-   * Thois ensures any accounts still pointing at hrm-test go to the right bucket for their assets
+   * This ensures any accounts still pointing at hrm-test go to the right bucket for their assets
    */
   if (environment === 'test') {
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/definitionVersions.ts
```diff
@@ -32,7 +32,7 @@ const getEnvironmentFromHrmBaseUrl = (manager: Flex.Manager) => {
 
   // Filter out E2E
   if (environment === 'test') {
-    return 'development';
+    return 'staging';
```
robert-bo-davis, line 35: The e2e env is in development, not staging right? I still don't completely understand what was going on in Dee's test failure, but it seems like deploying form change to development for e2e test that reside there may be the correct process instead of looking at staging for e2e test definitions?

stephenhand: So the E2E env is dev, and it points at `hrm-development` - it always has

`hrm-test` is an alias for the staging environment, which some of our older accounts still point to in our Service Configuration.

I'll update the inaccurate comment

nickhurlburt: For what it's worth, `hrm-test` was created as the staging endpoint when we first set up the terraformized infrastructure, because there was already an `hrm-staging` endpoint set up in the original Elastic Beanstalk architecture (which is now deleted). It will be good to get back to just calling it staging.

PR #1253, opened 2023-02-14
[Bug fix]: Audio notifications ring for transferred task

GPaoloni:
## Description
This PR fixes two bugs:
- The one reported in the ticket, where after a task was transferred, the original counselor will still receive message notifications that take place in the underlying conversation.
- Another bug that I spotted while working on this, where a rejected transferred task will sometimes (pretty often) hang on ringing, making **_really_** annoying to keep the tab open. I suspect this was caused because the calls to `AudioPlayerManager.start` and `AudioPlayerManager.stop` were very close to each other, causing the audio to be forever enqueued. To fix this issue I'm changing the approach to the "ring while reserved": instead than using a `repeatable` audio, we'll use a simple one, and every 500ms check if the task is still pending. If it is, just ring again and set a timeout for another 500ms. 

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1676)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Confirm that both issues are occurring in the version deployed to development (or `master`).
- `npm run dev` and confirm that none of the two issues are occurring in this version. Note: you'll need two counselor sessions, but only one of them in local server is enough.

### DIFF
--- a/plugin-hrm-form/src/utils/setUpTaskRouterListeners.ts
+++ b/plugin-hrm-form/src/utils/setUpTaskRouterListeners.ts
@@ -15,34 +15,67 @@
  */
 
 import { Manager, TaskHelper, StateHelper, ITask } from '@twilio/flex-ui';
+import type { Conversation } from '@twilio/conversations';
 
 import { FeatureFlags } from '../types/types';
+import * as TransferHelpers from './transfer';
 
-const removeConversationListeners = (task: ITask) => {
+const removeConversationListeners = (conversation: Conversation) => {
+  const safelyRemoveListeners = (eventName: string) => {
+    try {
+      if (conversation.listenerCount(eventName)) {
+        conversation.removeAllListeners(eventName);
+      }
+    } catch (err) {
+      console.error(`Failed to safelyRemoveListeners with event ${eventName}`, err);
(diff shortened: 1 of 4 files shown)


plugin-hrm-form/src/utils/setUpTaskRouterListeners.ts
```diff
@@ -15,34 +15,67 @@
...
 
 import { Manager, TaskHelper, StateHelper, ITask } from '@twilio/flex-ui';
+import type { Conversation } from '@twilio/conversations';
 
 import { FeatureFlags } from '../types/types';
+import * as TransferHelpers from './transfer';
 
-const removeConversationListeners = (task: ITask) => {
+const removeConversationListeners = (conversation: Conversation) => {
+  const safelyRemoveListeners = (eventName: string) => {
```
stephenhand, line 24: Sucks that there isn't a built in method for this

stephenhand (commented):
LGTM

We should probably look at adding some test coverage in another PR

stephenhand (approved):

PR #1254, opened 2023-02-14
normalizeEmail refactor and document visibility change

mythilytm:
Primary reviewer: @GPaoloni 

## Description
- Audio alerts play only when document is NOT visible
- Email normalization to account for several instances of `@` and `.`

### Related Issues
Fixes #....


GPaoloni:
@mythilytm I've addressed this two in [my current PR](https://github.com/techmatters/flex-plugins/pull/1253), with a slight difference: I'm using the `conversationsClient` `identity`, which should always match with exact format of the `messageInstance.author`. Can you give a few rounds of testing to that change and see if this fixes the issue without the need of normalizing the email?

mythilytm:
Closing this - changes are addressed in https://github.com/techmatters/flex-plugins/pull/1253

PR #1262, opened 2023-02-15
CHI-1421: Run a local HTTP server to load form definitions for local dev

stephenhand:
## Description

* Adds an option to set the form definition base URL from an environment variable or Service Configuration
* For `npm run dev`, runs an instance of `http-server` pointing at the form definitions root folder and points the flex instance at it for form defs
* Added `dev:s3-forms` script which gives the old `dev` behaviour

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes #....

### Verification steps

* Run `npm run dev` and verify form defs come from a local server
* Run `npm run dev s3-forms` and verify form defs come from the remote assets URL
* Deploy flex to Aselo Development and check it still works


plugin-hrm-form/package.json
```diff
@@ -11,7 +11,9 @@
     "preinstall": "cd ../hrm-form-definitions && npm ci && npm run build",
     "postinstall": "flex-plugin pre-script-check",
     "prestart": "flex-plugin pre-start-check",
-    "dev": "cross-env UNBUNDLED_REACT=true NO_MONITORING=true npm start",
+    "dev:local-form-definition-server": "npx http-server ../hrm-form-definitions/form-definitions -p 8090 --cors",
```
stephenhand, line 14: These hardcoded ports are probably going to start causing issues eventually, but using ephemeral ports in scripts is a bit of a PITA, I think we're alright yet.

murilovmachado (approved):
Works fine for me!

plugin-hrm-form/package.json
robert-bo-davis: I started putting together a doc when I first began sorting through all of the hardcoded ports across projects, but I'm not sure I actually put it up publicly anywhere. I'll see if I can pull it up so we at least have something to reference when we spin up a new service. I can't think of any (reasonable) strategies for managing ephemeral ports across our various services. If you have ideas I'd love to hear them!

robert-bo-davis (approved):
Didn't test, but LGTM!

PR #1265, opened 2023-02-16
CHI-1421: Custom contact components

stephenhand:
## Description

* Support for custom components in forms:
* Adds a 'CustomContactComponent' form item type. This allows you to specify a custom component to include in your form - the type of component is specified using the name it was registered under (see below)
  * Adds a custom component registry to the plugin. This allows you to 'register' react components that you wish to use in your forms.
  * You register components by wrapping them in a generator function that passes in a taskSid or a contact ID iinto the component (the idea of the CustomContactComponent is that the component works with contact information, so expects either a task for new contacts or a contact for existing ones)
  * Adds a 'context' parameter to the form generation custom hook, allowing you to pass in contextual data like the contact or task ID associated with the form - this is so it can be pasdsed into the custom components
  * Adds an example component. This will eventually be the ReferralList, the first 'real' custom component, but is just a stub for now
*  Deduplicates missing definition version IDs before requesting the JSON from s3. Before there was a case if the case list loaded up several items with the same missing definition, duplicate requests for the same form definitions could be sent - not a huge issue but it's an easy optimisation to do.

Unit tests in #1272 

### Checklist
- [ ] New tests added
- N/A Feature flags added (covered under one big Resources flag
- [X] Strings are localized

### Related Issues
Supports CHI-1421

### Verification steps
* Check the placeholder is present on Aselo Dev new case information tab forms but not for editing existing contacts
* Try adding this component to other contact forms, check it works
* Try adding it to the case editing forms, it should render an error because there is no contact / task ID in the context


plugin-hrm-form/src/components/forms/inputGenerator.tsx
```diff
@@ -65,6 +71,53 @@ export const createInput = ({
...
+          </div>
+        );
+      } catch (err) {
+        return (
+          <div>
+            Error rendering custom contact form component &lsquo;{formItemDefinition.component}&rsquo;: {err.message}
+          </div>
+        );
+      }
+    }
```
GPaoloni, line 120: Does it makes sense to abstract this conditions/branching in a component, to keep `createInput` as simple as possible?

plugin-hrm-form/src/components/tabbedForms/TabbedForms.tsx
```diff
@@ -49,6 +49,8 @@ import CSAMAttachments from './CSAMAttachments';
 import { forTask } from '../../states/contacts/issueCategorizationStateApi';
 import { newCSAMReportAction } from '../../states/csam-report/actions';
 import { CSAMReportTypes } from '../../states/csam-report/types';
+// Ensure ww import any custom components that might be used in a form
+import '../contact/ReferralList';
```
GPaoloni, line 53: Why is this needed? :thinking:

plugin-hrm-form/src/utils/definitionVersions.ts
```diff
@@ -23,12 +23,12 @@ import { getDefinitionVersions } from '../hrmConfig';
 // eslint-disable-next-line import/no-unused-modules
 const getMissingDefinitionVersions = async (versions: DefinitionVersionId[]) => {
   const { definitionVersions } = getDefinitionVersions();
-  const missingDefinitionVersions: DefinitionVersionId[] = versions.filter(
-    v => Object.values(DefinitionVersionId).includes(v) && !definitionVersions[v],
+  const missingDefinitionVersions = new Set(
+    versions.filter(v => Object.values(DefinitionVersionId).includes(v) && !definitionVersions[v]),
```
GPaoloni, line 27: :rocket:

GPaoloni (approved):
Looks good to me!

(Starting a non-blocker thread) I get the idea of what you are doing here, but why is it better to have the "custom contact components" abstracted one level on top of the form generators, rather than just letting them be top-level custom inputs? Wont it work just the same if `referral-list` is a possible type?

stephenhand:
> Looks good to me!
> 
> (Starting a non-blocker thread) I get the idea of what you are doing here, but why is it better to have the "custom contact components" abstracted one level on top of the form generators, rather than just letting them be top-level custom inputs? Wont it work just the same if `referral-list` is a possible type?

I considered that option, and it could still be a valid way to go. 

My thinking was that any custom component that deals with a contact can be treated generically by the form generation framework, rather than each component having to be called out individually in the code. Basically the form generator doesn't care what the component is, it just knows it has to pass it in a contact ID or a task ID, and any 'custom-contact-component' is the same. It's the same for people writing form definitions, they know that any form item of this type is a custom component and it doesn't add data to the form it's placed in, it does the same thing wherever it is placed.

I'm also trying to avoid a proliferation of 'single use' form input types that make the system harder to work with in general by introducing something more generic. I know we could wait until the second 'custom component' comes along to do that, but if the pattern is already in place it's more likely people will follow it

plugin-hrm-form/src/components/forms/inputGenerator.tsx
stephenhand: I moved the generator logic out to it's own function. Didn't really seem to be worth making a full fledged component but the code is separate now

PR #1268, opened 2023-02-16
Fix: prevent bell ring for offline contact reservations

GPaoloni:
## Description
This PR fixes the bug where a ring bell will fire for offline contact generated tasks (reservations actually). 

### Verification steps
Save a few offline contacts and confirm the ring is no more :)


mythilytm (approved):
Looks good!

mythilytm:
I was also going to case the task as ITask type in `trySubscribeAudioAlerts`

GPaoloni:
> I was also going to case the task as ITask type in trySubscribeAudioAlerts

I don't think this is needed, as it's already wrapped in a `TaskHelper.isChatTask` conditional.

PR #1285, opened 2023-02-25
Update `fetchApi` to handle plaintext responses

mythilytm:

## Description
- This PR ensures fall back to parsing the response as plaintext to ensure `fetchApi` handles plaintext response

### Checklist
- [n/a] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Related Issues
Fixes # [1719](https://tech-matters.atlassian.net/browse/CHI-1719)

### Verification steps
1. Receive a contact and fill out all required fields
2. Click on `Add Contact to New Case`
3. Click on `Cancel Contact and Return to Case`
4. Ensure you can see 'Case' view

### DIFF
--- a/plugin-hrm-form/src/services/fetchApi.ts
+++ b/plugin-hrm-form/src/services/fetchApi.ts
@@ -75,10 +75,7 @@ export const fetchApi = async (baseUrl: URL, endpointPath: string, options: Requ
     }
     throw new ApiError(`Error response: ${response.status} (${response.statusText})`, { response, body });
   }
-  let result;
-  try {
-    result = (await response.json()) as Promise<any>;
-  } finally {
-    return result;
-  }
+
+  const contentType = response.headers?.get('Content-Type');
+  return contentType && contentType.includes('json') ? response.json() : response.text();
 };
--- a/plugin-hrm-form/src/services/fetchApi.ts
+++ b/plugin-hrm-form/src/services/fetchApi.ts
@@ -76,6 +76,8 @@ export const fetchApi = async (baseUrl: URL, endpointPath: string, options: Requ
     throw new ApiError(`Error response: ${response.status} (${response.statusText})`, { response, body });
   }
 
-  const contentType = response.headers?.get('Content-Type');
-  return contentType && contentType.includes('json') ? response.json() : response.text();
+  if ((response.headers?.get('Content-Type') ?? '').toLowerCase().includes('json')) {
+    return response.json();
+  }
+  return response.text();
 };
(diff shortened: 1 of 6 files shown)


plugin-hrm-form/src/services/fetchApi.ts
```diff
@@ -75,6 +75,10 @@ export const fetchApi = async (baseUrl: URL, endpointPath: string, options: Requ
     }
     throw new ApiError(`Error response: ${response.status} (${response.statusText})`, { response, body });
   }
-
-  return response.json() as Promise<any>;
+  let result;
+  try {
+    result = (await response.json()) as Promise<any>;
+  } finally {
+    return result;
```
stephenhand, line 82: We should never put a `return` statement in a `finally` block. Most languages that support a `try ... catch ... finally ...`  construct consider it illegal syntax. JS doesn't for some reason, but it doesn't make it a good idea. It means it's possible to write code that returns in the `try` block and the `finally` bloc, meaning your function has returned twice??

`finally` blocks should only be used for cleanup code that we always want to run after we either return or throw an exception, like unsubscribing from an event, closing a connection etc. The `return` should always be in the `try` and / or the `catch` block

plugin-hrm-form/src/services/fetchApi.ts
```diff
@@ -75,6 +75,10 @@ export const fetchApi = async (baseUrl: URL, endpointPath: string, options: Requ
     }
     throw new ApiError(`Error response: ${response.status} (${response.statusText})`, { response, body });
   }
-
-  return response.json() as Promise<any>;
+  let result;
```
stephenhand, line 78: This approach means that we parse all responses as json, and if it fails for any reason, it swallows the exception and returns `undefined`? I don't think this is the best strategy at such a low level in our API. It will hide issues with malformed JSON where the response is supposed to parsable as JSON, making it throw a harder to diagnose error elsewhere. It will also clobber any plaintext responses where we actually care about the content.

If a response isn't shouldn't be treated as JSON, it should be reflected in it's `Content-Type` header, and having experimented, the `DELETE` endpoint does indeed return `text/plain` as a content type. So instead of swallowing errors I would suggest something like this:

```suggestion
  if ((response.headers.get('Content-Type') ?? '').toLowerCase().includes('json')) {
    return response.json();
  }
  return response.text();
```

So we are a bit smarter than previously about treating JSON as JSON and text as text - this way we till get parser errors if something is _supposed_ to be JSON but isn't, and we get the content of plain text.

(I did a brief test of the suggested change above and it seems to work btw)

stephenhand (changes requested):
So there's one endpoint in our API surface that returns text instead of JSON? 🤦 Sorry for missing that & thanks for taking this on @mythilytm .

Whilst your change addresses the issue, I suggested an alternative approach that I think might set us up better in the longer term if you take a look at the comments

plugin-hrm-form/src/services/fetchApi.ts
mythilytm: Thank you!

stephenhand (approved):
Thanks Mythily!

PR #1291, opened 2023-02-28
Alejandro in update insights mapping

janorivera:
Primary reviewer:
@Humairaa127  just FYI I'm remapping some the fields that are being saved into Insights and I'm also adding some new fields in the caseInformation tab 
This data is the data that the helpline is currently saving manually in a spreadsheet.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### DIFF
--- a/hrm-form-definitions/form-definitions/in/v1/LayoutDefinitions.json
+++ b/hrm-form-definitions/form-definitions/in/v1/LayoutDefinitions.json
@@ -5,6 +5,7 @@
     "childInformation": {
     },
     "caseInformation": {
+        "splitFormAt": 6
     }
   },
   "case": {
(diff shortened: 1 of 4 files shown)


Humairaa127 (commented):
Question: Why have you removed the state>district mapping from oneToMany to oneToOne?

janorivera:
@Humairaa127  ah yes, because the way how data appears is "state; district", Uma and Sid want these in separated columns. To separate those 2 fields in columns I'd need to create a metric for each district and each state which is doable but it seems that is not worth the effort if we can just map those 2 in 2 different fields.

PR #1295, opened 2023-03-08
View Contact/Case Permissions

murilovmachado:
## Description
This PR adds the following permissions:
- `viewContact`
- `viewCase`

The counselor should never be able to view a Contact or Case if he/she doesn't have permission for it.

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-1642
Backend PR: https://github.com/techmatters/hrm/pull/316

### Verification steps
If the backend PR is not merged:
- Set `viewContact` permission to `[["isSupervisor"], ["isOwner"]]`
- Search for contacts
- Check that you cannot view details of contacts you don't own
- Check that you can view details of contacts you own
- Repeat the same for Cases

If the backend PR is merged:
- Set `viewContact` permission to `[["isSupervisor"], ["isOwner"]]`
- Search for contacts
- All the contacts displayed are the ones you own
- Check that you can view their details
- Repeat the same for Cases


plugin-hrm-form/src/permissions/search-permissions.ts
```diff
@@ -0,0 +1,59 @@
+/**
```
stephenhand, line 1: This logic might be a good candidate for some unit tests (in this PR or a follow up)

plugin-hrm-form/src/components/search/SearchResults/index.tsx
```diff
@@ -241,13 +242,16 @@ const SearchResults: React.FC<Props> = ({
...
               {contacts &&
                 contacts.length > 0 &&
-                contacts.map(contact => (
-                  <ContactPreview
-                    key={contact.contactId}
-                    contact={contact}
-                    handleViewDetails={() => handleViewDetails(contact)}
-                  />
-                ))}
+                contacts.map(contact => {
```
stephenhand, line 245: Is this change required if the results are being filtered on the backend? (same q for the case preview & the cast list row)

stephenhand (approved):
Couple of qs but nothing blocking this PR

plugin-hrm-form/src/components/search/SearchResults/index.tsx
murilovmachado: No. This scenario only happens if the backend is not in place:
- In case the backend is not merged
- or if there're bugs on the backend

stephenhand: Ah I see, it's compatibility code, we can probably remove it once HRM is updated?

PR #1296, opened 2023-03-09
fix download error for non-ascii encodings

acedeywin:

## Description

- Bug fix: fix download error when downloading attachment with non-ascii encodings file name
### Checklist
- [N/A] New tests added
- [N/A] Feature flags added
- [N/A] Strings are localized

### Related Issues
Fixes C˝HI-1713

### Verification steps

- Navigate to Cases
- Select a Case that was created by you
- Scroll down to Document
- Add and upload a file  (files that are named both in English and any other language - Thai, for example)
- Go back to Case and click to view a downloaded file
- Download file

### DIFF
--- a/plugin-hrm-form/src/components/common/forms/DownloadFile.tsx
+++ b/plugin-hrm-form/src/components/common/forms/DownloadFile.tsx
@@ -34,6 +34,7 @@ const DownloadFile: React.FC<Props> = ({ fileNameAtAws }) => {
   useEffect(() => {
     if (preSignedUrl) {
       downloadLink.current.click();
+      URL.revokeObjectURL(preSignedUrl);
     }
   }, [preSignedUrl, downloadLink]);
 
--- a/plugin-hrm-form/src/components/common/forms/DownloadFile.tsx
+++ b/plugin-hrm-form/src/components/common/forms/DownloadFile.tsx
@@ -15,7 +15,7 @@
  */
 
 /* eslint-disable react/prop-types */
-import React, { useState, useRef, useEffect } from 'react';
+import React from 'react';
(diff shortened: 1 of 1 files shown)


stephenhand:
@acedeywin - I think I can see what you are doing from the code but can you write up an explanation of specifically how your are changing the download flow and why in the PR description

I'm interested to know why we need to move to creating an  `a` tag directly in the DOM rather than using the hidden JSX element and an effect like we were previously. 

On the face of it it seems like the previous approach would still work with the object URL - was there a specific issue you came up against trying to do that? I just think we need to be wary on direct DOM manipulation in React apps, I know we delete the element again right away but I'd still be worried about the potential for unexpected side effects

acedeywin:
> @acedeywin - I think I can see what you are doing from the code but can you write up an explanation of specifically how your are changing the download flow and why in the PR description
> 
> I'm interested to know why we need to move to creating an `a` tag directly in the DOM rather than using the hidden JSX element and an effect like we were previously.
> 
> On the face of it it seems like the previous approach would still work with the object URL - was there a specific issue you came up against trying to do that? I just think we need to be wary on direct DOM manipulation in React apps, I know we delete the element again right away but I'd still be worried about the potential for unexpected side effects

@stephenhand Yes, I could implement it with what we have. I was having issues trouble some unknown error so I had to refactor to how I could implement it. But I have fixed it now and refactored it to what we previously had.

stephenhand (commented):
Looks good - only thing is we should probably still call `URL.revokeObjectURL(url);` like we did in your first pass, to prevent us leaking memory on the browser.

If you put that line in after the click invoke in the `useEffect` function it should work I think?

stephenhand (approved):
LGTM

PR #1297, opened 2023-03-09
Terraform linting and security tools

robert-bo-davis:
## Description

This adds some common terraform linting and security tools to the makefile configuration. They generate a *lot* of output right now and it will be a massive change set to fix everything they find. I'm adding this for team insight. We should probably plan to slowly introduce each of these tools into our CI/CD process and do the code cleanup at some point during future lulls in active TF development.

The suggested order of rollout:

1. fmt-check/fmt-fix since this will fix issues automatically.
2. tflint will require manual changes
3. tfsec will require manual changes.

### Verification steps

From any terraform helpline directory run the following (these run global checks even though you are in a specific start directory):

```
make fmt-check
make tflint
make tfsec
```

### DIFF
--- a/twilio-iac/terraform-poc-account/main.tf
+++ b/twilio-iac/terraform-poc-account/main.tf
@@ -87,7 +87,6 @@ module flex {
   operating_info_key = var.operating_info_key
   permission_config = "zm"
   definition_version = var.definition_version
-  serverless_url = module.serverless.serverless_environment_production_url
   hrm_url = "https://hrm-development-eu.tl.techmatters.org"
   multi_office_support = var.multi_office
   feature_flags = var.feature_flags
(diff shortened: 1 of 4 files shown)


twilio-iac/terraform-poc-account/main.tf
```diff
@@ -87,7 +87,6 @@ module flex {
   operating_info_key = var.operating_info_key
   permission_config = "zm"
   definition_version = var.definition_version
-  serverless_url = module.serverless.serverless_environment_production_url
```
stephenhand, line 90: Intended change for this PR?

stephenhand (approved):
Not tested these but tested the same changes in infrastructure-config & look good there

twilio-iac/terraform-poc-account/main.tf
robert-bo-davis: Yes. serverless_url is duplicated in that block. Has been showing red in my IDE forever, but causes full failure that shortcuts the rest of the checks for all of the above tools, so I went ahead and fixed it.

PR #1302, opened 2023-03-14
Handles EmojiPicker initial state on narrow screens

murilovmachado:
## Description
The EmojiPicker was previously resized when the panel was resized (listener).
But this didn't cover the scenario where the initial state the screen width was small: it needed a resize event until the EmojiPicker width was recalculated.

This PR does 2 things to fix that and make it more robust:
- Triggers the EmojiPicker resize function the first time it's opened
- Also listens to `window.resize` event (not only Twilio's one)

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-1770

### Verification steps
1) Make the screen very narrow
2) Accept a chat task
3) Open EmojiPicker and check it's not cut/hidden (without resizing the panel or the screen)

### DIFF
--- a/plugin-hrm-form/src/components/emojiPicker/index.tsx
+++ b/plugin-hrm-form/src/components/emojiPicker/index.tsx
@@ -52,6 +52,7 @@ const EmojiPicker: React.FC<TaskContextProps & ConnectedProps<typeof connector>>
   const [inputText, setInputText] = useState('');
   const [isOpen, setIsOpen] = useState(false);
   const [emojisPerLine, setEmojisPerLine] = useState(EMOJIS_PER_LINE_DEFAULT);
+  const [firstTimeOpen, setFirstTimeOpen] = useState(false); // Tracks if the EmojiPicker was opened for the first time
 
   const { blockedEmojis } = definitionVersion ?? { blockedEmojis: [] };
   const conversationSid = getConversationSid(task);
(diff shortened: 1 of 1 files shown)


GPaoloni (approved):
The change works good! 

Now that I'm testing this on my laptop (14" screen), I've noticed that the emoji picker has a lesser z index or an overflow or something, that is hiding the topmost chunk of it (the controls and search).

Screenshot from 2023-03-24 16-15-35

murilovmachado:
> Now that I'm testing this on my laptop (14" screen), I've noticed that the emoji picker has a lesser z index or an overflow or something, that is hiding the topmost chunk of it (the controls and search).

I've tried playing with this issue, but there's no easy CSS trick for it. I guess Twilio on purpose doesn't want anything to cover the Chat/Info tabs or the End Chat button. I'll merge it as it is, and we can come back to this if any user find this could be affecting them.

murilovmachado:
> Now that I'm testing this on my laptop (14" screen), I've noticed that the emoji picker has a lesser z index or an overflow or something, that is hiding the topmost chunk of it (the controls and search).

I've tried playing with this issue, but there's no easy CSS trick for it. I guess Twilio on purpose doesn't want anything to cover the Chat/Info tabs or the End Chat button. I'll merge it as it is, and we can come back to this if any user finds this could be affecting them.

PR #1303, opened 2023-03-15
Sinekhaya zm formupdates

sinehome:
Primary reviewer: Dee Luo

## Description

Updates to Childline Zimbabwe forms
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### DIFF
--- a/hrm-form-definitions/form-definitions/zw/v1/caseForms/HouseholdForm.json
+++ b/hrm-form-definitions/form-definitions/zw/v1/caseForms/HouseholdForm.json
@@ -201,12 +201,12 @@
         "label": ""
       },
       {
-        "value": "Boy",
-        "label": "Boy"
+        "value": "Male",
+        "label": "Male"
       },
       {
-        "value": "Girl",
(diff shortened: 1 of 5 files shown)


dee-luo (changes requested):
@sinehome A few notes!

- Double checking - I don't remember the request to remove "Region" or "Address Description" on ChildInformation? 
- Can we add leading 0's to the age values in the CallerInformation tab? For the single digit ages

Everything else LGTM!

sinehome:
@dee-luo I remember having a conversation about region not being relevant to their context (it had our default values - we can confirm maybe?). The address description is duplicated.

dee-luo (approved):
Okay got it! Yes I think on Region probably something to confirm - I thought I remembered them liking it but could be remembering incorrectly. 

For the leading 0's - sorry should've clarified, we only need them on the "values" to save to the backend. We can keep the labels as is with the single digits. 

Feel free to merge and deploy once you're done!

PR #1307, opened 2023-03-22
Hangup call before dropping rejected transfer

GPaoloni:
This PR is related to https://github.com/techmatters/serverless/pull/426

## Description
This PR fixes the bug where warm transfers are unable to reject transfer (see ticket below). This issue does not occurs when the transfer-target worker rejects the initial reservation, but when joins the call and then hits the "Reject Transfer" button. The fix is pretty simple: before completing the task that links the conference, invoke a "Hangup" action.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1729)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
Follow the reproduction steps in the ticket and confirm that the behavior is now the expected one.

### DIFF
--- a/plugin-hrm-form/src/utils/transfer.ts
+++ b/plugin-hrm-form/src/utils/transfer.ts
@@ -15,7 +15,7 @@
  */
 
 // eslint-disable-next-line no-unused-vars
-import { Actions, ITask, TaskHelper, StateHelper } from '@twilio/flex-ui';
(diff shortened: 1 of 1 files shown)


murilovmachado (approved):
🚀

PR #1309, opened 2023-03-22
Sinekhaya mt tf updates

sinehome:
Primary reviewer:
Alejandro
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,12 @@
 {
-  "name": "flex-plugins",
-  "lockfileVersion": 2,
+  "name": "@tech-matters/flex-plugins",
   "requires": true,
-  "packages": {}
+  "lockfileVersion": 1,
+  "dependencies": {
+    "typescript": {
+      "version": "4.9.5",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.9.5.tgz",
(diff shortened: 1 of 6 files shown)


package-lock.json
janorivera, line 1: This file should not be changed.

package.json
janorivera, line 1: This file should not be changed.

twilio-iac/terraform-modules/channels/flow-templates/language-mt/with-chatbot.tftpl
janorivera, line 1: Have you deployed the studio flow already? so I can take a look? 
One thing to have in mind, is that if we are capturing the data like gender, age, etc, using the webchat form, we shouldn't redirect the webchat flow to the chatbot that is going to ask for the same info, so that path will need to change (but just for the webchat flow).

janorivera (changes requested):

GPaoloni (commented):
@sinehome The changes on the prodution bot look good to me! Consider this to be an approved for that, but I'll defer to @janorivera for the approval of the PR since he had some comments, it will be good for him to give another pass.

PR #1311, opened 2023-03-24
CHI-1602: Migrate resources UI to call CloudSearch endpoint

stephenhand:
Primary reviewer:

## Description

* Migrates the flex resources client to use the payload expected by the new CloudSearch powered resource search endpoint
* Renames internal variables that include 'omnisearch' to use the term 'general search' instead 

### Checklist
- N/A New tests added
- N/A (all behind an overall resource feature flag prior to v1) Feature flags added
- N/A Strings are localized

### Related Issues
CHI-1602

### Verification steps

Regression test search endpoint (it should be able to search on resource information as well as name now)


mythilytm (approved):
This looks great. Thanks Steve.

PR #1312, opened 2023-03-24
CHI-1569 refactor makefile system for terragrunt support

robert-bo-davis:
Primary reviewer:

## Description

These are the basic makefile system changes to support both terragrunt stages and the older terraform hl directories simultaneously.


twilio-iac/aarambh-production/Makefile
```diff
@@ -1,3 +1,2 @@
-TF_VER = 1.3.6
-
 include ../makefiles/*.make
+include ../makefiles/terraform/*.make
```
robert-bo-davis, line 2: You'll see these additional includes throughout the PR. There are now two "sub-configurations" for the make system. One for the new system that uses terragrunt as the default and one that uses terraform. This include is the "switch" that tells the system which underlying software to use for bare commands like make plan.

twilio-iac/aarambh-production/Makefile
```diff
@@ -1,3 +1,2 @@
-TF_VER = 1.3.6
```
robert-bo-davis, line 1: The individual TF_VER is no longer needed unless you want to override the makefile default.

stephenhand (approved):
LGTM

twilio-iac/makefiles/00-setup.make
```diff
@@ -1,25 +1,24 @@
...
-    #TODO: fix this to navigate up two dirs instead of one
-    MY_PWD := $(shell powershell Split-Path -Path (Get-Location) -Parent)
     MY_ENV := $(shell powershell Split-Path -Path (Get-Location) -Leaf)
     DIND_ARG = -v "//var/run/docker.sock:/var/run/docker.sock"
 else
     # We're NOT running windows, assume bash is available
-    SECRETS = -e AWS_DEFAULT_REGION -e AWS_SECRET_ACCESS_KEY -e AWS_ACCESS_KEY_ID -e TWILIO_ACCOUNT_SID -e TWILIO_AUTH_TOKEN -v ~/.aws:/root/.aws -e GITHUB_OWNER -e GITHUB_TOKEN
-    MY_PWD := $(shell cd ../../; pwd)
     MY_ENV := $(shell basename $(CURDIR))
+    MY_RELATIVE_PATH := $(shell echo $(CURDIR) | sed -e "s|$(MY_PWD)||")
```
GPaoloni, line 15: Why is this needed vs the github function for the full path?

GPaoloni (approved):

twilio-iac/makefiles/00-setup.make
robert-bo-davis: Because the workdir can be at any level below the git root.

We could be runninng `make plan` from a regular helpline so the path we are running from would be: `/some/dir/on/host/github_root/terraform/as-production` and  the path we want to work from on the container would be `/app/terraform/as-production`. Or we could be calling it from a specific terragrunt stage which would be `/app/terraform/stages/provision`. Or we could be calling a `make plan-all HL=as HL_ENV=production` from the stages directory and it would b e`/app/terraform/stages`. We must have the full relative path to the current directory from the git root so that we can correctly set the workdir within the container.

PR #1313, opened 2023-03-24
CHI-1569: terragrunt stages

robert-bo-davis:
Primary reviewer:

## Description

This is the meat of the configuration driven terragrunt stages refactor.


twilio-iac/stages/Makefile
```diff
@@ -0,0 +1,20 @@
+MY_PWD := $(shell git rev-parse --show-toplevel)
+
+include $(MY_PWD)/twilio-iac/makefiles/*.make
+include $(MY_PWD)/twilio-iac/makefiles/terragrunt/*.make
+
+init_args := HL=$(HL) HL_ENV=$(HL_ENV)
+
+init-all: init-provision init-chatbot init-configure
```
robert-bo-davis, line 8: These `-all` commands won't work correctly right now because the dependency blocks have all been commented out in the various terragrunt.hcl files because they make state migration using local state more difficult than they are worth in the short term.

In the long term, these will allow us to run commands like make HL=as HL_ENV=staging plan-all from the stages directory and terragrunt will run plan in all of the stages in the correct order.

twilio-iac/stages/provision/terragrunt.hcl
```diff
@@ -0,0 +1,42 @@
...
+  * This sets the variables that are fed into the stage module to be the combined config at local.config
+  */
+inputs = local.config
+
+/**
+  * This is the main terragrunt block that defines the stage module and the hooks that run before it.
+  */
+terraform {
+  // TODO: remove this once we've migrated all secrets
+  before_hook "migrate_tf_secrets" {
```
robert-bo-davis, line 30: `before_hook` and `after_hook` allows execution of shell commands before and/or after the terraform command for the terraform commands specified in `commands`. This is similar to things we can do with the makefile system, but a bit more robust.

stephenhand (approved):
Think I understand it, at least a little after the walkthrough of the simpler infrastructure_config terragrunt setup

twilio-iac/terraform-modules/stages/configure/main.tf
```diff
@@ -0,0 +1,143 @@
...
+  source               = "../../flex/service-configuration"
+  twilio_account_sid   = local.secrets.twilio_account_sid
+  short_environment    = var.short_environment
+  operating_info_key   = var.operating_info_key
+  permission_config    = local.permission_config
+  definition_version   = var.definition_version
+  serverless_url       = local.serverless_url
+  multi_office_support = var.multi_office
+  feature_flags        = var.feature_flags
+  hrm_url              = "https://hrm-staging-eu.tl.techmatters.org"
```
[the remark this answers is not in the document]
robert-bo-davis: Good catch. This is caught and fixed [further down the PR chain](https://github.com/techmatters/flex-plugins/blob/CHI-1569-terraform-refactor-update-scripts/twilio-iac/terraform-modules/stages/configure/main.tf#L21).

GPaoloni (approved):
Haven't done a per-line-review, but I get the idea and LGTM! Thanks! :rocket:

PR #1316, opened 2023-03-24
CHI-1569 state migration scripts

robert-bo-davis:
Primary reviewer:

## Description

This PR includes the simple bash based state migration system that will exist temporarily while we migrate helplines to the new system.


twilio-iac/scripts/migration/migrateTFSecrets.sh
```diff
@@ -0,0 +1,44 @@
...
+    --value "$(aws ssm get-parameter --name $old_key --with-decryption --query Parameter.Value --output text)" \
+    --tags Key=Environment,Value=${environment} Key=Helpline,Value=${short_helpline}
+  printf "Done!\n\n"
+fi
+
+if [ $old_key_exists -eq 0 ]; then
+  read -p "The old key (${old_key}) still exists. Do you want to delete it?(y/N)" -n 1 -r
+  if [[ $REPLY =~ ^[Yy]$ ]]; then
+    echo "Deleting old secrets from ${old_key}"
+    # aws ssm delete-parameter --name $old_key
```
robert-bo-davis, line 40: This is commented out for now because I want to avoid anything that could be destructive to the old TF config until we are definitely ready to be destructive. 😉

stephenhand (approved):
Are these scripts intended to be run directly, or do they get used in the make terraform system (or some other docker container)? 

If they need to be run directly that could be problematic on Windows - although looking at them the scripts don't appear to use docker themselves, so running them in WSL or in a container should be ok.

robert-bo-davis:
> Are these scripts intended to be run directly, or do they get used in the make terraform system (or some other docker container)?

They are definitely meant to be run within the TG system. Currently the state migration is run via `make migrate-state` and the secrets manager is run as a [before_hook in the provision step](https://github.com/techmatters/flex-plugins/blob/CHI-1569-terraform-refactor-update-scripts/twilio-iac/stages/provision/terragrunt.hcl#L30).

PR #1317, opened 2023-03-24
CHI-1569 catchall for the rest of the giant PR split

robert-bo-davis:
Primary reviewer:

## Description

This is a catchall for the changes I missed in the other PRs. The branch related to this PR should be testable for all commands except apply in the new system.

### DIFF
--- a/scripts/package.json
+++ b/scripts/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "preinstall": "cd ../hrm-form-definitions && npm run build",
+    "preinstall": "cd ../hrm-form-definitions && npm ci && npm run build",
     "lint": "eslint --ext js .",
     "lint:fix": "npm run lint -- --fix",
     "copyFlowLegacy": "node copyFlow.js",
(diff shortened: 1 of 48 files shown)


scripts/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "preinstall": "cd ../hrm-form-definitions && npm run build",
+    "preinstall": "cd ../hrm-form-definitions && npm ci && npm run build",
```
robert-bo-davis, line 6: twilio-iac only runs from a fresh checkout were failing because this was missing.

twilio-iac/terraform-modules/aws/default/variables.tf
```diff
@@ -88,3 +83,9 @@ variable "serverless_url" {
   description = "URL used to access Aselo Twilio serverless functions"
   type        = string
 }
+
+variable "post_survey_bot_sid" {
+  description = "Internal Twilio resource SID provided by another module"
+  type        = string
+  default     = ""
```
robert-bo-davis, line 90: This param is now optional so that this module will work with both the old and new terraform systems.

twilio-iac/terraform-modules/github/default/main.tf
```diff
@@ -1,7 +1,9 @@
-
-
-provider "github" {
```
robert-bo-davis, line 3: moved github provider out of module so that it could be used with a "count" option.

GPaoloni (approved):

PR #1320, opened 2023-03-27
Add some missing configurations in terraform- emoji picker flag and assets bucket url

mythilytm:
Primary reviewer: @

## Description
- There were some older service config state that was run (`terraform apply`) for an account. This PR adds emoji_picker and assets bucket url 

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


twilio-iac/aselo-as-production/main.tf
```diff
@@ -171,6 +172,7 @@ module aws {
   flex_proxy_service_sid = module.services.flex_proxy_service_sid
   post_survey_bot_sid = module.chatbots.post_survey_bot_sid
   survey_workflow_sid = module.survey.survey_workflow_sid
+  assets_bucket_url = module.assets.assets_bucket_url
```
mythilytm, line 175: What is the path for assets bucket?

robert-bo-davis: Adding here as well as slack for future reference

```
assets_bucket_url = "https://assets/-${lower(var.environment)}.[tl.techmatters.org](http://tl.techmatters.org/)"
```

or 

```
assets_bucket_url = "https://assets/-${lower(local.environment)}.[tl.techmatters.org](http://tl.techmatters.org/)"
```

depending on how the helpline is configured.

twilio-iac/aarambh-production/main.tf
```diff
@@ -121,5 +121,5 @@ module aws {
   flex_proxy_service_sid = module.services.flex_proxy_service_sid
   post_survey_bot_sid = module.chatbots.post_survey_bot_sid
   survey_workflow_sid = module.survey.survey_workflow_sid
-
+  assets_bucket_url = "https://assets-${lower(var.environment)}.tl.techmatters.org"
```
robert-bo-davis, line 124: Instead of passing this in as a variable, it can be generated in a local in the aws module [here](https://github.com/techmatters/flex-plugins/blob/master/twilio-iac/terraform-modules/aws/default/main.tf#L23).

That locals block will change to:

```
locals {
  docs_s3_location = "tl-aselo-docs-${lower(var.short_helpline)}-${lower(var.environment)}"
  chat_s3_location = "tl-public-chat-${lower(var.short_helpline)}-${lower(var.short_environment)}"
  assets_bucket_url = "https://assets-${lower(var.environment)}.tl.techmatters.org"
}
```

And can then be referenced with `local.assets_bucket_url`.

I don't see any modification to the aws module to actually do anything with that value (or accept the variable), yet.

Since this is a new variable, it only needs to be added to the "new style" of SSM param like in the [ssm.tf file](https://github.com/techmatters/flex-plugins/blob/master/twilio-iac/terraform-modules/aws/default/ssm.tf).

Let me know if you need anymore pointers!

robert-bo-davis (changes requested):

PR #1325, opened 2023-03-30
Bug fix: Welcome Message not getting triggered in Dev and Kek Vonal Staging

acedeywin:
Primary reviewer:

## Description

- Bug fix: Welcome Message not getting triggered in Dev and Kek Vonal Staging

### Checklist
- [N/A] New tests added
- [N/A] Feature flags added
- [N/A] Strings are localized

### Related Issues
Fixes CHI-1768


stephenhand (commented):
Can you explain what you changed in this PR - the diffs are not very clean and it's very difficult to see what you changed inside the try catch blocks

acedeywin:
> Can you explain what you changed in this PR - the diffs are not very clean, and it's very difficult to see what you changed inside the try-catch blocks

The welcome message not getting triggered because the `getUnknownOption` function was breaking, and the error was not correctly handled. So the `try-catch` block was added to handle the error graciously and not break the entire code.

The main error was that `getUnknownOption` couldn't determine a valid unknown option for key. In this case, `unknownOption` didn't have the right value or `e.value === 'Unknown'` does not exist within the `json` files for Kek Vonal (Line 31 in prepopulateForm.ts). So I did a bit of a refactor on Line 31 in prepopulateForm.ts and added values for `unknownOption` in some of the `json` files.

plugin-hrm-form/src/utils/setUpActions.ts
```diff
@@ -206,7 +206,7 @@ export const afterAcceptTask = (featureFlags: FeatureFlags, setupObject: SetupOb
   else prepopulateForm(task);
 
   // If this is the first counsellor that gets the task, say hi
-  if (TaskHelper.isChatBasedTask(task) && !TransferHelpers.hasTransferStarted(task)) {
+  if (TaskHelper.isChatBasedTask(task)) {
```
stephenhand, line 209: What is this change for? Is it part of you PR or a change from elsewhere?

stephenhand (changes requested):
Please revert all the `try .. catch ... log` blocks. Swallowing errors is poor practice and should only be used in exceptional circumstances.

We should have proper error reporting for when prepopulation fails, i.e. catch the error at the top level and display an error to the user that there was a problem prepopulating the form, but that is out of scope for this ticket - for now, letting the errors throw & stopping the welcome message is better than swallowing them & hiding issues

plugin-hrm-form/src/utils/setUpActions.ts
acedeywin: @stephenhand, Ooh, I removed it while troubleshooting. Thank you for spotting it. I will add it back.

acedeywin:
> Please revert all the `try .. catch ... log` blocks. Swallowing errors is poor practice and should only be used in exceptional circumstances.
> 
> We should have proper error reporting for when prepopulation fails, i.e. catch the error at the top level and display an error to the user that there was a problem prepopulating the form, but that is out of scope for this ticket - for now, letting the errors throw & stopping the welcome message is better than swallowing them & hiding issues

@stephenhand, I have fixed another bug in the `getAnswerOrUnknown` function. In the Kek Vonal case, the `key` (gender and age) is undefined since users don't answer any preEngagement question that involves it. So since `key` is undefined, the `isUnknownAnswer` variable throws an error because `isUnknownAnswer` only checks for two conditions - `answers[key].error` or `answers[key].answer === unknown`. So I added a third condition, `answers[key] === undefined`, which fixed the issue.

plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -72,7 +74,7 @@ const getAnswerOrUnknown = (
 
   if (itemDefinition.type === 'select') {
     const unknown = getUnknownOption(key, definition);
-    const isUnknownAnswer = answers[key].error || answers[key].answer === unknown;
+    const isUnknownAnswer = answers[key] === undefined || answers[key].error || answers[key].answer === unknown;
```
stephenhand, line 77:
```suggestion
    const isUnknownAnswer = !answers[key] || answers[key].error || answers[key].answer === unknown;
```

plugin-hrm-form/src/utils/setUpActions.ts
stephenhand: You might have added back to the wrong if statement?

plugin-hrm-form/src/utils/setUpActions.ts
```diff
@@ -198,15 +198,15 @@ export const afterAcceptTask = (featureFlags: FeatureFlags, setupObject: SetupOb
 ) => {
   const { task } = payload;
 
-  if (TaskHelper.isChatBasedTask(task)) {
+  if (TaskHelper.isChatBasedTask(task) && !TransferHelpers.hasTransferStarted(task)) {
```
stephenhand, line 201: See below

plugin-hrm-form/src/utils/mappers.ts
```diff
@@ -72,23 +72,28 @@ export const mapChannelForInsights = (channel: string) => {
 };
 
 export const mapAge = (ageOptions: string[]) => (age: string) => {
-  const ageInt = parseInt(age, 10);
+  try {
```
stephenhand, line 75: Can we remove this try ... catch ... log as well? Or is there a specific reason you are keeping it?

plugin-hrm-form/src/utils/setUpActions.ts
acedeywin: Fixed.

plugin-hrm-form/src/utils/prepopulateForm.ts
acedeywin: Done!

plugin-hrm-form/src/utils/mappers.ts
acedeywin: Done!

stephenhand (approved):
LGTM

PR #1329, opened 2023-04-04
Resource View QA revisions

mythilytm:
Primary reviewer:

## Description
- This PR fixes the Resource View and Preview components with the requested qa revisions and newer data structure changes

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes # 1727


plugin-hrm-form/src/components/resources/convertKHPResourceAttributes.ts
```diff
@@ -112,20 +126,39 @@ const extractDescriptionInfo = (description, language: Language) => {
   return descriptionByLanguage && descriptionByLanguage.info ? descriptionByLanguage.info.text : null;
 };
 
-export const convertKHPResourceAttributes = (attributes, language: Language): KhpUiResource['attributes'] => {
+const extractRequiredDocuments = (documentsRequired, language: Language) => {
+  const documents = [];
+
+  for (const key in documentsRequired) {
```
stephenhand, line 132: I generally fine the `Object.entries(...)`, `Object.keys()` & `Object.values()` functions makes logic involving iterating over keys cleaner.

For example here you implement the same function as:

```
  Object.values(documentsRequired)
    .map(document => document.find(item => item.language === language || item.language === '')?.value)
    .filter(languageDoc => languageDoc)
```

Also - do we need to fall back to English if the specified language doesn't exist, rather than omitting it?

plugin-hrm-form/src/components/resources/search/resourceView/ResourceAttribute.tsx
```diff
@@ -44,18 +44,22 @@ const ResourceAttribute: React.FC<Props> = ({ description, children, isExpandabl
       return <ExpandableAttributeContent expandLinkText="ReadMore" collapseLinkText="ReadLess" content={children} />;
     }
 
+    // For Documents Required attribute
+    if (Array.isArray(children)) {
```
stephenhand, line 48: Might it be better to convert the documents to a comma separated string in the conversion function, rather than adding a special case in the component?

plugin-hrm-form/src/components/resources/types.ts
```diff
@@ -0,0 +1,77 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+
+export type KhpUiResource = {
```
stephenhand, line 17: Nice

stephenhand (approved):
Couple of small nits & queries, looks good otherwise

plugin-hrm-form/src/components/resources/convertKHPResourceAttributes.ts
mythilytm: I am getting this error on `find` - `Property 'find' does not exist on type 'unknown'.ts(2339)` I will take a deeper dive in a latter PR

murilovmachado: @stephenhand Do you think this should make it into the coding standards repo?

PR #1331, opened 2023-04-05
MT Insights Mapping

dee-luo:
Update Insights according to this file: https://app.box.com/file/826297333105

### DIFF
--- a/hrm-form-definitions/form-definitions/mt/v1/insights/oneToManyConfigSpecs.json
+++ b/hrm-form-definitions/form-definitions/mt/v1/insights/oneToManyConfigSpecs.json
@@ -7,3 +7,4 @@
     ],
     "saveForNonDataContacts": true
   }
+]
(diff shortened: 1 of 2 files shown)


hrm-form-definitions/form-definitions/mt/v1/insights/oneToManyConfigSpecs.json
```diff
@@ -1 +1,10 @@
-[]
\ No newline at end of file
+[
+  {
+    "attributeName": "conversation_attribute_9",
```
janorivera, line 3: This is new! what does it do? is this going to save the contact ID?

janorivera (approved):
LGTM!

hrm-form-definitions/form-definitions/mt/v1/insights/oneToManyConfigSpecs.json
dee-luo: Yes it does! We already save Case ID as well in oneToOne, but I didn't know what Rachel meant by ChatID so I included this here.

PR #1335, opened 2023-04-06
CHI-1785: Experimenting with the 'nuclear option' of replacing the text

stephenhand:
This is an experiment demoing the 'nuclear option' if we can't get Twilio to optimise their native message input, which is to replace the message input control

This is a POC hack, if we were to do this, on top of this we would need to reimplement:

* Proper styling
* Attachment support
* Rejig emoji picker & canned responses (this one should be easy)
* Reimplement the auto-growing text area that is the cause of the performance issues in the Twilio box, but do it performantly


plugin-hrm-form/src/components/CannedResponses.tsx
```diff
@@ -23,11 +23,13 @@ import { MessageInputChildrenProps } from '@twilio/flex-ui-core/src/components/c
 import { selectCannedResponses } from '../states/selectors/hrmStateSelectors';
 import { CannedResponsesContainer, FormSelect, FormSelectWrapper, FormOption } from '../styles/HrmStyles';
 import { getTemplateStrings } from '../hrmConfig';
+import useTraceUpdate from '../hooks/useTraceUpdate';
 
 type Props = Partial<MessageInputChildrenProps>;
 
 const CannedResponses: React.FC<Props> = props => {
   const cannedResponses = useSelector(selectCannedResponses);
+  useTraceUpdate(props, 'CannedResponses');
```
GPaoloni, line 32: What's this one for?

GPaoloni (approved):

murilovmachado (approved):

plugin-hrm-form/src/utils/setUpComponents.tsx
```diff
@@ -357,12 +359,19 @@ export const removeActionsIfTransferring = () => {
...
+  Flex.MessageInputV2.Content.replace(<AseloMessageInput key="textarea" />, { sortOrder: -1 });
+  Flex.MessageList.Content.replace(<AseloMessageList key="list" />);
+};
+
 /**
  * Canned responses
  */
 export const setupCannedResponses = () => {
   Flex.MessageInput.Content.add(<CannedResponses key="canned-responses" />);
-  Flex.MessageInputV2.Content.add(<CannedResponses key="canned-responses" />);
```
murilovmachado, line 365: Why are we removing it from `MessageInputV2`, but leaving it at `MessageInput`?
Isn't V2 the one being used now by all accounts?

plugin-hrm-form/src/components/AseloMessageInput.tsx
```diff
@@ -0,0 +1,140 @@
...
+}) => {
+  const { register, handleSubmit, setValue, getValues } = useForm();
+
+  useEffect(() => {
+    setValue('messageInputArea', draftText);
+    return () => {
+      updateDraftMessageText(getValues('messageInputArea'));
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [conversationSid]);
```
murilovmachado, line 71: Isn't it dangerous to disable dependencies-related lint rules?

It feels VERY hacky to include `conversationSid` as a dependency, even though it's not being used by this hook, while at the same time, we're not including `draftText, getValues, setValue, updateDraftMessageText`, which are actually being used. 

I think I get why we're doing this way, but this seems like using the wrong tool/solution for a problem.

plugin-hrm-form/src/utils/setUpComponents.tsx
stephenhand: Good catch, thx - fixed now

PR #1337, opened 2023-04-07
chore: migrate pl staging and fixes

robert-bo-davis:
Primary reviewer:

## Description

These are the fixes for a fully migrated simple helpline (PL)

### Related Issues
Fixes #CHI-1637

### Verification steps

Run `make plan HL=pl HL_ENV=staging` for all three stages.


twilio-iac/scripts/migration/state/stage-common.sh
```diff
@@ -52,10 +52,8 @@ function moveResources {
         continue
       fi
 
-      set -e
```
robert-bo-davis, line 55: Some moved resources may not exist so we don't actually want to error if the move fails

twilio-iac/terraform-modules/stages/chatbot/main.tf
```diff
@@ -16,9 +16,11 @@ locals {
 
   // Default chatbot SIDs can be overwriten by custom chatbot SIDs included from the helpline's additional.chatbot.tf file
   default_chatbot_sids = {
-    post_survey = var.default_autopilot_chatbot_enabled ? module.chatbots[0].post_survey_chatbot_sid : ""
+    // UGH, this is a dirty hack to get a value from outputs of a module behind count without getting and empty tuple error.
+    pre_survey  = var.default_autopilot_chatbot_enabled ? split(",", join(",", module.chatbots.*.pre_survey_bot_sid))[0] : ""
+    post_survey = var.default_autopilot_chatbot_enabled ? split(",", join(",", module.chatbots.*.post_survey_bot_sid))[0] : ""
```
GPaoloni, line 21: 😳

GPaoloni (approved):
Haven't tried running the stages, but changes looks good to me.

Is this already applied?

twilio-iac/terraform-modules/stages/chatbot/main.tf
robert-bo-davis: Yeah... I just simplified it a little bit... but hacks around TF limitations are really ugly sometimes.

robert-bo-davis:
> Is this already applied?

Not yet. Was waiting for review. The stages have been applied with `-refresh-only` flag so outputs are available. You can run plans for each stages and the output matches the current output from a plan in master.

PR #1345, opened 2023-04-12
docs: update docs and fix terragrunt setup-new-environment task

robert-bo-davis:
Primary reviewer:

## Description

Updates new helpline documentation with terragrunt instructions. This is not fully tested, but should work. @janorivera is setting up a new helpline and I will address any issues as he encounters them.


janorivera (approved):
lgtm!

PR #1346, opened 2023-04-12
[PoC] addWrapper with virtualized list on MessageList

GPaoloni:
## Description
This would be an ideal solution to https://github.com/techmatters/flex-plugins/pull/1335, using a [virtualized list wrapper](https://github.com/bvaughn/react-virtualized-auto-sizer). Unfortunately, this method appears in the code but is not supported at the time of writing. Message from Twilio support:
>Unfortunately, at this time, the list you found in the docs is the full list of components that are able to utilize the addWrapper() API.  I just double-checked with MessageList and TaskCanvas (also not supported, but has the addWrapper() method), and neither yielded any changes.  Engineering is working to expand the list of components that can utilize the API, but I don't have any sort of timetable on when those changes will be put into effect.

Leaving around in case this ever arrives :)

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### DIFF
--- a/plugin-hrm-form/package-lock.json
+++ b/plugin-hrm-form/package-lock.json
@@ -22620,6 +22620,11 @@
         "@reach/observe-rect": "^1.1.0"
       }
     },
+    "react-virtualized-auto-sizer": {
+      "version": "1.0.14",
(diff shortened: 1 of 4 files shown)


GPaoloni:
https://github.com/techmatters/flex-plugins/pull/1347

GPaoloni:
https://github.com/techmatters/flex-plugins/pull/1347

PR #1351, opened 2023-04-15
CHI-1868: Fix transfer listeners

stephenhand:
## Description

Fixes CHI-1868

* No longer strips all listeners from a text conversation when it is transferred away.
* Adds our custom listeners to a registry when we add them to a conversation
* 'Deactivates' those listeners when a conversation is transferred, i.e. removes them from the conversation but leaves them in the registry
* 'Reactivates' them if the conversation is ever transferred back, i.e. looks them up from the registry and adds them back to the convo

Other ideas I tried:

* @GPaoloni suggested changing the User Notification level for the conversation to 'muted'. From the API docs this appeared to be what we needed, but in reality it appeared to have no effect, even when muted the audio notifications for the conversation still came through
* Leaving the convo when transferring away then rejoining when transferred back - didn't work at all, the counsellor still wasn't considered a participant until the caller sent a message, messages from the conversation with the other counselor were still absent, new messages didn't appear in the counselors message view, all sorts of issues

### Checklist
- [ ] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-1868

### Verification steps

See ticket

### DIFF
--- a/plugin-hrm-form/src/conversationListeners.ts
+++ b/plugin-hrm-form/src/conversationListeners.ts
@@ -1,20 +1,20 @@
 import { Conversation } from '@twilio/conversations';
 
-type ConversationListener = Parameters<Conversation['on']>;
+type ConversationListenerParams = Parameters<Conversation['on']>;
 
-const aseloListeners: Record<string, ConversationListener[]> = {};
+const aseloListeners: Record<string, ConversationListenerParams[]> = {};
 
-export const addAseloListener = (conversation: Conversation, ...[event, listener]: ConversationListener) => {
-  conversation.on(event, listener);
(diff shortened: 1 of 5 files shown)


GPaoloni (approved):
Looks great to me!

I still wonder why everyone is subscribed to conversations when they are created. Have you opened a ticket for that? I can do that if not.

stephenhand:
> Looks great to me!
> 
> I still wonder why everyone is subscribed to conversations when they are created. Have you opened a ticket for that? I can do that if not.

I was thinking about this, one thing worth doing might be to compare behaviour of agents vs admins. I'm basing my observations on an admin account, where getting events from all convos l, whilst still being odd from a design perspective, at least makes sense from a security/ permissions angle

PR #1352, opened 2023-04-16
CA configuration and Task router module

robert-bo-davis:
…outer
Primary reviewer:
@janorivera 
## Description

Tested example of passing in template to configure taskRouter.

### DIFF
--- a/twilio-iac/terraform-modules/taskRouter/v1/main.tf
+++ b/twilio-iac/terraform-modules/taskRouter/v1/main.tf
@@ -1,0 +1,101 @@
+terraform {
+  required_providers {
+    twilio = {
+      source  = "twilio/twilio"
+      version = "0.17.0"
+    }
+  }
+}
+
+locals {
+  helplines                      = var.helplines == null ? [var.helpline] : var.helplines
+  helplines_filter               = var.helplines == null ? "1==1" : "helpline IN [${join(", ", formatlist("'%s'", local.helplines))}]"
+  task_routing_filter_expression = var.custom_task_routing_filter_expression != "" ? var.custom_task_routing_filter_expression : "${local.helplines_filter} OR channelType ==\"web\" OR isContactlessTask == true"
+  event_callback_url             = "${var.serverless_url}/webhooks/taskrouterCallback"
+}
+
+// This module assumes that we only work with one Workspace
+// Workspace
+resource "twilio_taskrouter_workspaces_v1" "flex_task_assignment" {
+  friendly_name      = "Flex Task Assignment"
+  multi_task_enabled = true
+  event_callback_url = local.event_callback_url
+  events_filter      = join(",", var.events_filter)
+}
+
+// Task Queue
+resource "twilio_taskrouter_workspaces_task_queues_v1" "task_queue" {
+  for_each       = { for tq in var.task_queues : tq.name => tq }
+  friendly_name  = each.value.friendly_name
+  workspace_sid  = twilio_taskrouter_workspaces_v1.flex_task_assignment.sid
+  target_workers = each.value.target_workers
+}
+
+moved {
+  from = twilio_taskrouter_workspaces_task_queues_v1.helpline_queue
+  to   = twilio_taskrouter_workspaces_task_queues_v1.task_queue["master"]
+}
+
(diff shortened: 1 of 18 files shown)


twilio-iac/terraform-modules/taskRouter/v1/main.tf
```diff
@@ -0,0 +1,101 @@
...
+  to   = twilio_taskrouter_workspaces_task_queues_v1.task_queue["master"]
+}
+
+// Workflow
+resource "twilio_taskrouter_workspaces_workflows_v1" "workflow" {
+  for_each      = { for w in var.workflows : w.name => w }
+  friendly_name = each.value.friendly_name
+  workspace_sid = twilio_taskrouter_workspaces_v1.flex_task_assignment.sid
+  configuration = templatefile(
+    each.value.templatefile,
```
robert-bo-davis, line 45: This is the important part. We are passing in a path to a template file and then *only* feeding in variables that come from locals and other data that we don't have access to in the configuration layer.

This allows us to have very complex and customizable json per helpline that gets filled in with relevant data about other resources that are managed by terraform.

twilio-iac/terraform-modules/taskRouter/v1/main.tf
```diff
@@ -0,0 +1,101 @@
...
+resource "twilio_taskrouter_workspaces_v1" "flex_task_assignment" {
+  friendly_name      = "Flex Task Assignment"
+  multi_task_enabled = true
+  event_callback_url = local.event_callback_url
+  events_filter      = join(",", var.events_filter)
+}
+
+// Task Queue
+resource "twilio_taskrouter_workspaces_task_queues_v1" "task_queue" {
+  for_each       = { for tq in var.task_queues : tq.name => tq }
```
robert-bo-davis, line 28: Another note is to try to use machine friendly names for your for_each map ids. If you have spaces, accent marks, or apostrophes in them that might be required in "friendly_name" you may make your life difficult elsewhere.

robert-bo-davis: if it isn't obvious what i did here.. you had :

```
{ for tq in var.task_queues : tq.friendly_name => tq.target_workers }
```

Which meant you had to do lookups like `task_queues["some long name with spaces"]` This is not the end of the world with simple names, but if we create for_each lists over things in another language, it will be harder to look up those names. Best to get in the habit of adding a "machine/english friendly name" in lower_snake_case and using that for your keys.

twilio-iac/terraform-modules/taskRouter/v1/main.tf
```diff
@@ -0,0 +1,101 @@
+terraform {
```
robert-bo-davis, line 1: Also notice that I named my sub-module `v1`. There is always a chance we will need to introduce breaking changes at some point and will (hopefully temporarily) have different versions of the same module in use concurrently. We should version these modules instead of naming them helpline specific things, default, or anything else to allow for this when we need to do it, but try to enforce the idea that there is only one primary code path.

There are other patterns for doing this if things get more complicated (mostly using git tags and remote modules), but I don't think we will get to that point for a long while.

twilio-iac/terraform-modules/taskRouter/v1/main.tf
robert-bo-davis: This could also be changed so that workspaces and task_queues all the others are already a map that doesn't need to be converted like

```
task_queue = {
	some_name : {
		friendly_name : "some friendly name"
		target_workers : "1=1"
	}
}
```

and:

```
variable "task_queue" {
	type = map(object({
		friendly_name = string
		target_workers = string
	}))
}
```

janorivera:
This looks beautiful thanks Bo! :D 
My only question is why there is a difference of format between the helplines template (json) and the ca template (tftpl) ? 
I'm going to go ahead and merge this.

robert-bo-davis:
> My only question is why there is a difference of format between the helplines template (json) and the ca template (tftpl) ? 

Oops! They should both be .tftpl. Could you fix that in your branch, please?

janorivera:
sure no problem!

PR #1356, opened 2023-04-18
CHI-1785: Replace message list (virtualized)

GPaoloni:
## Description
This PR:
- Factors out the message list component used in the transcripts section, also making the list a "virtualized" one.
- Reuses above component to replace the actual message list for chat contacts.

Note: styles are not matching the styles of Flex v2, we may need another pass for that.

Explanation of this change:
List virtualization or windowing is a performance optimization consisting of "reducing the amount of items attached to the DOM", by only rendering what's visible in the viewport. When you scroll, the elements that move out from the viewport are removed. This helps reducing the initial load and scroll perfomance issues.
However, none of this two issues was actually related to the slowness of the input typing. The actual issue is that on each key press, the browser re-paints every message attached to the DOM (probably because of the dynamic-size nature of the message bubles). This repainting causes a lot "reflows" in the browser, which is a blocking task. The longer the message list, the more repaints are needed and the worst the slowness gets.
By virtualizing this list, we reduce the ammount of messages in the DOM to a reasonable size. They are still being repainted, so there's still some overhead, but the time spent on each repaint is noticeable reduced.

### Checklist
- [ ] New tests added
- [x] Feature flags added
- [x] Strings are localized

### Verification steps
- `npm run dev`.
- Accept a new webchat task.
- Confirm that the message list looks good.
- Confirm that the typing indicator shows as expected if "child" starts typing (only webchat).
- Confirm that the transcripts view for a saved contact are still working as expected.


murilovmachado (approved):
Works great!!🚀

PR #1361, opened 2023-04-20
Chat Input Dynamic Resizing

murilovmachado:
## Description
Handles dynamic resizing of the text area.
It starts with showing **1 line** of text and grows to **5 lines**. After that, a scrollbar is displayed.
When erasing text, it shrinks appropriately, also handling whether the scrollbar is shown or not.

text-area

### Verification steps
Play with writing and erasing text and check that the text area dynamic resizing works according to the specs.

### DIFF
--- a/plugin-hrm-form/src/components/AseloMessageInput.tsx
+++ b/plugin-hrm-form/src/components/AseloMessageInput.tsx
@@ -30,6 +30,21 @@ import {
 } from '../states/conversations';
 import asyncDispatch from '../states/asyncDispatch';
 
+/**
+ * The following CSS attributtes should be set in here
+ * so the dynamic resizing will work properly:
+ * - padding (vertical: top or bottom)
+ * - border-width
+ * - line-height
+ */
+const PADDING_VERTICAL = 8;
+const BORDER_WIDTH = 1;
+const LINE_HEIGHT = 20;
+const MIN_LINES = 1;
+const MAX_LINES = 5;
+const MIN_HEIGHT = MIN_LINES * LINE_HEIGHT + PADDING_VERTICAL * 2 + BORDER_WIDTH * 2;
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/components/AseloMessageInput.tsx
```diff
@@ -75,6 +94,72 @@ const AseloMessageInput: React.FC<Props> = ({
...
+    if (isNotDisplayingScrollbar) {
+      textarea.style.overflow = prevOverflow;
+    }
+
+    return scrollHeight;
+  };
+
+  const handleDynamicSize = (textarea: HTMLTextAreaElement) => {
+    const currentScrollHeight = getPureScrollHeight(textarea);
+    const shouldExpand = currentScrollHeight > prevScrollHeight && height < MAX_HEIGHT;
```
stephenhand, line 139: Separating the 'shouldExpand' and 'shouldShrink' cases seems a little odd to me, would it not make more sense to just determine if the height needs to change at all and then set it if it does?

stephenhand (approved):
LGTM - couple of non blocking comments & qs

plugin-hrm-form/src/components/AseloMessageInput.tsx
```diff
@@ -105,7 +190,8 @@ const AseloMessageInput: React.FC<Props> = ({
     }
   });
 
-  const handleChange = () => {
+  const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+    handleDynamicSize(e.target);
```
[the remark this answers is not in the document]
murilovmachado: Yeah, let's leave debouncing it for later, especially because it can make the auto-grow experience poor. Maybe looking at other optimization instead of debouncing would be a better fit for this feature.

plugin-hrm-form/src/components/AseloMessageInput.tsx
murilovmachado: It's because to determine the height, we need to know if it's shrinking or growing, there's some specific logic for each scenario

PR #1374, opened 2023-04-25
CHI-1899: All account runner helper function for scripts

stephenhand:
## Description

* A function in scripts that will scrape the SSM parameters for all the account creds in an environment and allow you to pass in an action to run against all of them
* An example script that uses this function to patch the service configuration of every account

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-1899

### Verification steps

Verified by patching asset bucket urls in all accounts

### DIFF
--- a/scripts/src/terraformSupplemental/updateFlexServiceConfiguration.ts
+++ b/scripts/src/terraformSupplemental/updateFlexServiceConfiguration.ts
@@ -1,15 +1,22 @@
 import axios from 'axios';
-import { logDebug, logSuccess } from '../helpers/log';
+import { logDebug, logInfo, logSuccess, logWarning } from '../helpers/log';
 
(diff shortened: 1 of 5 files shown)


scripts/src/terraformSupplemental/updateFlexServiceConfiguration.ts
```diff
@@ -23,29 +30,60 @@ export async function updateFlexServiceConfiguration(payload: unknown) {
   }
 }
 
-export async function patchFeatureFlags(flags: Record<string, boolean>) {
-  logDebug('Setting the following feature flags:', flags);
-
+export async function patchFeatureFlags(
```
GPaoloni, line 33: Probably deserves a rename?

PR #1383, opened 2023-04-26
Address styling diffs between Twilio & Aselo Messages List

GPaoloni:
## Description
This PR:
- Reworks styles for AseloMessageList (message list replacement), to make them look like the Flex 2 styles.
- Fixes a small bug where sending a message won't revert the text area to it's initial height.
- Adds debouncing to the "resize text area" behavior, as I've seen some degradation when not doing so:
  - No debounce:
    Screenshot from 2023-04-26 21-18-12
  - 300ms debounce:
    Screenshot from 2023-04-26 21-22-01
  - 500ms debounce:
    Screenshot from 2023-04-26 21-14-06

Screenshots of the changes (what to look at):
- Transcript of a saved contact:
  Screenshot from 2023-04-26 20-07-39
- Message list in a live chat
  Screenshot from 2023-04-26 20-05-01
  Screenshot from 2023-04-26 20-04-52

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1897)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Enable the feature flag, can be done via `src/hrmConfig.ts` by returning
```
featureFlags: {
  ...featureFlags,
  enable_aselo_messaging_ui: true,
},
```
- `npm run dev`  to start local server.
- Start and accept a new chat based contact.
- Confirm that the styles are looking like those in default Flex setup (or in the screenshots). Ideally test this in different monitor sizes to confirm nothing goes crazy.


plugin-hrm-form/src/components/AseloMessaging/AseloMessageInput/AseloMessageInput.tsx
```diff
@@ -197,14 +190,22 @@ const AseloMessageInput: React.FC<Props> = ({
...
+      setHeight(initialHeight);
     }
   });
 
-  const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
-    handleDynamicSize(e.target);
-    triggerTyping();
-    updateSendButtonState();
-  };
+  const handleChange = debounce(
```
stephenhand, line 197: Does a leading call for all 3 not impact the UX at all? I was concerned doing anything that updated the screen as a leading call could hamper immediate typing updates for slow machines, make it 'skip' when you first started typing, or make it choppy when someone is typing slowing?

If you've tested these cases and it's ok then I'm ok with it

stephenhand (approved):
Code looks good, just one Q. I've not had chance to test yet, will give it a go ASAP

stephenhand:
Spotted one minor styling issue in testing: 
image

A very long word gets a horizontal scrollbar rather than breaking across multiple lines, which I assume is the desired behaviour?

plugin-hrm-form/src/components/AseloMessaging/AseloMessageInput/AseloMessageInput.tsx
GPaoloni: Since one is a simple state update, and the other is a non-awaited promise, the only relevant one is the only one that was also happening at start (resize calculation), so I don't think there's an issue. But I can split them all if you think is wiser!

PR #1386, opened 2023-04-27
CHI-1907: Resize on canned response fix

stephenhand:
## Description

* Fix issue where canned response selection doesn't resize the text input. Setting the text via redux doesn't trigger an onChange event, so now the resize check is explicitly triggered as part of the effect handling redux text updates

### Checklist
- N/A New tests added
- [X] Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-1907

### Verification steps

Select a lengthy canned response
Try editing the canned response afterwards, pasting text etc.


GPaoloni (approved):
Thanks!

PR #1398, opened 2023-05-04
increase message bubble text size

acedeywin:
Primary reviewer: @GPaoloni  

## Description

- Modification - Cleanup styling: message bubble text has a size of 12px. The goal was to increase it to Twilio's default font-size of 14px.

### Checklist
- [N/A] New tests added
- [N/A] Feature flags added
- [N/A] Strings are localized

### Related Issues
Fixes CHI-1925

### Verification steps

- Initiate a chat conversation
- Verify that the message bubble text size matches the Twilio default (14px)

### DIFF
--- a/plugin-hrm-form/src/components/Messaging/MessageItem/styles.tsx
+++ b/plugin-hrm-form/src/components/Messaging/MessageItem/styles.tsx
@@ -116,7 +116,7 @@ export const MessageBubbleBody = styled('div')`
 MessageBubbleBody.displayName = 'MessageBubbleBody';
 
 export const MessageBubbleBodyText = styled(FontOpenSans)<{ isCounselor: boolean }>`
-  font-size: 12px;
+  font-size: 14px;
   line-height: 1.54;
   overflow-wrap: break-word;
   color: ${({ isCounselor }) => (isCounselor ? '#FFFFFF' : '#222222')};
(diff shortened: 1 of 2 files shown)


GPaoloni (commented):
@acedeywin Sorry to be annoying, but since we own this components now, you don't need to use the CSS overrides, instead you can directly edit the style element. Seems like [this one](https://github.com/techmatters/flex-plugins/blob/dda72cd7be3a590838757dc8d6d41c79014ee5c7/plugin-hrm-form/src/components/Messaging/MessageItem/styles.tsx#L119) is overriding [it's parent's](https://github.com/techmatters/flex-plugins/blob/dda72cd7be3a590838757dc8d6d41c79014ee5c7/plugin-hrm-form/src/components/Messaging/MessageItem/styles.tsx#L113) style, so removing that line should fix the issue.

acedeywin:
> @acedeywin Sorry to be annoying, but since we own this components now, you don't need to use the CSS overrides, instead you can directly edit the style element. Seems like [this one](https://github.com/techmatters/flex-plugins/blob/dda72cd7be3a590838757dc8d6d41c79014ee5c7/plugin-hrm-form/src/components/Messaging/MessageItem/styles.tsx#L119) is overriding [it's parent's](https://github.com/techmatters/flex-plugins/blob/dda72cd7be3a590838757dc8d6d41c79014ee5c7/plugin-hrm-form/src/components/Messaging/MessageItem/styles.tsx#L113) style, so removing that line should fix the issue.

@GPaoloni - Thank you for pointing this out. The changes have been implemented.

GPaoloni (approved):
Thanks!

PR #1402, opened 2023-05-04
Alejandro/cl/tf migration 2

janorivera:
Primary reviewer:

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


twilio-iac/terraform-modules/stages/configure/variables.tf
```diff
@@ -181,3 +181,8 @@ variable "enable_post_survey" {
   type    = bool
   default = false
 }
+
+variable "slack_webhook_url" {
```
robert-bo-davis, line 185: This probably needs to be pulled out of SSM. Let's discuss this in slack tomorrow. Everything other than this looks good!

robert-bo-davis (changes requested):

robert-bo-davis (approved):
LGTM. A few minor nit notes.

One thing I'd like to add into the process is to make sure we go ahead and run plan/apply against all of the helplines that have been migrated for any stages that have changed to make sure they all maintain a clean apply state before every merge.

It could be a lot, but if we try to do it *every time* it won't end up being a huge chore in the future. Eventually, we'll set up a CI/CD job that runs plan-all for ALL terragrunt helplines on every PR before it can be merged so we ensure that we maintain clean applies, but for now, if we do it manually while there are only a few, it will keep us from breaking things unexpectedly.

Feel free to merge this and perform the final "plan all helplines" check after your next one. I just wanted to go ahead and put the idea in your head.

PR #1408, opened 2023-05-09
Alejandro/nz/stg tg setup

janorivera:
Primary reviewer:
@robert-bo-davis 

## Description

@robert-bo-davis I'm not quite sure if you need to review these PRs, but I'm tagging you here mainly because of the change on the below file. I think that now that we can create workflows dynamically it doesn't make much sense to save just the master_workflow_id or survey_workflow_id, to be honest I don't why we save them, but in any case we shouldn't make the survey one a required/default workflow since just a few helplines use it. For now I've added a default value of 123, but we might want to improve this in the future. What do you think?
[twilio-iac/terraform-modules/stages/provision/main.tf](https://github.com/techmatters/flex-plugins/pull/1408/files#diff-74fcf09d3801ed2a558f902cef03f1fee76fe47d8fd98e5f01bbb3c44af369a9)

This configuration has been applied and tested on the staging account.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


twilio-iac/terraform-modules/stages/provision/main.tf
```diff
@@ -73,7 +75,9 @@ module "aws" {
   flex_proxy_service_sid             = module.services.flex_proxy_service_sid
   # TODO: manually delete this resource from SSM after migration
   # post_survey_bot_sid = module.chatbots.post_survey_bot_sid
-  survey_workflow_sid = module.survey.survey_workflow_sid
+  # This is not pretty, ideally we will not create a variable if we don't create the survey resource
+  # I'll refactor this module in the future but for now I'll assign a fixed 123 id since it doesn't accept empty strings.
+  survey_workflow_sid = lookup(module.taskRouter.workflow_sids, "survey", "123")
```
robert-bo-davis, line 80: We really need to lean up the aws module eventually... but for now, would [try()](https://developer.hashicorp.com/terraform/language/functions/try#examples) make more sense?

ex:
```
survey_workflow_sid = try(module.taskRouter.workflow_sids.survey, null)
```

robert-bo-davis: We could also just pass down `module.taskRouter.workflow_sids` and use the `try()` in a local to only create the new style SSM param (`/<environemnt>/twilioi/...`) if it is needed. The old style ssm params REALLY need to go away eventually.

janorivera:
ok @robert-bo-davis  OK I made some changes according to our conversation. I've applied it and tested it and it is working fine now.  This is ready for review. :)

twilio-iac/terraform-modules/aws/default/ssm.tf
```diff
@@ -137,6 +137,7 @@ resource "aws_ssm_parameter" "twilio_flex_proxy_service_sid" {
 }
 
 resource "aws_ssm_parameter" "twilio_survey_workflow_sid" {
+  count       = var.survey_workflow_sid == "" ? 0 : 1
```
robert-bo-davis, line 140: We are going to need this one to exist as well to prevent the same issues we encountered with the old style SSM params in the upcoming migration to the new SSM params.

robert-bo-davis (changes requested):

janorivera:
ok, good catch! thanks @robert-bo-davis

robert-bo-davis (approved):
LGMT! Thanks

PR #1414, opened 2023-05-15
POC: Flex isolated UI tests

stephenhand:
This is a POC for Flex UI tests, i.e. test written in playwright, that test the whole Flex application running together, but isolated from the backend - both the Twilio and Aselo backends that is

Includes example tests that:

* Test the task counts window, to give a very simple demonstration of how we might manipulate the mocked backed to simulate various Twilio driven scenarios in our plugin code
* Run AXE tests on some of the Cases UI, to demonstrate how we might write tests for our own code (which is much more straightforward than trting to mock out Twilio stuff.

The new code is almost all under `./e2e-tests/ui-test` - in the long term it would probably live in it's own package, but since it reuses some of the e2e page objects, and that's a pattern we probably want to continue with, it makes sense to keep them together in the same package for the POC. We would probably need a 3rd 'common page objects' package if we split them to facilitate this reuse.

The mocks are divided between `.e2e-tests/ui-test/flex-in-a-box and ./e2e-tests/ui-test/aselo-service-mocks`

`flex-in-a-box` is generic Flex UI mocking code which could be reused by anybody trying to run isolated Playwright tests on a Flex project. It could theoretically be moved out to it's own repo as a stanndalone lib

`aselo-service-mocks` is higher level mocking code (a lot of which is built on top of `flex-in-a-box`) which mocks services and customised Twilio types specific to Asedlo

Also includes an extra CI check that runs the new tests

To test locally:

From the `e2e-test` directory run: 

```
npm run test:ui
```

TODO: Some more commentary


mythilytm (approved):
This is really great work. PR is massive so I haven't gotten a chance to review evertthing

PR #1419, opened 2023-05-16
refactor: configuration driven tf updates for aws lex v1

robert-bo-davis:
Primary reviewer:

## Description

These changes have been applied. You will have to click the [Return to the V1 console](https://console.aws.amazon.com/lex/home?region=us-east-1#bots:) button to see the changes in the aws console. I don't promise it is perfect, and there is likely more that needs to be done

I left the awscc work I had done in place so we don't loose that work if we need to go back to it.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


twilio-iac/terraform-modules/lex/v1/main.tf
```diff
@@ -0,0 +1,134 @@
+terraform {
+  required_providers {
+    aws = {
+      source                = "hashicorp/aws"
+      configuration_aliases = [aws.hl-region]
```
robert-bo-davis, line 5: @janorivera We need to convert the aws module to use this passed in provider alias pattern instead of creating its own provider eventually. Just wanted you to be aware of it. Can search the PR for `hl-region` to see how it is used.

acedeywin (approved):
LGTM

robert-bo-davis:
Replacing with #1484

PR #1427, opened 2023-05-18
adding resources_base_url

janorivera:
Primary reviewer:
@stephenhand 
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### DIFF
--- a/twilio-iac/terraform-modules/flex/service-configuration/main.tf
+++ b/twilio-iac/terraform-modules/flex/service-configuration/main.tf
@@ -9,7 +9,7 @@ terraform {
 
 locals {
   hrm_url            = var.hrm_url == "" ? (var.short_environment == "PROD" ? "https://hrm-production.tl.techmatters.org" : (var.short_environment == "STG" ? "https://hrm-staging.tl.techmatters.org" : "https://hrm-development.tl.techmatters.org")) : var.hrm_url
-  resources_base_url = var.resources_base_url == "" ? "https://hrm-${lower(var.environment)}.tl.techmatters.org" : var.resources_base_url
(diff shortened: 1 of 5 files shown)


twilio-iac/terraform-modules/flex/service-configuration/main.tf
```diff
@@ -8,9 +8,10 @@ terraform {
 }
 
 locals {
-  hrm_url           = var.hrm_url == "" ? (var.short_environment == "PROD" ? "https://hrm-production.tl.techmatters.org" : (var.short_environment == "STG" ? "https://hrm-staging.tl.techmatters.org" : "https://hrm-development.tl.techmatters.org")) : var.hrm_url
-  assets_bucket_url = "https://assets-${lower(var.environment)}.tl.techmatters.org"
-  permission_config = var.permission_config == "" ? var.operating_info_key : var.permission_config
+  hrm_url            = var.hrm_url == "" ? (var.short_environment == "PROD" ? "https://hrm-production.tl.techmatters.org" : (var.short_environment == "STG" ? "https://hrm-staging.tl.techmatters.org" : "https://hrm-development.tl.techmatters.org")) : var.hrm_url
```
robert-bo-davis, line 11: Really just a nit, but since we are passing the full `var.environment` in now, so all of these defaults could just use variable substitution instead of the janky ternary operator chains like the assets_bucket_url does below.

janorivera: yeah I agree... I'll create a another PR for that... I always see that when I'm doing something else

stephenhand (approved):

PR #1431, opened 2023-05-25
chore: add ssm param for transcript retention override

robert-bo-davis:
Primary reviewer:

## Description

This adds an SSM parameter to set the CA Production ssm param for 90 days of transcript retention. It has been targeted applied.

### Verification steps

Verify ssm param exists


stephenhand (approved):
LGTM

PR #1432, opened 2023-05-25
bg: Fix Household Member/Perpetrator Duplication

acedeywin:
Primary reviewer:

## Description

- Bug fix: Household Member/Perpetrator Duplication

**The Problem:** In the Case forms, while adding household members or perpetrators, when we add two or more households or perpetrators consecutively (Save and add Another…), it gets duplicated (even though the checkbox to add them to the perpetrator/household is not selected). When we add just one household or perpetrator and exit, it doesn’t happen. This only happens when we add two or more households or perpetrators consecutively.

### Checklist
- [N/A] New tests added
- [N/A] Feature flags added
- [N/A] Strings are localized

### Related Issues
Fixes [CHI-1963](https://tech-matters.atlassian.net/browse/CHI-1963)

### Verification steps

- Navigate to case-list and select a case you created
- Click on the `Perpetrator` or `Household Member` button
- Input data, but don't select the checkbox. Click on the `Save and add Another…` button
- Repeat step 3
-  Repeat step 3 again, but this time, select the checkbox
- Navigate back to case-list (click on the `Cancel` button)
- Verify that the added household member/perpetrator is only duplicated where the checkbox was selected.

### DIFF
--- a/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
+++ b/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
@@ -43,7 +43,7 @@ import { transformValues } from '../../services/ContactService';
 import { updateCase } from '../../services/CaseService';
 import { createStateItem, CustomHandlers, disperseInputs, splitAt, splitInHalf } from '../common/forms/formGenerators';
 import { useCreateFormFromDefinition } from '../forms';
-import type { CaseInfo, CaseItemEntry, CaseItemFormValues, CustomITask, StandaloneITask } from '../../types/types';
+import type { CaseInfo, CaseItemEntry, CustomITask, StandaloneITask } from '../../types/types';
 import {
   AddCaseSectionRoute,
   AppRoutes,
--- a/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
+++ b/plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
@@ -191,8 +191,23 @@ const AddEditCaseItem: React.FC<Props> = ({
       };
       newInfo = sectionApi.upsertCaseSectionItemFromForm(info, newItem);
       formDefinition.forEach(fd => {
+        /*
+         * In the first instance when fd.type === 'copy-to', rawForm[fd.name] returns a boolean.
+         * Then subsequently rawForm[fd.name] returns an array
+         * First, we check when rawForm[fd.name] is a boolean and assign the value validateCopyTo.
+         * Next, we check when rawForm[fd.name] is array,
+         * validates if its length is greater than zero and assign the value to validateCopyTo
+         */
+
+        let validateCopyTo = rawForm[fd.name];
+        const isArray = Array.isArray(rawForm[fd.name]);
+
+        if (isArray) {
+          validateCopyTo = (rawForm[fd.name] as string).length > 0;
+        }
+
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
```diff
@@ -191,8 +191,23 @@ const AddEditCaseItem: React.FC<Props> = ({
       };
       newInfo = sectionApi.upsertCaseSectionItemFromForm(info, newItem);
       formDefinition.forEach(fd => {
+        /*
```
stephenhand, line 194: Can we get to the root of why the value in `rawForm[fd.name]` changes type in this scenario? The value should be a boolean every time. It could be that some internal state that needs to be cleared isn't so we are getting multiple values in there or something?

This fix feels a bit like we are working around an issue we don't fully understand. This may turn out to be the best fix, but I want to decide that based on a full understanding of the problem

acedeywin: @stephenhand - From my investigations, the problem rests on the mutation happening in react-hook-form. So the mutated value gets passed down to the `workingCopy`. The solution I came up with was to reset the form state using the reset method in `react-hook-form` `method.reset()`. So this method is called in the `saveAndStay()` function, which resets the entire form state, fields reference, and subscriptions.

stephenhand: The difference between the quick fix I tried to do with methods.reset and this one is that it puts the call right at the end of `saveAndStay` - I wouldn't have thought that would make a difference but apparently it did, any thoughts as to why?

Either way, nice work finding the root of the issue

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
```diff
@@ -43,7 +43,7 @@ import { transformValues } from '../../services/ContactService';
 import { updateCase } from '../../services/CaseService';
 import { createStateItem, CustomHandlers, disperseInputs, splitAt, splitInHalf } from '../common/forms/formGenerators';
 import { useCreateFormFromDefinition } from '../forms';
-import type { CaseInfo, CaseItemEntry, CustomITask, StandaloneITask } from '../../types/types';
+import type { CaseInfo, CaseItemEntry, CaseItemFormValues, CustomITask, StandaloneITask } from '../../types/types';
```
stephenhand, line 46: Not sure if this change is needed?

stephenhand (commented):
One minor nit then LGTM

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
acedeywin: @stephenhand  - Unused type has been removed.

plugin-hrm-form/src/components/case/AddEditCaseItem.tsx
acedeywin: @stephenhand - Where were you calling `method.reset()`? Just after `save()` or right before it? It has to be at the end in order to set `geValue()` to the default value. By then, every other method (removeCaseSectionWorkingCopy(), initialiseNewCaseSectionWorkingCopy()) would have been called. It's more like doing a cleanup at the end.

stephenhand (approved):

PR #1440, opened 2023-06-01
Conferencing: preserve dialog across states

GPaoloni:
## Description
This PR adds a `conferencing` reducer to our Redux store, that will keep track of the conferencing state for a given voice task. The idea is to preserve the partial input provided for the external call, in case the counselor change the view and comes back to the task.

Waiting on the core UI to wire this.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1941)
- [x] New tests added
- [ ] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Start development server (`npm run dev`).
- Start a voice contact and accept in your end.
- Play around with the "add external party" dialog and confirm state is preserved if you change to a different view like Search (dialog kept open, partial input preserved).
- Confirm that a new participant placeholder is added when you add an external party. Note you can't interact with it until the party either accepts or rejects the call.


stephenhand (approved):
Looks good so far! I do sometimes wonder if the way we structure our state, with lots of separate context specific task maps is the best way to go - or if we should try to merge all the state that's contextual to a single task under one map in the state? 

It might help us address memory leaks etc. more easily etc. if, when a task is done, we just have to remove an entry from a single map rather than trying to find all the state related to that task in various contexts.

That's a much bigger question for another day though, the separate maps approach is working fine for us so far.

plugin-hrm-form/src/components/Conference/ConferenceActions/ToggleMute.tsx
```diff
@@ -51,13 +51,16 @@ const ToggleMute: React.FC<Props> = ({ call, task, conference }) => {
...
+  const buttonText = `Microphone ${isMuted ? 'On' : 'Off'}`;
 
   return (
     <StyledConferenceButtonWrapper>
       <StyledConferenceButton disabled={!isLiveCall} onClick={handleClick}>
         {isMuted ? <MicOffOutlined /> : <MicNoneOutlined />}
       </StyledConferenceButton>
-      <span>Microphone {isMuted ? 'On' : 'Off'}</span>
+      <span>
+        <Template code={buttonText} />
```
mythilytm, line 62: Thanks for adding all the template strings I missed 👍

plugin-hrm-form/src/components/Conference/ConferenceActions/ConferencePanel.tsx
```diff
@@ -14,51 +14,92 @@
...
+    setIsLoading(true);
     const from = Manager.getInstance().serviceConfiguration.outbound_call_flows.default.caller_id;
-    const to = targetNumber;
-    await conferenceApi.addParticipant({ from, conferenceSid, to });
-
-    setIsAdding(false);
+    const to = phoneNumber;
+    const result = await conferenceApi.addParticipant({ from, conferenceSid, to });
+    addNewParticipantToState(result.participant);
+    setIsLoading(false);
```
mythilytm, line 90: Nice to see all the improvements and refactors

mythilytm (approved):
Works great!

PR #1457, opened 2023-06-14
Disable transfers when there's 3 or more participants

murilovmachado:
## Description
Disable transfers when there are 3 or more participants.

~TODO:~
- ~Disable button when adding a participant~
  - ~This will require to move Conference local state into redux~
- ~Check that there's only 2 participants when triggering the transfer action~
- ~tests~

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-1940

### Verification steps
- Make a voice call
- Check the `transfer` button is enabled
- Add a participant
- Check the `transfer` button is disabled


mythilytm (approved):
Looks great!

GPaoloni (approved):

PR #1458, opened 2023-06-14
Conferencing: continue 3-way call after 1 person hangs up

GPaoloni:
This PR is related to https://github.com/techmatters/serverless/pull/474.

## Description
This PR adds logic to handle 3+ participants in a conference, not causing the conference to end if one of them hangs.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1960)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Make sure the above branch of Serverless is deployed.
- `npm run dev` to start local Flex.
- Start a new call contact.
- Accept with the Flex instance.
- Start a conference with a 3rd participant.
- Confirm that if there are more than 2 parties in conference, one hanging won't end the conference.
- Confirm that if there are 2 parties, one hanging will end the conference.


plugin-hrm-form/src/components/Conference/ConferenceMonitor/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+import { conferenceApi } from '../../../services/ServerlessService';
+
+const isJoinedWithEnd = (p: ConferenceParticipant) => p.status === 'joined' && p.mediaProperties.endConferenceOnExit;
+const isJoinedWithoutEnd = (p: ConferenceParticipant) =>
+  p.status === 'joined' && !p.mediaProperties.endConferenceOnExit;
+
+type Props = TaskContextProps;
+
+const ConferenceMonitor: React.FC<Props> = ({ conference }) => {
```
stephenhand, line 28: Sorry for the basic questions but this feels like something we should be handling on the backend - perhaps using this API https://www.twilio.com/docs/voice/twiml/conference#attributes-statusCallback , what's the thinking behind handling it here - is there no way to set up the webhook on Flex conference calls, or is this just a bit of tech debt to move quickly and deliver the feature (either is a good reason TBF)?

Also, an invisible component feels icky, but I trust that you've looked into it and concluded that this is neater than other options (redux, action listeners or w/e). 

Perhaps we could move the core logic out into a separate file? Looking at it it's hard to tell immediately that this is doesn't do any user facing UI work because there's a fair bit of code in there. If the component was just a few lines long with an effect & a state variable it would be more obvious it's just a stub. Unit tests would be easier to write as well.

stephenhand (approved):
Few questions but nothing blocking. Not had chance to test, will get to it in the afternoon hopefully.

stephenhand:
FYI sorry I'm not sure I'll get time to test this locally soon - given than this feature is getting pretty intensive QA you might just want ot merge & push on

PR #1468, opened 2023-06-21
Prepopulate forms from Lex pre-survey

murilovmachado:
## Description
Adapt prepopulate forms for handling pre-survey coming from Lex, as well.
A feature flag `enable_lex_prepopulate` is added to tell if it's using Lex or not.

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-1791

### Verification steps
Pre-population should work on both scenarios:
- Using **Autopilot** and `FeatureFlags.enable_lex_prepopulate = false`, or
- Using **Lex** and `FeatureFlags.enable_lex_prepopulate = true`


plugin-hrm-form/src/utils/prepopulateForm.ts
```diff
@@ -177,10 +208,10 @@ export const prepopulateForm = (task: ITask) => {
     return;
   }
 
-  const { answers } = memory.twilio.collected_data.collect_survey;
+  const answers = getAnswers(featureFlags.enable_lex_prepopulate, memory);
 
-  const isAboutSelf = answers.about_self.answer === 'Yes';
-  const callType = isAboutSelf || !answers.about_self ? callTypes.child : callTypes.caller;
+  const isAboutSelf = answers.aboutSelf === 'Yes';
+  const callType = isAboutSelf || !answers.aboutSelf ? callTypes.child : callTypes.caller;
```
GPaoloni, line 214: I know this is not your change, but why the `|| !answers.aboutSelf` here :thinking: ?

GPaoloni (approved):
LGTM!

plugin-hrm-form/src/utils/prepopulateForm.ts
murilovmachado: I agree this is not very readable, I needed to think to figure out the answer:
This line is a safe guard for a scenario where `aboutSelf` is not present on the `answers` object (in case the bot is misconfigured, for example)

PR #1474, opened 2023-06-23
CHI-1846" Get the 'log in to Flex' E2E test running against KHP production

robert-bo-davis:
## Description

The primary problem solved by this PR is a refactor of the e2e-test configuration system to make it more future proof for running e2e tests agains various staging/production helplines. While configuring this I also moved SSM configuration loading into the configuration system.

While testing some fixes, I deployed webchat for the development e2e user and that hadn't been done in over a year. This fixes the locator issues that surfaced after that deployment.

While testing qa build e2e tests, i found a regression that was introduced by the e2e-ui tests adding a dependency on plugin-hrm-form. I DRYed out the isntall path for plugin-hrm-form and created PRs in all of our other repos that depend on e2e tests for qa tag cutting.

The list of sibling PRs to this are:

techmatters/webchat#260
techmatters/hrm#404
techmatters/serverless#477

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### Verification steps

From your local env
Run `npm run test:staging:khp`
Run `npm run test:local`

Test deploying a qa release from this branch.


.github/actions/install-flex-plugin/action.yml
```diff
@@ -0,0 +1,16 @@
+name: 'install-flex-plugin'
```
robert-bo-davis, line 1: We were installing all kinds of different versions of flex tooling. I DRYed out the basic install steps for reuse in actions.

e2e-tests/config.ts
```diff
@@ -0,0 +1,252 @@
...
+  /**
+   * This probably seems weird, and it is. Playwright doesn't run in a
+   * single process, so we can't just set a global variable. Instead,
+   * we depend on environment variables to be the global source of truth
+   * and to help us to only load SSM values once.
+   * When we explicitly set a config value, we also set the env var
+   * which will be always be the first source loaded when attempting to
+   * init a config value.
+   */
+  process.env[configOptions[key].envKey] = value as string;
```
stephenhand, line 166: Good to know this works for the future - I was never sure if writing to process.env actually wrote an OS environment variable, I thought it might just write it to the map for this process but not persist it to the OS.

stephenhand (approved):
Looks good, the flex plugin action is a nice refactor

e2e-tests/config.ts
robert-bo-davis: YMMV in other use cases. Playwright all runs in one shared shell and multiple processes and this is an "official-ish" work around. I'm not positive how far up the OS shell tree it will propagate in other javascript systems.

PR #1476, opened 2023-06-24
feat: lambda for running playwright

robert-bo-davis:
Primary reviewer:

## Description

Running multiple tests in a lambda using the jest-like syntax runs into some issues. The `--single-threaded`  chromium flag is required (along with a bunch of others) to run chromium in a lambda. Single threaded does some strange things when closing pages/contexts between tests that causes the browser to shut all the way down. It then has problems starting back up.

After banging my head against these issues for a couple of days, I refactored to just run individual tests for each lambda invocation. This workaround is tested working.

In addition to getting the docker container working in a lambda, I added an untested POC for github action based deploymen and did a bit more DRYing of the e2e test system in github actions.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### Verification steps

View run information for the `e2e` service in the infrastructure/serverless section of datadog.


e2e-tests/lambda/handler.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+
```
robert-bo-davis, line 16: Using TS and adding in the build/deploy complexity of transpilation for this very simple lambda handler that is really just passing work off to un-transpiled playwright seemed like overkill.

.github/actions/run-e2e-tests/action.yml
```diff
@@ -0,0 +1,86 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see https://www.gnu.org/licenses/.
+
+name: 'run-e2e-tests'
```
stephenhand, line 15: Nice, this has been on my mental TODO list for a while now

stephenhand (approved):
Great work getting it running in a lambda. I assume you found https://github.com/JupiterOne/playwright-aws-lambda whilst investigating these issues - did it not help in our scenario?

One question, if we are using the --single-process flag are we going to be ok on tests that use multiple browser sessions simultaneously, like the webchat test? The flag still allows multiple sessions right? It just ensures they are all executing in the same process. Just asking in case there were any gotchas you were aware of for that case.

robert-bo-davis:
> I assume you found [JupiterOne/playwright-aws-lambda](https://github.com/JupiterOne/playwright-aws-lambda) whilst investigating these issues - did it not help in our scenario?

It helped in that I stole some of the custom chromium settings from it. It doesn't support the jest style `test()` format we are using. I am striving to have one set of tests with switches/configuration options for the various ways we run them, so I ddidnt think it made sense to move away.

> if we are using the --single-process flag are we going to be ok on tests that use multiple browser sessions simultaneously, like the webchat test?

We are going to be okay. The webchat tests are currently successfully running against as e2e/dev in a lambda alongside the login test. I also wanted to verify this as part of the POC. ;)

PR #1477, opened 2023-06-27
CHI-1599: update ff patch script

stephenhand:
Primary reviewer:

## Description

* Reverts deletion of 'regular' SSM script functions that run with the base privileges of the user whose AWS creds are provided because these are still used.
* Fix patch-feature-flags and allow feature flags to be patched across all accounts in an environment - not so useful now we do it the TF way but might still come in handy for non TF managed accounts or perhaps in an emergency.

### Related Issues
CHI-1599

### Verification steps

Use the patch-feature flags script the 'classic' way, i.e. with Twilio creds in the environment vars, then for the whole staging env

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -155,14 +155,51 @@ will create a new file `my_custom_bot.tf` that contains the definition of the bo
 
 This script allows you to safely update the feature flags specified in a Twilio Account's Service Configuration without affecting other settings
 
+#### Providing Twilio Credentials
+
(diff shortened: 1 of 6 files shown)


robert-bo-davis:
Definitely keep in mind that very soon nobody will have personal IAM tokens that have access to production SSM params. At that point these scripts will need to be run from the devops instance which will need to assume a role with that access.

murilovmachado (approved):
Looks good

PR #1481, opened 2023-06-28
Fix Chat Transfer button

murilovmachado:
## Description
Fix for chat transfers: enable button.

### Verification steps
Check that the transfer button is enable on chat tasks.


mythilytm (approved):
Looks great!

PR #1482, opened 2023-06-28
Fix for issue exposed by new KHP data upload.

stephenhand:
## Description

Some resources have no mainContact at all when imported from the latest data upload, so we've added a small piece of defensive code to prevent the dreaded 'white screen' breaking search results

Also removes feature flag to switch from the placeholder SQL based search and elasticsearch, the placeholder search is being removed so we will use ES all the time from now on


acedeywin (approved):
LGTM

PR #1484, opened 2023-06-28
CHI-1786 Post survey lex bot

robert-bo-davis:

## Description

- New feature: Post-survey lex bot terraform implmentation

### Checklist
- [N/A] New tests added
- [N/A] Feature flags added
- [N/A] Strings are localized
- N/A ] Tested for chat contacts

### Related Issues
Fixes CHI-1786


robert-bo-davis:
@acedeywin This is the new clean branch with (hopefully) all of your actual changes.

In addition to adding the json file, you'll need to update the config for a helpline so the lex_bot module knows to use that json. An example config is [here](https://github.com/techmatters/flex-plugins/pull/1419/files#diff-bf7390badd975b10ebf8a1e6dcdbe291d252c986f114e53f5b78ea4e464be35dR10).

If you modify that same file, you'll need to update it to something like:

```
lex_bots                          = merge(
  jsondecode(file("/app/twilio-iac/helplines/configs/lex/config.json")),
  jsondecode(file("/app/twilio-iac/helplines/configs/lex/post_survey.json")),
)
```

We'll probably also want to rename the `config.json` file from the rough POC to `pre_survey` or something similar so it isn't confusing in the future.

Once that is done, you should be able to test a plan of the lex stage and see your changes in the plan.

acedeywin:
> @acedeywin This is the new clean branch with (hopefully) all of your actual changes.
> 
> In addition to adding the json file, you'll need to update the config for a helpline so the lex_bot module knows to use that json. An example config is [here](https://github.com/techmatters/flex-plugins/pull/1419/files#diff-bf7390badd975b10ebf8a1e6dcdbe291d252c986f114e53f5b78ea4e464be35dR10).
> 
> If you modify that same file, you'll need to update it to something like:
> 
> ```
> lex_bots                          = merge(
>   jsondecode(file("/app/twilio-iac/helplines/configs/lex/config.json")),
>   jsondecode(file("/app/twilio-iac/helplines/configs/lex/post_survey.json")),
> )
> ```
> 
> We'll probably also want to rename the `config.json` file from the rough POC to `pre_survey` or something similar so it isn't confusing in the future.
> 
> Once that is done, you should be able to test a plan of the lex stage and see your changes in the plan.

@robert-bo-davis  - Changes have been implemented.

twilio-iac/helplines/configs/lex/post_survey.json
```diff
@@ -0,0 +1,103 @@
...
+                    "n",
+                    "no",
+                    "nope",
+                    "nah",
+                    "not"
+                  ]
+                }
+              }
+          }
+        },
```
GPaoloni, line 44: Note that:
- This is not the `value_selection_strategy` we want here, it should be like the one I left in the pre survey bot, and
- This is an exact copy of the pre survey `YesNo` slot type. Won't it be better to have something like a template string that abstracts this shared slot type, so we can reuse the exact same clone in both bots?

GPaoloni (commented):
I have not tested the behavior of this bot in Lex console. From code looks mostly correct to me.

twilio-iac/helplines/configs/lex/post_survey.json
acedeywin: @GPaoloni  - I also think creating a reusable `slot_type` for both surveys would be better, but I needed to figure out how to proceed. @robert-bo-davis - any idea on how to create a template string that abstracts this shared slot type in other to keep the `slot_type` DRY? 
@GPaoloni  - about the `value_selection_strategy`, I am not sure what you mean?

robert-bo-davis: Slot types and intents are global in lex v1. They aren't specific to a bot. I would suggest that we refactor the POC json config structure to leverage this instead of just drying them out via template strings.

I suggest we divide configs into 3 sections: slot_types, intents, and bots under a helpline/language combo which will mirror the pages in the AWS console instead of grouping all three sections under a specific bot.

Screen Shot 2023-07-04 at 7 43 02 AM

This would not only dry out the terraform code, it would also reduce the number of lex configs while being extremely flexible for future use.

I've added a PR into this branch that is a POC of the pattern that I suggest. It has only been tested up to the plan stage, but should give you an example of the suggested pattern. #1488

GPaoloni: @acedeywin In [this commit](https://github.com/techmatters/flex-plugins/pull/1419/commits/e15c068cd33a56872cb2374fb0b7c297c80e7429) I changed the `value_selection_strategy` to `TOP_RESOLUTION` as that will result in a response like `Y` to evaluate to `Yes`. Test the current behavior on the bots you added here and you will see that they don't do that (i.e. a `Y` is saved as a `Y`, and not converted to a `Yes`). You can read more about this [here](https://docs.aws.amazon.com/lexv2/latest/APIReference/API_SlotValueSelectionSetting.html).

acedeywin: @GPaoloni - `value_selection_strategy` has been updated to `TOP_RESOLUTION`

robert-bo-davis (commented):
LGTM! 

I think we should clean up the unused lex v2 code and we may need some tweaks once we tie in @murilovmachado 's changes and test them. But this PR is already too big, let's get it merged and then handle those in follow up PRs.

PR #1492, opened 2023-07-04
fix: e2e lambda fail on list lambda error

robert-bo-davis:
Primary reviewer:

## Description

Adds a block so e2e lambda deploy action correctly errors if the lambda list function fails.


stephenhand (approved):
LGTM

PR #1498, opened 2023-07-10
Creates Lex Bot development_as_es_CO_pre_survey

murilovmachado:
## Description
This PR creates the bot `development_as_es_CO_pre_survey`.
Lex doesn't support locale `en-CO`, so `es-419` was used (it's the code for Latin American Spanish).
Also, in order to support `es-419`, it was needed to enable `enable_model_improvements`.

### Verification steps
See Lex Bot created at AWS

### DIFF
--- a/twilio-iac/helplines/as/common.hcl
+++ b/twilio-iac/helplines/as/common.hcl
@@ -19,6 +19,20 @@ locals {
     jsondecode(file("/app/twilio-iac/helplines/configs/lex/en_US/bots/pre_survey.json")),
   )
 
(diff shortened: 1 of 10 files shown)


robert-bo-davis (approved):
LGTM

stephenhand:
I think these made up 'ISO like' culture codes are going to cause use more and more headaches as we try to integrate with 3rd party systems that expect real ones. I put an item for overhauling our translations systems with a multi layered approach similar to what's been prototyped in webchat in our engineering roadmap, that separates translations from text customisations

PR #1504, opened 2023-07-17
CHI-1608: PoC read-only manageServiceConfiguration system

robert-bo-davis:
Primary reviewer:

## Description

In the spirit of trying to keep PRs manageable and fostering early input, this is an initial POC with basic working read only features for the new manageServiceConfiguration system.

Continuing to tack on features to the sprawling `/scripts` system in typescript feels like the wrong direction to go to me. I chose python because it is a much more well adopted "cli DevOps tooling" language than TypeScript. It doesn't suffer from the same "have to transpile locally before running" problems that we keep having in the larger system (even though `tsx` is solving some of these issues).

I recognize that this is a further departure from established team norms and expansion into an addition language. If there are blocking concerns, I'm willing to rewrite this in TypeScript, but we need to make that choice before I get much further.

This does not contain the make system, real updates, or some of the more advanced features yet. It is the scaffolding and basic readonly layout.

There is a sibling PR here: techmatters/infrastructure-config#250

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes CHI-2057

### Verification steps

After techmatters/infrastructure-config#250 is merged and new remote terraform image is deployed,

1. Set your aws creds to env variables
2. Run `make shell` from any helpline directory (ex: `twilio-iac/aselo-as-production`)
3. `cd ../scripts/python_tools` from shell in container
4. `python ./manageServiceConfiguration.py -h` to output usage documenation
5. run various commands to test them


twilio-iac/scripts/python_tools/src/aws/ssm_client.py
```diff
@@ -0,0 +1,150 @@
+import boto3
```
stephenhand, line 1: This is the kind of duplication we'll just need to keep an eye on going forward with a blend of JS / python scripts - obvs 150 lines is nothing to worry about though

stephenhand (approved):
The diff shows it deleting a bunch of Twilio top level configuration items that it will actually be ignoring, but I'm guessing that's a TODO for further PRs

Also I needed to run `python ./manageServiceConfiguration.py -h` to run it so I updated the instructions.

Also I needed to set AWS_DEFAULT_REGION either on the host or on the container, just setting AWS_REGION still failed looking for a region, not sure if this is expected

LGTM though

robert-bo-davis:
> The diff shows it deleting a bunch of Twilio top level configuration items that it will actually be ignoring, but I'm guessing that's a TODO for further PRs

Yup. That is expected at the current stage. To be addressed in the next iteration.

> I needed to run python ./manageServiceConfiguration.py -h to run it so I updated the instructions.

Hmmm. That shouldn't be required since there is a shebang line. Calling it directly has been working for me. Maybe related to some weird build things while I was struggling with techmatters/infrastructure-config#251.

> Also I needed to set AWS_DEFAULT_REGION either on the host or on the container, just setting AWS_REGION still failed looking for a region, not sure if this is expected

Thanks for pointing that out. Will fix in the next iteration.

PR #1507, opened 2023-07-17
CHI-2049 hide vertical line that was introduced in flex-ui upgrade

robert-bo-davis:
## Description

This overrides the new setting in flex-ui that caused a vertical line to appear between the split panes.

### Checklist

### Related Issues
Fixes CHI-2049

### Verification steps

View main flex-plugin page and there is no longer a vertical line through the page.


mythilytm (approved):
Works! LGTM

PR #1520, opened 2023-07-20
Sinekhaya/mt/mapping

sinehome:
Mapping "Nickname" field Webchat.


janorivera (approved):
lgtm! 
I see some changes on the tfpl file, does that mean that we need to do an apply?

sinehome:
@janorivera yes, it's just the change of 'Support Mentor' to 'Professional'. No update to the actual flow(s)

PR #1522, opened 2023-07-21
AS Staging Setup (Terraform) 

sinehome:
Setting up staging environment through Terraform. 

Adding new staging.hcl file and updating the common.hcl file.


twilio-iac/helplines/as/common.hcl
```diff
@@ -50,5 +55,48 @@ locals {
         bots       : local.es_co_bots
       }
     }
+
+    workflows = {
+      master : {
+        friendly_name : "Master Workflow"
+        templatefile : "/app/twilio-iac/helplines/cl/templates/workflows/master.tftpl"
```
GPaoloni, line 62: Why are we using the `/cl` config here? If it is shared, makes sense to move it to a "shared location"?

twilio-iac/helplines/as/common.hcl
```diff
@@ -50,5 +55,48 @@ locals {
...
+    }
+    task_channels = {
+      default : "Default"
+      chat : "Programmable Chat"
+      voice : "Voice"
+      sms : "SMS"
+      video : "Video"
+      email : "Email"
+      survey : "Survey"
+    }
```
GPaoloni, line 100: I feel like this chunk of code here will affect the development and production accounts, since both use `helplines/as/common.hcl`, but at the same their are not provisioned from terraform, so I'm not sure. What do you think would be the best @robert-bo-davis? Moving this into `helplines/as/staging.hcl` or leaving it here?

GPaoloni (commented):
I left some questions and asking Bo's opinion. The changes for staging look good to me, I'm just doubtful on how much code we should share with development account.

twilio-iac/helplines/as/common.hcl
robert-bo-davis: Not sure, the workflow changes have mostly been handled by @janorivera

twilio-iac/helplines/as/common.hcl
robert-bo-davis: Ohhhh.... none of the as helplines have actually been migrated to terragrunt yet. This is all just placeholder. The lex provisioning is the first thing we will be adding in the new TG setup for the AS helplines, so everything unrelated to that is just copypasta that can be ignored until we start migrating them. If you want to clean them up, feel free, but it shouldn't actually impact anything unless it is impacting you're lex plan.

twilio-iac/helplines/as/common.hcl
robert-bo-davis: Ale, please disregard the tag and see note below.

twilio-iac/helplines/as/common.hcl
robert-bo-davis: I think it is fine in common until we migrate another environment. When we do, we'll need to sort out what goes where. I have a feeling that we will start from scratch (a brand new helpline) when we eventually provision development via terraform and that these common values are likely to be common across staging/production and that development will be the outlier so they will probably stay here anyway.

twilio-iac/helplines/as/common.hcl
sinehome: I had copied over the configurations from CL, will update that appropriately. For now we'll use the main template in `twilio-iac/helplines/templates/workflows/master.tftpl` instead of `twilio-iac/helplines/cl/templates/workflows/master.tftpl`

twilio-iac/helplines/as/common.hcl
```diff
@@ -37,6 +37,11 @@ locals {
     helpline           = "Aselo"
     old_dir_prefix     = "aselo-as"
     definition_version = "as-v1"
+
+    default_autopilot_chatbot_enabled = false
+    helpline_language                 = "en-US"
+    contacts_waiting_channels         = ["web"]
```
janorivera, line 43: contacts_waiting_channels can be removed now since this is part of a service configuration

twilio-iac/helplines/as/common.hcl
```diff
@@ -37,6 +37,11 @@ locals {
...
+    enable_post_survey                = true
     lex_bot_languages  = {
```
janorivera, line 45: When deploying I would hold off on deploying the lex bots until my lex PR is merged. This will probably look different.

twilio-iac/helplines/as/common.hcl
```diff
@@ -50,5 +55,37 @@ locals {
...
+      },
+      survey : {
+        friendly_name : "Survey Workflow"
+        templatefile : "/app/twilio-iac/helplines/templates/workflows/survey.tftpl"
+      }
+    } 
+
+    task_queues = {
+      messaging : {
+        "target_workers" = "routing.skills HAS 'Messaging'",
```
janorivera, line 72: Just confirm if we want to work with skills for this account?

twilio-iac/helplines/as/staging.hcl
```diff
@@ -0,0 +1,40 @@
...
+    #Studio flow
+    flow_vars = {
+
+    }
+
+    ui_editable = true
+    #Chatbots
+
+    #Feature flags
+    feature_flags = {
```
janorivera, line 17: I think this is already out of date after Bo's PR got merged.

janorivera (changes requested):
Also, I don't see any Channel definitions, like the web channel and the studio flow that it will use. Is that planned for a second stage?

janorivera:
Things I think are missing:

- Channel definitions
- Update lex bot definitions
- remove feature flags

PR #1523, opened 2023-07-21
[UI] Autocomplete

acedeywin:

- New feature: Resource search autocomplete

**Goal**: To see relevant suggestions for multiple individual search terms based on what users type in the Resource Search bar to find relevant resources more quickly for a specific service user.

### Checklist
- [x] New tests added
- [N/A] Feature flags added
- [x] Strings are localized

### Related Issues
Fixes [CHI-2009](https://tech-matters.atlassian.net/browse/CHI-2009)

### Verification steps

- Go to Resource
- Search for a resource

**Hint**: you can search for terms like `Centres` or `Services`. These would bring out some suggested search. 

[CHI-2009]: https://tech-matters.atlassian.net/browse/CHI-2009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugin-hrm-form/src/services/ResourceService.ts
```diff
@@ -65,3 +72,7 @@ export const searchResources = async (
     results: fromApi.results,
   };
 };
+
+export const suggestResources = async (prefix: string, size: string): Promise<TaxonomyLevelNameCompletion> => {
```
stephenhand, line 76: Naming nit - It's suggesting search terms, not resources

stephenhand (commented):
If you mock out the data down in the `suggestResources` method in the `ResourcesService` rather than up in the component will allow you to build out more of your solution whilst the backend is fixed

For example you don't have anything to take the result of the API call and store it in the redux state, or to then feed it up into the UI - this could all be built with a mocked out API call

plugin-hrm-form/src/services/ResourceService.ts
acedeywin: @stephenhand  - I renamed it to `suggestSearch`

stephenhand (changes requested):
As a basic starting point it works well, however it needs some debouncing, because at the moment the keyboard input is choppy if you are typing quickly, and it spams a LOT of requests to the backend if you are typing fast.

I would start off with a component level debounce similar to the one in AseloMessageInput: https://github.com/techmatters/flex-plugins/blob/master/plugin-hrm-form/src/components/AseloMessaging/AseloMessageInput/AseloMessageInput.tsx

The difference is that this uses a hook form where as you will need a separate state item in the component for the latest version of the text

Reach out if you want more details about this

There also seems to be a large gap under the search box now?

acedeywin:
> https://github.com/techmatters/flex-plugins/blob/master/plugin-hrm-form/src/components/AseloMessaging/AseloMessageInput/AseloMessageInput.tsx

@stephenhand  - all feedback has been implemented.

plugin-hrm-form/src/components/resources/search/SearchResourcesForm.tsx
```diff
@@ -130,6 +139,19 @@ const SearchResourcesForm: React.FC<Props> = ({
     }
   };
 
+  const handleSearch = debounce(
```
stephenhand, line 142: You are debouncing the wrong action for this - we want the `setGeneralSearchTermBoxText` call to be synchronously for each `onChangeSearch`. 

We want the call to `suggestSearch` to be a debounced handler for the `onChangeSearch` event, not an effect that gets run every time the form gets rerendered like it is currently.

acedeywin: @stephenhand  - Yes, you are correct. I have implemented the debounce on action dispatch. Thank you.

stephenhand (approved):
Looks good and works well now!

PR #1528, opened 2023-07-24
ServiceConfig - 3 - CHI-2058 - Keep track of all service config versions and changes in S3

robert-bo-davis:
Primary reviewer:

## Description

This also refactors the config/run system to allow any combo of env/helpline or to run against all helplines (with user validation) and uses a lock file in S3 to keep folks from stepping on each other.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### Verification steps

Don't verify anything here checkout the branch in the lowest level of the PR chain and verify there.


twilio-iac/scripts/python_tools/src/service_configuration/versions.py
```diff
@@ -0,0 +1,83 @@
...
+        self.skip_lock = kwargs.get('skip_lock') or False
+        self.s3_client = S3Client(AWS_ROLE_ARN)
+        self.init_ip()
+        self.init_sha()
+        self.init_s3_paths()
+        self.lock()
+        self.init_version()
+
+    def init_ip(self):
+        self.ip = requests.get('https://api.ipify.org').text
```
GPaoloni, line 31: Only whitelisted IPs will be able to run this? Or what's the purpose of this?

GPaoloni: Ah, it is used to keep an audit track I believe, right?

GPaoloni (approved):
LGTM!

Questions here: 
- The lock works at an account level?
- The lock is determined by the existence or absence of a file, right?
- The IP is tracked so we know who has the lock?

robert-bo-davis:
> The lock works at an account level?
Correct. And at an env level. Since a ServiceConfig object is specific to a helpline/environment code and each ServiceConfig creates a Version which creates a lock for actions that require locking, each helpline env pair is locked individually.

> The lock is determined by the existence or absence of a file, right?
Correct. We are just using S3 keys for locking. This isn't perfect because s3 doesn't offer a perfect "create a key if it doesn't exist already" like redis or dynamodb (which is why terraform uses dyanmodb), but it is good enough for our purposes. The chances of us having two users actively working on the same service config so close in time that they run a locking command within microseconds of each other is relatively slim.

> The IP is tracked so we know who has the lock?

Yup. Just meant as a away to know if somebody else has the lock or if you do. Was already using it for the audit trail, so went ahead and stuck it in the lock file as well.

PR #1535, opened 2023-07-26
ServiceConfig - 4 - CHI-2058 working service config update, cleanup, and fixes

robert-bo-davis:
Primary reviewer:

## Description

This adds the actual call to update a service config.
This includes quite a bit of productionization, some fixes, and cleanup.

Some fixes:
env based template fields
adds the ability to remove missing attributes correctly
deals with complex values in syncer correctly
cleans up locks properly on exception or ctrl-c
Allows sync and service-config actions to run agains all helplines, a specific env (service-config only), a specific helpline in all envs, or a specific helpline (service-config only)

Again, this PR looks a lot larger than it really is because it includes config updates for every helpline. You can safely ignore all of the changes in the `twilio-iac/helplines` and just look at the output from running `make service-config-plan` from the `./twilio-iac` directory to make sure the change plans all look reasonable based on the migrated configuration. 

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### Verification steps

run `make service-config-plan` from the `./twilio-iac` directory


twilio-iac/scripts/python_tools/src/service_configuration/service_configuration.py
```diff
@@ -164,14 +210,30 @@ def init_new_state(self):
...
+                set_nested_key(self.new_state, field, local_value)
+
+        self.init_ssm_fields()
+        self.init_template_fields()
+        for key, value in self.template_config.items():
+            new_value = get_nested_key(self.new_state, key)
+            if new_value and new_value != value:
+                continue
+
+            set_nested_key(self.new_state, key, value)
```
GPaoloni, line 226: Why skipping the case where the new value differs here? :thinking:

twilio-iac/scripts/python_tools/src/service_configuration/service_configuration.py
```diff
@@ -196,16 +258,37 @@ def init_version(self):
...
             skip_lock=self.skip_lock,
         )
 
     def get_config_path(self, type: str) -> str:
         return self.local_configs[type].get('path')
 
+    def apply(self):
+        if not self.version:
+            raise Exception(
+                'Cannot apply changes without a version. Something went wrong.')
```
GPaoloni, line 274: Why would this happen and how should we proceed for recovery? No need for a big answer, I just want to have a sense at what to look at.

twilio-iac/scripts/python_tools/src/service_configuration/service_configuration.py
```diff
@@ -196,16 +258,37 @@ def init_version(self):
...
+
+        self._twilio_client.update_flex_configuration(new_state)
+
+        # add a new version for the new state
+        Version(
+            environment=self.environment,
+            helpline_code=self.helpline_code,
+            state=self.new_state,
+            skip_lock=True,
+        )
```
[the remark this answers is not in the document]
GPaoloni: NVM, I just re read the code for it, yes it does.

GPaoloni (approved):
LGTM!

twilio-iac/scripts/python_tools/src/service_configuration/service_configuration.py
```diff
@@ -31,9 +31,32 @@
...
+OVERRIDE_FIELDS = [
+    'attributes',
+    'ui_attributes',
+]
+
+REGION_URL_POSTFIX_MAP = {
+    'us-east-1': '',
+    'eu-west-1': '-eu',
+    'ca-central-1': '-ca',
 }
```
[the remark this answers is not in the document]
robert-bo-davis: we should probably just use the full region instead of the weird prefix in our urls so we don't have to do these kinds of weird lookups. It isn't like these are user facing URLs, there is no great reason not to use the full url safe aws region instead of abbreviating it and forcing lookups other places.

But yes, good call! I updated techmatters/infrastructure-config#251 with the instructions.

twilio-iac/scripts/python_tools/src/service_configuration/service_configuration.py
robert-bo-davis: This is a developer facing exception to try to catch a developer change that would make the lock fail to process.

twilio-iac/scripts/python_tools/src/service_configuration/config.py
```diff
@@ -330,13 +323,21 @@ def validate_json(self):
...
+            print(f'ERROR: The {self.action} action does not accept an environment argument.')
+            exit(1)
 
-    def validate_environment_or_helpline(self):
+    def validate_all_helplines(self):
         if not (self.environment or self.helpline_code):
             confirm = input(
                 'No environment or helpline code provided. '
-                'Do you want to run this command for all helplines in environments? (y/N): '
+                'Do you want to run this command for ALL helplines in ALL environments? (y/N): '
```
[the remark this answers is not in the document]
robert-bo-davis: Not THAT scary. I'll reply in more detail in one of your other comments.

twilio-iac/scripts/python_tools/src/service_configuration/service_configuration.py
robert-bo-davis: Good catch! This is a bug. Updating to this in a later PR:

```
        for key, value in self.template_config.items():
            local_value = get_nested_key(self.local_state, key)
            # We want to allow the user to override the template value with a
            # local value.
            if local_value:
                continue

            set_nested_key(self.new_state, key, value)
```

PR #1536, opened 2023-07-26
ServiceConfig - 5 - CHI-2060 Documentation and auth fixes

robert-bo-davis:
Primary reviewer:

## Description

This adds documentation for running the system and fixes a regression related to authenticating the show action with `TWILIO_ACCOUNT_SID` and `TWILIO_AUTH_TOKEN` env vars.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### Verification steps

run `make service-config-show-remote` with twilio env vars set on your terminal


twilio-iac/docs/service_configuration.md
```diff
@@ -0,0 +1,56 @@
...
+
+See requirements in [twilio-iac/README.md](../README.md)
+
+## Usage
+
+All make commands are run from the `twilio-iac` directory.
+
+### Targeting a Helpline and/or An Environment via Environment Variables
+
+These env vars are optional. If they are not set, the commands will run against ALL helplines in ALL environemnts.
```
GPaoloni, line 13: I think this is too error prone, even if we have a "type yes" validation, this being the default scares me. Can we make it erropr if no env vars are passed, and make this to require an explicit parameter like `--all` (or whatever).

twilio-iac/docs/service_configuration.md
```diff
@@ -0,0 +1,56 @@
...
+These commands are used to manage the service configuration for a helpline.
+
+`make service-config-plan` - outputs a plan for the service configuration update
+`make service-config-apply` - applies the service configuration update
+
+### Remote Sync Commands
+
+These commands are used to update the local json config files based on the remote service configuration to ensure that the local config is up to date when changes have been made outside of the service configuration manager.
+
+*Note: these commands do not accept an HL_ENV argument as they must load the configurations for all environments in order to compare them.*
```
GPaoloni, line 30: Is this something we could change? I feel like having to fetch all configs is unnecessary burden if we know exactly which account changed? But maybe there's a reason why we can't

GPaoloni (approved):
Thanks for the clear documentation!

twilio-iac/docs/service_configuration.md
robert-bo-davis: It won't actually hurt anything if a user types yes. It will just load data for every single helpline, which doesn't take *that* long. Any apply action will still request individual helpline/env approval with a plan for each change. So there is a *lot* more validation that will happen before a user could actually break anything.

twilio-iac/docs/service_configuration.md
robert-bo-davis: Since the sync system needs to compare the plan for changes from every environment to create a common.json, it won't work with a call that specifies a single helpline/env. If you specify a helpline, it will only load all envs for THAT helpline. The comment isn't totally clear, sorry.

PR #1537, opened 2023-07-26
ServiceConfig - 6 - CHI-2060 terraform/terragrunt updates to move to new service configuration manager

robert-bo-davis:
Primary reviewer:

## Description

This sets up the new service configuration manager to run as a post apply hook in the configure stage for terragrunt heplines.

It removes the old serviceConfiguration tool calls from terraform and all of the related extraneous variables.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


janorivera:
ok something odd... for CL I have the staging.json file as
{
    "attributes": {
        "resources_base_url": ""
    }
}
The attribute right now look:
"resources_base_url": "https://hrm-staging.tl.techmatters.org/",
But when I try running a plan it returns:
Initializing service configuration for cl

------------------
Environment: staging
Helpline Code: cl
Account SID: AC6ca34b61e7bf2d7cf8b8ca24e7efe65f

Plan:
No changes.
Cleaning up...
(edited)

twilio-iac/scripts/python_tools/src/service_configuration/manager.py
```diff
@@ -82,7 +81,13 @@ def print_plan(plan: DeepDiff):
...
             elif diff_type == 'values_changed':
                 output.append(f'Update {path} from {change.t1} to {change.t2}')
-
+            elif diff_type == 'iterable_item_added':
+                output.append(f'Added item to {path} with value {change.t2}')
+            elif diff_type == 'iterable_item_removed':
+                output.append(
+                    f'Removed item from {path} with value {change.t1}')
+            else:
+                output.append(f'Unknown change: {change}')
```
GPaoloni, line 90: Do we want to handle this way, or maybe loudly erroring is better? What's the consequence of having an unknown change here?

twilio-iac/thailand-production/main.tf
```diff
@@ -133,7 +91,7 @@ module "twilioChannel" {
   pre_survey_bot_sid       = twilio_autopilot_assistants_v1.pre_survey_bot_TH.sid
   target_task_name         = local.target_task_name
   channel_name             = each.key
-  janitor_enabled          = !local.enable_post_survey
+  janitor_enabled          = true
```
GPaoloni, line 94: Is this correct :thinking:?

GPaoloni (approved):

twilio-iac/scripts/python_tools/src/service_configuration/manager.py
robert-bo-davis: The plan will show a weird change in a weird format that is still hopefully useful to the end user. It won't actually break anything if it is applied (the user could use `make service-config-show` to look at the generated json if they want to verify that it is valid.

My thinking is that if we error and run into an edge case that the user actually wants to apply, then they have to escalate to a developer to fix the problem. If we just inform them that a weird unhandled part of the plan exists and let them make an educated decision about whether they should apply the plan or not, then they can work around the issue without blocking their work.

twilio-iac/thailand-production/main.tf
robert-bo-davis: I manually looked at the value of `local.enable_post_survey` before removing that setting and then set the `janitor_enabled` attribute to the inverse, so yes, it should be correct.

PR #1538, opened 2023-07-27
 CHI-2001: Exclude siteId from hours displays

stephenhand:
## Description

* Filter out the 'siteId' property from operating hours sets in resources for display to the user
* Reorder content on resource view page in line with mocks

### Related Issues
CHI-2001

### Verification steps

Visually evaluate resource view page


acedeywin (approved):
LGTM

PR #1543, opened 2023-07-28
QA feedback for styling on resources pages

stephenhand:
Primary reviewer:

## Description

* Fixed left alignment on resource search result item and details
* Updated 'Copy ID' button tooltip
* Ensured 'Not Listed' appears for empty Hours of Operation
* Increased padding between filter checkboxes and their labels
* Lined up the 'back' buttons for the resource view and search results

### Checklist

### Related Issues
CHI-2001

### Verification steps

See ticket

### DIFF
--- a/plugin-hrm-form/src/components/resources/ResourceIdCopyButton/index.tsx
+++ b/plugin-hrm-form/src/components/resources/ResourceIdCopyButton/index.tsx
@@ -21,6 +21,7 @@ import { useState } from 'react';
 import { Template } from '@twilio/flex-ui';
 
 import { Button } from './styles';
+import { getTemplateStrings } from '../../../hrmConfig';
 
 type OwnProps = {
   resourceId: string;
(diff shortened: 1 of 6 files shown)


acedeywin (commented):
> Lined up the 'back' buttons for the resource view and search results

Does the above mean the "Search Result" should align with the back button?  If yes, then there is still a little bit of margin to add for these to align.

Screenshot 2023-07-28 at 11 25 46 AM

stephenhand:
> > Lined up the 'back' buttons for the resource view and search results
> 
> Does the above mean the "Search Result" should align with the back button? If yes, then there is still a little bit of margin to add for these to align.
> 
> Screenshot 2023-07-28 at 11 25 46 AM

It means that the back button on the search results page and the back button on the resource details page should be in the same place

acedeywin (approved):
LGTM

PR #1545, opened 2023-07-29
CHI-2062: webchat test updates to prevent writing data to hrm

robert-bo-davis:
Primary reviewer:

## Description

This expands the webchat scripts to be customizable per helpline/environment. This system will likely be a bit more of a cascading config before we are done with it. 

It configures the flexchat system to work with the default twilio chat system our our custom implementation.

It also fixes the webchat cleanup to cleanup all tasks with the e2eTestMode flag set to true instead of specifying info about the task that could change between helplines.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### DIFF
--- a/e2e-tests/chatScripts.ts
+++ b/e2e-tests/chatScripts.ts
@@ -1,0 +1,39 @@
+import {
+  botStatement,
+  callerStatement,
+  counselorAutoStatement,
+  counselorStatement,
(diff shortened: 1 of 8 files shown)


mythilytm (approved):
Tested and LGTM

PR #1547, opened 2023-08-01
Fix Conference Button

murilovmachado:
## Description
This PR:
- Updates `@twilio/flex-ui` to version `2.3.3`
- Makes input number dialog appear when clicking Conference button

image

Context:
- Twilio's recently updated `@twilio/flex-ui` to version `2.3.3`
- On our package-lock file, we had `@twilio/flex-ui` on version `2.3.1`
- And we have a script that was updating the deployed version to `~2.3.1`, which bumps to `2.3.3`
- That's acceptable because patch changes aren't supposed to cause any kind of breaking changes
- But this time it actually had changes that affected Conferencing 
  - The actions buttons were previously on`div.Twilio-TaskDetailsPanel-Buttons`
  - And now they are placed in an inner div `div.Twilio-TaskDetailsPanel-Buttons > div.Twilio-CallCanvasActions` 
  - This inner div has the CSS property `overflow-x: hidden`, that was causing the dialog to be hidden
- This PR explicitly updates `@twilio/flex-ui` to version `2.3.3`
- And makes the necessary styles changes so the input number dialog is visible again

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-2093

### Verification steps
1. Call Aselo
2. Answer call
3. Click conferencing button
4. See the the number input dialog shows up

### DIFF
--- a/plugin-hrm-form/package-lock.json
+++ b/plugin-hrm-form/package-lock.json
@@ -10880,9 +10880,9 @@
       }
     },
     "node_modules/@dnd-kit/accessibility/node_modules/tslib": {
-      "version": "2.6.0",
-      "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.6.0.tgz",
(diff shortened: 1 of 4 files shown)


stephenhand (approved):
LGTM!

PR #1548, opened 2023-08-01
externalRecordings - 1 - CHI-2097 external recordings stage

robert-bo-davis:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### Verification steps

run `make plan HL=as HL_ENV=development` from `twilio-iac/stages/external-recordings`


GPaoloni (approved):
LGTM!

Can you write down (if you didn't further down the PR chain) why is it better to have this in it's own stage? Is it so we can use this in non-terraformed accounts?

PR #1550, opened 2023-08-01
Gian chi 2046

GPaoloni:
## Description
This PR fixes the bug where a webchat chat won't load after it was ended by the service user. See the ticket linked below for more information.

It looks like the cause of this issue is one of the event listeners subscribed to the `participantLeft` event, fired on a conversation (chat channel) when the task is wrapped (due to the Chat Channel Orchestrations). 
A different approach to go about this would be to stop removing the workers from the channel on the wrapup event.

Commit https://github.com/techmatters/flex-plugins/pull/1550/commits/be3a172ca1c5711d75c62edba88fd7cbab2637ce in this branch is just showing that even if we don't add any code to the Flex plugin, this issue is still occurring under the same conditions.
You can also reproduce this bug by, 
- While counselor in the active chat window, end the chat from webchat, then change the counselor ciew (e.g. to an offline contact) and then go back to the webchat chat. 
- While using the development local instance pointing to Aselo Development account, comment out lines 78-83 of `src/utils/setUpTaskRouterListeners.ts`, which means this bug is reproducible in Aselo Development.

In both cases, you'll notice the messages list is gone. This makes me wonder if this isn't actually intended Flex behavior. I tried to setup a clean Flex account with webchat but that ended up being more effort than it deserved. I'm happy to budget some time and do so, just to confirm that this occurs in an entirely blank Flex instance, if that's considered to be worthy.

A support ticket has been open with Twilio to ask about this behavior.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2046)
- [x] New tests added
- [n/a ] Feature flags added
- [n/a ] Strings are localized

### Verification steps
Follow the repro steps in the above ticket and confirm the bug not reproducible anymore (this branch needs to be deployed to the target account where this is intended to be tested, or point to one of those accounts from a local Flex instance).

### DIFF
--- a/plugin-hrm-form/src/utils/setUpTaskRouterListeners.ts
+++ b/plugin-hrm-form/src/utils/setUpTaskRouterListeners.ts
@@ -35,13 +35,19 @@ const removeConversationListeners = (conversation: Conversation) => {
   conversation.eventNames().forEach(safelyRemoveListeners);
 };
 
-const removeConversationListenersForTask = (task: ITask) => {
+const runCallbackIfChatBasedTask = (callback: (conversation: Conversation) => void) => (task: ITask) => {
   if (TaskHelper.isChatBasedTask(task)) {
     const conversationState = StateHelper.getConversationStateForTask(task);
 
-    removeConversationListeners(conversationState.source);
+    // eslint-disable-next-line callback-return
+    callback(conversationState.source);
   }
 };
+
+const removeParticipantLeftConversationListeners = runCallbackIfChatBasedTask((conversation: Conversation) => {
+  conversation.removeAllListeners('participantLeft');
+});
+const removeConversationListenersForTask = runCallbackIfChatBasedTask(removeConversationListeners);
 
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/utils/setUpTaskRouterListeners.ts
```diff
@@ -71,6 +77,9 @@ export const setTaskWrapupEventListeners = (featureFlags: FeatureFlags) => {
    */
   if (featureFlags.enable_post_survey) {
```
stephenhand, line 78: Might it make this code sightly simpler if you moved this feature flag check inside the handler? That you could have a single handler with the `runCallbackIfChatBasedTask` logic inlined rather than having to wrap 2 callbacks - I think anyway

Just a suggestion though, not blocking.

stephenhand (approved):
LGTM, 1 non blocker comment

PR #1551, opened 2023-08-02
CHI-2043: Ensure search term is updated in the state when a search is initially submitted

stephenhand:
## Description

A couple of cases were occurring where the contents of the search box were going out of sync , so I added a couple of fixes to ensure the 2 were 'resynced' more reliably:

* A new effect will reset the contents of the box to the redux value when the redux value changes.
* Things that change the box contents other than typing / pasting (using the 'clear' cross, clear button or selecting an autocomplete suggestion) now update the redux value rather than the box text state - the box text is then updated via the new effect above
* Submitting a search updates the current search parameter in redux to the value submitted as a search. This ensures the 2 are in sync after the user submits a search by hitting 'Enter' (this doesn't blur the text box like clicking the 'Search' button does, so they weren't being synced in this case previously)

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-1989

### Verification steps

* Use the search UI in lots of different ways, setting filters and search terms in different orders, using the mouse & keyboard, the clear button on the search input box, the clear form button at the bottom, autocomplete suggestions etc.
* When searches are submitted, ensure these 3 things are correct - The search results look to be for the last submitted search (rather than a previous one), the search result description matches the submitted search, and the search form is in the expected search when you return to it from the results
* Try several searches with and without resetting the search form, try paginating etc. and ensure the 3 things outlined above are preserved

### DIFF
--- a/plugin-hrm-form/src/___tests__/states/resources/search.test.ts
+++ b/plugin-hrm-form/src/___tests__/states/resources/search.test.ts
@@ -88,7 +88,15 @@ describe('actions', () => {
       const startingState = getState();
       dispatch(searchResourceAsyncAction({ generalSearchTerm: 'hello', pageSize: 42 }, 1337, true));
       const state = getState();
-      expect(state).toStrictEqual({ ...startingState, status: ResourceSearchStatus.ResultPending, results: [] });
+      expect(state).toStrictEqual({
+        ...startingState,
+        parameters: {
+          ...startingState.parameters,
+          generalSearchTerm: 'hello',
+        },
+        status: ResourceSearchStatus.ResultPending,
+        results: [],
+      });
     });
 
     test("'newSearch' flag not set - dispatches pending action that sets status to ResultPending but leaves results array as is", async () => {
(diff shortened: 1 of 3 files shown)


murilovmachado:
@stephenhand I think I got a bug, possibly related to `Location` not being cleared.

Steps to reproduce:
1. Fill search term "a"
2. Search results: MANY RESULTS
3. Go back to the search page and add province and city
4. Search results: Few results
5. Go back to the search page and clear the province (just leave the search term "a")
6. Search results: still few results (should be a lot more)

P.S.: On the search result pages, the label "Showing x results for..." is correct. But the results are not.

image

murilovmachado (approved):
I'm pre-approving this, so feel free to merge when you have that mentioned issue fixed. 🚀 

One thing that I've noticed, but out of the scope of this PR, is that the search form is handling 99% of its state through redux, but for the **search term input** only it handles the state locally with `useState`(in addition to redux)

```
const [generalSearchTermBoxText, setGeneralSearchTermBoxText] = React.useState(generalSearchTerm);
```

My gut feeling thinks we could live we the redux state only, but I really didn't stop to see why we handle this one with `useState`.

stephenhand:
> I'm pre-approving this, so feel free to merge when you have that mentioned issue fixed. 🚀
> 
> One thing that I've noticed, but out of the scope of this PR, is that the search form is handling 99% of its state through redux, but for the **search term input** only it handles the state locally with `useState`(in addition to redux)
> 
> ```
> const [generalSearchTermBoxText, setGeneralSearchTermBoxText] = React.useState(generalSearchTerm);
> ```
> 
> My gut feeling thinks we could live we the redux state only, but I really didn't stop to see why we handle this one with `useState`.

This is an optimisation - we do store the search term back to redux, but not on every keypress, because a redux cycle every keypress makes it noticeably unresponsive - so we update component state every keypress, then sync that back to redux at key points, such as bluring the text input or submitting a search

PR #1568, opened 2023-08-08
chore: update prettierrc and eslint so IDE format on save works

robert-bo-davis:
## Description

This follows the recommended fixes [here](https://github.com/twilio-labs/twilio-style/tree/main/packages/eslint-config-twilio#prettier) to enable IDE auto-formatting to pick up prettier changes from the underlying twilio eslint plugins.

### Verification steps

Make a change (remove a semicolon) and depend on the eslint prettier plugin to fix it on save via the prettier/eslint VS code plugin. Make sure the semicolon is fixed and nothing else in the file changes.

Verify that, if format on save was already working as expected in your IDE, this doesn't break it.

### DIFF
--- a/plugin-hrm-form/.eslintrc.json
+++ b/plugin-hrm-form/.eslintrc.json
@@ -1,15 +1,12 @@
 {
-  "extends": [
-    "twilio-react",
-    "plugin:sonarjs/recommended"
-  ],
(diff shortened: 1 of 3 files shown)


stephenhand (approved):
LGTM

PR #1569, opened 2023-08-08
CHI-1989: Add 'Languages Serviced' and 'Coverage' to the Resource Details

stephenhand:
Primary reviewer:

## Description

Address a couple of issues raised as part of the UI tidy up but needed backend mapping changes

* Coverage is now shown at resource level & per site
* Coverage descriptions should be compatible with the current coverage data and the planned upcoming format
* 'Languages Serviced' is also shown for the resource

### Checklist
- N/A New tests added
- N/A Feature flags added
- [X] Strings are localized

### Related Issues
CHI-1989

### Verification steps

View various resources and look for the 'Languages Serviced' and 'Coverage' entry on the resource. I don't think the old format Coverage puts coverage on the site level, so until we have that data, coverage on each site will always show as 'Not Listed'


murilovmachado (approved):
Looks good to me!

PR #1576, opened 2023-08-09
Alejandro/lex migration/zm_staging

janorivera:
Primary reviewer:

## Description

ZM staging migration.  This is an old account and it doesn't have any old TF repo, so changes on Twilio were done manually.
There is an update as well to share slot_types, bots and intents on cases were the language is different but everything else is the same.

### Checklist
- [x] Successfully applied
- [x] Tested a normal flow on Webchat and Whatsapp

### Related Issues
Fixes #....


robert-bo-davis (commented):
LGTM! Not blocking merge, but could you add a `twilio-iac/stages/lex/README.md` that explains this cascading config load behavior so folks don't have to figure it out from context?

robert-bo-davis (approved):

PR #1579, opened 2023-08-10
chore: console log save contact form errors

robert-bo-davis:
## Description

This adds some console errors on hrm form save failure that we can search for in datadog.


mythilytm (approved):
Thanks so much for this!

PR #1580, opened 2023-08-11
Alejandro/lex migration/co staging

janorivera:
Primary reviewer:

## Description

CO staging Lex migration.

1. Pre survey is implemented but without lex bots, the modifications were done on the Studio flow
2. Post survey was created but not yet enabled, since this is not actually enabled on CO Prod. This needs more work to get it running but is not worth doing now together with the migration.
3. Added a short readme file under stages/lex

The TF is a bit old and for that reason is not very pretty, I expect this to change when migrate these accounts to the current TG structure.

- [x] TF applied
- [x] Pre survey tested on webchat

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


robert-bo-davis (approved):
LGTM

PR #1589, opened 2023-08-16
CHI-1989: Allow multiple Target Populations to be displayed on resource details

stephenhand:
They are now shown as a CSV


stephenhand:
Simple change, merging without approval

PR #1592, opened 2023-08-16
CHI-1989: Remove support for legacy coverage format in resources

stephenhand:
## Description

Also change validation criteria for determining if a resource / site has displayable coverage


stephenhand:
Simple change, will not wait for approvals

PR #1603, opened 2023-08-21
CHI-2146 - getSignedS3Url path refactor

robert-bo-davis:
Primary reviewer:

## Description

This updates to the new path for signed s3 urls and sets up overridable lambdaUrl separate from hrmUrl for local testing.

### Verification steps

Run tests


robert-bo-davis:
@GPaoloni when you get a moment, could you give this a quick review, please.

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -397,14 +398,18 @@ const getInsightsUpdateFunctionsForConfig = (
 };
 
 const generateUrlProviderBlock = (externalRecordingInfo: ExternalRecordingInfoSuccess, contact: HrmServiceContact) => {
-  const { accountSid, hrmBaseUrl } = getHrmConfig();
+  const { lambdaBaseUrl } = getHrmConfig();
 
   const { bucket, key } = externalRecordingInfo;
+  const url_provider = generateUrl(
```
robert-bo-davis, line 404: @mythilytm you'll need to abstract this url generator away somewhere so you can reuse it in the contact external media player system.

PR #1606, opened 2023-08-22
Updated locations with new data from KHP

stephenhand:
## Description

Update cities list for dropdown


stephenhand:
Data update only - merging without approval

PR #1607, opened 2023-08-23
CHI-1989: Trim & remove double quotes from suggestions

stephenhand:
## Description

So they still work if the user puts the quotes in themselves - this has tripped up KHP QA so would probably trip up counselors too.

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A] Strings are localized

### Related Issues
CHI-1989

### Verification steps

Add leading spaces, trailing spaces and double quotes to search terms - the suggestions & highlights should appear as if those characters were absent

Regression test that the terms are still suggested correctly without any of these


murilovmachado (approved):
🚀

PR #1613, opened 2023-08-25
CHI-2166: Get rid of zero appearing when there are no referrals in the contact view.

stephenhand:
## Description

* Checking an array length without casting it to a boolean or null / undefined to determine if a react element should be included results in a '0' being written in the negative case

### Related Issues
CHI-2166

### Verification steps

See ticket

@GPaoloni


stephenhand:
Trivial change, merging without approval

PR #1627, opened 2023-09-01
Exclude chat orchestration under flag condition

GPaoloni:
This PR is related to https://github.com/techmatters/serverless/pull/507

## Description
This PR moces the "exclude DeactivateConversation Orchestration" logic to be executed under two circumstances: 
- If `enable_post_survey` feature flag is enabled, preserving the existing logic.
- If `backend_handled_chat_janitor`is enabled, which will be turned on for those accounts where the post survey is not enabled but we will want to move this logic to be handled by the backend services. This is a change in line with the fix of [this bug](https://tech-matters.atlassian.net/browse/CHI-2202), and is intended to be temporarily enabled until all accounts have been updated with this PR and the corresponding Serverless code, at which point we'll want to remove them - this change should be the default way to handle the orchestrations moving forward at that point.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2202)
- [x] New tests added
- [x] Feature flags added
- [n/a] Strings are localized

### Verification steps
See related PR.

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.tsx
+++ b/plugin-hrm-form/src/HrmFormPlugin.tsx
@@ -143,7 +143,7 @@ const setUpActions = (
   setupObject: ReturnType<typeof getHrmConfig>,
   getMessage: (key: string) => (language: string) => Promise<string>,
 ) => {
-  ActionFunctions.setUpPostSurvey(featureFlags);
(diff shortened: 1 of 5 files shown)


stephenhand (approved):
LGTM

PR #1630, opened 2023-09-04
CHI-2169: Fire an internal action to clear state after accepting a transfer

stephenhand:
## Description

In order to work around the issue where the flex state can go out of sync if the conversation is already loaded prior to it being transferred, we are explicitly clearing the conversation being transferred from the flex redux state when accepting a text transfer

This forces it to be reloaded from the source and the correct state is then loaded into the UI

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-2169

### Verification steps

See ticket

### DIFF
--- a/plugin-hrm-form/.eslintrc.json
+++ b/plugin-hrm-form/.eslintrc.json
@@ -11,6 +11,7 @@
   "rules": {
     "no-console": "off",
     "no-warning-comments": "off",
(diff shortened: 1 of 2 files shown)


murilovmachado (approved):
Great that you've put a comment for the issue and fix!

PR #1633, opened 2023-09-05
RATI(IN) 'add' Permissions

sinehome:
Update RATI (IN)'s permissions config to allow other users to add to the case details, not just the creator and/or supervisor.


janorivera (approved):
lgtm! just remember that whenever you change permissions you have to change it here and also in the HRM repo. 
For the HRM repo you need to create a PR and add Gian as a reviewer. I can't approve PRs in that repo.

PR #1635, opened 2023-09-06
Replace use of SearchAPIContact for HrmServiceContact

murilovmachado:
Primary reviewer:

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


plugin-hrm-form/src/types/types.ts
```diff
@@ -134,24 +134,28 @@ export type ContactRawJson = {
...
 
 export type HrmServiceContact = {
   id: string;
+  accountSid: string;
   twilioWorkerId: string;
   number: string;
   conversationDuration: number;
   csamReports: CSAMReportEntry[];
   referrals?: ResourceReferral[];
+  conversationMedia?: ConversationMedia[];
```
stephenhand, line 151: Why are we duplicating the `conversationMedia` property for `rawJson` and the top level? I assume we still need it on `rawJson` for POSTing / PATCHing to HRM? What about contacts connected to cases, I assume those would still be coming through on the rawJson? Or is HRM now writing / reading them in both places?

Happy to move them up to the top level, makes more sense given the data restructure on the backend, but I just want to be sure we've covered the compatibility scenarios. Also, we should only have the property in one place in the general purpose type - all the compatibility should be handled in the service layer, then there's no ambiguity about which property you should be working with in the rest of the app

plugin-hrm-form/src/components/case/ViewContact.tsx
```diff
@@ -44,7 +44,7 @@ const mapStateToProps = (state: RootState, { task, contactId }: OwnProps) => {
 };
 
 const mapDispatchToProps = {
-  loadRawContactIntoState: loadRawContact,
+  loadRawContactIntoState: loadContact,
```
stephenhand, line 47: Could this just be removed? Doesn't look to be used

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -156,7 +206,7 @@ export const searchResultToContactForm = (def: FormDefinition, information: Reco
   );
 };
 
-export function transformCategories(
+function transformCategoriesOld(
```
stephenhand, line 209: Are we still using this?

stephenhand (commented):
One Q about `conversationMedia`, just minor stuff otherwise

plugin-hrm-form/src/components/case/ViewContact.tsx
murilovmachado: Removing

plugin-hrm-form/src/services/ContactService.ts
murilovmachado: Removing

plugin-hrm-form/src/types/types.ts
murilovmachado: HRM is already moving `conversationMedia` from `rawJson` to the top level. And I think it already sending this information duplicated. The idea is that when the front-end is ready, they can just stop sending from `rawJson` and keep the top level only. 

At least this is what it seams when looking at HRM code.

stephenhand: Can we just remove the `conversationMedia` from `rawJson` then? 

If we ensure HRM gets deployed before this version of Flex (and we normally deploy HRM before Flex in prod deploys anyway) we should just be able to forget about the old prop here.

murilovmachado: It makes sense to remove it, yes. As soon as I fix one last failing E2E tests I'll go back to this and check if it's really safe to remove it from `rawJson`.

murilovmachado: Ok, I think I got everything working and we can't remove `conversationMedia` from `rawJson` just yet.

In summary, when saving a contact, the backend expects `conversationMedia` to come from either:
- top level; or
- `rawJson`

But the gotcha is that when reading from `rawJson`, the backend does an additional [transformation](https://github.com/techmatters/hrm/blob/6e1a2012556c8d864f3c94ddab3ebab896e3198e/hrm-domain/hrm-service/src/contact/contact.ts#L206):
```
const getNewContactPayload = (
  newContact: CreateContactPayload,
): {
  newContactPayload: NewContactRecord;
  csamReportsPayload?: CSAMReport[];
  referralsPayload?: ReferralWithoutContactId[];
  conversationMediaPayload?: NewConversationMedia[];
} => {
  if (usesFormProperty(newContact)) {
    const {
      csamReports: csamReportsPayload,
      referrals: referralsPayload,
      conversationMedia,
      form,
      ...rest
    } = newContact;

    const conversationMediaPayload = conversationMedia
      ? conversationMedia
      : form.conversationMedia
      ? form.conversationMedia.map(intoNewConversationMedia) // <===== Transformation here
      : []; // prioritize new format, but allow legacy Flex clients to send conversationMedia

    return {
      newContactPayload: {
        ...rest,
        rawJson: newContact.form,
      },
      csamReportsPayload,
      referralsPayload,
      conversationMediaPayload,
    };
  }
```

This transformation is not done if the `conversationMedia` comes from the top level. That means the front-end would need to send data already transformed, but currently it doesn't. So I guess this is still on-going work from the other pod.

Not sure if I could summarized everything clearly, but TLDR: for now, we'll `saveContact` passing `conversationMedia` on `rawJson`, but when `searchContact`, we'll use what comes from the root level.

stephenhand: Ok - I think I understand.

But can we isolate this incompatibility to the `ContactService.ts`, when we use any of the APIs that still provide the version we want on the `rawJson` attribute, can we move it up to the top level before returning the object to the caller, and when we need to send the array on `rawJson`, copy it from the top level? And remove the `conversationMedia` array from `rawJson` on the shared `HrmServiceContact` type?

We keep the majority of the code working with contacts clear that way, and when HRM is sorted out, it will be easy to align Flex because we only need to change 1 file

murilovmachado: @stephenhand done, `conversationMedia` is no longer duplicated in our type, it's now a top-level attribute only. 

I have an adapter function in place called just before using `saveContact` so that we send data in the format `saveContact` expects.

stephenhand (approved):

PR #1636, opened 2023-09-07
Chi 1436 refactor case saving logic redux promise

acedeywin:

## Description

**Feature Update**: Refactor case saving logic to use redux promise

**Goal**: To refactor the current saving logic to be triggered in the redux layer off the back of a save action, not in the component.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [x] Strings are localized

### Related Issues
Fixes [CHI-1436](https://tech-matters.atlassian.net/browse/CHI-1436)

### Verification steps

- Go to Case List
- Select a case
- Edit case summary
- Verify that that update was successful
- Go to Offline Contacts
- Fill out the forms
- Click "Add Contact to New Case"
- Verify that contact was successfully added to new case

[CHI-1436]: https://tech-matters.atlassian.net/browse/CHI-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugin-hrm-form/src/states/case/saveCase.ts
```diff
@@ -0,0 +1,122 @@
...
+  createReducer(initialState, handleAction => [
+    handleAction(updateCaseAsyncAction.pending as typeof updateCaseAsyncAction, state => {
+      return {
+        ...state,
+        tasks: {
+          [taskId]: {
+            ...state.updateCase.tasks,
+          },
+        },
+        status: SavedCaseStatus.ResultPending,
```
stephenhand, line 58: I think this status needs to be tracked per case rather than globally. Get the basic functionality working first though, then we can revisit this

stephenhand (commented):
I spotted a couple of issues, fixing those might unblock you

plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -481,7 +479,7 @@ const mapDispatchToProps = (dispatch, { task }: OwnProps) => {
     loadContact: bindActionCreators(ContactActions.loadContact, dispatch),
     cancelNewCase,
     updateCaseAsyncAction: (caseId: CaseType['id'], body: Partial<CaseType>) =>
-      searchAsyncDispatch(updateCaseAsyncAction(caseId, task.taskSid, body)),
+      dispatch(updateCaseAsyncAction(caseId, task.taskSid, body)),
```
stephenhand, line 482: Why was this changed to a regular dispatch, was it causing an issue? If you don't wrap them in `asyncDispatch` it will mean any exceptions thrown from the promise will be difficult to handle properly

plugin-hrm-form/src/states/case/saveCase.ts
```diff
@@ -40,84 +46,67 @@ export const updateCaseAsyncAction = createAsyncAction(
...
-//       status: SavedCaseStatus.ResultPending,
-//     };
-//   }),
+const handleRejectedAction = (handleAction, asyncAction) =>
+  handleAction(asyncAction, ({ state, rootState }, { payload }) => {
+    return {
+      state: {
+        ...state,
+        error: payload,
+        status: SavedCaseStatus.ResultReceived,
```
stephenhand, line 97: The status and error properties should be per task rather than global. Otherwise we could get strange, difficult to debug edge cases when more than one case is being saved at the same time

plugin-hrm-form/src/states/case/saveCase.ts
```diff
@@ -0,0 +1,112 @@
...
+      },
+      rootState,
+    };
+  });
+
+export const updateCaseReducer = (initialState: SaveCaseReducerState) =>
+  createReducer(initialState, handleAction => [
```
stephenhand, line 104: We should have some unit tests for the new reducer logic

acedeywin: @stephenhand  - I am getting the below error when writing tests for async actions in savecase reducer

   ```
"reducer" is a required argument, and must be a function or an object of functions that can be passed to combineReducers

      68 |
      69 | const testStore = (stateChanges: SaveCaseReducerState) =>
    > 70 | configureStore({
         | ^
      71 |   preloadedState: { ...initialState, ...stateChanges },
      72 |   reducer: boundSaveCaseReducer,
      73 |   middleware: getDefaultMiddleware => [

      at configureStore (node_modules/@reduxjs/toolkit/src/configureStore.ts:164:11)
      at testStore (src/___tests__/states/case/saveCase.test.ts:70:1)
      at _callee$ (src/___tests__/states/case/saveCase.test.ts:126:40)
      at tryCatch (node_modules/@babel/runtime/helpers/regeneratorRuntime.js:86:17)
      at Generator._invoke (node_modules/@babel/runtime/helpers/regeneratorRuntime.js:66:24)
      at Generator.next (node_modules/@babel/runtime/helpers/regeneratorRuntime.js:117:21)
      at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at _next (node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
      at node_modules/@babel/runtime/helpers/asyncToGenerator.js:32:7
      at Object.<anonymous> (node_modules/@babel/runtime/helpers/asyncToGenerator.js:21:12)

```

I have already mocked the reducer in configureStore as it was done in the resources reducers test, but I wonder why this error is coming up.

```
const testStore = (stateChanges: SaveCaseReducerState) =>
configureStore({
  preloadedState: { ...initialState, ...stateChanges },
  reducer: boundSaveCaseReducer,
  middleware: getDefaultMiddleware => [
    ...getDefaultMiddleware({ serializableCheck: false, immutableCheck: false }),
    promiseMiddleware(),
  ],
});
```

**Comment by stephenhand** [comment]
@acedeywin - there were 2 issues. One was that some of the stuff you were trying to test was being mocked out, so breaking the functionality. One of the great things about redux actions / reducers is that they are pure functions, they don't have side effects. This means you can test them easily without having to use mocks,. The only mocking you should need to do are the calls to the external services

The second issue was some nasty circular dependencies meaning some stuff was trying to reference your reducer before it was defined. A lot of the time these issues don't show in production because the offending dep was already loaded by something else, but the more isolated test environment shows the issue. These circular dependency issuers can be spotted by paying close attention to the call stacks of setup errors in your tests

The tests are still broken but they look to be more straightforward logic errors now

File: plugin-hrm-form/src/___tests__/states/case/saveCase.test.ts [hunk 5]
```diff
@@ -56,57 +76,153 @@ const nonInitialState: SaveCaseReducerState = {
...
 
 describe('actions', () => {
-  test('Calls the updateCaseAsyncAction service with the correct arguments', async () => {
+  test('Calls the updateCase service, and update a case', () => {
+    updateCaseAsyncAction(234, mockPayload.taskSid, mockPayload.case);
+    expect(updateCase).toHaveBeenCalledWith(234, mockPayload.case);
+  });
+
+  test('Calls the createCase service, and create a case', () => {
+    createCaseAsyncAction(contactForm, mockPayload.taskSid, workerSid, definitionVersion);
```
stephenhand, line 175: Why is the test passing in a contact form to the create case action? Shouldn't it be a case?

stephenhand (approved):
Couple of minor comments - I'll pre-approve assuming you will address them, otherwise LGTM!

plugin-hrm-form/src/___tests__/states/case/saveCase.test.ts
stephenhand: Also since I merged #1655 the `TaskEntry` type that contactForm uses is gone, so there will be a conflict

acedeywin: @stephenhand  - The `createCase` service takes in contactForm as one of its parameters.

stephenhand: Oh I see now, thanks. You will need to use a HrmServiceContact instead of a TaskEntry now, because that's what CaseService takes since the refactor.

You can base the test contact on VALID_EMPTY_CONTACT and just overwrite the properties you want to specify to make the setup code less verbose

stephenhand (approved):

acedeywin:
@stephenhand - All conflict has been resolved, and feedback implemented.

stephenhand:
> @stephenhand - All conflict has been resolved, and feedback implemented.

Just need to fix the failing tests and you are GTG

PR #1640, opened 2023-09-08
CHI-1987 refactor: get rid of funky unneeded return type

robert-bo-davis:
## Description

This refactors the funky unneeded return type in Contact service


mythilytm (approved):
LGTM

PR #1642, opened 2023-09-11
CHI-1850 cleanup orphaned chat channels in e2e webchat system

robert-bo-davis:
Primary reviewer:

## Description

Deleting tasks doesn't cleanup the channel or disconnect the chat user from the channel. We were orphaning channels with the e2e chat user attached and hitting the 1000 channel per user limit. This adds a job to cleanup those orphaned channels and that job is run on a schedule every hour in [a sibling pr](https://github.com/techmatters/infrastructure-config/pull/260). 

The job takes over 30 seconds to run because there are so many anonymous chat users and the call to twilio is VERY VERY slow. So we use a scheduled job instead of calling it inline with the tests to keep the e2e test time to a minimum.

### Verification steps

View successful invocations in datadog: https://app.datadoghq.com/functions?cloud=aws&entity_view=lambda_functions&panel_end=1694444434244&panel_paused=false&panel_start=1694440834244&selection=aws-lambda-functions%2Bstaging-use1-ca-e2e-webchat%2Bus-east-1%2B712893914485&start=1694442998119&end=1694443898119&paused=false


stephenhand (approved):
LGTM!

PR #1646, opened 2023-09-11
[Bug Fix] Sound alert does not play on first contact

GPaoloni:
## Description
This PR proposes a solution for the below linked bug ticket.
The proposed solution is to play a sound in plugin initialization, in a way that the Audio Player is already loaded and working from the beginning. 
One thing we could/should do is to upload a new "empty sound" to play, but as this requires uploading to every bucket I'll wait for @robert-bo-davis assistance.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2209)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
Start a local server on this branch, and confirm that the bug described in the ticket is not reproducible anymore.

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.tsx
+++ b/plugin-hrm-form/src/HrmFormPlugin.tsx
@@ -210,7 +210,7 @@ export default class HrmFormPlugin extends FlexPlugin {
     subscribeReservedTaskAlert();
     subscribeNewMessageAlertOnPluginInit();
     // Force one notification on init so AudioPlayer is eagerly loaded
-    playNotification('bell');
+    playNotification('silence');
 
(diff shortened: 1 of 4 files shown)


mythilytm (commented):
@GPaoloni I don't see call for how the silent tone will be consumed by AudioPlayerManager. Is that not necessary?

robert-bo-davis (approved):
LGTM once we switch out for silent mp3

mythilytm:
> One thing we could/should do is to upload a new "empty sound" to play, but as this requires uploading to every bucket
I am on this

PR #1647, opened 2023-09-11
[Bug Fix] Contacts not being properly updated on Insights

GPaoloni:
## Description
This PR is the result of the all-hands we held with Podraces (@mythilytm and @robert-bo-davis), where after the great investigation Mythily did on this, we were able to isolate and fix (we think) the issue.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2000)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### DIFF
--- a/plugin-hrm-form/src/services/InsightsService.ts
+++ b/plugin-hrm-form/src/services/InsightsService.ts
@@ -31,7 +31,7 @@ import { ITask } from '@twilio/flex-ui';
 import { isNonDataCallType } from '../states/validationRules';
 import { mapChannelForInsights } from '../utils/mappers';
 import { getDateTime } from '../utils/helpers';
-import { Case, CustomITask, HrmServiceContact } from '../types/types';
+import { Case, CustomITask, HrmServiceContact, isTwilioTask } from '../types/types';
 import { formatCategories } from '../utils/formatters';
 import { getDefinitionVersions, getHrmConfig } from '../hrmConfig';
 import { shouldSendInsightsData } from '../utils/setUpActions';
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -433,13 +433,7 @@ export const buildInsightsData = (
 ) => {
   const previousAttributes = typeof task.attributes === 'string' ? JSON.parse(task.attributes) : task.attributes;
 
-  if (
-    !shouldSendInsightsData({
-      ...task,
-      attributes: previousAttributes,
```
nickhurlburt, line 439: Was the issue that `previousAttributes` was incorrectly overwriting `attributes.skipInsights` to be `false`?

GPaoloni: I'll add a more detailed description here, but I don't know if this is the PR that will get merged or if Mythily is working on a parallel one. However the TL;DR is that destructing a Twilio task results in an incomplete object. This is caused because of how classes are implemented in JS (being really just [prototypal inheritance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain)). This resulted in 
```
{
  ...task, 
  attributes: previousAttributes,
}
```
being an incomplete representation of a Twilio task. When this "recreated task" was passed down the functions chain, one of the nested calls in the stack uses `task.sid`, which was actually undefined, because the `task.sid` is not really a "top level property of the task", but an inherited property instead. A weird, and hard to find one :p.

GPaoloni: A fun example that can be reproduced in the Chrome console to explain why this happens:
```
const asd = { a: 1 } 
> undefined

const asd2 = { }
> undefined

asd2.__proto__ = asd
> {a: 1}

asd2.a
> 1

const asd3 = { ...asd2 }
> undefined

asd3.a
> undefined
```

nickhurlburt: Wow. Not something I would have thought of. Nice work, Podracers!

mythilytm (approved):
This looks great. I don't see anything additional I should add.

robert-bo-davis (approved):
I added #1654 which moves to guard pattern and gets rid of the as any, mostly as a though experiment and to start negotiation around removing the lint rule that prevents guard patterns for complex bool returns. Can merge this with or without that change as far as I'm concerned.

PR #1654, opened 2023-09-13
refactor: use guard pattern for shouldSendInsightsData

robert-bo-davis:
## Description

Quick refactor to get rid of `as any` and move to guard pattern for complex bool

### DIFF
--- a/plugin-hrm-form/src/utils/setUpActions.ts
+++ b/plugin-hrm-form/src/utils/setUpActions.ts
@@ -62,8 +62,7 @@ export const shouldSendInsightsData = (task: CustomITask) => {
   if (isOfflineContactTask(task)) return false;
   if (task.attributes?.skipInsights) return false;
 
-  const hasTaskControl = !featureFlags.enable_transfers || TransferHelpers.hasTaskControl(task);
-  if (!hasTaskControl) return false;
+  if (featureFlags.enable_transfers && !TransferHelpers.hasTaskControl(task)) return false;
 
   return true;
 };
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/utils/setUpActions.ts
```diff
@@ -54,12 +54,20 @@ export const loadCurrentDefinitionVersion = async () => {
 /**
  * @param task
  */
+/* eslint-disable sonarjs/prefer-single-boolean-return */
 export const shouldSendInsightsData = (task: CustomITask) => {
   const featureFlags = getAseloFeatureFlags();
+
+  if (!featureFlags.enable_save_insights) return false;
+  if (isOfflineContactTask(task)) return false;
```
nickhurlburt, line 62: I am a little confused by this addition, as it seems to block us from ever saving offline contacts to insights, which will break things worse. But maybe I am misunderstanding and @GPaoloni may be better to review

GPaoloni: @robert-bo-davis I agree with Nick on this one. It would cause the same kind of error that the original PR is fixing, but for the offline contacts :P

I'd suggest we loose a bit the type of the offline contacts or add `skipInsights` as an optional flag that may or may not come for an offline contact.

GPaoloni (approved):
After the note made by Nick is addressed, good to merge!

plugin-hrm-form/src/utils/setUpActions.ts
robert-bo-davis: Ohhh. You are right. Have we tested that the original fix PR doesn't break offline contacts for helplines with enable_transfers set to true?

PR #1662, opened 2023-09-19
CHI-2238: Ensure date filters include the full local day

stephenhand:
## Description

This change ensures that date filters specified in Case Lists, Case Searches & Contact Searches now cover the full LOCAL days specified, e.g. specifying from 2023-08-10 to 2023-08-11 in a contact search returns contacts created between 2023-08-10 00:00:00 and 2023-08-11 23:59:59 in the local time zone of the browser making the request.

In order for this to be a full fix, https://github.com/techmatters/hrm/pull/463 also needs to be deployed. If one or the other are deployed by themselves, the boundaries of date filtering will be incorrect, but they are incorrect already, so it's not a big compatibility issue

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-2238

### Verification steps
Deploy a version of HRM with this PR applied: https://github.com/techmatters/hrm/pull/463, and a version of Flex with this PR applied.
Test that the date filters for Case Search, Contact Search and all 3 Case List date filters are returning all the correct records but no more for the date range specified


plugin-hrm-form/src/states/search/actions.ts
```diff
@@ -40,15 +42,24 @@ export const handleSearchFormChange = (taskId: string) => <K extends keyof t.Sea
...
 
-    const searchResultRaw = await searchContactsApiCall(searchParams, limit, offset);
+    const { dateFrom, dateTo, ...rest } = searchParams ?? {};
+    const searchParamsToSubmit: SearchParams = rest;
+    if (dateFrom) {
+      searchParamsToSubmit.dateFrom = formatISO(startOfDay(parseISO(dateFrom)));
+    }
+    if (dateTo) {
+      searchParamsToSubmit.dateTo = formatISO(endOfDay(parseISO(dateTo)));
+    }
```
murilovmachado, line 60: Question: both on `searchContacts` and `searchCases` actions, I see we're doing this transformation to ISO.
By looking at the other code changes, it seemed to me that the other components are passing `searchParams.dateFrom` and `searchParams.dateTo` in ISO format as well, so we ended up with a duplicate transformation. Did I get it correctly?

I'm not suggesting changing anything, just wanted to make sure I understood the code changes completely.

murilovmachado (approved):
🚀

plugin-hrm-form/src/states/search/actions.ts
stephenhand: It's because we are taking in an ISO format date, but we then need to adjust it to be the beginning / end of the day as a `Date` object, then put it BACK to being an ISO string, so you are correct, we do ISO format the date twice, but there is a logic to it.

At the moment I think passing around dates internally as ISO strings is the lesser of the 2 evils, this whole fix highlights the limitations of the JS `Date` object which is ignorant of time zones.

When `Temporal` is implemented in all major browsers & our node version we should look to start using that internally and only convert to strings to display to users or send to APIs, but for now sticking with strings seems like the best option

PR #1666, opened 2023-09-20
[filesUrlMigration] Implement `generateSignedUrlPath` to case document upload, download, and delete 

mythilytm:
Primary reviewer:

## Description
- This PR primarily focuses on disabling serverless upload, download, and delete operations for case documents and calling lambda to validate permissions. It incorporates the refactored `generateSignedUrlPath` function and also includes various cleanups and bug fixes.

### Checklist
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Related Issues
Fixes # https://tech-matters.atlassian.net/browse/CHI-2110

### Verification steps
- In the network tab, check that the call is made directly to the lambda.


plugin-hrm-form/src/hrmConfig.ts
```diff
@@ -47,6 +47,9 @@ const readConfig = () => {
   const assetsBucketUrl = manager.serviceConfiguration.attributes.assets_bucket_url;
   const getFormDefinitionsBaseUrl = buildFormDefinitionsBaseUrlGetter(new URL(configuredFormDefinitionsBaseUrl));
 
+  const { helpline_code: helplineCode, environment } = manager.serviceConfiguration.attributes;
+  const docsBucket = `tl-aselo-docs-${helplineCode}-${environment}`;
```
robert-bo-davis, line 51: Someone updated the helplineCode attributes to be upper case.

We need to figure out who did that and why and if it is required, then change this to force the code into lower case.

plugin-hrm-form/src/components/common/forms/DownloadFile.tsx
```diff
@@ -19,15 +19,17 @@ import React, { useState, useRef, useEffect } from 'react';
...
 
+import { getHrmConfig } from '../../../hrmConfig';
 import { Flex, Box, StyledNextStepButton, HiddenText } from '../../../styles/HrmStyles';
 import { formatFileNameAtAws } from '../../../utils';
-import { getFileDownloadUrl } from '../../../services/ServerlessService';
+import { fetchHrmApi, generateSignedURLPath } from '../../../services/fetchHrmApi';
 
 type Props = {
   fileNameAtAws: string;
+  objectId: string;
```
robert-bo-davis, line 29: Instead passing objectId alone, 
we probably want:

```
targetObject: {
	id: string
	type: SomeTypeRepresentingAvailableObjectTypes
}
```

and to pass that through the entire prop chain... until we can move to a context.

plugin-hrm-form/src/components/case/documentUploadHandler.ts
```diff
@@ -43,28 +44,56 @@ const uploadDocument = async (file: File, preSignedUrl: string, mimeType: string
...
+  );
 };
 
-export default fileUploadCustomHandlers;
+export const bindFileUploadCustomHandlers = (caseId: number) => (): CustomHandlers => {
+  return {
+    onFileChange: bindOnFileChange(caseId),
+    onDeleteFile: bindOnDeleteFile(caseId),
+  };
+};
```
GPaoloni, line 99: Extremely small note: 
You can 
```suggestion
export const bindFileUploadCustomHandlers = (caseId: number): CustomHandlers => {
  return {
    onFileChange: bindOnFileChange(caseId),
    onDeleteFile: bindOnDeleteFile(caseId),
  };
};
```
then when you call it you can remove one set of parenthesis.

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -485,10 +485,12 @@ export async function connectToCase(contactId, caseId) {
...
-  `/files/urls?method=getObject&objectType=contact&objectId=${contactId}&fileType=${mediaType}&bucket=${bucket}&key=${key}`;
+// export const generateExternalMediaPath = (
+//   contactId: string,
+//   mediaType: ContactMediaType,
+//   bucket: string,
+//   key: string,
+// ) =>
+//   `/files/urls?method=getObject&objectType=contact&objectId=${contactId}&fileType=${mediaType}&bucket=${bucket}&key=${key}`;
+// const externalMediaPath = generateSignedURLPath('getObject', 'contact', contactId, 'recording', bucket, key);
+// console.log(externalMediaPath); // Output should match that of generateExternalMediaPath
```
GPaoloni, line 496: Can be removed?

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -356,22 +355,30 @@ const getInsightsUpdateFunctionsForConfig = (
...
+    const mediaUrl = fetchHrmApi(
+      generateSignedURLPath({
+        method: 'getObject',
+        objectType: 'contact',
+        objectId: contact.id.toString(),
+        fileType: 'recording',
+        location: { bucket, key },
+      }),
+    );
+    console.log('>>> mediaUrl', mediaUrl);
```
robert-bo-davis, line 370: console.log

robert-bo-davis (changes requested):

GPaoloni (commented):
Two comments small notes

plugin-hrm-form/src/components/case/ViewCaseItem.tsx
```diff
@@ -73,6 +73,11 @@ const ViewCaseItem: React.FC<Props> = ({
     changeRoute({ ...routing, action: CaseItemAction.Edit }, task.taskSid);
   };
 
+  const caseObjS3Download: { caseId: number; objectType: ObjectType } = {
+    caseId: connectedCase?.id,
+    objectType: 'case',
+  };
```
robert-bo-davis, line 79: I don't think the form this high in the stack needs to have any awareness of *how* the caseObj is going to be used further down the stack and we definitely don't want the calls further down the stack to have a difference between a case, contact, resource, csam or whatever other type we may pass through.

This is really just a temporary fix until we can move to a context, but I think it is important to keep generic components as generic as possible.

I have an example refactor PR up at #1688. Let's pair to go over the why's behind my changes when you are available.

robert-bo-davis (changes requested):
See example refactor in #1688

robert-bo-davis (approved):
LGTM once final test is complete.

PR #1669, opened 2023-09-21
CHI-2106 follow on refactors

stephenhand:
## Description

* Moved some 'transform' logic for mixed checkboxes into the redux layer
* Removed some 'detransform' logic that looks to be redundant

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-2166

### Verification steps

Regression test create / edit contact functions, especially using the 'mixed checkbox' type form item


plugin-hrm-form/src/states/contacts/contactDetailsAdapter.ts
```diff
@@ -14,31 +14,36 @@
  * along with this program.  If not, see https://www.gnu.org/licenses/.
  */
 
-/**
```
stephenhand, line 17: I moved 'transformValues' into here as it seemed like the place for contact transforming utility functions, but now it's the only one in here because all the rest have gone :-P

plugin-hrm-form/src/states/contacts/contactDetailsAdapter.ts
```diff
@@ -14,31 +14,36 @@
...
+import { DefinitionVersion, FormDefinition, FormItemDefinition, isNonSaveable } from 'hrm-form-definitions';
 
-/**
- * @param {{ [category: string]: string[] }} accumulator
- * @param {[string, { [subcategory: string]: boolean }]} currentValue
- */
-const catsReducer = (accumulator, [cat, subcats]: [string, Record<string, boolean>]) => {
-  const subcatsList = Object.entries(subcats).reduce(subcatsReducer, []);
+import { ContactRawJson } from '../../types/types';
+import { RecursivePartial } from '../../___tests__/RecursivePartial';
```
murilovmachado, line 20: It feels very strange to me to import a type from `__tests__` here.
Can we move this type out of `__tests__`?

murilovmachado (approved):
Thanks steve, the code looks 500% more clean with this PR.

murilovmachado:
Didn't find any issue while testing it locally, except the one Stephen is already working on.

plugin-hrm-form/src/states/contacts/contactDetailsAdapter.ts
stephenhand: Oh I didn't notice that - it's only one line, I'll copy the implementation - thanks!

PR #1672, opened 2023-09-21
Bring back Twilio recordings overlay

GPaoloni:
## Description
This PR is a hotfix for the 2023-09-21 production deploy, to bring back the Twilio recordings overlay if external recordings is not enabled for a given account (also use for older contact records where external recordings was not implemented).


robert-bo-davis (approved):
LGTM based on mob session

PR #1676, opened 2023-09-22
Terragrunt migration CO Staging

janorivera:
Primary reviewer:

### Checklist

- [x]  Updates to main.tf to use latest version of modules
- [x]  Updates to staging.hcl file (also to production.hcl but that one is incomplete - I need a function id that will come with the next prod release)
- [x]  Work with the State to import existing resources or to move them to the correct module (survey resources)
- [x]  Migrate state executed for all stages
- [x]  Clean apply done to all stages
- [x]  Test that everything works OK

### Related Issues
Fixes #....


robert-bo-davis (approved):
LGTM once ssm param case issue is merged around.

PR #1686, opened 2023-09-27
Chi 1435 refactor contact saving logic redux promise

acedeywin:

## Description

- **Feature refactor:** Refactor contact save logic into redux-promise middleware - part 1 (`updateContactsFormInHrm`).

The ticket handles the refactoring of network calls for contact to be handled in our Redux layer rather than the component layer. This is the first PR for the `updateContactsFormInHrm` service layer method.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues

### Verification steps

- Search for a contact
- Select a contact you can edit
- Edit any aspect of the contact
- Verify that the edit was successful


plugin-hrm-form/src/states/contacts/saveContact.ts
```diff
@@ -0,0 +1,93 @@
...
+    replaceExisting: boolean,
+    contactId: string,
+    body: Partial<ContactRawJson>,
+    helpline: string,
+    reference?: string,
+  ): Promise<{ contacts: Partial<HrmServiceContact>[]; replaceExisting: boolean; reference?: string }> => {
+    const contact = await updateContactsFormInHrm(contactId, body, helpline);
+    return {
+      contacts: [contact],
+      replaceExisting,
```
stephenhand, line 36: I don't think the 'replaceExisting' flag is relevant to saving a contact, you will always want to replace the contact in redux with the version you just saved

plugin-hrm-form/src/states/contacts/saveContact.ts
```diff
@@ -0,0 +1,93 @@
...
+      reference,
```
stephenhand, line 37: I don't think the idea of references are relevant to saved contacts either, you should leave any existing references in place on a contact that's already loaded into redux and not create any for one that isn't

stephenhand (approved):
LGTM except for a couple of comments. I haven't tested it though

plugin-hrm-form/src/states/contacts/saveContact.ts
acedeywin: @stephenhand  - Won't removing `reference` and `replaceExisting` contradict our current pattern where they are used within the `updateEntries`, which is subsequently used to update the `existingContacts`?

stephenhand: I don't mean remove them from state, just leave them untouched in the save contact action, ensure they are copied from the pre-existing state - so the action doesn't need those properties

acedeywin: @stephenhand  - i have updated the PR

stephenhand: We should remove the `references` and `replaceExisting` properties from the UPDATE_CONTACT_ACTION completely rather than hardcoding them, we can just put the required logic in the reducer

acedeywin: @stephenhand - PR has been updated.

plugin-hrm-form/src/states/contacts/saveContact.ts
```diff
@@ -0,0 +1,80 @@
...
+                expanded: {},
+                gridView: false,
+              },
+            };
+            return [
+              c.id,
+              {
+                ...currentContact,
+                savedContact: replaceExisting || !current.references.size ? c : state[c.id].savedContact,
+                ...state[c.id].references,
```
stephenhand, line 68: We should remove this line, this will copy the references properties directly into the contact. The existing references will be copied into the correct place by line 66

plugin-hrm-form/src/states/contacts/saveContact.ts
```diff
@@ -0,0 +1,80 @@
...
+              categories: {
+                expanded: {},
+                gridView: false,
+              },
+            };
+            return [
+              c.id,
+              {
+                ...currentContact,
+                savedContact: replaceExisting || !current.references.size ? c : state[c.id].savedContact,
```
stephenhand, line 67: This ternary is redundant now if replaceExisting is always true

```suggestion
                savedContact:  c,
```

plugin-hrm-form/src/states/contacts/saveContact.ts
```diff
@@ -0,0 +1,80 @@
...
+export const saveContactReducer = (initialState: ExistingContactsState) =>
+  createReducer(initialState, handleAction => [
+    handleAsyncAction(handleAction, updateContactsFormInHrmAsyncAction.pending),
+
+    handleAction(
+      updateContactsFormInHrmAsyncAction.fulfilled,
+      (state, { payload }): ExistingContactsState => {
+        const replaceExisting = true;
+        const updateEntries = payload.contacts
+          .filter(c => state[c.id]?.references && replaceExisting)
```
stephenhand, line 54:
```suggestion
          .filter(c => state[c.id]?.references)
```

plugin-hrm-form/src/states/contacts/saveContact.ts
```diff
@@ -0,0 +1,80 @@
...
+import { UPDATE_CONTACT_ACTION } from './types';
+import { ExistingContactsState } from './existingContacts';
+
+export const updateContactsFormInHrmAsyncAction = createAsyncAction(
+  UPDATE_CONTACT_ACTION,
+  async (
+    contactId: string,
+    body: Partial<ContactRawJson>,
+    helpline: string,
+  ): Promise<{ contacts: Partial<HrmServiceContact>[] }> => {
```
stephenhand, line 30: Why does this promise return an array of contacts? We only ever save one at a time. We can simplify the reducer logic if we only have a single contact as well.

stephenhand (commented):
Still some clean up needed

plugin-hrm-form/src/states/contacts/saveContact.ts
acedeywin: @stephenhand  - That was how the `loadContactReducer` was done and the array was for the purpose of manipulating the contact update entries, where the different values were passed to the state and subsequently passed to `existingContact`

stephenhand: loadContact is a different action that just loads contacts into redux and supports loading multiple contacts. Your save action only ever updates a single contact and the API should reflect that. The 'updateEntries' is only used in loadContact because multiple categories can be updated in one action. For saveContact it isn't needed, you can just update the single contact via it's id in the usual way

stephenhand (approved):

PR #1687, opened 2023-09-27
chore: bump checkout action version

robert-bo-davis:
Primary reviewer:

## Description

This bumps checkout action form v2 to v3 so that it is using a version of node that isn't unsupported.


nickhurlburt (approved):
Thanks!

stephenhand (approved):

PR #1688, opened 2023-09-27
refactor: use generic targetObject to pass through form target

robert-bo-davis:
Primary reviewer:

## Description

Refactors to generic targetObject for case passthrough.


mythilytm (approved):
LGTM

PR #1692, opened 2023-09-28
Chi 1435: Refactor contact saving logic redux promise - Part 2

acedeywin:

## Description

- **Feature refactor**: Refactor contact save logic into redux-promise middleware - part 2 (`connectToCase` and `submitContactForm`).
The ticket is the second part of the contact save logic refactor. It handles the refactoring of network calls for contact to be handled in our Redux layer rather than the component layer. This is the first PR for the `connectToCase` and `submitContactForm` services layer method.

- **Bug fix**: It also fixes the `Save and End Case` error.
- TODO: Implement units for the newly implemented actions and reducers. 

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues

### Verification steps

- Create an offline contact
- Fill in all the necessary details
- Click on "Save Contact"
- Verify that the contact was saved successfully
- Repeat steps 1 and 2
- Click on "Add Contact to New Case"
- Click on "Save and End"
- Verify that contact was created and successfully added to the case


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -187,6 +193,28 @@ const Case: React.FC<Props> = ({
...
+  // We want to validate that savedContact state is not returning an empty object
+  const validateSavedContact = () => typeof savedContact === 'object' && Object.keys(savedContact).length > 0;
+
+  /**
+   * Always check to see when savedContact state has been updated.
+   * Then go ahead and execute the connectToCaseAsyncAction
+   */
+  useEffect(() => {
+    if (validateSavedContact()) {
+      connectToCaseAsyncAction(savedContact.id, connectedCase.id);
```
stephenhand, line 205: I don't know if this is working the way to think it is - submitting the async action won't wait for the network call to complete - on the next render cycle, the task will be completed, regardless of whether the connectToCase call has completed or not.

If the idea is to do them sequentially, creating a separate async action which does both in sequence would be the simplest way. Otherwise you will have to put a state in the redux store and use that to control the sequence

plugin-hrm-form/src/components/tabbedForms/BottomBar.tsx
```diff
@@ -85,7 +87,7 @@ const BottomBar: React.FC<
     setSubmitting(true);
 
     try {
-      await submitContactForm(task, contact, metadata, caseForm as Case);
+      submitContactFormAsyncAction(task, contact, metadata, caseForm as Case);
```
stephenhand, line 90: Again, this logic probably isn't sequential any more, and should either be part of a single action or the sequence should be managed by state

plugin-hrm-form/src/states/contacts/reducer.ts
```diff
@@ -149,10 +149,12 @@ export const initialState: ContactsState = {
   },
   editingContact: false,
   isCallTypeCaller: false,
+  savedContact: {} as Contact,
```
stephenhand, line 152: We can't have global contact state like this,  the submitFormContact reducer needs to work with the tasks map for unsaved contacts

stephenhand (changes requested):

plugin-hrm-form/src/states/contacts/reducer.ts
acedeywin: @stephenhand  - This has been handled.

plugin-hrm-form/src/components/case/Case.tsx
acedeywin: @stephenhand  - Wouldn't it be better to implement a cleanup in the second `useEffect` that calls the `handleCompleteTask` method?

stephenhand: It isn't cleanup, it's a second step in the save operation. The idea of this refactor is to move the logic for handling this into redux, rather than relying on components to manage it, so it's better to handle the sequencing in the action creator or the redux state.

The component should just be able to send off a case for saving and be told when it's done, it shouldn't need to know about the separate backend calls involved in the save operation

acedeywin: @stephenhand  - Noted.

stephenhand (approved):
These look ok, we will probably need to be able to connect to case without submitting a contact in the future, but for now combining them is probably simplest

PR #1707, opened 2023-10-05
Chi 1435 refactor contact saving logic redux promise test

acedeywin:

## Description

- Test Implementation: Refactor contact save logic into redux-promise middleware

This is the unit test for the [CHI-1435](https://tech-matters.atlassian.net/browse/CHI-1435)

### Checklist
- [x] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues

### Verification steps


plugin-hrm-form/src/___tests__/states/contacts/saveContact.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  createdAt: '12-05-2023',
+  updatedAt: '12-05-2023',
+  connectedContacts: [baseContact] as Contact[],
+};
+
+const dispatch = jest.fn();
+
+describe('actions', () => {
+  test('Calls the updateContactsFormInHrmAsyncAction action, and update a contact', async () => {
+    dispatch(updateContactsFormInHrmAsyncAction(baseContact.id, baseContact.rawJson, 'demo-v1'));
```
stephenhand, line 95: Can we verify that the response gets added to the state correctly as well?

stephenhand: I updated the case redux tests with something to test the effect of the state, you can use that as a guide

plugin-hrm-form/src/___tests__/states/contacts/saveContact.test.ts
```diff
@@ -88,12 +91,29 @@ const baseCase: Case = {
...
+} as const;
+
 const dispatch = jest.fn();
 
 describe('actions', () => {
   test('Calls the updateContactsFormInHrmAsyncAction action, and update a contact', async () => {
     dispatch(updateContactsFormInHrmAsyncAction(baseContact.id, baseContact.rawJson, 'demo-v1'));
+    const existingContacts = boundSaveContactReducer(
+      baseState,
+      updateContactsFormInHrmAsyncAction(baseContact.id, baseContact.rawJson, 'demo-v1'),
```
stephenhand, line 112: This logic looks like you are updating the contact with the same data it already has, and then checking that it is still the same afterwards, which would be true whether it worked or not.

Also it looks like you aren't waiting for the async action to be fulfilled, which is when the state should actually change, before validating the state

stephenhand (approved):
LGTM

PR #1711, opened 2023-10-06
Turn Fullstory on for Chile Staging

annaliseirby090:
Primary reviewer:

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


robert-bo-davis:
If we turn on full story for both prod and staging, we should probably move the flag up into common.json

PR #1721, opened 2023-10-12
CHI-1434: Remove transformation of search results

stephenhand:
## Description

* Get standard contacts from search endpoint instead of converting them
* Make missing shared state not throw transfer errors - this is so it will be compatible with the next version of Flex that won't use the sync service

### Checklist
- [X] New tests added
- [N/A] Feature flags added
- [X] Strings are localized

### Related Issues
CHI-1434

### Verification steps

None - folded into upcoming QA run

### DIFF
--- a/plugin-hrm-form/src/services/ContactService.ts
+++ b/plugin-hrm-form/src/services/ContactService.ts
@@ -15,7 +15,6 @@
  */
 
 import { TaskHelper } from '@twilio/flex-ui';
-import { CallTypes, DefinitionVersion } from 'hrm-form-definitions';
 
 import { isNonDataCallType } from '../states/validationRules';
 import { getQueryParams } from './PaginationParams';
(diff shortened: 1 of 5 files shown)


plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -61,114 +51,7 @@ export async function searchContacts(
...
-    conversationDuration: number;
-    createdBy: string;
-    taskId: string;
-    updatedBy?: string;
-    updatedAt?: string;
-  };
-  details: ContactRawJson;
-  csamReports: CSAMReportEntry[];
-  referrals?: ResourceReferral[];
-};
```
murilovmachado, line 95: 🎉 🎉

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -363,13 +246,28 @@ export const saveContact = async (
...
+    await saveContactToExternalBackend(task, {
+      ...payloads.response,
+      rawJson: {
+        ...payloads.response.rawJson,
+        caseInformation: {
+          ...payloads.response.rawJson.caseInformation,
+          categories: legacyCategories,
+        },
+      },
+    });
```
murilovmachado, line 263: good catch

murilovmachado (approved):
🚀

PR #1722, opened 2023-10-13
refactor: use insights player for external recordings

robert-bo-davis:
Primary reviewer:

## Description

Because of the task modifications to support for the insights player in insights, the insights player also works from the contact page which reduces complexity, gets rid of the "download recording" button from the browser player, and provides a more standard user experience.

### Verification steps

Play back an external recording


stephenhand (approved):
Looks great, nice that we can do this.

PR #1728, opened 2023-10-16
CHI-1434 use redux contact save

stephenhand:
## Description

* Integrates the redux driven save logic with the new contact saving workflow
* Deprecates some initialisation actions and the initialization reducer logic hangs off load / save actions inst4ead

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-1434

### Verification steps

Regression around all contact save functionality


plugin-hrm-form/src/states/contacts/findContactByTaskSid.ts
```diff
@@ -14,10 +14,12 @@
  * along with this program.  If not, see https://www.gnu.org/licenses/.
  */
 
-import { contactFormsBase, namespace, RootState } from '..';
+import { RootState } from '..';
 import { ContactState } from './existingContacts';
 
 const findContactByTaskSid = (state: RootState, taskSid: string): ContactState =>
-  Object.values(state[namespace][contactFormsBase].existingContacts).find(cs => cs.savedContact?.taskId === taskSid);
+  Object.values(state['plugin-hrm-form'].activeContacts.existingContacts).find(
```
murilovmachado, line 21: The reason we're using the hardcoded value `plugin-hrm-form` here instead of the const `namespace` is related to the circular dependencies you were talking about?

plugin-hrm-form/src/utils/transfer.ts
```diff
@@ -193,7 +193,7 @@ export const canTransferConference = (task: ITask) => {
   }
 
   const isLiveCall = TaskHelper.isLiveCall(task);
-  const { callStatus } = (Manager.getInstance().store.getState() as RootState)[namespace][conferencingBase].tasks[
+  const { callStatus } = (Manager.getInstance().store.getState() as RootState)['plugin-hrm-form'].conferencing.tasks[
```
murilovmachado, line 196: same question here

murilovmachado (approved):
Looks good! Just a very minor question. 🚀

plugin-hrm-form/src/states/contacts/findContactByTaskSid.ts
stephenhand: Yes we can use the constant again once the dependency is broken. I've been thinking we probably don't need to use the constants now we are using TS though? If we make sure the state is typed as `RootState` Typescript will check we are using the correct name won't it?

PR #1729, opened 2023-10-16
Chi 1434 move redux constants

stephenhand:
#Do not review until #1728 is merged

## Description

This PR moves the constants we use for the various redux namespaces into their own file

Them being defined in the same file as the root state was causing circular dependency headaches, because every time you used any of these constants, you have to pull in the whole state handling code hierarchy

Now these constants can be used in isolation.

We should probably start deprecating the use of these in favour of using the correct types now we are in TS land, but this solves the immediate issue.

There aren't really any 'real' changes here, aside from a little bit of tidying the edits are all related to moving those constants

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes #....


murilovmachado (approved):
looks good!

PR #1730, opened 2023-10-16
fix: always cleanup e2e account tests

robert-bo-davis:
Primary reviewer:

## Description

This fixes a regression where e2e test tasks weren't being cleaned up properly on failure for the e2e account which doesn't use the special e2e webchat client page that sets e2e test mode to true on tasks for use in staging/production accounts.

This doesn't fix ALL of the current e2e test issues, but it does fix the pileup of orphaned tasks.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


stephenhand (approved):
Thx!

PR #1732, opened 2023-10-17
CHI-1994: HRM form refactor

stephenhand:
## Description

Minor refactor to used mapped dispatchers in TabbedForms.tsx & HrmForm.tsx,
Replaced direct save with redux action in HrmForm.tsx

### Checklist
- [X] Corresponding issue has been opened
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized
- [X] Tested for chat contacts
- [X] Tested for call contacts

### Related Issues
CHI-994

### Verification steps

None required, automated tests should cover it

PR #1737, opened 2023-10-18
CHI-1994: Modal routing

stephenhand:
# Review after #1732 is merged

## Description

* Significantly enhances the capabilities of the routing system. Now every task has a 'stack' of routes which gets added to with forward navigation, and removed from when the user 'goes back'
* Added first class support for modal navigation - now a route can have an 'activeModal' property, which is another stack of routes. Routes in activeModalStacks can have their own activeModals, thus allowing us to stack modal navigation
* changeRoute now automatically adds routes to the top level active modal. going back pops from the current top modal routes
* closing a modal drops the whole stack of routes, removing the parent route's 'activeModal' property
* Components base their rendering on the latest route of the top modal, a helper function is included for this
* Case navigation is reworked as a modal to demonstrate

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-1994

### Verification steps

Regression test navigation around the application


plugin-hrm-form/src/states/routing/types.ts
```diff
@@ -15,16 +15,34 @@
...
+export type RouteWithModalSupport = {
+  route: 'tabbed-forms' | 'case' | 'case-list' | 'search';
+  activeModal?: AppRoutes[];
+};
+
+export type TabbedFormRoute = {
+  route: 'tabbed-forms';
+  subroute?: TabbedFormSubroutes;
+  autoFocus?: boolean;
+} & RouteWithModalSupport;
```
murilovmachado, line 39: Written like this, the `route` gets overwritten from `tabbed-forms` to `tabbed-forms' | 'case' | 'case-list' | 'search`.

If that's intentional, then I would suggest removing line 36 `route: 'tabbed-forms';`
If not, this type would have to be defined like:
```
export type TabbedFormRoute = RouteWithModalSupport & {
  route: 'tabbed-forms';
  subroute?: TabbedFormSubroutes;
  autoFocus?: boolean;
};
```

murilovmachado (approved):
The code looks great, Steve.

I haven't tried it locally yet, but I don't want you to be blocked by it. Feel free to merge.

stephenhand:
FYI - since #1745 will be merged very soon after this I wouldn't bother testing it, better to test it all together

plugin-hrm-form/src/states/routing/types.ts
stephenhand: The '&' means that the type has to satisfy both sides of the operator, meaning that it should only allow `tabbed-forms` but still add support for `activeModal` shouldn't it?

plugin-hrm-form/src/states/routing/reducer.ts
```diff
@@ -83,13 +85,97 @@ const contactUpdatingReducer = (state: RoutingState, action: ContactUpdatingActi
...
+    if (nextStack) {
+      return [...routeStack.slice(0, -1), { ...currentRoute, activeModal: nextStack }];
+    }
+    // closeTopModal returned undefined, currentRoute's activeModal is the top of the stack and should be removed
+    const { activeModal, ...currentRouteWithoutModal } = currentRoute;
+    return [...routeStack.slice(0, -1), currentRouteWithoutModal];
+  }
+  // Empty activeModalStack should be removed, empty base route should be left alone (probably in a bad state though...
+  return parent ? undefined : routeStack;
+};
```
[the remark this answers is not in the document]
stephenhand: Yes, basically there are some parts of the behaviour that vary based on whether this is the base route or not - but it should only be used internally, external callers should always use the base route as the parameter and not set the parent

I am absolutely going to write unit tests for these routing reducer helpers in a follow on PR ASAP!

plugin-hrm-form/src/states/routing/types.ts
murilovmachado: You're right! I didn't know that

PR #1740, opened 2023-10-19
ft: configure lex for e2e account

acedeywin:

## Description

- New feature: Migrate E2E account from Autopilot to Lex
- Task: Configure lex for e2e account on Terraform.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes [CHI-2213](https://tech-matters.atlassian.net/browse/CHI-2213)

### Verification steps

Run `make HL=e2e HL_ENV=development plan` and verify everything is fine

[CHI-2213]: https://tech-matters.atlassian.net/browse/CHI-2213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


janorivera:
Is it not easy here to rename this account to something like "ee" to follow our standard 2 letter code convention? Then we could keep the code "cleaner" without having to add code just for one exception. 
@robert-bo-davis  what do you think?

twilio-iac/terraform-modules/lex/v1/main.tf
```diff
@@ -50,7 +50,7 @@ resource "aws_lex_intent" "this" {
 
   provider = aws.hl-region
 
-  name        = "${local.name_prefix}_${each.key}"
+  name        = replace("${local.name_prefix}_${each.key}", "2", "TO")
```
stephenhand, line 53: You should add a comment to explain why we are doing something so weird

stephenhand (approved):
Have you also migrated E2E to terragrunt, nice, thx!

twilio-iac/terraform-modules/lex/v1/main.tf
murilovmachado: I don't understand 100% this piece. Can you include in your comment an example of this scenario? (or just reply here)

acedeywin:
@janorivera - Do you think `eTe` would be better?

janorivera:
@acedeywin any 2 letter code will work!, I think that was the agreement, I don't mind making exceptions but we should have good reasons to make those exceptions.  I thought that "ee" could be a good option since it doesn't use numbers and it's 2 letters.  But as I said before this might require more work if there are already aws parameters created so I don't know if it's worth it. That is why I also wanted to see what @robert-bo-davis thinks. 
@stephenhand I don't think that this account has been migrated, I think this is just for Lex

robert-bo-davis:
I don't think trying to totally change the e2e account code globally at this point is worth the effort. The two letter country code system is going to break down the first time we get 2 helplines in the same country anyway, so we should be looking towards an expanded system. Knowing that numbers won't work in lex is useful. ;) I think ee should probably be the e2e code eventually, so I agree that just stripping the 2 for the lex bot name is probably fine here, but lets not make this relatively small task any larger by adding in a full rename of the code for all resources.

janorivera:
ok @acedeywin so not worth it now! you can disregard my previous comments :)

PR #1745, opened 2023-10-20
CHI-1994: Search in a modal modal

stephenhand:
## Description

* Refactors search routing to use standard routing system rather than its own thing
* Search now opens in a modal, and contacts / cases results open in a stacked modal
* Standard 'NavigableContainer' which automatically determines what back / close buttons are needed based on the context of the route
* Move dialog state for edit contacts to redux

TODO: styling is only partially complete

### Checklist
- [X] New tests added
- N/A Feature flags added
- [ ] Strings are localized

### Related Issues
CHI-1994

### Verification steps

* See ticket - also regression test routing


stephenhand:
I think the E2E test failure might be due to the E2E chatbot change @acedeywin? Looks like the chatbot never starts and the chat goes straight into the queue, which breaks the test

plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -165,33 +173,36 @@ const EditCaseSummary: React.FC<Props> = ({
...
   );
 
-  const checkForEdits = () => {
+  const checkForEdits = (closeModal: boolean) => {
     if (isEqual(workingCopy, savedForm)) {
-      closeActions();
-    } else setOpenDialog(true);
+      closeActions(closeModal);
+    } else setDialogState(closeModal ? DialogState.OpenForClose : DialogState.OpenForBack);
   };
```
murilovmachado, line 194: My brain goes crazy when I see `if` clause using brackets and the `else` doesn't. I know this bracket choice was already in the code, so feel free to ignore this comment.

murilovmachado (approved):
Looks good to me, also tried it locally and it works fine!

Couldn't do an in-depth review, but the code changes are aligned with what I expected. great work.

PR #1763, opened 2023-10-25
refactor: move profile modal into hrmForm and begin refactor of HRM f…

robert-bo-davis:
…orm router

Primary reviewer:

## Description

This moves profile routing into HRM form so it works from the categorize call menu and also refactors route handling a bit.


mythilytm (approved):
This looks great

PR #1764, opened 2023-10-26
CHI-2319 cancel offline contacts

stephenhand:
## Description

* Draft offline contacts get fully removed when cancelled
* In progress offline contacts get reloaded on browser session reloads

As an artefact of saving offline contacts under a single ID per user, 'draft offline contacts' may exist at unexpected times and cause confusion in search results, it might be best to just filter out all draft offline contacts from search results in HRM until the search experience is updated

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
FixesCHI-2319

### Verification steps

* Cancel an offline contact, then start another - the data should all be reset to blanks / defaults
* Refresh the browser halfway through a contact, it should reload with the contact still in progress (minus and changes since you last changed tabs)
* Regression test offline contacts basic saving & searching

### DIFF
--- a/e2e-tests/browser.ts
+++ b/e2e-tests/browser.ts
@@ -16,6 +16,7 @@
 
 import { Browser, BrowserContext, Page, chromium } from '@playwright/test';
 import { logPageTelemetry } from './browser-logs';
+import { getConfigValue } from './config';
 
 export type SetupPageReturn = {
   context: BrowserContext;
@@ -45,7 +46,9 @@ export const setupContextAndPage = async (browser: Browser): Promise<SetupPageRe
   await waitForBrowser(browser);
 
   console.log('Launching page');
-  const context = await browser.newContext();
+  const context = await browser.newContext({
+    storageState: getConfigValue('storageStatePath') as string,
+  });
   const page = await context.newPage();
 
   logPageTelemetry(page);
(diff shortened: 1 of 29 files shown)


murilovmachado (approved):
Code looks good to me.

Trying it locally, it worked 90% fine. Didn't have time to investigate it properly, but sometimes after refreshing the page some information from the first tab (contact tab) was missing, more specifically `Channel` and `Counsellor`. Not sure if it just a front-end issue (prob yes, but didn't investigate).

stephenhand:
> Code looks good to me.
> 
> Trying it locally, it worked 90% fine. Didn't have time to investigate it properly, but sometimes after refreshing the page some information from the first tab (contact tab) was missing, more specifically `Channel` and `Counsellor`. Not sure if it just a front-end issue (prob yes, but didn't investigate).

Did you definitely switch tabs away from the contact tab before refreshing? It wouldn't save if you don't do that because it saves on tab change

stephenhand:
> Code looks good to me.
> 
> Trying it locally, it worked 90% fine. Didn't have time to investigate it properly, but sometimes after refreshing the page some information from the first tab (contact tab) was missing, more specifically `Channel` and `Counsellor`. Not sure if it just a front-end issue (prob yes, but didn't investigate).

Good catch - 2 actions were overwriting each other's draft contacts, pushed a fix

PR #1766, opened 2023-10-26
CHI-2320: Resume case

stephenhand:
## Description

* Loading a contact into redux via the 'create' or 'load by task' endpoint will also load a case into redux if it is attached to the contact
* If you close a browser session when adding a new contact with a case open, reopening the contact will open the case so it can be saved / cancelled

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-2320

### Verification steps

Create a contact
Add a case
Refresh the browser

You should see the case open on the home page with any saved contacts

Repeat with offline contact

Ensure that save & cancel case functionality work after a browser refresh


plugin-hrm-form/src/states/case/reducer.ts
```diff
@@ -39,7 +39,12 @@ import { RootState } from '..';
 import { getAvailableCaseStatusTransitions } from './caseStatus';
 import { SaveCaseReducerState, saveCaseReducer } from './saveCase';
 import { CaseListContentStateAction } from '../caseList/listContent';
-import { configurationBase } from '../storeNamespaces';
+import { configurationBase, namespace } from '../storeNamespaces';
+import {
+  ContactUpdatingAction,
+  CREATE_CONTACT_ACTION_FULFILLED,
+  LOAD_CONTACT_FROM_HRM_BY_TASK_ID_ACTION_FULFILLED,
+} from '../contacts/types';
```
GPaoloni, line 47: If this actions are shared among more than one reducer, does it makes sense to move them up in the folder structure?

GPaoloni (approved):
LGTM! :rocket: 

Silly question: the case gets mounted of the existence of a "related case", without the need for routing changes?

murilovmachado (approved):
Looks good!

plugin-hrm-form/src/states/case/reducer.ts
stephenhand: Maybe? They are still contact related actions, even though they influence other systems. I always get confused by 'REMOVE_CONTACT_STATE' being a top level action, I always look for in in contacts.  I'd prefer a few horizontal deps for making things easier to find.

Another longer term question is do we still want to have multiple task maps in our redux store? We could restructure things so everything scoped within a task is stored under the same map entry? Not sure if it's better or worse. Perhaps with Bo's selector PR it becomes less important because reducers have visibility of the whole state

stephenhand:
> LGTM! 🚀
> 
> Silly question: the case gets mounted of the existence of a "related case", without the need for routing changes?

We route to 'new-case' in the routing reducer in a case is linked on a new contact

PR #1782, opened 2023-11-01
Client Profiles - 11 - initial work on contact details profile link

robert-bo-davis:
Primary reviewer:

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


plugin-hrm-form/src/components/contact/ContactDetails.tsx
```diff
@@ -75,9 +76,17 @@ const ContactDetails: React.FC<Props> = ({
   clearContactDraft,
   currentRoute,
   openFormConfirmDialog,
+  loadContactFromHrm,
   ...otherProps
   // eslint-disable-next-line sonarjs/cognitive-complexity
 }) => {
+  console.log('>>> ContactDetails', contactId);
```
GPaoloni, line 83: Remove?

plugin-hrm-form/src/components/profile/ProfileCases.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const onClickViewCase = () => {
+      // load case modal? or page?
+    };
+
+    return (
+      <CasePreview
+        key={`CasePreview-${cas.id}`}
+        currentCase={cas}
+        counselorsHash={counselorsHash}
+        onClickViewCase={can(PermissionActions.VIEW_CASE) && onClickViewCase}
```
GPaoloni, line 46: What about abstracting `can(PermissionActions.VIEW_CASE)` inside of `onClickViewCase` like
```
const onClickViewCase = () => {
  if (!can(PermissionActions.VIEW_CASE)) { return; }
  ...actual code...
}
```

We could even inform somehow in the UI that the user does not have permissions for the action.

plugin-hrm-form/src/components/profile/ProfileCases.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ConnectedProps<typeof connector>;
+
+const ProfileCases: React.FC<Props> = ({ profileId, counselorsHash }) => {
+  const renderItem = (cas: Case) => {
+    const { can } = getPermissionsForCase(cas.twilioWorkerId, cas.status);
+    const onClickViewCase = () => {
+      // load case modal? or page?
+    };
```
GPaoloni, line 39: Do current thing and we can override later? Just to be consistent with the below rendering for contacts.

plugin-hrm-form/src/components/profile/ProfileIdentifierBanner/ProfileIdentifierBanner.tsx
```diff
@@ -0,0 +1,152 @@
...
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ConnectedProps<typeof connector>;
+
+const ProfileIdentifierBanner: React.FC<Props> = ({ task, openProfileModal, modalOpen }) => {
+  let formattedIdentifier: string = '';
+  let identifierIdentifier: Identifier['identifier'] = '';
+  if (isTwilioTask(task)) {
+    formattedIdentifier = getFormattedNumberFromTask(task);
+    identifierIdentifier = getNumberFromTask(task);
+  }
```
GPaoloni, line 50: Prefer constants over variables when possible?

plugin-hrm-form/src/components/profile/ProfileIdentifierBanner/ProfileIdentifierBanner.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const { identifier } = useIdentifierByIdentifier({ identifierIdentifier, shouldAutoload: true });
+  const profileId = identifier?.profiles?.[0]?.id;
+
+  // Ugh. The previous contacts count is off by one because we immediately create a contact when a task is created.
+  // contacts should really have a status so we can filter out the "active" contact on the db side.
```
GPaoloni, line 55: Fair, should be simple to do.

plugin-hrm-form/src/components/profile/ProfileIdentifierBanner/ProfileIdentifierBanner.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const contactsCountState = useProfileProperty(profileId, 'contactsCount') || 0;
+  const contactsCount = contactsCountState ? contactsCountState - 1 : 0;
+  const casesCount = useProfileProperty(profileId, 'casesCount') || 0;
+
+  // Hooks can't be called conditionally, so we have to do this here
+  if (!isTwilioTask(task)) return null;
```
GPaoloni, line 61: Maybe move this statement one level up in the React tree so the `ProfileIdentifierBanner` is not rendered at all? Seems simpler and cheaper in term of resources :p

plugin-hrm-form/src/components/profile/ProfileRelationshipList.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const renderLoadMore = () => {
+    if (loading) return <div>Loading...</div>;
+    if (exhausted) return null;
+
+    return (
+      <button type="button" onClick={incrementPage}>
+        Load more
+      </button>
+    );
+  };
```
GPaoloni, line 73: I know we chatted about this but I don't remember the outcome haha. Are we moving to pagination for consistency or is this the new way?

plugin-hrm-form/src/components/profile/ProfileRouter.tsx
```diff
@@ -0,0 +1,84 @@
...
+  task: RouterTask;
+};
+
+// eslint-disable-next-line no-use-before-define
+type Props = OwnProps & ConnectedProps<typeof connector>;
+
+const PROFILE_ROUTES = {
+  profile: {
+    routes: ['profile'],
+    renderComponent: props => <ProfileTabs {...props} />,
```
GPaoloni, line 40: Adding types to this various subroute props parameters in the render functions may save us from some weird bugs, forcing to verify the props validity before the actual render. Just a thought, not something we have to do for now :p

plugin-hrm-form/src/states/profile/hooks/useIdentifier.ts
```diff
@@ -0,0 +1,93 @@
...
+  identifier?: Identifier;
+};
+
+/**
+ * Load a identifier by id into redux
+ * @param {UseIdentifierByIdentifierLoaderParams} identifierId - The id of the identifier to load
+ * @returns {UseIdentifierByIdentifierLoaderReturn} - State and actions for the identifier
+ */
+export const useIdentifierByIdentifierLoader = ({
+  identifierIdentifier,
```
GPaoloni, line 49: Identifier-ception is real xD

plugin-hrm-form/src/states/profile/reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+    shouldUseReducer: shouldUseRelationshipReducer,
+    reducer: relationshipReducer,
+  },
+];
+
+export function reduce(state = t.initialState, action: t.ProfileActions): t.ProfileState {
+  const reducer = reducers.find(r => r.shouldUseReducer(action));
+  if (reducer) {
+    return reducer.reducer(t.initialState)(state, action);
+  }
```
GPaoloni, line 46: I'm confused about this chunk here. Don't we want all of this reducers to be mounted at once? Does this means that parts of Redux are "unloaded" after being needed? Or that they are not "loaded" until they are?

plugin-hrm-form/src/states/routing/types.ts
```diff
@@ -167,13 +199,19 @@ type OtherRoutes =
   | TabbedFormRoute
   | SearchRoute
   | ContactRoute
-  | CaseListRoute;
+  | CaseListRoute
+  | ProfileRoute
+  | ProfileEditRoute
+  | ProfileFlagEditRoute
+  | ProfileNoteEditRoute;
```
GPaoloni, line 206: Won't it be better if this are all grouped as subroutes of a single parent route, since this are all specific to viewing profile list / profile details?

plugin-hrm-form/src/types/types.ts
```diff
@@ -332,14 +336,49 @@ export function isOfflineContact(contact: Contact): boolean {
...
+export type Identifier = {
+  id: number;
+  identifier: string;
+  accountSid: string;
+  createdAt?: string;
+  updatedAt?: string;
+  profiles: Profile[];
+};
+
+export type ProfileNote = {
```
GPaoloni, line 360: ProfileSection might be a better name? Or we prefer to hide that in the backend side and treat them as "notes" here, even tho only some of them will be of `sectionType: note`? BTW the property is `type` here but `sectionType` in the backend.

GPaoloni (commented):
NOTE: I changed the base to `master` because it was easier to me review this way rather than the PRs chain. Base should be change to [bo/chore/CHI-2260-relationshipPagination](https://github.com/techmatters/flex-plugins/tree/bo/chore/CHI-2260-relationshipPagination), but I'm not sure if that will hide feedback left here :p 

Looks good to me so far, but I'm not sure if this is the "final form" or if it needs more iteration before merging. Sorry for the many comments but this is huge haha

plugin-hrm-form/src/components/profile/ProfileCases.tsx
robert-bo-davis: Yes! This case file is currently an incomplete placeholder since cases were broken I didn't finish it. But these are all good suggestions.

plugin-hrm-form/src/components/profile/ProfileRelationshipList.tsx
robert-bo-davis: I believe we are moving to "regular pagination" I just didn't get around to it yet.

plugin-hrm-form/src/types/types.ts
robert-bo-davis: I believe that Mythily is handling rename of all of my misnamed note->section in her followup PR

plugin-hrm-form/src/components/profile/ProfileIdentifierBanner/ProfileIdentifierBanner.tsx
robert-bo-davis: This was all last minute and hacky in prep for vacation. Thanks for calling it out. I've refactored a bit in #1807 so that the underlying hooks and getNumber functions handle type checking correctly so it is properly abstracted away and the component can just deal with the logic it actually cares about. Thoughts on that approach?

plugin-hrm-form/src/components/profile/ProfileRouter.tsx
robert-bo-davis: Good call! Moved to a common type for all of these components in #1807 that should ALWAYS just receive a task and a profile ID and then rely on redux to figure out what props they need.

plugin-hrm-form/src/states/profile/reducer.ts
robert-bo-davis: I was also confused and making this problem overly complicated. I forgot that, even though my gut says it is confusing, it is recommended that [multiple reducers should be able to respond to a single action](https://redux.js.org/style-guide/#allow-many-reducers-to-respond-to-the-same-action) and handle updates to their chunk of the state. I've updated to this pattern in #1807 which simplifies this chunk of code significantly. Let me know if you have questions about the new approach.

plugin-hrm-form/src/states/routing/types.ts
robert-bo-davis: That sounds correct to me. I don't think there should be big complex global types if we move to the pattern where individual domain specific routers output all of the types and consts needed to route to themselves. This is really just me trying to follow the existing pattern as well as possible. I think we should deal with it in a future refactor.

GPaoloni (approved):
LGTM

For anybody else reading this, we discussed the state of this with Bo and while there are some incoming changes, we think is better to merge this since it's behind a feature flag.

PR #1786, opened 2023-11-01
Added a placeholder 'could not load' page if contact fails to create

stephenhand:
## Description

Following on from yesterday's discussion I reworked the error handling case around not being able to create a contact

Now if a contact is not created when starting a task for any reason, a placeholder UI is put in place of the forms with 2 buttons, one to retry creating the contact, and another to complete the task

### Checklist
- [ ] New tests added
- N/A Feature flags added
- [X] Strings are localized

### Related Issues
CHI-2320

### Verification steps

To test this you need to disable contact creation in HRM somehow, possibly just override the HRM URL in flex with an invalid oner
You should then check that the UI appears when acceopting a new task and the buttons work as expevcted


murilovmachado (approved):
The code looks good to me!

One thing not to forget is to continue this conversation with Jana and the product team. For me, as a dev, the error message and action buttons are very clear. But I'm not sure if the counselor will understand this ContactNotLoaded screen. There may be things we can think design-wise to make it easier for the counselor.

PR #1788, opened 2023-11-01
Summary tab - Ability to add contact to case

acedeywin:

## Description

- New feature: Modifying the existing “Add to New Case” button on the Summary tab of an Active Contact.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes CHI-1972

### Verification steps

- Create an offline contact 
-  Verify that we have a new "Add to New Case” drop-up button with "New Case" and "Existing Case" as options 
- The "New Case” button should behave the same as the previous “Add Contact to Case” button
- The “Existing Case” button should take you to the Search tab


stephenhand (commented):
You should merge up from master because the other code that adds a caseId property so that conflict can be resolved, because I think that's causing the E2E test fails

plugin-hrm-form/src/types/types.ts
```diff
@@ -168,6 +168,7 @@ export type Contact = {
   accountSid: string;
   twilioWorkerId: string;
   number: string;
+  caseId?: number | null;
```
stephenhand, line 171:
```suggestion
  caseId?: number;
```

You only need the ? for optional parameters

plugin-hrm-form/src/components/tabbedForms/BottomBar.tsx
```diff
@@ -140,33 +168,68 @@ const BottomBar: React.FC<
...
-                >
-                  <FolderIcon style={{ fontSize: '16px', marginRight: '10px' }} />
-                  <Template code="BottomBar-AddContactToNewCase" />
-                </StyledNextStepButton>
+            <AddNewCaseDropdown
+              handleNewCaseType={handleSubmitIfValid(handleOpenNewCase, onError)}
+              handleExistingCaseType={openSearchModal}
+              dropdown={dropdown}
+            />
+            {isAddedToExistingCase ? (
```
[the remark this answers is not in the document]
stephenhand: You will also need to add a check on the `enable_case_merging` feature flag here once my PR is merged, if you merge this first you can do it in a separate PR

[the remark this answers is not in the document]
acedeywin: @stephenhand  - Feedback has been implemented.

plugin-hrm-form/src/components/tabbedForms/BottomBar.tsx
```diff
@@ -140,33 +168,70 @@ const BottomBar: React.FC<
...
+                      {!dropdown && (
+                        <KeyboardArrowDownIcon
+                          style={{ fontSize: '20px', marginLeft: '10px', width: '24px', height: '24px' }}
+                        />
+                      )}
+                    </StyledNextStepButton>
+                  </Box>
+                )}
+            {isAddedToCase ? (
+              <SaveAndEndContactButton
```
stephenhand, line 213: I think from the mocks the save button needs to be red whether the contact has been added to a case or not, so you don't need 2 different buttons depending on the state

Remember that the red button also needs the ` {isSubmitting ? <CircularProgress size={12} style={{ position: 'absolute' }} /> : null}` spinner code from the previous save button including

acedeywin: @stephenhand  - It's done.

stephenhand (approved):

PR #1792, opened 2023-11-03
CHI-1909: Add to case search

stephenhand:
## Description

* Add the 'Add To Case' button as specified by CHI-1909
* Added front end permissions & feature flag
* Updated search redux state to have flags indicating that the case or contact search need refreshing from the back end. These get set when actions are processed in the reducer that will modify data potentially returned by the search in HRM.
* Updated the search results page to dispatch the relevant reload action when these flags are set
* Removed some of the component level specific workarounds to refresh search results now we use this flag 

### Checklist
- [ ] New tests added
- [X] Feature flags added
- [ ] Strings are localized

### Related Issues
CHI-1909

### Verification steps

See ticket


murilovmachado:
The error above is actually happening on Aselo Development as well, so it's not related to this PR.

plugin-hrm-form/src/states/search/reducer.ts
```diff
@@ -110,9 +117,35 @@ const contactUpdatingReducer = (state: SearchState, action: ContactUpdatingActio
...
+      action.type,
+    )
+  ) {
+    state = {
+      ...state,
+      tasks: Object.fromEntries(
+        Object.entries(state.tasks).map(([key, value]) => [key, { ...value, caseRefreshRequired: true }]),
+      ),
+    };
+  }
```
murilovmachado, line 148: These 2 `if` clauses here break a little bit the pattern of how we implement our `reduce` functions, by modifying `state` before-hand. 

Although it doesn't look super nice to me, I bet you did this because the other options could be more confusing to understand (these same changes scattered throughout the `switch` cases, mixed with other state changes). 

So I'm okay with this.

murilovmachado (approved):
The code looks good to me. I couldn't test it because of the 401 error, though.

plugin-hrm-form/src/states/search/reducer.ts
stephenhand: Yes, these are kind of 'pre pass reducers' where many actions can have the same effect on the state

PR #1801, opened 2023-11-07
Unmask a phone number for a specific helpline

mythilytm:
Primary reviewer:

## Description
- By default, the caller's phone number is masked. KHP requested cases where the phone number needs to be revealed in an emergency. 
- Added a button that will toggle the visibility of this phone number. 
- Used MUI component Popper instead of a custom dialog because it gets cut off by the adjacent panels. 

### Checklist
- [n/a] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized
- [] Tested for chat contacts
- [] Tested for call contacts

### Related Issues
Fixes # https://tech-matters.atlassian.net/browse/CHI-2242

### Verification steps
Screenshot 2023-11-07 at 1 13 41 PM
 1. Call the voice channel. 
 2. Run this branch, and test the behavior of the button and the pop-up.


plugin-hrm-form/src/HrmFormPlugin.tsx
```diff
@@ -77,7 +77,27 @@ const setUpComponents = (
   translateUI: (language: string) => Promise<void>,
 ) => {
   const { canView } = getPermissionsForViewingIdentifiers();
-  const maskIdentifiers = !canView(PermissionActions.VIEW_IDENTIFIERS);
+  // const maskIdentifiers = !canView(PermissionActions.VIEW_IDENTIFIERS);
+  const maskIdentifiers = true;
```
mythilytm, line 81: Temporary for testing

plugin-hrm-form/src/HrmFormPlugin.tsx
```diff
@@ -77,7 +77,27 @@ const setUpComponents = (
...
+      theirFriendlyNameOverride: false,
+      yourFriendlyNameOverride: true,
+    };
+    Flex.MessageList.Content.remove('0');
+    // Masks TaskInfoPanelContent - TODO: refactor to use a react component
+    const strings = getTemplateStrings();
+    strings.TaskInfoPanelContent = strings.TaskInfoPanelContentMasked;
+    strings.CallParticipantCustomerName = strings.MaskIdentifiers;
+
+    if (setupObject.helplineCode === 'as' || setupObject.helplineCode === 'ca') Components.setUpViewMaskedVoiceNumber();
```
mythilytm, line 99: Temporary for testing

plugin-hrm-form/src/components/common/ViewTaskNumber.tsx
```diff
@@ -0,0 +1,68 @@
...
+const ViewTaskNumber = ({ task }: Props) => {
+  const [viewNumber, setViewNumber] = useState(false);
+  const viewNumberRef = useRef(null);
+
+  const toggleViewNumber = () => {
+    setViewNumber(!viewNumber);
+  };
+
+  const handleClose = () => {
+    setViewNumber(!viewNumber);
```
robert-bo-davis, line 38: Would a single toggleViewNumber function make more sense than duplicating the same code in a handler?

plugin-hrm-form/src/HrmFormPlugin.tsx
robert-bo-davis: I think we should probably use a configuration flag for this feature instead of a helplineCode switch. If it were me, I'd probably move the "switch" logic into the setUpViewMaskedVoiceNumber function so that all responsibility for the feature is delegated to that chunk of code, but I think that is more of a personal preference thing. We should definitely avoid switching on helplineCode though. 

The switch should also probably take into account if the user has `VIEW_IDENTIFIERS` permission as well. There is no reason to show the component if the user can already see the identifier, right?

plugin-hrm-form/src/components/common/ViewTaskNumber.tsx
```diff
@@ -0,0 +1,68 @@
...
+        </Paper>
+      </Popper>
+    ) : null;
+
+  return (
+    <>
+      <StyledUnmaskButton onClick={toggleViewNumber} ref={viewNumberRef}>
+        {viewNumber ? <EyeOpenIcon /> : <EyeCloseIcon />}
+      </StyledUnmaskButton>
+      {renderTaskNumberPopper()}
```
robert-bo-davis, line 63: Render functions are always a balancing act. I probably tend to overuse them and don't want to infect the rest of the team with my bad habits. @GPaoloni do you have an opinion about when to use them and when to avoid them?

robert-bo-davis (changes requested):
A few non critical notes. I think avoiding switching on helplineCode and only showing the component if the identifiers are masked are the only MUSTs. All of the rest are SHOULDs or COULDs most of which I'd like to have Gian's opinion on.

plugin-hrm-form/src/components/common/ViewTaskNumber.tsx
robert-bo-davis: I tend to use them for chunks of code that I think: Hmm... this chunk will probably be a separate component in the near future, may as well go ahead and segment it out. Or if there is more complexity to the "what should I display" than a `?: null` ternary. But those are just some general rules of thumb and this could very well be a remnant of my POC that I don't remember writing. 🤣

plugin-hrm-form/src/styles/HrmStyles.tsx
```diff
@@ -544,6 +544,15 @@ export const TransferStyledButton = styled('button')<TransferStyledButtonProps>`
 `;
 TransferStyledButton.displayName = 'TransferStyledButton';
 
+export const StyledUnmaskButton = styled(TransferStyledButton)`
```
robert-bo-davis, line 547: I lost a comment, I think. Hmm. 

Do we want to bundle all of this into a component folder instead of spreading very related code across a bunch of global spaces?

Something like creating a `components/common/ViewTaskNumber` directory that has an icons directory, an index.tsx, and a styles.tsx so all of the related code lives together? @GPaoloni is there an existing standard for something like this?

mythilytm: Good point. I am curious what @GPaoloni suggests

GPaoloni (approved):
Code looks to me. One thing is that the client profiles identifier can still be seen

Screenshot from 2023-11-14 19-30-11

mythilytm:
To test this, the permission for `canViewIdentifiers` should be set to false and the newly added configFlag `enableUnmaskingCalls` should be set to true.
https://github.com/techmatters/flex-plugins/blob/CHI2242-mask-number/plugin-hrm-form/src/utils/setUpComponents.tsx#L234

mythilytm:
> Code looks to me. One thing is that the client profiles identifier can still be seen

For testing, the `canViewIdentifiers` permission should be set to false and the newly added configFlag `enableUnmaskingCalls` should be changed to true

mythilytm:
@GPaoloni 
I am unable to figure out why [separating the masking Identifiers logic](https://github.com/techmatters/flex-plugins/pull/1801/commits/03755c52c442828b3e94e9d4b688d4545f9014a8) is causing a unit test failure related to transfer button. 
https://github.com/techmatters/flex-plugins/actions/runs/6879656796/job/18712071058

I have tried debugging this but haven't figured it out. Behavior works fine when run locally. Curious if you have any thoughts.

plugin-hrm-form/src/channels/setUpChannels.tsx
```diff
@@ -28,6 +28,7 @@ import SmsIcon from '../components/common/icons/SmsIcon';
 import * as TransferHelpers from '../utils/transfer';
 import { colors, mainChannelColor } from './colors';
 import { getTemplateStrings } from '../hrmConfig';
+// import { maskIdentifiersByChannel } from '../components/common/MaskingIdentifiers';
```
robert-bo-davis, line 31: This probably needs to be re-enabled, right?

robert-bo-davis (approved):
LGTM once LineChatChannel masking is verified working

plugin-hrm-form/src/channels/setUpChannels.tsx
```diff
@@ -169,8 +169,6 @@ export const setupLineChatChannel = maskIdentifiers => {
     main: icon,
   };
 
-  if (maskIdentifiers) maskIdentifiersByChannel(LineChatChannel);
```
robert-bo-davis, line 172: Is this change going to cause LineChatChannels to always be unmasked, even when they are meant to be masked?

PR #1810, opened 2023-11-09
Chi 1972 summary tab ability add contact case

acedeywin:

## Description

- New Feature: This is the final part for this [PR](https://github.com/techmatters/flex-plugins/pull/1788) that addresses how the search results that appear should be.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes [CHI-1972](https://tech-matters.atlassian.net/browse/CHI-1972)

### Verification steps

- Create an offline contact
- Verify that we have a new "Add to New Case” drop-up button with "New Case" and "Existing Case" as options
- The "New Case” button should behave the same as the previous “Add Contact to Case” button
- The “Existing Case” button should take you to the Search tab where the header would be "Search for Cases"
- Search for a Case
- Verify that the search results only return cases

[CHI-1972]: https://tech-matters.atlassian.net/browse/CHI-1972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugin-hrm-form/src/components/NavigableContainer.tsx
```diff
@@ -31,12 +31,14 @@ import * as CaseActions from '../states/case/actions';
 import * as RoutingActions from '../states/routing/actions';
 import { HeaderCloseButton, HiddenText, Row, StyledBackButton } from '../styles/HrmStyles';
 import { LargeBackIcon, NavigableContainerBox, NavigableContainerTitle } from '../styles/NavigableContainerStyles';
+import { setSearchExistingCase } from '../states/search/actions';
```
stephenhand, line 34: NavigableContainer is a generic navigation component, it should not have stuff specific to search in it. There is already a mechanism for injecting custom close behaviour via a function, we should use that

plugin-hrm-form/src/states/search/reducer.ts
```diff
@@ -279,6 +281,19 @@ export function reduce(
         },
       };
     }
+    case t.SEARCH_EXISTING_CASE: {
```
stephenhand, line 284: I think this property would be better implemented as an additional property of the navigation route (e.g. `{ route: 'search', subroute: 'form', action: 'select-case' }` & `{ route: 'search', subroute: 'case-result', action: 'select-case' }` rather than a property that needs to be switched on and off with a separate action

stephenhand (changes requested):

plugin-hrm-form/src/states/search/reducer.ts
acedeywin: @stephenhand  - All feedback has been implemented

plugin-hrm-form/src/states/routing/types.ts
```diff
@@ -39,9 +39,10 @@ export type TabbedFormRoute = {
   autoFocus?: boolean;
 } & RouteWithModalSupport;
 
-export type SearchRoute = RouteWithModalSupport & {
```
stephenhand, line 42: Why is `RouteWithModalSupport` removed? Search routes need to support modals because cases and contact details appear as modals over search results

stephenhand (commented):
Code now looks good except for one q

Testing it though, it doesn't seem to change the state of the 'Add to case' button to 'Added To Case' immediately after adding to a case. You can still add to case, and the search results show the case you added in the correct state, but still allows you to connect to other cases.

This leads me to think that the search results in redux state are being updated correctly, but the contact under existingcontacts is not? It's strange because adding a case via the search tab updates the state immediately.

plugin-hrm-form/src/states/routing/types.ts
acedeywin: @stephenhand  - I can't recall removing this. It might have been when I was fixing the merge conflict. I will pull the update from master now.

acedeywin:
> Code now looks good except for one q
> 
> Testing it though, it doesn't seem to change the state of the 'Add to case' button to 'Added To Case' immediately after adding to a case. You can still add to case, and the search results show the case you added in the correct state, but still allows you to connect to other cases.
> 
> This leads me to think that the search results in redux state are being updated correctly, but the contact under existingcontacts is not? It's strange because adding a case via the search tab updates the state immediately.

@stephenhand - It’s not really clear to me because it works fine while testing it. Once a contact gets added to a case and you click on the Cancel button, you will see that the contact is already added to a case and you won’t be able to search or add a case to it. Unless maybe you mean something different?

stephenhand (approved):
LGTM

PR #1815, opened 2023-11-09
Connect Contact to Case Banners

murilovmachado:
## Description
Case Merging Banners:
- Banner for when a contact is added to a case
- Banner for when the contact is removed from a case

TODO:
- Unit tests
- Translations
- Banner inside the case itself (when creating a new case)
- Double check feature flags

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-1861

### Verification steps
Connect a contact to a case, and check the banner appears. Interact with the banner(s).


plugin-hrm-form/src/states/case/findCaseByTaskSid.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import { RootState } from '..';
+import { Case } from '../../types/types';
+import { connectedCaseBase, namespace } from '../storeNamespaces';
+
+const findCaseByTaskSid = (state: RootState, taskSid: string): Case =>
+  state[namespace][connectedCaseBase].tasks[taskSid]?.connectedCase;
+
+export default findCaseByTaskSid;
```
stephenhand, line 24: I think our agreed naming convention for these sort of state helpers is `selectXXX` - so `selectCaseByTaskId` in this case

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -280,6 +280,17 @@ export async function connectToCase(contactId: string, caseId: number) {
   return fetchHrmApi(`/contacts/${contactId}/connectToCase`, options);
 }
 
+export async function removeFromCase(contactId: string, caseId: number) {
```
stephenhand, line 283: Thanks for implementing the client for this! Was meaning to get around to it

plugin-hrm-form/src/components/caseMergingBanners/state.ts
```diff
@@ -0,0 +1,87 @@
+/**
```
stephenhand, line 1: Should this be under the `states` directory?

plugin-hrm-form/src/components/caseMergingBanners/state.ts
```diff
@@ -0,0 +1,87 @@
...
+}));
+
+type CaseMergingBannersAction =
+  | ReturnType<typeof showConnectedToCaseBannerAction>
+  | ReturnType<typeof showRemovedFromCaseBannerAction>
+  | ReturnType<typeof closeRemovedFromCaseBannerAction>;
+
+type CaseMergingBannersState = {
+  [contactId: string]: {
+    showConnectedToCaseBanner: boolean;
```
stephenhand, line 54: Is managing whether we show the connected banner with a separate flag necessary? Can we not always infer whether we need to show it from the status of the contact? For the logic around showing buttons or not, I locate the 'taskContact' in the redux state (i.e. a contact with the taskId matching the current task), and use it's caseId to determine the connected state: https://github.com/techmatters/flex-plugins/pull/1827

I can see why you need to track the removed banner state though, because that can be dismissed by the user.

Certainly not a blocker, if what you have works, that's great - just a suggestion

stephenhand (approved):
Some non blocking comments - I haven't tested it yet, but will do and let you know if I have any issues

plugin-hrm-form/src/states/case/findCaseByTaskSid.ts
murilovmachado: Yeah, I remember we agreed on `selectXXX`, but then I only saw `findByXXX` in the code, that's why I used `findCaseByTaskSid`. But I'm happy to change it for `selectCaseByTaskSid`, it's inline with what we agreed and with redux's official suggestion.

plugin-hrm-form/src/components/caseMergingBanners/state.ts
murilovmachado: I think it's more DDD-ish to keep all case merging banners stuff together, but our code is separating the state to the `states` folder. I have a preference for keeping it here, but I can move it to `states` if you think it'd be better to follow the pattern we were already using.

plugin-hrm-form/src/components/caseMergingBanners/state.ts
murilovmachado: What you say is true for now, but I can see we needing to manage `showConnectedToCaseBanner` in the scenario of adding an **existing** contact to a case. Unless we change the expectation of this blue banner to actually be always show for contacts connected to cases (regardless if it was just added or not).

PR #1830, opened 2023-11-15
Client Profiles - Fix case viewing and pagination bugs

robert-bo-davis:
## Description

This fixes viewing case from the profile relationship lists and some pagination nits in the profile paginator.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


plugin-hrm-form/src/components/tabbedForms/TabbedForms.tsx
```diff
@@ -198,9 +199,25 @@ const TabbedForms: React.FC<Props> = ({
   }
 
   if (currentRoute.route === 'case') {
+    // This is a dirty hack so that case viewing works for the create case form and
```
robert-bo-davis, line 202: I do not like this, but it is the best I could come up with quickly without a major refactor. Wanted to draw @stephenhand's attention here.

plugin-hrm-form/src/states/profile/profiles.ts
```diff
@@ -32,13 +32,21 @@ type CommonRelationshipParams = {
...
-    const offset = page * PAGE_SIZE;
-    const limit = PAGE_SIZE;
+  async ({ profileId, type, page, loadedPage }: LoadRelationshipAsyncParams): Promise<any> => {
+    /*
+    This handles skipping forward several pages at once and loads the pages in between into state
+    in a single request. This could end up with a large payload if there are dozens of pages and
+    hundreds of records.
+    If the tradeoff is not worth it, we can change this to only load the current page and only store
+    it in state like the current behavior.
+    */
```
robert-bo-davis, line 47: This is also not my favorite chunk of code ever. Some of this is from moving from a "load more" style loader back to a "regular paginator". This could be be refactored so that each page is a key on an object so you can skip around and only reload data if needed. It has been a long time since I've done numbered page pagination and I've never tried to keep it in state. Most of the projects I've worked on have used order bys, filters, and load more style loading so folks could get at the data they want. In my experience, folks *rarely* care about the data in the middle of the list enough to try to jump straight to it.

GPaoloni: The only thing I want to know here is, what happens if more records are added to the list? If user goes to page 2, then back to page 1, but in the middle another record appeared, this new record (the most recent one), should appear in the page 1. Is this behavior working that way? 
One thing I imagined when you said that we want to separate "the records of the db" from "the ids that are shown in a list list" is that we'll have the ids that correspond to one page (plus the minimal info for that page), and that is updated each time you paginate by call the DB, and in a separate reducer the actual objects that were accessed and loaded from the DB (like the entire profile). Don't know if I explain the idea, happy to chat about this, but is that the direction you are aiming here?

GPaoloni (approved):
I think I get the PR but probably worth chatting about this with you. Anyways, if it fixes the issues it still LGTM!

PR #1841, opened 2023-11-16
Banner on created Case

murilovmachado:
## Description
Shows a banner when a case is created with:
- Text: Case # 1234 created and contact added
- Button: Cancel Case

TODO:
- ~Fix visual issues~ (done)
- Tests will come in a follow up PR

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-2380

### Verification steps
- Fill a contact
- Create a new Case
- Verify the banner shows up
- Verify the cancel button works


stephenhand:
Are the banners for existing cases coming in a further PR? At the moment we only get the 'Added to Case' banner for new cases, we don't see it when you add an existing case, or go back into a new case after you added it then go back to the contact

One minor thing I also noticed - the 'Add To Case' button on the summary tab doesn't seem to update state immediately when you click the 'Cancel Case' link. It still says 'Added To Case' - if you go to another tab and come back it has changed back to the 'Add To Case' button as expected

murilovmachado:
@stephenhand I've fixed the issue that you pointed on the bottom bar button, and also fixed the styling.
I'll keep this PR related to New Case only, and talk to the product team how it should be on existing cases.

murilovmachado:
Resolving these conflicts is not very trivial. Please don't review this until the conflicts are resolved.

plugin-hrm-form/src/components/case/CaseHome.tsx
```diff
@@ -243,7 +244,7 @@ const CaseHome: React.FC<Props> = ({
         )}
         <Box marginTop="13px">
           <CaseDetailsComponent
-            caseId={id.toString()}
+            caseId={id}
```
murilovmachado, line 247: Why did we change this be a `string`?

plugin-hrm-form/src/components/caseMergingBanners/CaseCreatedBanner.tsx
```diff
@@ -0,0 +1,96 @@
...
+  navigateBack,
+}) => {
+  const handleCancelCase = async () => {
+    /**
+     * TODO: should we move the next 3 lines to a single redux action?
+     * How can we dispatch other actions from within an action?
+     */
+    const contactIds = cas.connectedContacts.map(c => c.id);
+    await Promise.all(contactIds.map(id => removeContactFromCase(id, caseId)));
+    await cancelCase(caseId);
```
murilovmachado, line 75: I think I should change at least the last line to be an action dispatch as part of this PR (instead of a direct service call)

plugin-hrm-form/src/components/caseMergingBanners/CaseCreatedBanner.tsx
```diff
@@ -0,0 +1,97 @@
...
+  navigateBack,
+}) => {
+  const handleCancelCase = async () => {
+    const contactIds = cas.connectedContacts.map(c => c.id);
+    await Promise.all(contactIds.map(id => removeContactFromCase(id, caseId)));
+
+    // Navigating back before removing the case provides a better user experience.
+    navigateBack(taskSid);
+    showRemovedFromCaseBanner(contactId);
+    await cancelCase(caseId, taskSid);
```
murilovmachado, line 78: `cancelCase` is now a redux action.

One thing that I've tried was just removing the case, without disconnecting the contacts before-hand, hoping that the backed would "cascade" the changes. But it didn't work, the backend just returns 500 when a contact has a foreign key pointing to the case being deleted

plugin-hrm-form/src/components/caseMergingBanners/CaseCreatedBanner.tsx
murilovmachado: Done

plugin-hrm-form/src/components/caseMergingBanners/CaseCreatedBanner.tsx
stephenhand: Is that on deleting the case or saving the contact?

plugin-hrm-form/src/states/case/saveCase.ts
```diff
@@ -123,5 +131,9 @@ export const saveCaseReducer = (initialState: SaveCaseReducerState) =>
     handleFulfilledAction(handleAction, createCaseAsyncAction.fulfilled),
     handleRejectedAction(handleAction, createCaseAsyncAction.rejected),
 
+    handlePendingAction(handleAction, cancelCaseAsyncAction.pending),
```
stephenhand, line 134: I guess all the necessary redux updates are already taken care of in the 'removeFromCase' action?

stephenhand (approved):
LGTM!

plugin-hrm-form/src/components/caseMergingBanners/CaseCreatedBanner.tsx
murilovmachado: on deleting the case

plugin-hrm-form/src/states/case/saveCase.ts
murilovmachado: For the feature to work, yes, everything is already handled by `removeFromCase`. But there's more redux state that is updated but this action (on connectedCase reducer)

PR #1858, opened 2023-11-21
Chi 1974 case merging design cleanup

acedeywin:
Primary reviewer:

## Description

- The mouseover state on the menu options for “new case” and “existing case” now span the entire row. 
- The button text is now “Add to Case,” instead of “Add to New Case.”
- The header of the page returning the list of cases that satisfy the search results now says “Select Case” instead of  “Search Results.”
- When selecting “add contact to a new case” and then select "cancel case and return to contact," it now return to the default “Add to Case” dropdown - Update: This is no longer relevant as the "Cancel case and return to contact" button has been removed.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes [CHI-1974](https://tech-matters.atlassian.net/browse/CHI-1974)

### Verification steps

[CHI-1974]: https://tech-matters.atlassian.net/browse/CHI-1974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugin-hrm-form/src/components/case/Case.tsx
```diff
@@ -352,7 +352,7 @@ const mapDispatchToProps = (dispatch, { task }: OwnProps) => {
     loadContacts: bindActionCreators(ContactActions.loadContacts, dispatch),
     updateCaseAsyncAction: (caseId: CaseType['id'], body: Partial<CaseType>) =>
       caseAsyncDispatch(updateCaseAsyncAction(caseId, task.taskSid, body)),
-    disconnectFromCase: (contactId: string) => caseAsyncDispatch(connectToCaseAsyncAction(contactId, null)),
+    disconnectFromCase: (contactId: string) => caseAsyncDispatch(removeFromCaseAsyncAction(contactId, null)),
     submitContactFormAsyncAction: (
```
stephenhand, line 356: We should remove the caseId from the removeFromCaseAsyncAction action and stop sending a body with the DELETE method in ContactService.ts - it is redundant - not blocking though

stephenhand (approved):
LGTM - one comment but not a blocker for this PR

plugin-hrm-form/src/components/case/Case.tsx
acedeywin: @stephenhand  - It's done.

PR #1860, opened 2023-11-21
CHI-2395: ability to get a shell with an assumed role

robert-bo-davis:
## Description

This adds the ability to get a shell with an assumed role via the make command:

```
make shell-assume-role ROLE=${someRoleName}
```


nickhurlburt (approved):
Thanks!

PR #1861, opened 2023-11-21
CHI-2396 fix case merging off

stephenhand:
## Description

* Convert the 'Add to Case' button bac to a single action to add new case when flag is off
* Validates form when adding a case if flag is off
* The close cross on the new case modal removes & deletes the case
* An explicit Cancel button with the same effect is added to the bottom bar

### Checklist
- N/A New tests added
- N/A Feature flags added
- [X] Strings are localized

### Related Issues
CHI-2396

### Verification steps

* Disable case merging for the helpline you are testing
* Refresh the browser
* Verify that the old workflow of being forced to add only a new case at the end is in place (with some cosmetic changes)
* Reenable the flag & refresh the browser
* Ensure that everything for case merging still works as expected

### DIFF
--- a/plugin-hrm-form/src/___tests__/states/case/actions.test.ts
+++ b/plugin-hrm-form/src/___tests__/states/case/actions.test.ts
@@ -15,7 +15,7 @@
  */
 
 import { Case } from '../../../types/types';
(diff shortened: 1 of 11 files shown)


murilovmachado (approved):
LGTM

PR #1865, opened 2023-11-21
CHI-2291: POC flag edit improvements

robert-bo-davis:
## Description

This includes improved inline editing so that clicking into the profiles section open edit mode and clicking out exits edit mode.

It also moves away from the hacky select implementation to a custom component for flag editing.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


plugin-hrm-form/src/components/profile/profileFlag/ProfileFlagSection.tsx
```diff
@@ -0,0 +1,62 @@
...
+          profileId={profileId}
+          task={task}
+          handleClose={closeEditProfileFlags}
+          modalRef={profileFlagsModalRef}
+        />
+      ) : (
+        <ProfileFlagsView
+          title="Edit Statuses"
+          type="button"
+          style={{ display: 'flex', width: '-webkit-fill-available' }}
```
GPaoloni, line 52: I prefer to have the styles moved to the styled components unless this is used in other places. I'm seeing a lot of inlined styles in this PR, I'm fine with it but let's move them when possible.

GPaoloni (approved):
LGTM, thanks for the work on this!

The flags input is not very accesible (e.g. can't tab into it), but since we'll continue iterating on it don't spend any more time on that, just noting so we remember to come back to it in the near future.

PR #1866, opened 2023-11-21
CHI-1971: Added To Case banner state inferred from contact state

stephenhand:
## Description

Since we now know we always want to show the 'Added To Case' banner when a contact is part of a case, we can infer whether to show it from the state of the contact rather than a separate state. This should address some inconsistencies in the banner state that are arising in QA

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-1971

### Verification steps

Try to reproduce point 1 in https://tech-matters.slack.com/archives/C05HT9K8602/p1700585180670849


murilovmachado (approved):
🚀

PR #1869, opened 2023-11-22
CHI-2301: Moved from search component state current page to routing redux

stephenhand:
## Description

* Move the 'current page' of search results from component state to the redux route - this should fix the issue described in the bug & also prevent the current page you are on being forgotten when switching tasks
* Update the save contact button on tabbed forms to be disabled using the redux 'updating' state, rather than its own component state

### Checklist
- [ ] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-2301

### Verification steps

See ticket


murilovmachado (approved):
🚀

PR #1892, opened 2023-11-29
CHI-2223: Fix the nag on browser close logic

stephenhand:
## Description

* The draft copy handling logic has changed, breaking the logic to display that window. To prevent this going forward, the confirm now appears if any contact is currently saving, not just if there are working copies
* The working copies for cases were not being fully cleaned up, meaning the confirm for cases kept appearing after changes were saved, this is now fixed

### Checklist
- [ ] New tests added
- [X] Feature flags added
- N/A Strings are localized

### Related Issues
Fixes #....

### Verification steps

* Make changes to a contact form - refreshing with unsaved changes on a tab should show the warning, after changing tabs, it should not
* Make changes to a case, saved or cancelled changes should not trigger the warning, but unsaved edits should
* Try to refresh whilst a contact is saving, it should trigger the warning, but not after save is complete


murilovmachado (approved):
🚀

PR #1897, opened 2023-11-30
CHI-1974: Figma fixes 7 & 8

stephenhand:
## Description

In order to address points 7 & 8 on here: https://www.figma.com/file/3m7IJjYbussNcTs2VRhfYz/case-merging---QA?type=design&node-id=67-760&mode=design&t=3GgjXswJVBnOrhTJ-0

* Prevent 'double banners' when removing a contact from a case then adding to another
* Add green tick to 'Added To Case' button.

### Related Issues
Fix CHI-1974

### Verification steps

See figma file


murilovmachado (approved):
looks good!

PR #1904, opened 2023-12-04
Fix when accepting Voice Calls

murilovmachado:
## Description
This may fix an issue when accepting voice calls.
This issue was happening in ZW Staging with Nick.
I could also reproduce it in Aselo Development using the network setting `Fast 3G`.


stephenhand (approved):
LGTM - is there an underlying issue we need to chase about the task being undefined aty this point, or is that expected?

PR #1908, opened 2023-12-05
CHI-2317: Refactor case redux

stephenhand:
## Description

* Connected Case now has a map property called 'cases' where all cases being used in memory are stored against their HRM ID (not their task)
* Case Search and Case List now load all their cases into the connectedCase store, and just keep IDs in their own store to track what cases to display in the result lists
* Case List now makes service requests via redux actions rather than directly from the component
* Moved the CaseCreatedBanner up into CaseHome so it is consistent with the 'Add To Case' banner - I'm not sure which of these is correct but this way was easier & cleaner to implement so going with this until told otherwise :-)

### Checklist
- [X] New tests added (more needed)
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes #....

### Verification steps

* General regression of case related functions


plugin-hrm-form/src/components/case/CaseHome.tsx
```diff
@@ -84,22 +85,22 @@ const CaseHome: React.FC<Props> = ({
   // eslint-disable-next-line sonarjs/cognitive-complexity
 }) => {
   if (!connectedCaseState) return null; // narrow type before deconstructing
-
+  const caseId = connectedCaseState.connectedCase.id.toString();
```
murilovmachado, line 88: By looking at this PR direction, I was expecting that `connectedCase.id` would already be of type`string`. Couldn't we change the type of `connectedCase` so we don't need to call `.toString()` here?

plugin-hrm-form/src/components/case/Timeline.tsx
```diff
@@ -72,27 +78,32 @@ const Timeline: React.FC<Props> = ({
...
   openViewCaseSectionModal,
   openAddCaseSectionModal,
+  connectedCase,
 }) => {
   const [mockedMessage, setMockedMessage] = useState(null);
 
+  if (!connectedCase || !timelineActivities) {
+    return null;
+  }
+  const caseId = connectedCase.id.toString();
```
murilovmachado, line 88: Same comment here. I wish any casting would happen way before in the chain, so that we can have `connectedCase.id` as type `string`, just like the type `Case`.

murilovmachado (approved):
Fantastic!
Just one minor comment on a type casting.

PR #1913, opened 2023-12-06
CHI-2454: Fix contact redux refs

stephenhand:
## Description

Mixed bag of fixes around contact loading

* Fixes referencing of contacts in redux so active contacts have a different reference to other contacts you view, so the active contact can't be deleted from state
* Adds a 'ResetRoute' changeRoute type, which resets the entire route, not just the modal - this fixes odd behaviour in the standalone search navigation
* Fixes 'View Contacts' button on case search reults page
* Moves responsibility for loading / dereferencing viewed contacts to the View Contact componnent. Also fixes loading logic to not spam requests

### Checklist
- [ ] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-2454

### Verification steps

See ticket,also check that 'View Contacts' switches to contact results as expected


acedeywin (approved):
LGTM

PR #1922, opened 2023-12-07
refactor: split tabbed-forms meta component into a router with sub components

robert-bo-davis:
Primary reviewer:

## Description

This splits the tabbed-forms meta component into separate components with a common Router type and abstracts the shared functionality between the TabbedFormsRouter and the ProfileRouter into a new Router component that can be used for all task based routers. (like search in the near future)

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


plugin-hrm-form/src/components/tabbedForms/TabbedFormsContact.tsx
```diff
@@ -0,0 +1,52 @@
+/**
```
stephenhand, line 1: Whilst I like the approach of using dedicated routers that follow a standard pattern, these intermediate glue components make me a little uneasy. I understand they are necessary in order to adhere to a 'single responsibility components' pattern - I just worry that spreading functionality for 1 part of the UI over too many different components will start to be just as confusing and hard to grok as the amorphous blob god components we are attempting to move away from

I suppose we just need to be super clear about WHY these guys exist, what is supposed to go in them, and what isn't. It's the sort of thing that looks pretty clear all together in 1 PR like this but can get lost as the months roll by after it's merged

robert-bo-davis: I agree wholeheartedly. In my mind the complexity of the glue components is really just a stop-gap to keep this PR as small and focused as possible while continuing to use `connect` and prop drilling. I think that in the long run the glue components should probably be replaced with custom hooks/context providers and all of the configuration logic for deeply nested child components should be abstracted away into custom hooks.

robert-bo-davis:
Should I keep resolving conflicts in this PR, or does the team prefer to leave the routing as-is?

stephenhand (approved):
Sorry @robert-bo-davis - I think it was in draft when I last looked at it - I'm happy to merge this, if you can resolve the conflicts that would be awesome, but I can do it too if you don't have time

robert-bo-davis:
No worries! Just making sure.

PR #1923, opened 2023-12-08
CHI-2461:  Show 'Add to Case' button for nondata contacts in search results

stephenhand:
## Description

* Adding non data contacts to cases should be permitted from the case search results, so the condition excluding those has been removed
* enable_case_management feature flag has been removed (not applied)

### Checklist
- N/A New tests added
- [X] Feature flags added
- N/A Strings are localized

### Related Issues
CHI-2461

### Verification steps

See ticket


acedeywin (approved):
LGTM

PR #1931, opened 2023-12-10
docs: add help system to makefiles

robert-bo-davis:
## Description

This adds a make help system and cleans up some make variable naming.

Screenshot 2023-12-11 at 6 09 41 AM
Screenshot 2023-12-11 at 6 09 17 AM


nickhurlburt (approved):
Cool!

twilio-iac/makefiles/terragrunt/terragrunt.make
```diff
@@ -1,57 +1,61 @@
...
 
-apply-tg:
+verify-env:
+	ifeq ($(HL),)
+	$(error No helpline specified. please be sure to set HL=<helpline> before running make)
+	endif
+
+	ifeq ($(HL_ENV),)
+	$(error No environment specified. please be sure to set HL_ENV=<environment> before running make)
+	endif
```
GPaoloni, line 12: Nice to have! Are we still able to run "stage wide applies" somehow? Maybe this check is only for tg and service copnfig scripts are the same as before? :thinking:

robert-bo-davis: This actually fixes the service-config so you can run them from any tg dir now instead of just from the top level. 

Before this change any makefile that included this file would ALWAYS require an env/hl_env var. Now it is only required for the tg commands.

PR #1938, opened 2023-12-12
[Client Profiles] Redux functionality for listing profiles

mythilytm:
Primary reviewer:

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


plugin-hrm-form/src/services/ProfileService.ts
```diff
@@ -63,3 +63,29 @@ export const updateProfileSection = (profileId: ProfileId, sectionId: ProfileSec
     body: JSON.stringify({ content }),
   });
 };
+
+type GetProfileListParams = {
+  offset?: number;
+  limit?: number;
+  sortBy?: 'id' | 'name' | 'createdAt' | 'updatedAt';
+  sortDirection?: 'asc' | 'desc';
+  profileFlagIds?: string[];
```
GPaoloni, line 72: Maybe this should be `number` (since ids are really numbers) and the service takes care of converting into the "comma separated string"?

plugin-hrm-form/src/states/profile/hooks/useIdentifier.ts
```diff
@@ -86,7 +85,7 @@ export const useIdentifierByIdentifier = (params: UseIdentifierByIdentifierParam
   const identifier = useSelector(
     (state: RootState) => ProfileSelectors.selectIdentifierByIdentifier(state, identifierIdentifier)?.data,
   );
-
+  console.log('>>> useIdentifierByIdentifier identifierIdentifier', { identifierIdentifier, identifier });
```
GPaoloni, line 88: :axe:

plugin-hrm-form/src/states/profile/profileList.ts
```diff
@@ -0,0 +1,68 @@
...
+  return loadProfileListEntryIntoRedux(state, update);
+};
+
+const handleLoadProfileListFulfilledAction = (state: t.ProfileListState, action: any) => {
+  const update = {
+    loading: false,
+    data: action.payload.profiles?.map((profile: any) => profile.id),
+  };
+
+  console.log('>>>handleLoadProfileListFulfilledAction', update);
```
GPaoloni, line 56: :axe:

plugin-hrm-form/src/states/profile/profiles.ts
```diff
@@ -323,6 +324,24 @@ export const updateProfileSectionAsync = createAsyncAction(
...
+        ...newState[profile.id]?.data,
+        ...profile,
+      },
+    };
+
+    newState = loadProfileEntryIntoRedux(newState, profile.id, profileUpdate);
+  }
+
+  return newState;
+};
```
GPaoloni, line 343: What this is doing is: given the profiles list, for each profile, loads the partial data of it (retrieved from the list endpoint), plus the defaults? 
If the above is correct, we might want to add some indicator to know if the profile was "fully loaded" or "pre-view loaded" (e.g. `fullView: false` property when the profile is `loadProfileEntryIntoRedux` from client list, but `fullView: true` when is called by the profile details endpoint).

plugin-hrm-form/src/states/profile/selectors.ts
```diff
@@ -64,3 +64,12 @@ export const selectProfileRelationshipsByType = (
...
 ) => selectProfileById(state, profileId)?.[type];
+
+export const selectProfileListState = (state: RootState) => {
+  const profileState = selectProfileState(state);
+  return {
+    data: profileState?.profilesList?.data,
+    error: profileState?.profilesList?.error,
+    loading: profileState?.profilesList?.loading,
+  };
+};
```
GPaoloni, line 75: Probably the same as
```suggestion
export const selectProfileListState = (state: RootState) => {
  const profileState = selectProfileState(state);
  return profileState?.profilesList || {};
};
```
?

plugin-hrm-form/src/states/profile/hooks/useProfileList.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+import { useSelector } from 'react-redux';
+
+import { Profile } from '../types';
+import * as ProfileSelectors from '../selectors';
+
+type UseProfileListReturn = {
+  profileIds: Profile['id'][] | undefined;
+  loading: boolean | undefined;
+};
```
GPaoloni, line 24: Why not including the profile list error property here?

plugin-hrm-form/src/states/profile/hooks/useProfileListLoader.ts
```diff
@@ -0,0 +1,68 @@
...
+import { useCallback, useEffect } from 'react';
+
+import { RootState } from '../..';
+import asyncDispatch from '../../asyncDispatch';
+import * as ProfileSelectors from '../selectors';
+import { loadProfileListAsync } from '../profileList';
+
+type UseProfileListLoaderParams = {
+  skipAutoload?: boolean;
+  options?: any; // filters, pagination, etc
```
GPaoloni, line 27: Type it? :point_right::point_left:

GPaoloni (approved):
The code looks great to me! I can't see any profiles in the list, I see they are loaded into state but the UI doesn't seem to be consuming them - probably something addressed in a follow up PR?

Also, if you could start adding some better descriptions to the PRs, that would be great, as it makes easier to understand what is changing and what should I look for, and is also good for the history of the repo :P (not a complaint, just a reminder for next time!).

plugin-hrm-form/src/services/ProfileService.ts
robert-bo-davis: Good call... but maybe: `ProfileFlag['id'][]` and the service handles the split. ;)

plugin-hrm-form/src/states/profile/profiles.ts
robert-bo-davis: The [profile details page](https://github.com/techmatters/flex-plugins/blob/master/plugin-hrm-form/src/components/profile/ProfileTabs.tsx#L64) should always [load the full profile](https://github.com/techmatters/flex-plugins/blob/master/plugin-hrm-form/src/states/profile/hooks/useProfileLoader.ts#L53-L58) in the background regardless of the state so there isn't really a great reason to keep track of what has and hasn't been loaded. In the pattern the underlying components in the details page should be written to handle the hydration after initial render without failing on missing data which makes navigation clicks feel fast.

mythilytm: I am just seeing this comment by Bo. I haven't gotten to adding getting the action to open the profile details page but given this info, I will merge this PR

PR #1939, opened 2023-12-13
bg: fix case Id link issue

acedeywin:
Primary reviewer:

## Description

- Bug fix: this PR address the issue as stated in [this](https://tech-matters.atlassian.net/browse/CHI-2451?focusedCommentId=13642) ticket comment Sara.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes [CHI-2451](https://tech-matters.atlassian.net/browse/CHI-2451?focusedCommentId=13642)

### DIFF
--- a/plugin-hrm-form/src/components/caseMergingBanners/ContactAddedToCaseBanner.tsx
+++ b/plugin-hrm-form/src/components/caseMergingBanners/ContactAddedToCaseBanner.tsx
@@ -39,8 +39,8 @@ type Props = OwnProps & ReturnType<typeof mapStateToProps> & ReturnType<typeof m
 
 const mapStateToProps = (state: RootState, { taskId }: OwnProps) => {
   const contact = selectContactByTaskSid(state, taskId);
-  const connectedCase = selectCaseByTaskSid(state, taskId);
   const caseId = contact?.savedContact?.caseId;
+  const connectedCase = state[namespace].activeContacts?.caseConnectedToContact
   return {
     contact: contact.savedContact,
     connectedCase,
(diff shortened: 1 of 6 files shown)


stephenhand (changes requested):
The overall approach works as a bandaid, but there are still some issues with the current fix, outlined in comments

As a general rule, we almost never want global properties in redux state, especially when dealing with contacts & cases, new state almost always should be scoped either to a task, a contact, or after the case redux refactor, a case

plugin-hrm-form/src/states/contacts/types.ts
```diff
@@ -66,6 +67,7 @@ export type ContactsState = {
   contactsBeingCreated: Set<string>;
   contactDetails: ContactDetailsState;
   isCallTypeCaller: boolean;
+  caseConnectedToContact: Case;
```
stephenhand, line 70: We should refactor the actions, reducers and selectors to use the `caseConnectedToContact` property defined here, scoped to an individual contact, rather than the global property they are currentyly updating

This way, when a user is simultaneously completing 2 tasks, there isn't a chance that the case could get mixed up between the 2 contacts as is possible now

acedeywin: @stephenhand  - The refactors have been implemented.

plugin-hrm-form/src/components/caseMergingBanners/ContactAddedToCaseBanner.tsx
```diff
@@ -39,8 +39,8 @@ type Props = OwnProps & ReturnType<typeof mapStateToProps> & ReturnType<typeof m
 
 const mapStateToProps = (state: RootState, { taskId }: OwnProps) => {
   const contact = selectContactByTaskSid(state, taskId);
-  const connectedCase = selectCaseByTaskSid(state, taskId);
   const caseId = contact?.savedContact?.caseId;
+  const connectedCase = state[namespace].activeContacts?.caseConnectedToContact[contact.savedContact.id];
```
stephenhand, line 43:
```suggestion
  const connectedCase = state[namespace].activeContacts.caseConnectedToContact[contact?.savedContact?.id];
```

actiuveContacts is always created, so we don't need the null guard there, but the current contact state or the saved contact attached to it might not be defined, so they do need null guards

stephenhand (approved):
One small change then LGTM - not tested it though

PR #1954, opened 2023-12-19
Chi-2357 add link contact case

acedeywin:
Primary reviewer:

## Description

- This PR adds the blue banner to link from contact to case as well as an Undo button on the orange banner after a case has been removed from the contact.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes [CHI-2357](https://tech-matters.atlassian.net/browse/CHI-2357)

### Verification steps

- Search for contact
- Click on a contact that has been connected to a case
- Verify that you can see the blue banner when you preview the contact
- Click the "Remove from Case" button to remove the case from the contact
- Verify that there is an "Undo" button on the orange banner
- Click on the "Undo" button to undo the previous action

[CHI-2357]: https://tech-matters.atlassian.net/browse/CHI-2357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugin-hrm-form/src/components/caseMergingBanners/ContactAddedToCaseBanner.tsx
```diff
@@ -28,21 +28,28 @@ import InfoIcon from './InfoIcon';
...
+  caseId?: string;
 };
 
 type Props = OwnProps & ReturnType<typeof mapStateToProps> & ReturnType<typeof mapDispatchToProps>;
 
-const mapStateToProps = (state: RootState, { taskId }: OwnProps) => {
-  const { savedContact } = selectContactByTaskSid(state, taskId);
-  const connectedCase = selectCaseByCaseId(state, savedContact.caseId ?? '')?.connectedCase;
-  const caseId = savedContact?.caseId;
+const mapStateToProps = (state: RootState, { taskId, contactId }: OwnProps) => {
```
stephenhand, line 42: We should change the ContactAddedToCaseBanner to always have a contactId passed in, it's available on TabbedForms.tsx which is the other place where this is used, so we may as well pass it in there too.

We can then simplify this logic to always expect the contact ID and always use that to look up the contact

plugin-hrm-form/src/states/contacts/reducer.ts
```diff
@@ -222,6 +223,14 @@ export function reduce(
         },
       };
     }
+    case t.SET_REMOVED_CASE_ID: {
```
stephenhand, line 226: We should enhance the CaseMergingBannersState that @murilovmachado created in `state/case/caseBanners.ts` rather than creating another state structure. We can then also reuse those actions

plugin-hrm-form/src/components/caseMergingBanners/ContactRemovedFromCaseBanner.tsx
```diff
@@ -22,33 +22,62 @@ import { Close } from '@material-ui/icons';
...
+import { contactFormsBase, namespace } from '../../states/storeNamespaces';
+import { Contact } from '../../types/types';
+import { connectToCaseAsyncAction } from '../../states/contacts/saveContact';
+import asyncDispatch from '../../states/asyncDispatch';
+import { RootState } from '../../states';
 
 type OwnProps = {
   taskId: string;
+  savedContact?: Contact;
+  showRemovedFromCaseBanner?: boolean;
```
stephenhand, line 36: This name is confusing, would `showUndoButton` be more clear?

plugin-hrm-form/src/components/caseMergingBanners/ContactRemovedFromCaseBanner.tsx
```diff
@@ -22,33 +22,62 @@ import { Close } from '@material-ui/icons';
...
 import { closeRemovedFromCaseBannerAction } from '../../states/case/caseBanners';
+import { contactFormsBase, namespace } from '../../states/storeNamespaces';
+import { Contact } from '../../types/types';
+import { connectToCaseAsyncAction } from '../../states/contacts/saveContact';
+import asyncDispatch from '../../states/asyncDispatch';
+import { RootState } from '../../states';
 
 type OwnProps = {
   taskId: string;
+  savedContact?: Contact;
```
stephenhand, line 35: Again, I think we should simplify this logic by making the banner always expect a contactId, which it then uses to look up the contact. `TabbedForms.tsx` which is the other place this banner appears already has the ID loaded from state, so it's trivial to pass it in their too and simplify this logic

We should expect a contactId rather than a full contact object, since we only use the ID anyway, anmd this is conistent with the other banner

stephenhand (commented):
A few things suggested that would streamline the implementation, but the logic looks correct

plugin-hrm-form/src/states/contacts/reducer.ts
acedeywin: The `SET_REMOVED_CASE_ID` is a new action type that would hold the caseId of the connected contact after the contact has been removed from the case. This caseId will subsequently be used when the user wants to Undo the initial action

plugin-hrm-form/src/components/caseMergingBanners/ContactRemovedFromCaseBanner.tsx
acedeywin: @stephenhand  - `showRemovedFromCaseBanner` state is for the blue banner state and was previously implemented by @murilovmachado. I was only reusing it. I think this was what to meant in your previous comment.

stephenhand: In this code, the part of the banner toggled by this variable isn't the whole banner, but the just the undo button. The banner itself, the warning triangle and main message are displayed whether this variable is set or not, which is why this is a misleading name in this context

plugin-hrm-form/src/components/caseMergingBanners/ContactRemovedFromCaseBanner.tsx
stephenhand: You only look to use the ID in the code? Even if you do use the rest of the contact data it should be mapped from the state via it's ID in this component rather than prop drilled whole. This is consistent with the other banner interface and more aligned with our general move away from prop drilling and more direct use of redux state

plugin-hrm-form/src/states/contacts/reducer.ts
stephenhand: I understand that - I'm saying we already have a redux state & associated actions for managing whether a 'removed from case' banner should be displayed on a contact. We don't want 2 different states for managing it on 2 different pages, a contact should either be showing a 'removed from case' banner or it shouldn't.

The only thing missing from the existing state is the case ID used in the undo functionality

plugin-hrm-form/src/components/caseMergingBanners/ContactAddedToCaseBanner.tsx
```diff
@@ -28,21 +28,28 @@ import InfoIcon from './InfoIcon';
...
+  const offlineSavedContact = selectContactByTaskSid(state, taskId)?.savedContact;
+  const existingSavedContact = state[namespace][contactFormsBase].existingContacts[contactId]?.savedContact;
```
stephenhand, line 44: [Sorry, missed this in the first pass] We have a `selectContactStateByContactId` selector function we should use for this. 

We should be using selector functions to map state to components in all new code now- in this case we already have one, but where they don't exist yet we should add them

acedeywin: @stephenhand  - All feedback has been implemented.

stephenhand (approved):
Updates look good - a small issue came up in testing that if you open a contact from the case timeline, remove it, then go back to the case, the contact is still there. I made a small update to the logic used to select the connected contacts for a case from redux that fixed this

So LGTM now.

stephenhand:
> Updates look good - a small issue came up in testing that if you open a contact from the case timeline, remove it, then go back to the case, the contact is still there. I made a small update to the logic used to select the connected contacts for a case from redux that fixed this
> 
> So LGTM now.

Except I broke it 🤦 - fixing it now

stephenhand:
All good now!

PR #1955, opened 2023-12-19
CHI-2138: Removed twilio sync based transfers

stephenhand:
## Description

* Removed legacy form transfer method using twilio sync
* Deleted associated legacy types
* Reorganised transfer code a little
* Updated transfer form state tests to match updated functionality

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-2138:

### Verification steps

* Regression test chat & call transfers 2.12.x -> 2.13.x 2.13.x -> 2.12.x, 2.13.x -> 2.13.x


plugin-hrm-form/src/transfer/formDataTransfer.ts
```diff
@@ -0,0 +1,70 @@
...
+export const saveFormSharedState = async (contactState: ContactState, { taskSid }: ITask): Promise<void> => {
+  if (!getAseloFeatureFlags().enable_transfers) return;
+  const { draftContact, savedContact } = contactState;
+  const asyncDispatcher = asyncDispatch(Manager.getInstance().store.dispatch);
+  if (draftContact) {
+    await asyncDispatcher(updateContactInHrmAsyncAction(savedContact, draftContact, `task-${taskSid}`));
+  }
+  if (savedContact.caseId) {
+    await asyncDispatcher(connectToCaseAsyncAction(savedContact.id, undefined));
+  }
```
murilovmachado, line 40: It's not very clear to me why we disconnect the contact from the case. Are we doing this because:
- that's the expected business rule we've agreed on; or
- just to keep the exact same behaviour that we previously had (when using twilio sync)

murilovmachado (approved):
LGTM! Just one question

plugin-hrm-form/src/transfer/formDataTransfer.ts
stephenhand: Because there are issues around transferring cases, because the case is owned by the original creator, so there are permissions issues so we are avoiding the problem by disconnecting cases on transfer for now, but we plan to bring transferring cases in once we know what to do in the various permission scenarios

PR #1964, opened 2023-12-27
refactor forms

GPaoloni:
Primary reviewer:

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....


stephenhand (commented):
Can we leave the unit tests in ___tests__ ? I don't want to be in a situation where some tests are under there and some tests are in the same folder as the code. 

If we want to do the lift of moving all the existing tests up front that's one thing, but doing it piecemeal just seems like we'll be in a worse, inconsistent situation forever

PR #1965, opened 2023-12-28
CHI-2431: chat transfer fix

stephenhand:
## Description

* Added a HRM contact ID to tasks now we create them up front
* Moved transfer receiving logic to be part of the contact creation action, and able to resume from partially completes states on refresh

### Checklist
- [ ] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-2431

### Verification steps

Difficult - you need to add code to simulate accepting chat transfers ending partway through in 2 stages:

- The ownership is never updated in HRM
- The ownership is updated, but task control is never moved to the receiving task (chat transfers only)

Then you need to set the code back to the checked in version and refresh, the transfer should complete and leave the receivers UI in a state where the contact can be completed & chat can continue

Also regression cold & warm voice transfers


plugin-hrm-form/src/states/contacts/saveContact.ts
```diff
@@ -43,15 +43,38 @@ import { ContactDraftChanges } from './existingContacts';
...
+      const attributes = task.attributes ?? {};
+      const { contactId } = attributes;
+      if (contactId) {
+        // Setting the task id and worker id on the contact will be a noop in most cases, but when receiving a transfer it will move the contact to the new worker & task
+        contact = await updateContactInHrm(contactId, { taskId: taskSid, twilioWorkerId: workerSid }, false);
+      } else {
+        contact = await createContact(contactToCreate, workerSid, attributes.transferMeta?.originalTask ?? taskSid);
+        if (contact.taskId! !== taskSid || contact.twilioWorkerId !== workerSid) {
+          // If the contact is being transferred from a client that doesn't set the contactId on the task, we need to update the contact with the task id and worker id
+          contact = await updateContactInHrm(contact.id, { taskId: taskSid, twilioWorkerId: workerSid }, false);
```
murilovmachado, line 65: Thanks for the comments, but I still have one question on the why: We need this just because some counselors might still be running an old version of Aselo or there are other scenarios where this code is needed?

plugin-hrm-form/src/transfer/formDataTransfer.ts
```diff
@@ -33,38 +31,9 @@ export const saveFormSharedState = async (contactState: ContactState, { taskSid
...
- * Loads contact being transferred from HRM (if there is any)
- */
-export const loadFormSharedState = async ({ taskSid, attributes }: ITask): Promise<ContactState> => {
-  const { store } = Manager.getInstance();
-  const rootState = store.getState() as RootState;
-  if (!getAseloFeatureFlags().enable_transfers) return null;
-  if (!attributes.transferMeta) {
-    console.error('This function should not be called on non-transferred task.');
-    return null;
-  }
```
murilovmachado, line 53: Great!

murilovmachado (approved):
The code looks good to me!
I didn't run it to test, TBH, but you can go on and merge this 🚀

plugin-hrm-form/src/states/contacts/saveContact.ts
stephenhand: Purely for compatibility - or if something has gone wrong and the task has come through without an ID for some unanticipated reason (which would be why I would be hesitant to specifically comment it as compatibility code that can be removed next version)

PR #1966, opened 2023-12-29
CHI-2430: Fix issue with contact selector in TaskView. 

stephenhand:
## Description

- Fix issue with contact selector in TaskView, where the wrong contact could be selected from redux
- Ensure that all outstanding edits are included in active contacts, not just the ones from the current form (the helpline was being dropped from omitted from the draft changes sent to the backend in some cases previously)

### Checklist
- [ ] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-2430

### Verification steps

* Deploy to a staging account with multiple helplines / offices configured
* Take a contact using a worker who has a specific helpline assigned
* The same worker should be able to see the contact in the search results
* To be sure, you can check that the 'helpline' field is set to the worker's helpline in the record on the Contacts database table


murilovmachado (approved):
LGTM!!

PR #1973, opened 2024-01-04
fix: Add safeguard before reading participantsLabels in conference components

GPaoloni:
## Description
This PR fixes the bug where supervisors are "Unable to live monitor 3-way conference call". The cause of this bug is that the Redux state for conference is not mounted when the supervisor tries to initially monitor the conference, so accessing `.participantsLabels` property in participant label component and Redux action cause two different errors. To prevent this I'm adding a `?` operator so it defers to a null value initially, until the conference is actually fetched and loaded into the state. 

[Screencast from 04-01-24 13:57:20.webm](https://github.com/techmatters/flex-plugins/assets/15805319/e333bc77-c261-49ac-93c0-83d2f9ccfb39)

### Checklist
- [ ] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2445)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
See attached video or try repro steps in the linked bug ticket.


mythilytm (approved):
Looks good

PR #1974, opened 2024-01-04
[Modica POC] Add support for Modica

murilovmachado:
## Description
- This PR adds support for Modica-SMS on the front-end
- Modica's SMS will show up on Aselo as if it's a regular SMS

This PR can be merged with no harm.

TODO: 
- [IMPORTANT]: modify `getNumberFromTask` so the contact number is saved correctly in the backend
  - see whatsapp and facebook channels
- Offline contacts: When the user chooses SMS, should we change behind the scenes to be Modica-SMS? 
- Verify on other screens: cases, contact details, etc,  if Modica's contacts still shows up as SMS or if any additional work is necessary.

Serverless related PR: https://github.com/techmatters/serverless/pull/559

### DIFF
--- a/plugin-hrm-form/src/states/DomainConstants.ts
+++ b/plugin-hrm-form/src/states/DomainConstants.ts
@@ -28,6 +28,8 @@ export const customChannelTypes = {
   line: 'line',
 } as const;
 
+export const smsChannelTypes = ['sms', 'modica'];
+
 export const channelTypes = {
   ...defaultChannelTypes,
   ...customChannelTypes,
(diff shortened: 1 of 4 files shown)


plugin-hrm-form/src/states/DomainConstants.ts
```diff
@@ -28,6 +28,8 @@ export const customChannelTypes = {
   line: 'line',
 } as const;
 
+export const smsChannelTypes = ['sms', 'modica'];
```
nickhurlburt, line 31: Might be worth a comment here to specify that `sms` is Twilio's default SMS and which helplines use which other SMS types. 

It's very possible in the future that we could have more SMS aggregators here, though maybe not for awhile.

nickhurlburt (approved):
Thanks. I don't see a reason not to merge this.

This also brings to light an interesting issue. Twilio already has built-in support for SMS in its system and in the Flex UI, but that's currently separate from custom channels... so how do we handle SMS in a seamless way in our UI and Twilio stack?

For what it's worth, I think we can make the assumption for now that no helpline will use both Twilio SMS _and_ SMS via custom channel. Hopefully that simplifies things.

PR #1977, opened 2024-01-08
fix: make tg stage var validation check using shell instead of pre-pr…

robert-bo-davis:
…occesor logic

## Description

This fixes a regression introduced in the make system cleanup/help system pr where make vars required for terragrunt commands were being checked using make preprocessor commands `ifeq` instead of runtime shell commands.

### Verification steps

Try running a `make plan`


janorivera (approved):
lgtm and it works

PR #1980, opened 2024-01-08
Add Skills column with category pill to teams view

nickhurlburt:
Hack to test out some Teams View ideas:
- adds a Skills column with pill boxes after the Agent column (we can put arbitrary content inside each cell of the skills column)
- adds placeholder code (unimplemented) to modify the sort function for the Agents column
- In the image below, I've made the "Monitor Task" panel wider by editing the CSS directly via the browser. We would need to explore further... unclear if this can be set up to be dynamic

Related Twilio docs: https://www.twilio.com/docs/flex/developer/ui/v1/components#teams-view-components

Screenshot 2024-01-09 at 3 32 33 PM


nickhurlburt:
The sort function added will allow for a column sort on the Skills column.

nickhurlburt:
Closing in favor of https://github.com/techmatters/flex-plugins/pull/1997

PR #1987, opened 2024-01-10
Make Modica tasks look like SMS throughout Aselo

murilovmachado:
## Description
The goal of this PR is to make Modica tasks look like SMS throughout Aselo, at places like:
- Case Timeline
- Contact Details
- QueueStatus
- PreviousContactsBanner

Because of this specific behavior, just adding `modica` channel to `channelTypes` in our `DomainConstants` is not enough, we needed a way to differentiate these kinds of channels. So, we have two kinds of channel types now:
- `ChannelTypes`: just like before, it includes all channel types
- `CoreChannelTypes`: it doesn't include channels like Modica, because for this matter `modica` = `sms`

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-2393

### Verification steps
Send an SMS from a NZ device, and verify it show looks like an SMS throughout Aselo.
I'm fine with just code reviewing this, cause testing will be tricky.

### DIFF
--- a/plugin-hrm-form/src/states/DomainConstants.ts
+++ b/plugin-hrm-form/src/states/DomainConstants.ts
@@ -46,8 +46,7 @@ export const coreChannelTypes = {
  * This is the complete list of channels.
  */
 export const channelTypes = {
-  ...defaultChannelTypes,
-  ...customChannelTypes,
+  ...coreChannelTypes,
   ...customSmsChannelTypes,
 };
 
--- a/plugin-hrm-form/src/states/DomainConstants.ts
+++ b/plugin-hrm-form/src/states/DomainConstants.ts
@@ -14,6 +14,8 @@
  * along with this program.  If not, see https://www.gnu.org/licenses/.
  */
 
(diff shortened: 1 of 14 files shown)


plugin-hrm-form/src/states/DomainConstants.ts
```diff
@@ -28,19 +30,35 @@ export const customChannelTypes = {
...
+ */
+export const coreChannelTypes = {
+  ...defaultChannelTypes,
+  ...customChannelTypes,
+};
 
+/**
+ * This is the complete list of channels.
+ */
 export const channelTypes = {
```
stephenhand, line 48: Could spread coreChannelTypes here instead of the original 2 :-P

stephenhand (approved):
LGTM

nickhurlburt:
Thanks Murilo.  cc @GPaoloni for visibility

PR #1988, opened 2024-01-10
Vulnerable Groups: Updating to Multiselect field

janorivera:
Primary reviewer:
Reviewers for visibility.
@dee-luo FYI
## Description

We decided that it was best to move these checkbox fields to a multiselect field to be a better match to the CHI data schema.
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### DIFF
--- a/hrm-form-definitions/form-definitions/cl/v1/insights/oneToOneConfigSpec.json
+++ b/hrm-form-definitions/form-definitions/cl/v1/insights/oneToOneConfigSpec.json
@@ -16,12 +16,11 @@
         ]
       },
       {
-        "name": "LGBTQI+",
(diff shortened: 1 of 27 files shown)


dee-luo:
@janorivera you'll need to check the Insights mapping files as well, since I think helplines may be mapping the individual checkbox fields to Insights fields. Will be nice to have them all mapped to one Insights field instead, but that also means we'll have to update Insights metrics for any dashboards using the existing vulnerable groups fields (I think we have it in the Annual Summary dashboards).

janorivera:
you are right! thanks! I've updated the insights mappings

PR #1996, opened 2024-01-11
[Client Profiles] Navigation to individual profiles from client profiles page

mythilytm:
Primary reviewer:

## Description
- This PR implements navigation to individual profiles"

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### Verification steps
Screenshot 2024-01-11 at 2 52 49 PM


plugin-hrm-form/src/components/profileList/ProfileList.tsx
```diff
@@ -15,14 +15,42 @@
...
+type Props = OwnProps & ConnectedProps<typeof connector>;
+
+const mapStateToProps = (state: RootState) => {
+  return {
+    routing: selectCurrentTopmostRouteForTask(state, standaloneTask.taskSid) ?? {
+      route: 'profile-list',
+      subroute: 'profile-list',
+    },
+  };
+};
```
GPaoloni, line 44: Don't we have a selector for this? If no then just skip this for now.

plugin-hrm-form/src/utils/setUpComponents.tsx
```diff
@@ -334,7 +334,7 @@ export const setUpCaseList = () => {
 
 export const setUpClientProfileList = () => {
   Flex.ViewCollection.Content.add(
-    <Flex.View name="client-profiles" key="client-profiles-view">
+    <Flex.View name="profile-list" key="profile-list-view">
```
GPaoloni, line 337: Can you start changing to `profiles list` wherever possible? I noticed a lot of places where plural was used and other where singular was. It makes more sense the plural form IMO :)

GPaoloni (approved):
Looks good, thanks!
Can you merge the updates in `CHI-2330-profilesList`? I'd like to test the behavior with the changes in to check nothing breaks.

plugin-hrm-form/src/utils/setUpComponents.tsx
mythilytm: I started off with the plural form but have changed it to be similar to singular use in case. Happy with this change otherwise

plugin-hrm-form/src/components/profileList/ProfileList.tsx
mythilytm: Sometimes the routing is getting stuck at `case-list` if starting there, and this was an attempt to hack this.

PR #1999, opened 2024-01-11
CHI-2358: Add extra line for describing when the status was last updated

stephenhand:
## Description

* Add  to the EditSummary modal.
* Use dedicated updateCaseStatus endpoint.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes #....

### Verification steps

See ticket - requires latest HRM v1.14 & enable_last_case_status_update_info set to true


plugin-hrm-form/src/states/case/saveCase.ts
```diff
@@ -39,7 +39,18 @@ export const createCaseAsyncAction = createAsyncAction(
...
+    let updatedCase;
+    if (Object.keys(otherEdits).length) {
+      updatedCase = await updateCase(caseId, otherEdits);
+    }
+    if (status) {
+      updatedCase = await updateCaseStatus(caseId, status);
+    }
+    if (!updatedCase) {
+      updatedCase = await getCase(caseId); // If the update is empty, just refresh the case from the backend.
+    }
```
GPaoloni, line 52: Why would this be true? If none of `updateCase` nor `updateCaseStatus` has been called why is the action called in first place?

plugin-hrm-form/src/types/types.ts
```diff
@@ -277,7 +281,8 @@ export type FeatureFlags = {
   enable_client_profiles: boolean; // Enables Client Profiles
   enable_case_merging: boolean; // Enables adding contacts to existing cases
   enable_confirm_on_browser_close: boolean; // Enables confirmation dialog on browser close when there are unsaved changes
-  enable_separate_timeline_view: boolean; // Enables confirmation dialog on browser close when there are unsaved changes
+  enable_separate_timeline_view: boolean; // Enables a limited inline case timelinbe with a link to the full timeline
+  enable_last_case_status_update_info: boolean; // Enables showing the time, user and changed status of the most recent case status update on the 'Edit Case Summary' page
```
GPaoloni, line 285: Aren't this better suited as config flags?

GPaoloni (approved):
LGTM :rocket:

plugin-hrm-form/src/types/types.ts
stephenhand: I don't think so - I don't see anyone wanting to keep these disabled long term

plugin-hrm-form/src/states/case/saveCase.ts
stephenhand: It probably shouldn't be, but it could, so I think we should handle the possibility?

PR #2003, opened 2024-01-12
Fix channel type on in progress contacts

stephenhand:
## Description

Sets the channel type for a contact based on task info when creating them, rather than defaulting to 'web'. This means that the channel type will be correct for draft contacts

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-2252

### Verification steps

See ticket


murilovmachado (approved):
Thank you 🚀

PR #2005, opened 2024-01-12
[Client Profiles] Add status filter to Client List

mythilytm:
Primary reviewer:

## Description
- This PR includes the addition of a status filter as a multi select filter. The status filter allows for viewing clients based on various status flags configured for the helpline account. This addition closely aligns with the existing style and functionality of filters in the Case List.
- In redux, along with sort, filter is added to the profileList settings

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [x] Strings are localized

### Related Issues
Fixes #....

### Verification steps

Screenshot 2024-01-12 at 3 03 43 PM


plugin-hrm-form/src/states/profile/types.ts
```diff
@@ -43,6 +44,7 @@ export const CREATE_PROFILE_SECTION = 'profile/profileSections/CREATE';
 export const UPDATE_PROFILE_SECTION = 'profile/profileSections/UPDATE';
 export const PROFILES_LIST_UPDATE_PAGE = 'profile/profilesList/UPDATE_PAGE';
 export const PROFILES_LIST_UPDATE_SETTINGS = 'profile/profilesList/UPDATE_SETTINGS';
+export const PROFILES_LIST_CLEAR_FILTERS = 'profile/profilesList/CLEAR_FILTERS';
```
GPaoloni, line 47: Do we need this new action? Can't we instead dispatch an `UPDATE_SETTINGS` that only clears the filters? Not saying this is wrong, just thinking if we can reduce the amount of actions needed to the bare minimum.

GPaoloni (approved):
Looks good to me!

Note that there's a small UI bug: the state is correctly stored across "switching views", but the UI does not reflects the (correct) state when you go back to the clients list. This could be due to the "initial" value being passed to the filters are not what's in state?

[Screencast from 19-01-24 16:13:37.webm](https://github.com/techmatters/flex-plugins/assets/15805319/fbdd6dc8-f6a7-4643-9c8d-32c08e4eccbc)

mythilytm:
> Note that there's a small UI bug: the state is correctly stored across "switching views", but the UI does not reflects the (correct) state when you go back to the clients list. This could be due to the "initial" value being passed to the filters are not what's in state?

I fixed this UI bug

PR #2010, opened 2024-01-17
Update Forms

sinehome:
Update Forms for Childline ZW

- Add new "Suburb" field
- Update Helpline manager information


dee-luo (commented):
@sinehome did you decide on making the field not required? (Not required for caller/household/perpetrator make sense, but checking if they wanted required for child?) 

I didn't see a form update for "Update Helpline manager information" - flagging in case that also needs to be updated.

sinehome:
@dee-luo yes, it will be required for the child (missed that). 

For helpline manager info, I'm just confirming the phone number w/ Best but I have already updated the other info (just need to push)

PR #2011, opened 2024-01-17
refactor: align permissions framework with HRM + UX for time-based permissions

GPaoloni:
## Description
This PR
- Reconciles the permissions frameworks, replicating exactly how it is implemented in HRM.
- Adds support for time based permissions on the client side (replicating what was introduced in [this PR](https://github.com/techmatters/hrm/pull/520)).

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2527)
- [x] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Trust unit tests or
- Change development permissions rules to use the new predicates, as indicated in https://github.com/techmatters/hrm/pull/520, and confirm that the actions are evaluated accordingly (i.e. UI does not allows you to do stuff if the records are older than what you specify in the predicate).


stephenhand (approved):
Just gave it a quick scan over but don't see anything of concern

PR #2012, opened 2024-01-19
CHI-2359: Added case transition rules to terragrunt config, with the …

stephenhand:
## Description

* Added case transition rules config SSM parameter to terragrunt config,
* Added required NZ rule & a test rule in AS_STG
* Updated the 'Status last updated' line to show the rule description if the status was updated by a rule
* Refactored case & case item history state lookups to be redux selectors rather than component code

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
CHI-2359

### Verification steps

* The correct version of HRM needs to be deployed so us-east-1 staging, and the terraform updates in infrastructure_config need to be applied to set up the test
* Set a case to 'In Progress' in AS_STG
* Wait 3 hours (the test rule says 5 minutes but the task to check only runs every 3 hours
* Check the case, the status should be 'closed' and the EditCaseSummary.tsx page should say it was updated by the rule

### DIFF
--- a/plugin-hrm-form/src/components/case/CaseSummaryEditHistory.tsx
+++ b/plugin-hrm-form/src/components/case/CaseSummaryEditHistory.tsx
@@ -22,9 +22,9 @@ import { Row } from '../../styles';
 import { CaseActionDetailFont } from './styles';
 import ActionHeader from './ActionHeader';
 import { getAseloFeatureFlags } from '../../hrmConfig';
-import { CaseAuditDetails } from '../../states/case/selectCaseStateByCaseId';
+import { CaseHistoryDetails } from '../../states/case/selectCaseStateByCaseId';
 
-type OwnProps = CaseAuditDetails;
+type OwnProps = CaseHistoryDetails;
 
 type Props = OwnProps;
 
--- a/plugin-hrm-form/src/components/case/CaseSummaryEditHistory.tsx
+++ b/plugin-hrm-form/src/components/case/CaseSummaryEditHistory.tsx
@@ -39,8 +39,8 @@ const CaseSummaryEditHistory: React.FC<Props> = ({ sourceCase, counselorsHash, d
   const previousStatusLabel = previousStatus
     ? definitionVersion.caseStatus[previousStatus]?.label || 'Unknown'
     : 'None';
-  const statusLabel = previousStatus ? definitionVersion.caseStatus[status]?.label || 'Unknown' : undefined;
-  const statusUpdatingCounsellorName = statusUpdatedBy ? counselorsHash[statusUpdatedBy] || 'Unknown' : undefined;
+  const statusLabel = previousStatus ? definitionVersion.caseStatus[status]?.label || `Unknown (${status})` : 'None'; // Shouldn't ever be 'None'
+  const statusUpdatingCounsellorName = statusUpdatedBy ? counselorsHash[statusUpdatedBy] || 'Unknown' : statusUpdatedBy;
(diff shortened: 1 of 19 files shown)


plugin-hrm-form/src/components/case/CaseSummaryEditHistory.tsx
```diff
@@ -17,51 +17,41 @@
...
 
   return (
     <>
-      <ActionHeader
-        added={added}
-        addingCounsellor={addingCounsellorName}
-        updated={updated}
-        updatingCounsellor={updatingCounsellorName}
-      />
+      <ActionHeader added={createdAt} addingCounsellor={updatedBy} updated={updatedAt} updatingCounsellor={createdBy} />
```
murilovmachado, line 45: Shouldn't it be `addingCounsellor={createdBy}` and `updatingCounsellor={updatedBy}`?

GPaoloni:
Why is this stored in SSM rather than as part of the customization forms and/or service config?

plugin-hrm-form/src/components/case/CaseSummaryEditHistory.tsx
stephenhand: Good catch, thx

twilio-iac/terraform-modules/stages/configure/main.tf
```diff
@@ -65,3 +65,20 @@ resource "aws_ssm_parameter" "transcript_retention_override" {
...
 }
+
+
+
+resource "aws_ssm_parameter" "case_status_transition" {
+  count       = var.case_status_transition_rules != null ? 1 : 0
+  name        = "/${lower(var.environment)}/${var.helpline_region}/hrm/scheduled-task/case-status-transitionrules/${nonsensitive(local.secrets.twilio_account_sid)}"
+  type        = "SecureString"
+  value       = jsonencode(var.case_status_transition_rules)
+  description = "Twilio account - External Recordings Enabled"
```
nickhurlburt, line 76: Copy/paste mistake?

twilio-iac/terraform-modules/stages/configure/main.tf
```diff
@@ -65,3 +65,20 @@ resource "aws_ssm_parameter" "transcript_retention_override" {
   type  = "String"
   value = var.hrm_transcript_retention_days_override
 }
+
+
+
+resource "aws_ssm_parameter" "case_status_transition" {
```
nickhurlburt, line 71: It seems that with the above transcript retention change, and also with this change, `twilio-iac` is broadening to "helpline-specific configuration", wherever that lives. We also seem to be relying more on SSM parameters for backend helpline-specific configuration over things we've used before, like process.env. Have we landed on that as a specific plan going forward, and/or captured that anywhere? Not a blocker for this PR, and maybe a tech discussion follow-up.

nickhurlburt (approved):
One specific and one broader comment.

twilio-iac/terraform-modules/stages/configure/main.tf
stephenhand: I thought that was the way we were doing helpline specific config on the backend. process.env is hard to manage in terraform due to configuration for multiple helplines being in a single file, whilst the pathing / filtering features of SSM make it pretty flexible

As I raised in our engineering call, config is definitely something we need to put some though into, both to agree our current standards and to define a roadmap going forward that keeps configuration manageable and minimises fragmentation

nickhurlburt: Makes sense to me! I added it to our discussion list

PR #2014, opened 2024-01-19
Release 202401 service config changes STG

nickhurlburt:
See notes at https://techmatters.app.box.com/notes/1411807981335

* Adds `enable_last_case_status_update_info` and `enable_separate_timeline_view` to all active helpline staging configs, while explicitly disabling those flags for all NZ accounts.
* Alphabetizes feature flags in helpline staging config for easier reading (please let's maintain this)
* @annaliseirby090 has an outstanding action to determine where to put `enable_case_merging`

@stephenhand I'd like some guidance if I'm doing this correctly. As far as I can tell, there is no way to add a flag once for all helplines in a stage, so I did them one by one.

UPDATE: `enable_case_merging` has been added to all active helplines after confirmation with Annalise

### DIFF
--- a/twilio-iac/helplines/as/configs/service-configuration/staging.json
+++ b/twilio-iac/helplines/as/configs/service-configuration/staging.json
@@ -17,9 +17,7 @@
             "enable_separate_timeline_view": true,
             "enable_transcripts": false,
             "enable_twilio_transcripts": false,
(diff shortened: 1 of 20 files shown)


nickhurlburt:
This is now in sync with all staging helplines and changes have been merged

PR #2017, opened 2024-01-22
Add default option for environment name

nickhurlburt:
Possibly addresses issue of the All-staging deploy for Flex not working. I have not tested this - I wanted to see if this makes sense as the general right direction.


stephenhand (approved):
Looks right to me

PR #2023, opened 2024-01-22
Add "flex_instance_sid" to excluded fields of service config tool

nickhurlburt:
For whatever reason, there is a new field in the Flex service config that we are not allowed to change, and not excluding it from our changes is breaking updates.


GPaoloni (approved):
Thanks! :rocket:

PR #2024, opened 2024-01-22
Sets CL config to match master

nickhurlburt:
Sets master configuration to match live config for Chile production.


nickhurlburt:
fyi @annaliseirby090 @dee-luo . We can revert this if we want to make these changes when Alejandro returns.

PR #2030, opened 2024-01-25
CHI-2546: Ensure the tabbed form subcomponents use the parent RHF context 

stephenhand:
## Description

The refactored forms were creating new RHFs when trying to validate, rather than using the parent context that was being used in other scenarios - so the RHF used in validation was always blank so always passed validation

This change adds a useTabbedFormContext method for subcomponents to use that will share context as intended, restoring the form validation behaviour

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-2546

### Verification steps

See ticket


nickhurlburt (approved):
For the sake of fast progress to allow people to QA, approved. But we should also have Gian look at it even after merge.

PR #2034, opened 2024-01-26
CHI-2517: add logging to insight

acedeywin:
Primary reviewer:

## Description

- This ticket handles adding error logging for failures to save Insights data as stated in [this](https://tech-matters.atlassian.net/browse/CHI-2517) Jira ticket.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes [CHI-2517](https://tech-matters.atlassian.net/browse/CHI-2517)

### Verification steps

[CHI-2517]: https://tech-matters.atlassian.net/browse/CHI-2517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugin-hrm-form/src/services/InsightsService.ts
+++ b/plugin-hrm-form/src/services/InsightsService.ts
@@ -209,67 +209,78 @@ type InsightsCaseForm = {
  * customization framework.  When it is, we will need to change this function.
  */
 const convertCaseFormForInsights = (caseForm: Case): InsightsCaseForm => {
-  if (!caseForm || Object.keys(caseForm).length === 0) return {};
-  let perpetrator: { [key: string]: string } = undefined;
-  let incident: { [key: string]: string | boolean } = undefined;
-  let referral: { [key: string]: string } = undefined;
-  let household: { [key: string]: string } = undefined;
-  const topLevel = {
-    id: caseForm.id.toString(),
-  };
-  if (caseForm.info?.perpetrators && caseForm.info.perpetrators.length > 0) {
-    /*
-     * Flatten out the Perpetrator object. This can be changed after this is using the
-     * customization framework.
-     */
-    const thePerp = caseForm.info.perpetrators[0];
-    const untypedPerp: any = {
-      ...thePerp,
-      ...thePerp.perpetrator,
+  try {
+    if (!caseForm || Object.keys(caseForm).length === 0) return {};
+    let perpetrator: { [key: string]: string } = undefined;
+    let incident: { [key: string]: string | boolean } = undefined;
+    let referral: { [key: string]: string } = undefined;
+    let household: { [key: string]: string } = undefined;
+    const topLevel = {
+      id: caseForm.id.toString(),
     };
-    delete untypedPerp.perpetrator;
-    delete untypedPerp.name;
-    delete untypedPerp.location;
-    perpetrator = {
-      ...untypedPerp,
-    };
-  }
-  if (caseForm.info?.incidents && caseForm.info.incidents.length > 0) {
-    const theIncident = caseForm.info.incidents[0];
-    const untypedIncident: any = {
-      ...theIncident,
-      ...theIncident.incident,
-    };
-    delete untypedIncident.incident;
-    incident = {
-      ...untypedIncident,
-    };
-  }
-  if (caseForm.info?.households && caseForm.info.households.length > 0) {
-    const theHousehold = caseForm.info.households[0];
-    const untypedHousehold: any = {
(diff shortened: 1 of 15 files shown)


plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -209,107 +209,129 @@ type InsightsCaseForm = {
...
-      date: caseForm.info.referrals[0].date.toString(),
+  } catch (error) {
+    const errorObject = {
+      message: error,
+      contactsDetails: caseForm.connectedContacts.map(({ channel, taskId }) => ({ channelType: channel, taskId })),
+      accountSid: caseForm.accountSid,
+      twilioWorkerId: caseForm.twilioWorkerId,
     };
+    console.error(errorObject);
+    throw new Error(error);
```
stephenhand, line 282: Can you rethrow the original error rather than a new one so we retain the call stack? I'm not sure it's actually valid to construct an error with another error, TS is only allowing it because catch parameters come through as 'any' types in TS

plugin-hrm-form/src/services/InsightsService.ts
```diff
@@ -209,107 +209,129 @@ type InsightsCaseForm = {
...
-      ...caseForm.info.referrals[0],
-      date: caseForm.info.referrals[0].date.toString(),
+  } catch (error) {
+    const errorObject = {
+      message: error,
+      contactsDetails: caseForm.connectedContacts.map(({ channel, taskId }) => ({ channelType: channel, taskId })),
+      accountSid: caseForm.accountSid,
+      twilioWorkerId: caseForm.twilioWorkerId,
     };
+    console.error(errorObject);
```
stephenhand, line 281: Could we log a string as well as the error object, something like `[InsightsService] Error converting case form` or whatever. Without a consistent string it will be hard to search for these logs across fullstory sessions

stephenhand (commented):
Apart from the specific comments, could we log a `console.warn` after successful saves as well like we discussed in the estimation. We can reduce the level to 'info' after we track down the issue, but we want to track what was saved anyway, in case the issue isn't resulting in a Flex error, we still have some extra data to work with

plugin-hrm-form/src/services/InsightsService.ts
acedeywin: @stephenhand - The original error is what was thrown last in the  `throw new Error(error)` block. The `console.error` block contains some additional info including the original error as a message object

plugin-hrm-form/src/services/InsightsService.ts
acedeywin: @stephenhand - this makes lots of sense. I will go ahead and add that :)

acedeywin:
> Apart from the specific comments, could we log a `console.warn` after successful saves as well like we discussed in the estimation. We can reduce the level to 'info' after we track down the issue, but we want to track what was saved anyway, in case the issue isn't resulting in a Flex error, we still have some extra data to work with

@stephenhand  - I need more clarification on the above. What is the information that I need to pass into the `console.warn`? Will it be the same kind of info as the `console.error`? 

> We can reduce the level to 'info' after we track down the issue, but we want to track what was saved anyway, in case the issue isn't resulting in a Flex error, we still have some extra data to work with

Also, can you throw more light on the above?

stephenhand:
> > Apart from the specific comments, could we log a `console.warn` after successful saves as well like we discussed in the estimation. We can reduce the level to 'info' after we track down the issue, but we want to track what was saved anyway, in case the issue isn't resulting in a Flex error, we still have some extra data to work with
> 
> @stephenhand - I need more clarification on the above. What is the information that I need to pass into the `console.warn`? Will it be the same kind of info as the `console.error`?
> 
> > We can reduce the level to 'info' after we track down the issue, but we want to track what was saved anyway, in case the issue isn't resulting in a Flex error, we still have some extra data to work with
> 
> Also, can you throw more light on the above?

Yes, similar information, whatever would be useful to somebody trying to track down a problem with insights data saving on the backend in your opinion

For the second point, I'm just saying we will put it to 'warn' initially while we find this bug, but because this isn't really the cotrrect level, we are just setting it to that so Fullstory captures it, we should change the level to info after we find & fix the bug

acedeywin:
@stephenhand - All feedback has been implemented.

stephenhand (approved):
One tests are passing LGTM

PR #2035, opened 2024-01-26
CHI-2459: rework connect dialog

stephenhand:
## Description

Since the tabbed forms refactor, the connected dialog was broken, the implementation was set up in search routes but now contact details are routed to directly in tabbed forms

To fix it:

* ConnectDialog implementation is moved into ContactDetails, only the expected navigation behaviour after you confirm is passed in.
* The option to copy is determined by whether the current task has a contact attached rather than with a flag
* A bunch of stuff made redundant by the above was removed, included the global 'isCallTypeCaller' redux flag hack

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
CHI-2459

### Verification steps

See ticket. Pay particular attention that the copy link option isn't offered anywhere from standalone searches


GPaoloni (approved):
LGTM!

nickhurlburt:
CHI-2549

nickhurlburt:
CHI-2459

PR #2036, opened 2024-01-26
CHI-2458: Align UI permissions with HRM for draft contacts

stephenhand:
## Description

The edit button should only show for draft contacts when the user is either the owner or the creator of the contact, in line with the rules on the backend, so a non working edit button should never be shown.

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Related Issues
Fixes CHI-2458

### Verification steps

See ticket

### DIFF
--- a/plugin-hrm-form/src/permissions/canEditContact.ts
+++ b/plugin-hrm-form/src/permissions/canEditContact.ts
@@ -1,0 +1,32 @@
+/**
+ * Copyright (C) 2021-2023 Technology Matters
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+
+import { getInitializedCan, PermissionActions } from '.';
+import { Contact } from '../types/types';
+import { getHrmConfig } from '../hrmConfig';
+
+const getCanEditContact = (contact: Contact): (() => boolean) => {
+  if (!contact.finalizedAt) {
+    // If the contact is a draft, we use the hardcoded rule that only its owner or creator can edit it
+    const { workerSid } = getHrmConfig();
+    const permitted = contact.twilioWorkerId === workerSid || contact.createdBy === workerSid;
+    return () => permitted;
+  }
+  const initializedCan = getInitializedCan();
+  return () => initializedCan(PermissionActions.EDIT_CONTACT, contact);
+};
+
+export default getCanEditContact;
--- a/hrm-form-definitions/form-definitions/zw/v1/HelplineInformation.json
+++ b/hrm-form-definitions/form-definitions/zw/v1/HelplineInformation.json
@@ -6,9 +6,9 @@
         "value": "Helpline Centre",
         "default": true,
         "manager": {
-          "name": "Best Sibanda",
-          "phone": "263 73 211 6116",
-          "email": "it@childline.org.zw"
+          "name": "Munyaradzi Nhemachena",
+          "phone": "263 24 225 2000",
+          "email": "helplinehre@childline.org.zw"
         }
       }
     ]
--- a/hrm-form-definitions/form-definitions/zw/v1/caseForms/HouseholdForm.json
+++ b/hrm-form-definitions/form-definitions/zw/v1/caseForms/HouseholdForm.json
@@ -187,6 +187,11 @@
     "optionsReferenceKey": "districtOptions"
   },
   {
+    "name": "suburb",
+    "label": "Suburb",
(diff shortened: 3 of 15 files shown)


nickhurlburt:
@stephenhand I'm pretty sure you meant CHI-2548 instead of 2458, so I've edited this to reflect that so it's caught on the Jira ticket

plugin-hrm-form/src/permissions/canEditContact.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+
+import { getInitializedCan, PermissionActions } from '.';
+import { Contact } from '../types/types';
+import { getHrmConfig } from '../hrmConfig';
+
+const getCanEditContact = (contact: Contact): (() => boolean) => {
```
GPaoloni, line 21: Isn't there a way to achieve this by using a combination of `can` plus the intended action, to keep it consistent with how the rest is implemented? I would rather abstract special cases inside of the `can` function at initialization, rather than exposing more wrappers around it. I believe `can` already supports receiving a `contact` (or a `target` actually), and the `workerSid`.
There may be impediments to that so this is a more of a question.

plugin-hrm-form/src/permissions/canEditContact.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (!contact.finalizedAt) {
+    // If the contact is a draft, we use the hardcoded rule that only its owner or creator can edit it
+    const { workerSid } = getHrmConfig();
+    const permitted = contact.twilioWorkerId === workerSid || contact.createdBy === workerSid;
```
GPaoloni, line 25: This is kinda introducing the concept of `isCreator` for contacts. Wouldn't be good to have that in as a predicate rather than hardcoded? :thinking:

GPaoloni (approved):
Looks good to me from the functionality perspective, but I have some questions regarding why you choose to implement this way. You can safely dismiss my concerns if you already thought/tried that way.

plugin-hrm-form/src/permissions/canEditContact.ts
GPaoloni: To exemplify what I mean by this, and also to relate this comment with the below one, I mean something like (in `plugin-hrm-form/src/permissions/index.ts`):

```
else if (kind === 'contact') {
      const conditionsState: ConditionsState = {
        isSupervisor: performer.isSupervisor,
        isOwner: isContactOwner(performer, target),
        isCreator: isContactCreator(performer, target), <<<<<<<< new condition available in contact kind
        everyone: true,
        createdDaysAgo: false,
        createdHoursAgo: false,
        ...appliedTimeBasedConditions,
      };
```
where isContactCreator = (performer, target) => target.createdBy === performer
and then the condition `const permitted = contact.twilioWorkerId === workerSid || contact.createdBy === workerSid;` could be defined from the permission files, something like: 
`"editContact": [["isCreator"], ["isOwner"]]`

Again, happy to dismiss this convo if you found it difficult or if I'm missing something.

GPaoloni: Ah, I'm missing the `if (!contact.finalizedAt)` chunk, which could either also be defined as part of the contact predicates, resulting in something like
`"editContact": [["isCreator", "isNotFinalized"], ["isOwner", "isNotFinalized"], ...... the rest of the actual predicates]`

But if this is the default behavior expected from Aselo for everybody then yes, it makes no sense to try to move it here.

I would still consider wrapping the logic for the `!contact.finalizedAt` case inside the `setupAllow` function rather than exposing a different wrapper around the `can`. Just personal opinion at this point.

Sorry the spam on comments, I was answering myself while writing and re-reviewing the code haha.

stephenhand:
@GPaoloni - you raise valid points, the reason why it is done this way is that permissions for draft contacts are currently hardcoded on the backend, so I hardcoded the fronted for consistency.

We did this on the hoof because requirements for draft contact permissions weren't considered ahead of time. It's tech debt for sure but I think we need to think through with the product team to come up with a more systematic solution rather than whipping something up in a PR.

I imagine an `isDraft` condition would be part of the solution, but we have to be careful to provide guardrails in whatever we do, because if you deviate from the hardcoded perms I implemented then some basic functions like transfers or offline contacts will completely break.

To further complicate things there's an additional special hardcoded case purely on the backend needed to make transfers work 🙄, so definitely some debt to address...

Does this give you context for my approach & address your concerns? Happy to discuss changes if you still think some are needed

GPaoloni:
@stephenhand Yup it all makes sense. I still prefer the "`can` is the source of truth"  even if we need to hardcode the `isDraft` chunk, but that's totally a personal preference and need no changes in this PR. Thanks for the response!

stephenhand:
> @stephenhand Yup it all makes sense. I still prefer the "`can` is the source of truth" even if we need to hardcode the `isDraft` chunk, but that's totally a personal preference and need no changes in this PR. Thanks for the response!

My preference for the wrapper was to make it obvious that it was doing something over and above the standard permission check.

I imagine it could be quite hard to debug an issue if somebody sees a standard 'can' call in the code, they just expect it to implement the rules specified in the config file, and it's only if you go into the can implementation itself that you see that there are hardcoded special cases. At least using a wrapper you can see that the call is taking a different path.

The disadvantage is that it's easier to accidentally forget to use the wrapper in future code where the check is required, but people should check how things are implemented elsewhere  when reusing established functionality, so I thought this was the lesser issue

PR #2038, opened 2024-01-26
202401 feature flag release config for production

nickhurlburt:
In short, the feature flag configuration in this PR turns on `enable_last_case_status_update_info` and `enable_separate_timeline_view` for all helplines by default, but exempts NZ.

Included is removal of redundant settings, including staging-only flags that are no longer needed, and moves NZ staging flags into NZ common (there is no NZ prod).

**This has not been applied.** In order to keep the code in sync with the config, we should apply to:
- [x] Development: will enable both flags for Aselo Dev and E2E Testing
- [x] Staging: will enable both flags for UK and ZA (which were accidentally missed earlier) as well as unused helplines (F2, PL, RO)
- [x] Production: will enable both flags for all helplines


nickhurlburt:
Config applied everywhere

PR #2043, opened 2024-01-29
Alphabetize feature flag list

nickhurlburt:
Our `FeatureFlags` type definition is used by the entire team (including Product/SEs) as the source of truth for feature flags definitions. It is much easier to look through when alphabetized. Let's keep it alphabetical.

This change will have no effect on our system.


murilovmachado (approved):
Nice!

PR #2048, opened 2024-02-05
added fullstory ids to different case merging components

skomaiha:
Added fullstory ids to different case merging components


stephenhand (approved):
Thanks Sara!

PR #2064, opened 2024-02-08
CHI-2225 Align case related dates to show user's browser/machine time timezones

mythilytm:
Primary reviewer:

## Description
- This PR align all the case related dates to show the correct date based on the local browser/machine time zone. 
- Remove the use of `getLocaleDateTime` helper function to `parseISO(inputDate).toLocalDateString` 
- Followup ticket has been created [to consolidate date formatting code](https://tech-matters.atlassian.net/browse/CHI-2586?atlOrigin=eyJpIjoiYzY3NWE3NDYxMWVmNDI5ZThiNzA5MmM0OTI2MzE1MTQiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) See [related slack thread here](https://tech-matters.slack.com/archives/C05JDH53S65/p1707428720148229?thread_ts=1707428254.633059&cid=C05JDH53S65)

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Related Issues
Fixes # https://tech-matters.atlassian.net/browse/CHI-2225

### Verification steps
1. Install the "Spoof Timezone" extension from the following URL: https://webextension.org/listing/spoof-timezone.html. Ensure the extension is updated based on the VPN's new IP address.
2. Create a new case
3. Verify the consistency of the case's date across various displays. Specifically, ensure that the date shown in the table row listing the case, the detailed view of the case, and the case preview within search results are all identical.
4. Reset vpn and browser to your current timezone. Check the same displays.


stephenhand (changes requested):
I think this is doing the opposite of what's needed? The dates in the Case List and Search preview look to be correct, the problem is that the `getLocaleDateTime` function is broken, not that we need to be using it in more places.

What the `getLocaleDateTime` function is currently doing is taking an ISO Date Time, splitting it into it's constituent parts using text parsing and recreating a new Date object with the same display time as the date it was passed, but now as a local time - so it is _changing_ the time.

So for example, we give getLocaleDateTime this string

'2020-06-10T20:00:00.000Z' - 
In New Zealand it will internally convert it to '2020-06-10T08:00:00.000+12:00' before reformatting it. But '2020-06-10T08:00:00.000+12:00' is actually '2020-06-09T20:00:00.000Z', so it changed the time

The behaviour we want it for it to change the input time to the local time that is the same _moment_ in time as the input date, not the same display time - so in the above example it should internally convert to '2020-06-10T20:00:00.000+12:00', which is still '2020-06-10T20:00:00.000Z'

By using this function in more places your are spreading the issue to more parts of the code, rather than fixing it.

The correct fix is either to fix the function, or perhaps better, get rid of it, because the correct behaviour is produced by simply calling parseISO(inputDate).toLocaleDateString()

Time zones are hard, this is why any time zone fix should always be accompanied by comprehensive unit tests which define exactly what you expect in each scenario, because without them, you will almost certainly get it wrong. 

In fact a lot of the time, when you sit down write out all your expectations as actual test cases, it turns out even your initial expectations were incorrect!

PR #2099, opened 2024-02-21
[DO NOT MERGE] POC: Staging Look and Feel

murilovmachado:
image

This was achieved by:
- Modifying MainHeader background color
- Changing MainHeader logo
  - the trick here is that the `staging` copy is part of the image itself

### DIFF
--- a/plugin-hrm-form/src/utils/setUpComponents.tsx
+++ b/plugin-hrm-form/src/utils/setUpComponents.tsx
@@ -375,6 +375,37 @@ export const setLogo = url => {
   }
 };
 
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/utils/setUpComponents.tsx
```diff
@@ -375,6 +375,37 @@ export const setLogo = url => {
   }
 };
 
+export const applyStagingTheme = (environment: string) => {
+  // if (environment !== 'staging') return;
```
murilovmachado, line 379: Uncomment this line if merging this at some point

PR #2104, opened 2024-02-22
CHI-2539: permission fix for edge case

acedeywin:
## Description

- This PR handles the edge cases when counsellor A initiate a transfer to counsellor B. In this scenario a counsellor without permission to view the case added to a contact won't be able to view such cases. So the banner with the case ID will be shown to the counselor but with no access to view it. 

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None


plugin-hrm-form/src/services/CaseService.ts
```diff
@@ -73,11 +74,18 @@ export async function updateCaseStatus(caseId: Case['id'], status: Case['status'
 }
 
 export async function getCase(caseId: Case['id']): Promise<Case> {
-  const options = {
+  const options: FetchOptions = {
     method: 'GET',
+    returnNullFor404: true,
```
stephenhand, line 79: Do we still need the `try ... catch` block below if we are setting this option? Not a blocker

stephenhand (approved):
LGTM, 1 question

plugin-hrm-form/src/services/CaseService.ts
acedeywin: @stephenhand - For handling the 404 error, we don't need it. I have made the necessary adjustment.

PR #2116, opened 2024-03-01
fix: fetch case when not loaded into state [CHI-2456]

GPaoloni:
### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2456)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Edit `src/public/appConfig.js` like `const accountSid = 'ACxxxxxxxxxxxxxxxxxx'; // as staging` where `ACxxxxxxxxxxxxxxxxxx` is the account sid of Aselo staging.
- Follow the reproduction steps of the linked Jira ticket and confirm you can't replicate those bugs anymore.

### DIFF
--- a/plugin-hrm-form/src/states/case/case.ts
+++ b/plugin-hrm-form/src/states/case/case.ts
@@ -14,9 +14,6 @@
  * along with this program.  If not, see https://www.gnu.org/licenses/.
  */
 
-// TODO: REMOVE
-/* eslint-disable import/no-unused-modules */
-
 import { createAsyncAction } from 'redux-promise-middleware-actions';
 
 import * as CaseService from '../../services/CaseService';
@@ -25,7 +22,6 @@ import { loadCaseIntoState } from './loadCaseIntoState';
 import type { Case } from '../../types/types';
 import type { HrmState } from '..';
 
-// eslint-disable-next-line import/no-unused-modules
 export const loadCaseAsync = createAsyncAction(t.LOAD_CASE_ACTION, CaseService.getCase, (caseId: Case['id']) => ({
   caseId,
 }));
@@ -38,7 +34,7 @@ export const handleLoadCasePendingAction = (
   state: HrmState,
   action: ReturnType<LoadCaseAsync['pending']> & { meta?: LoadCaseAyncMeta },
 ): HrmState => {
-  const { caseId } = action.meta as { caseId: Case['id'] };
+  const { caseId } = action.meta;
 
   const cas = state.connectedCase.cases[caseId]?.connectedCase;
 
@@ -46,6 +42,7 @@ export const handleLoadCasePendingAction = (
     ...state,
     connectedCase: loadCaseIntoState({
       state: state.connectedCase,
+      caseId,
       definitionVersion: state.configuration.currentDefinitionVersion,
       newCase: cas,
       error: null,
@@ -66,6 +63,7 @@ export const handleLoadCaseFulfilledAction = (
     ...state,
     connectedCase: loadCaseIntoState({
       state: state.connectedCase,
+      caseId,
       definitionVersion: state.configuration.currentDefinitionVersion,
       newCase: cas,
       error: null,
@@ -86,6 +84,7 @@ export const handleLoadCaseRejectedAction = (
     ...state,
     connectedCase: loadCaseIntoState({
       state: state.connectedCase,
+      caseId,
       definitionVersion: state.configuration.currentDefinitionVersion,
       newCase: cas,
       error: action.payload,
--- a/plugin-hrm-form/src/states/case/case.ts
+++ b/plugin-hrm-form/src/states/case/case.ts
@@ -14,27 +14,33 @@
  * along with this program.  If not, see https://www.gnu.org/licenses/.
  */
 
-import { createAsyncAction } from 'redux-promise-middleware-actions';
+import { createAction, createAsyncAction } from 'redux-promise-middleware-actions';
 
 import * as CaseService from '../../services/CaseService';
 import * as t from './types';
 import { loadCaseIntoState } from './loadCaseIntoState';
 import type { Case } from '../../types/types';
 import type { HrmState } from '..';
+import { dereferenceCase, referenceCase } from './referenceCase';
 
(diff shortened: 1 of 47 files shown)


plugin-hrm-form/src/states/case/case.ts
```diff
@@ -0,0 +1,94 @@
+/**
```
stephenhand, line 1: Maybe name this something more explicit, like 'singleCase'?

plugin-hrm-form/src/states/case/selectors.ts
```diff
@@ -0,0 +1,27 @@
+/**
```
stephenhand, line 1: We shouldn't go down the path of generic bags of selectors, still trying to get rid of the old 'reducters' 'actions' etc. selectors should either be in separate files, or colocated with highly related selectors in a file named to indicate their shared functional responsibility

plugin-hrm-form/src/states/case/hooks/useCase.ts
```diff
@@ -0,0 +1,66 @@
+/**
```
stephenhand, line 1: I'm a little concerned with how this hook is managing the lifetime of resources. It isn't applying any references to a case it loads, so if anything else decides to release the cases while the component is mounted, it could clean the case from state out from under this component because it doesn't know something is still using it, causing this screen to lose its data until remounted

You could start passing a reference string, but then you really should release the case once the component unmounts. This would at least ensure the component has the case loaded all the time it is mounted, but to prevent a leak you should then release the case on unmount. If nothing else holds a less transient reference to the case, this could result in the case being loaded from HRM more often than expected, since it will reload it every remount.

Makes me wonder if custom hooks are simply too transient & myopic to handle the responsibility of ensuring entities are loaded into redux efficiently and there should be some sort of pure redux solution instead, where we trigger lazy loading and manage references based on the state of the routing system instead?

Or alternatively, we move away from reference counting and use an expiry instead, like we do with resources. Basically clear out any references over a certain age and reload them next time they are needed. For this to work, we would need to ensure all our code could handle redux resources arbitrarily being pulled from underneath them at any time, which we don't currently have.

If we transition to an expiry model, I suppose we could leave the reference counting model in place whilst we migrate - basically entities are only cleared up if they have no references specified AND they are older than the expiry time window. That way stuff build to expect arbitrary entity expiration could load stuff with no references, whereas legacy stuff could still use a reference and avoid nasty surprises.

Thoughts? For this PR you should at least start using a reference when loading the case, and releasing said reference on unmount

stephenhand:
One big & one small comment. Also - is this intended for the current release? If it is the PR should target the RC branch. Also, introducing a new custom hook feels like a big change to fix a bug on an RC branch. Could we perhaps consider a more conservative fix for the RC branch and keep this for master. Maybe you are doing this already & that's why this PR points at master, ignore me if so!

GPaoloni:
@stephenhand As discussed via Slack (or at least the understanding I have of that conversation :P)
- I've addressed the references as you pointed out (thanks for this!) in commit [12cefb6](https://github.com/techmatters/flex-plugins/pull/2116/commits/12cefb66ac672b1ef4413dc41e4dd3515c5f126c). I strongly suggest you reviewing that single commit as there are the important changes. Within it, I'm also leaving some debug `console.log`s that you can filter by the `>>>>>` string. If you play around the code based, you should be able to see when/how the updated code is trying to reference/dereference the case. I couldn't broke this code after doing some regression testing, but we might want to do more of that.
- The commit following that one ([12cefb6](https://github.com/techmatters/flex-plugins/pull/2116/commits/12cefb66ac672b1ef4413dc41e4dd3515c5f126c)) includes a lot of very small changes. Basically, I've changes all references to `caseId` from `string` to `number` type, which is the actual type that the case id have in our system. The fact that we used strings sometimes and the actual `case.id` (a number) in others, lead to a lot of weird effects and UI updates being triggered unnecessarily, as `caseId` would go from `123` to `'123'` when navigating different case views.
- > Also - is this intended for the current release? If it is the PR should target the RC branch. Also, introducing a new custom hook feels like a big change to fix a bug on an RC branch. Could we perhaps consider a more conservative fix for the RC branch and keep this for master.

  This PR is being merged against `master` but should probably go into the -rc branch instead. I will change the base and try to resolve merge conflicts if any. Do you foresee any problems merging against the -rc? This is a release blocking fix AFAIK, and I don't think the approach without hooks is simpler than this as I need to do the exact same - fetch, reference, etc - but in every place where the case needs to be in state when component mounts (`Case` and `ContactAddedToCaseBanner`). I can that way but will do only if you really think merging this PR is a bad idea.
- > You could start passing a reference string, but then you really should release the case once the component unmounts. This would at least ensure the component has the case loaded all the time it is mounted, but to prevent a leak you should then release the case on unmount. If nothing else holds a less transient reference to the case, this could result in the case being loaded from HRM more often than expected, since it will reload it every remount.

  If you play around with the 1st commit mentioned above, you can check that the case is not fetched between renders at all. It is released after the `Case`/`ContactAddedToCaseBanner` component is entirely unmounted, unless you are transitioning from one to another (in which cases the reference is held until the next state is computed) avoiding - to some extent - the scenario you describe here. The case is released if the parent component is unmounted and something else that doesn't references the case is mounted however. I think this is expected? But we could reduce the round-trips to HRM by adding some delayed-dereferences, let's say of a minute? Just in case the counselor opens-closes-open the case in rapid sucessions. If we want to do something like that, let's not do so in this PR please haha.

Things that we mentioned in our discussion but I am not addressing here:
- We are considering moving from references to "expiry timestamps". That is, the cases might get cleaned up from local store after a certain period of time not being accessed.
- Something I'm missing? :thinking:

stephenhand:
We want case ids to be strings everywhere. We should convert them in HRM really. Making case & contact IDs ascending IDs in the first place was a mistake, but we at least don't want to lock our type system out of doing things better in the future. 

Using numeric case IDs has zero benefit if code (they are just identifiers, its not like we'll be doing maths on them) and forcing them to be treated as number makes it harder if a helpline comes along that wants to start using their pre-existing identifier scheme to identify cases. The fact that most of your changes are to use ``Case['id']` mean they won't need backing out but I'm reverting all the changes to explicitly type them as numbers

GPaoloni:
Closing in favor of https://github.com/techmatters/flex-plugins/pull/2250

PR #2117, opened 2024-03-04
CHI-2616: llow save without media

stephenhand:
## Description

* Allows contacts to continue saving & finalizing even when the media saving portion of the process fails
* Logs out more info about the issue & context
* 

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Other Related Issues

None

### Verification steps

Test happy path by saving call & chat contacts
To test the error case you need to run broken code locally
The real test will be when the intermittent 'no call SID' occurs, and the contact still finalizes, but this is not reproducible RN

### DIFF
--- a/plugin-hrm-form/src/services/ContactService.ts
+++ b/plugin-hrm-form/src/services/ContactService.ts
@@ -78,47 +78,65 @@ export const handleTwilioTask = async (task): Promise<HandleTwilioTaskResponse> 
     return returnData;
   }
 
-  if (TaskHelper.isChatBasedTask(task)) {
-    // Store a pending transcript
+  try {
+    if (TaskHelper.isChatBasedTask(task)) {
+      // Store a pending transcript
+      returnData.conversationMedia.push({
+        storeType: 'S3',
+        storeTypeSpecificData: {
+          type: 'transcript',
+          location: undefined,
+        },
+      });
+    }
+
+    if (TaskHelper.isChatBasedTask(task) || TaskHelper.isCallTask(task)) {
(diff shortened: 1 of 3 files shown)


plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -78,47 +79,72 @@ export const handleTwilioTask = async (task): Promise<HandleTwilioTaskResponse>
...
+
+    if (TaskHelper.isChatBasedTask(task) || TaskHelper.isCallTask(task)) {
+      // Store reservation sid to use Twilio insights overlay (recordings/transcript)
+      returnData.conversationMedia.push({
+        storeType: 'twilio',
+        storeTypeSpecificData: {
+          reservationSid: task.sid,
+        },
+      });
+    }
```
murilovmachado, line 102: I don't understand almost nothing on `conversationMedia`.  But just asking here if this is intended:
When `TaskHelper.isChatBasedTask(task)` is true, it pushes 2 values into returnData
- one of store type `S3`
- and another of store type `twilio`

Is that it?

murilovmachado (approved):
I don't feel confident in reviewing the correctness of the logic, but code/styling wise it looks good to me.

plugin-hrm-form/src/services/ContactService.ts
stephenhand: Yea, that logic hasn't been changed in this PR, its just got a little mangled in the diff - it's a lot easier to read if you use the 'hide whitespace' option

PR #2118, opened 2024-03-04
fix: profiles filter translation [CHI-2631]

GPaoloni:
## Description
This PR fixes the bug described in the Jira ticket linked below
> Client profile tags like Abusive and Blocked are translatable, but not the filters labels on the client profiles page are not.

The bug was caused because in order to render the list of flags uppercased, a string manipulation was used. This caused that the filters, for a value like `abusive`, ended up receiving a string like `Abusive`. Since the translations are assumed to be on the `name` of the flags, this caused a missmatch between the value being passed to the filters and the translations. This bug has been present since this component was implemented ([see](https://github.com/techmatters/flex-plugins/blame/92102fcb862e34e102a87a904c686c4b247d8ae8/plugin-hrm-form/src/components/profileList/filters/ProfileFilters.tsx#L39)).

To fix this issue, the filters now receive the profile flag `name` without any string manipulation, and instead the UI components will capitalize the strings being passed on. 
In order to achieve this behavior, I also needed to do some small tweaks to the `src/components/caseList/filters/MultiSelectFilter` component: when the filter is not searchable, the string rendered can now be translated (via `Template` component). This would be harder to implement if we allow this behavior in "searchable" variants, but doable. I'm skipping that for now since it's not really related to the bug this PR intends to fix.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2631)
- [ ] New tests added
- [n/a] Feature flags added
- [x] Strings are localized

### Verification steps
- Start local server (`npm run dev`).
- Confirm that the flags are displayed in the UI as intended.
  Screenshot from 2024-03-04 17-02-19
- Change `src/translations/en-US/flexUI.json` adding an entry like:
  `"abusive": "Verbally Abusive"`.
- Confirm that the flags are now translated according to the above changes.
  image

### DIFF
--- a/plugin-hrm-form/src/components/caseList/filters/MultiSelectFilter.tsx
+++ b/plugin-hrm-form/src/components/caseList/filters/MultiSelectFilter.tsx
@@ -51,11 +51,12 @@ type ReactHookFormValues = {
 
 type OwnProps = {
   name: string;
-  text: string;
+  text: string | React.ReactElement;
   defaultValues: Item[];
   withSearch?: boolean;
   openedFilter: string;
   searchable?: boolean;
+  capitalizeOptions?: boolean;
   searchDescription?: string;
   applyFilter: (values: Item[]) => void;
   setOpenedFilter: (name: string) => void;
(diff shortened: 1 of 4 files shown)


mythilytm (approved):
Looks great!

PR #2123, opened 2024-03-07
CHI-2627 bg: fix bug with viewing contact

acedeywin:
## Description

- Bug fix: Opening contact linked to closed case brings a white screen.
- From the contact search view, while trying to open a contact linked to a closed case, the system brings a white screen.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### Verification steps

This should be replicated with NZ
- Search for contact with close case
- Click on the contact
- Verify that the contact details are displayed

### DIFF
--- a/plugin-hrm-form/src/permissions/index.ts
+++ b/plugin-hrm-form/src/permissions/index.ts
@@ -315,14 +315,14 @@ const applyTimeBasedConditions = (conditions: TimeBasedCondition[]) => (
       if (cond === 'createdHoursAgo') {
         return {
           ...accum,
-          [key]: differenceInHours(ctx.curentTimestamp, target && parseISO(target.createdAt)) < param,
+          [key]: target && differenceInHours(ctx.curentTimestamp, parseISO(target.createdAt)) < param,
         };
       }
 
       if (cond === 'createdDaysAgo') {
         return {
           ...accum,
-          [key]: differenceInDays(ctx.curentTimestamp, target && parseISO(target.createdAt)) < param,
+          [key]: target && differenceInDays(ctx.curentTimestamp, parseISO(target.createdAt)) < param,
         };
       }
 
--- a/plugin-hrm-form/src/permissions/index.ts
+++ b/plugin-hrm-form/src/permissions/index.ts
@@ -315,14 +315,14 @@ const applyTimeBasedConditions = (conditions: TimeBasedCondition[]) => (
       if (cond === 'createdHoursAgo') {
         return {
           ...accum,
-          [key]: differenceInHours(ctx.curentTimestamp, parseISO(target.createdAt)) < param,
+          [key]: differenceInHours(ctx.curentTimestamp, parseISO(target?.createdAt)) < param,
         };
       }
 
       if (cond === 'createdDaysAgo') {
         return {
           ...accum,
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/permissions/index.ts
```diff
@@ -315,14 +315,14 @@ const applyTimeBasedConditions = (conditions: TimeBasedCondition[]) => (
       if (cond === 'createdHoursAgo') {
         return {
           ...accum,
-          [key]: differenceInHours(ctx.curentTimestamp, parseISO(target.createdAt)) < param,
+          [key]: differenceInHours(ctx.curentTimestamp, parseISO(target?.createdAt)) < param,
```
stephenhand, line 318: What is the effect of calling `parseISO(undefined)`? An 'Invalid Date`? In which case what is the effect of calling `differenceInHours` with an `Invalid Date`? Regardless, I don't think it's a good idea to leave it up to the person reading the code to figure this out. Explicitly handling the scenario where `target` is undefined will make the code & logic much easier to follow

stephenhand (commented):
See inline comment. Also can you explain what you are changing and why in the PR description. Reviewers of this PR don't really know why the code change fixes the referenced bug

plugin-hrm-form/src/permissions/index.ts
acedeywin: @stephenhand  - So I should explicitly check for `target && target` rather than using `?` to check when it's `undefined`?

stephenhand: Yes, always return false if target is undefined

acedeywin: @stephenhand - It's done

plugin-hrm-form/src/permissions/index.ts
```diff
@@ -315,14 +315,14 @@ const applyTimeBasedConditions = (conditions: TimeBasedCondition[]) => (
       if (cond === 'createdHoursAgo') {
         return {
           ...accum,
-          [key]: differenceInHours(ctx.curentTimestamp, parseISO(target.createdAt)) < param,
+          [key]: differenceInHours(ctx.curentTimestamp, target && parseISO(target.createdAt)) < param,
```
stephenhand, line 318:
```suggestion
          [key]:  target && (differenceInHours(ctx.curentTimestamp,parseISO(target.createdAt)) < param),
```

plugin-hrm-form/src/permissions/index.ts
```diff
@@ -315,14 +315,14 @@ const applyTimeBasedConditions = (conditions: TimeBasedCondition[]) => (
...
         };
       }
 
       if (cond === 'createdDaysAgo') {
         return {
           ...accum,
-          [key]: differenceInDays(ctx.curentTimestamp, parseISO(target.createdAt)) < param,
+          [key]: differenceInDays(ctx.curentTimestamp, target && parseISO(target.createdAt)) < param,
```
stephenhand, line 325:
```suggestion
          [key]: target && (differenceInDays(ctx.curentTimestamp, parseISO(target.createdAt)) < param),
```

stephenhand (commented):
Needs to short circuit the whole check

stephenhand (approved):

PR #2124, opened 2024-03-07
refactor: use dedicate endpoints to fetch case and contact counts [CHI-2624]

GPaoloni:
## Description
This PR fixes the bug ticket linked below, where "Contact/Case count (number within parenthesis) doesn't match results". To solve this issue, we decouple the case and contacts count and instead we fetch the first page using the case and contact dedicated endpoints. This solves the missmatches in the counts as permissions are taken into consideration that way.
For more information about the root cause of this bug, refer to https://github.com/techmatters/hrm/pull/587.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2624)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Configure HRM permissions in such a way that `everyone` can see contacts and cases.
- Start HRM local server (`npm run start:light`), start Flex server pointing to the local HRM (`npm start dev:local`)
- Confirm that the case and contact count works as expected - count matches the lists views.
- Configure HRM permissions in such a way that no one can see contacts and cases.
- Confirm that the case and contact count works as expected - count matches the lists views.

### DIFF
--- a/plugin-hrm-form/src/states/profile/hooks/useProfileRelationshipsBytype.ts
+++ b/plugin-hrm-form/src/states/profile/hooks/useProfileRelationshipsBytype.ts
@@ -1,0 +1,76 @@
+/* eslint-disable import/no-unused-modules */
+/**
+ * Copyright (C) 2021-2023 Technology Matters
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+
+import { useCallback, useEffect } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+
+import { selectProfileRelationshipsByType } from '../selectors';
+import asyncDispatch from '../../asyncDispatch';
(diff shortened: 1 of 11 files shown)


plugin-hrm-form/src/states/profile/hooks/useProfileRelationshipsBytype.ts
```diff
@@ -0,0 +1,76 @@
...
+    loading,
+    loadProfileRelationshipsByTypeAsync,
+  };
+};
+
+export const useProfileRelationshipsByType = ({ profileId, type, page }: UseProfileRelationsByType) => {
+  // Triggerr a load on the profile relationships
+  useProfileRelationshipsLoaderByType({ profileId, type, page });
+
+  const data = useSelector((state: RootState) => selectProfileRelationshipsByType(state, profileId, type))?.data;
```
stephenhand, line 63: What's the reason to repeatedly call the selector rather than calling it once & then destructuring?

I'm guessing it's related to object stability & reducing rerenders?

stephenhand (approved):
Looks good to merge, just one Q but I think I know the answer anyway

plugin-hrm-form/src/states/profile/hooks/useProfileRelationshipsBytype.ts
GPaoloni: We may not have a profile state for the given id the first time this selector is called, and that results in a big kaboom-blank-page. This approach ensures that never happens. We could do `const data = useSelector((state: RootState) => selectProfileRelationshipsByType(state, profileId, type))?.data || {} ;` but then we might get fake "different states" (if ? returns null we get `undefined` as value, but if the reducer is initialized and the actual property is `null` we get an unnecessary update because `undefined !== null`). This approach ensures you get `null` until you get something for real. Not a big deal but I've had component level hooks being re-triggered because of this silly thing so I'm getting into the habit of thinking when `undefined !== null` can actually mean something.

PR #2132, opened 2024-03-11
CHI-2456: minimal fix

stephenhand:
## Description

Minimal required changes to fix the issues described in CHI-2456. There is a proper fix aligned with long term refactoring goals at https://github.com/techmatters/flex-plugins/pull/2116 - but the scope of this PR snowballed beyond what I feel should be merged onto an RC branch, so will be merged to master instead

* Added effects to load cases from HRM in Case.tsx & ContactAddedToCaseBanner.tsx
* Fixed routing so ProfileRouter.tsx correctly routes to a case launched from a contact even though it has no context
* Fixed logic to display Save and End button to not required populated connectedContacts to be loaded

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### DIFF
--- a/plugin-hrm-form/src/components/caseMergingBanners/ContactAddedToCaseBanner.tsx
+++ b/plugin-hrm-form/src/components/caseMergingBanners/ContactAddedToCaseBanner.tsx
@@ -78,7 +78,7 @@ const ContactAddedToCaseBanner: React.FC<Props> = ({
     if (caseId && !connectedCase) {
       loadCase(caseId);
     }
-  });
+  }, [caseId, connectedCase, loadCase]);
   /*
   TODO: Convert to a custom hook since it has been used in several places within
   the Flex-plugins repo?


mythilytm (approved):
Works great

murilovmachado (approved):

GPaoloni:
Don't consider this a review, just asking to know: 
- How does this handle case references?
- Is this one going to be omitted when merging rc into master or is intended to coexist with https://github.com/techmatters/flex-plugins/pull/2116?

stephenhand:
> Don't consider this a review, just asking to know:
> 
> * How does this handle case references?
> * Is this one going to be omitted when merging rc into master or is intended to coexist with [fix: fetch case when not loaded into state [CHI-2456] #2116](https://github.com/techmatters/flex-plugins/pull/2116)? (Assuming the later is merged as that's what I understand from the description :p)

Sorry - I was going to say on the PR - this minimal fix probably introduces a leak with references, but it's temporary and I doubt anyone views THAT many cases in a single session for it to cause major issues for the few weeks it will be live. I know I was a stickler about references with your PR but that is the long term fix :-P

Yes - I don't think we need to explicitly back this out when we merge this into master, just reject all the conflicts when merging back up

PR #2139, opened 2024-03-11
[bug] Fixes usage of commas in yellow banner 

mythilytm:
## Description
- This PR ensures consistency in the use of commas for the two versions of the yellow banner. 

### Checklist
- [n/a] Corresponding [issue](https://tech-matters.slack.com/archives/C06G61NK44E/p1709738111201399) has been opened - 
- [n/a] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Other Related Issues
- https://tech-matters.slack.com/archives/C06G61NK44E/p1709738111201399

### Verification steps
1. Start running the app and initiate a call / chat
2. In `plugin-hrm-form/src/components/profile/IdentifierBanner/ProfileIdentifierBanner.tsx`, change/hardcode `contactsCount` and `casesCount` to 0, and then 1 (4 scenaios), then look for the yellow banner to have the correct use of commas. 
3. Hardcode `enableClientProfiles` in `plugin-hrm-form/src/components/profile/IdentifierBanner/index.tsx` to false and run through the 4 scenarios in `plugin-hrm-form/src/components/profile/IdentifierBanner/PreviousContactsBanner.tsx`


plugin-hrm-form/src/components/profile/IdentifierBanner/PreviousContactsBanner.tsx
```diff
@@ -112,9 +106,11 @@ const PreviousContactsBanner: React.FC<Props> = ({
       </BannerLink>
       {casesCount > 0 && (
         <>
-          {contactsCount > 0 && ','}&nbsp;
-          <Template code="PreviousContacts-And" />
-          &nbsp;
+          {contactsCount > 0 && <>&nbsp;</>}
```
stephenhand, line 109: Honestly I've never liked that we build up sentences dynamically like this, it doesn't account for languages whose grammar might not be the same as English. Each translatable section of text should form a complete phrase, and in cases like this where we can't really use the template interpolation, we should break the information up into separate discrete phrases.

I know this isn't something you can do in this PR because it needs consultation with design & product, just want to call it out because we should be pushing back on it for future designs & changes

stephenhand (approved):
One non-actionable comment, so LGTM!

PR #2143, opened 2024-03-14
CHI-2654: Move 'copy to perpetrator' logic to redux

stephenhand:
## Description

The root cause of this issue was another 'weird values being returned by RHF for some reason' issue. These issues have proven tricky to fully diagnose in the past and usually result in less than satisfactory non intuitive fixes that are easy to break in future work.

Because of this, I decided to undertake a small piece of refactoring that is already on our 'TODO list', which is to move the 'copy to' logic in the redux layer.

This way the copy logic can all be run against the redux case 'workingCopy' which we know has correct sanitized values

As part of this work, I changed 'copyTo' fields in form definitions to be standard fields, rather than 'non saveable' fields so that they are persisted to the redux working copy. In their only current usage, in case sections, the redux layer strips the field from the data sent to HRM. 

The only effect would be if 'copyTo' fields were used in other Aselo forms where there is no specific handling for them, they will be persisted as a regular field. I think this is a feature rather than a bug. If this form item type is used in a form where it isn't properly supported, at least we record that somebody intended to copy something in the saved data. We could potentially use this later to fix up the data, unlike if we just dropped that value on the floor like we did previously.

This is probably a 'revealed attack' scenario, the bug existed for a while but until recently the feature wasn't working at all, so was only shown when the first issue was fixed. I think the initial issue was fixed as part of the work to migrate case sections to be updated via their own HRM endpoints rather than the old general endpoint

### Checklist
- [ ] New tests added
- N/A Feature flags added
- N/A Strings are localized

TODO: unit test coverage for the copy-to case in the redux state would be good, but the component level logic it replaces has no test coverage either, so not PR blocking I think :-P

### Other Related Issues

None

### Verification steps

See ticket

### DIFF
--- a/hrm-form-definitions/form-definitions/as/v1/caseForms/HouseholdForm.json
+++ b/hrm-form-definitions/form-definitions/as/v1/caseForms/HouseholdForm.json
@@ -166,7 +166,6 @@
     "name": "copyToPerpetrator",
     "label": "This household member is also a perpetrator",
     "type": "copy-to",
-    "target": "perpetrators",
-    "saveable": false
+    "target": "perpetrators"
   }
 ]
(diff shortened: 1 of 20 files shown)


murilovmachado (approved):
LGTM! Thanks for the root cause analysis described in the PR.

PR #2144, opened 2024-03-14
CHI-2614: Fix issue with the save and end button

acedeywin:
## Description

- Bug fix: Error when clicking the `Save and end` button multiple times.

- Root cause: This bug can be reproduced only when trying to save online contact. This is because the `completeTask` `async` function takes more time to resolve while saving online contact. By the time `submitContactFormAsyncAction` gets resolved, the `loadingStatus` is updated to loaded and as such enables the "Save and End" button, which can now be clicked, thus causing the error.

- Fix: To fix this, we need to always set the `loadingStatus` to loading and keep the "Save and end" button in a loading state until `completeTask` gets resolved.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### Verification steps

- Start a web chat
- Accept the task as a counsellor
- Fill in the necessary details and click on the `Save and end` button
- Verify that the `Save and end` button cannot be clicked a second time while contact is been saved.

### DIFF
--- a/plugin-hrm-form/src/states/contacts/saveContact.ts
+++ b/plugin-hrm-form/src/states/contacts/saveContact.ts
@@ -227,7 +227,6 @@ export const loadContactIntoRedux = (
   contact: Contact,
   reference?: string,
   newMetadata?: ContactMetadata,
-  loadingStatus?: boolean,
 ): ContactsState => {
   const { existingContacts } = state;
   const references = existingContacts[contact.id]?.references ?? new Set();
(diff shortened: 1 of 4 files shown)


stephenhand (commented):
How does the status get set back to 'LOADED' after the task work is completed? It seems like this would leave the contact in a LOADING state forever?

acedeywin:
@stephenhand - Update has been implemented.

plugin-hrm-form/src/states/contacts/saveContact.ts
```diff
@@ -243,6 +243,10 @@ export const loadContactIntoRedux = (
       ...existingContacts,
       [contact.id]: {
         ...existingContacts[contact.id],
+        /*
```
stephenhand, line 246: Need to remove this comment now?

stephenhand (approved):
1 clean up comment, otherwise LGTM

nickhurlburt:
CHI-2591

PR #2161, opened 2024-03-15
Potential fix for flaky E2E test failures

mythilytm:
## Description
- E2E tests have been failing and one of the issues being `Add Offline Contact` button was not getting clicked given it is disabled while loading. The test run is then skipping the check whether the button is clicked. This PR adds some time to wait and check if the tabbedForms will load.
- I also added a new script for ease of testing locally

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### Verification steps
- E2E test is passing!!

### DIFF
--- a/e2e-tests/agent-desktop.ts
+++ b/e2e-tests/agent-desktop.ts
@@ -29,7 +29,11 @@ export const agentDesktop = (page: Page) => {
     for (let i = 0; i < 3; i++) {
       try {
         await addOfflineContactButton.click();
-        await expect(addOfflineContactButton).not.toBeEnabled({ timeout: 2000 });
+        await expect(addOfflineContactButton).not.toBeEnabled({ timeout: 5000 });
+        const dataCallTypeButton = page.locator(
+          `//button[@data-testid='DataCallTypeButton-child']`,
+        );
+        await expect(dataCallTypeButton).toBeVisible({ timeout: 5000 });
         break;
       } catch (e) {
         if (i === 2) {
--- a/e2e-tests/package.json
+++ b/e2e-tests/package.json
@@ -11,6 +11,7 @@
     "test:local": "cross-env DEBUG=pw:api LOAD_SSM_CONFIG=true npm run test",
     "test:development:as": "cross-env DEBUG=pw:api LOAD_SSM_CONFIG=true HL_ENV=development HL=as SKIP_DATA_UPDATE=true npm run test",
     "test:development:e2e": "cross-env DEBUG=pw:api LOAD_SSM_CONFIG=true HL_ENV=development HL=e2e npm run test",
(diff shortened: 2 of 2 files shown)


e2e-tests/agent-desktop.ts
```diff
@@ -29,7 +29,11 @@ export const agentDesktop = (page: Page) => {
     for (let i = 0; i < 3; i++) {
       try {
         await addOfflineContactButton.click();
-        await expect(addOfflineContactButton).not.toBeEnabled({ timeout: 2000 });
+        await expect(addOfflineContactButton).not.toBeEnabled({ timeout: 5000 });
```
nickhurlburt, line 32: This is good as a temporary fix. As noted in offline conversation with Mythily, the core problem seems to be that the button is _visible_ by L28, but isn't _clickable_. Therefore this change isn't really what it seems. While the code says “after clicking, wait 5 seconds for the button to be disabled”, in actuality what we’re doing is “the button might not be clickable on the first go-around, so stick a 5-second wait in there, and hopefully the button will be loaded by the second go-around”. While I'm not sure how, we should make sure it's clickable before we get to L31.

It looks like an additional problem we have is that the Add Offline Contact button is not being disabled quickly enough. By trying this in my browser on Slow 3G, I can click on that button several times before it gets disabled. I would guess this stems from disabling logic that made sense when Add Offline Contact did not make any backend calls to start. Perhaps we need to change that logic so that it is disabled immediately, before the backend call is made. There's no reason to wait to disable it once it's clicked.

e2e-tests/package.json
```diff
@@ -11,6 +11,7 @@
     "test:local": "cross-env DEBUG=pw:api LOAD_SSM_CONFIG=true npm run test",
     "test:development:as": "cross-env DEBUG=pw:api LOAD_SSM_CONFIG=true HL_ENV=development HL=as SKIP_DATA_UPDATE=true npm run test",
     "test:development:e2e": "cross-env DEBUG=pw:api LOAD_SSM_CONFIG=true HL_ENV=development HL=e2e npm run test",
+    "test:development:e2e:local": "cross-env LOAD_SSM_CONFIG=true HL_ENV=development HL=e2e npm run test -- --headed",
```
nickhurlburt, line 14: Cool! What does this do differently than the `debug` script immediately below?

nickhurlburt (approved):
Good as a fix to get things moving, though it's surfacing bigger issues, captured in comments below

e2e-tests/package.json
mythilytm: They are similar but I find `DEBUG=pw:api` is not helpful because it 'pollutes' with a lot of unhelpful logs and makes it hard to read relevant debugging statements. The `headed` option helps see the test run in the browser. I did not want to delete the one below in case it is more helpful for some one with a different IDE or way to debug

PR #2168, opened 2024-03-18
CHI-2656: Cancel existing case bug

stephenhand:
## Description

* Fix logic to hide 'cancel banner' when a case is not created for this contact
* Silence console log error spam around tabbed forms

### Checklist
- N/A New tests added
- N/A Feature flags added
- [X] Strings are localized
- [ X] Tested for call contacts

### Other Related Issues

None

### Verification steps

See ticket

### DIFF
--- a/plugin-hrm-form/src/components/case/CaseHome.tsx
+++ b/plugin-hrm-form/src/components/case/CaseHome.tsx
@@ -57,7 +57,7 @@ import { selectDefinitionVersionForCase } from '../../states/configuration/selec
 import selectCaseHelplineData from '../../states/case/selectCaseHelplineData';
 import { selectCounselorName } from '../../states/configuration/selectCounselorsHash';
 import { contactLabelFromHrmContact } from '../../states/contacts/contactIdentifier';
-import { selectFirstContactByCaseId } from '../../states/contacts/selectContactByCaseId';
+import { selectContactsByCaseIdInCreatedOrder } from '../../states/contacts/selectContactByCaseId';
 
 export type CaseHomeProps = {
   task: CustomITask | StandaloneITask;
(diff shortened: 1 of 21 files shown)


plugin-hrm-form/src/components/case/CaseHome.tsx
```diff
@@ -74,15 +74,18 @@ const mapStateToProps = (state: RootState, { task }: CaseHomeProps) => {
   const taskContact = isStandaloneITask(task) ? undefined : selectContactByTaskSid(state, task.taskSid)?.savedContact;
   const routing = selectCurrentTopmostRouteForTask(state, task.taskSid) as CaseRoute;
   const { connectedCase, availableStatusTransitions = [] } = connectedCaseState ?? {};
+  const caseContacts = selectContactsByCaseIdInCreatedOrder(state, routing.caseId);
 
-  const contactForLabel = selectFirstContactByCaseId(state, routing.caseId)?.savedContact ?? taskContact;
-  const isCreating = Boolean(taskContact && taskContact.caseId === routing.caseId && !taskContact.finalizedAt);
+  const contactForLabel = caseContacts[0]?.savedContact ?? taskContact;
+  const isNewContact = Boolean(taskContact && taskContact.caseId === routing.caseId && !taskContact.finalizedAt);
+  const isNewCase = caseContacts.length === 1 && taskContact && taskContact.caseId === routing.caseId;
```
murilovmachado, line 81: Is this check strong enough to determine if this is a new case?

Just by reading this line it only checks:
- If the case has 1 contact
- And this contact is the taskContact

murilovmachado (approved):
Looks good to me. I had only one question.

plugin-hrm-form/src/components/case/CaseHome.tsx
stephenhand: Because there is no such thing as a 'new case' in Aselo, I thought the best thing to do is check if the case has no other contacts than the once you are currently creating. It's possible that an older case could get into this state by removing all the existing contacts from it except the one you are creating now - but do we really care if the user deletes the case at that point? It would just be a voided case anyway if it was left

plugin-hrm-form/src/components/tabbedForms/TabbedFormsTabs.tsx
```diff
@@ -212,7 +212,11 @@ const TabbedFormsTabs: React.FC<Props> = ({
   } = updatedContact;
 
   const tabsToIndex = mapTabsToIndex(savedContact, updatedContact.rawJson);
-  const tabIndex = tabsToIndex.findIndex(t => t === subroute);
+  // tabIndex -1 doesn't cause user facing errors but it does generate lots of console log noise
```
nickhurlburt, line 215: Yay!!! Thank you for cleaning up log spam!

PR #2174, opened 2024-03-19
CHI-2566 Filtering Counselor skills in Teams view

mythilytm:
## Description
- This PR adds the feature to filter through counsellor skills in the Teams view.
- Adds a new feature flag: `enable_teams_view`

### Checklist
- [n/a] New tests added
- [x] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### Verification steps
- Hardcode `enable_teams_view` to true in two places where it is used
- Filter through the Skills in the Filter side panel

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.tsx
+++ b/plugin-hrm-form/src/HrmFormPlugin.tsx
@@ -140,7 +140,7 @@ const setUpComponents = (
     if (featureFlags.enable_canned_responses) Components.setupCannedResponses();
   }
 
-  if (featureFlags.enable_teams_view) TeamsView.setUpSkillsColumn();
+  TeamsView.setUpSkillsColumn();
   TeamsView.setUpTeamViewFilters();
   TeamsView.setUpWorkerDirectoryFilters();
 
--- a/plugin-hrm-form/src/HrmFormPlugin.tsx
+++ b/plugin-hrm-form/src/HrmFormPlugin.tsx
@@ -48,7 +48,7 @@ import { setUpTransferActions } from './transfer/setUpTransferActions';
 import { playNotification } from './notifications/playNotification';
 import { namespace } from './states/storeNamespaces';
 import { setUpTransferComponents } from './components/transfer/setUpTransferComponents';
-import { setUpSkillsColumn } from './components/teamsView';
+import TeamsView from './components/teamsView';
 
 const PLUGIN_NAME = 'HrmFormPlugin';
 
@@ -140,9 +140,9 @@ const setUpComponents = (
     if (featureFlags.enable_canned_responses) Components.setupCannedResponses();
   }
 
-  Components.setupTeamViewFilters();
-  setUpSkillsColumn();
-  Components.setupWorkerDirectoryFilters();
+  TeamsView.setUpSkillsColumn();
+  TeamsView.setUpTeamViewFilters();
+  TeamsView.setUpWorkerDirectoryFilters();
 
   if (featureFlags.enable_conferencing) setupConferenceComponents();
 };
--- a/plugin-hrm-form/src/HrmFormPlugin.tsx
+++ b/plugin-hrm-form/src/HrmFormPlugin.tsx
@@ -140,7 +140,7 @@ const setUpComponents = (
     if (featureFlags.enable_canned_responses) Components.setupCannedResponses();
   }
 
-  TeamsView.setUpSkillsColumn();
+  if (featureFlags.enable_teams_view) TeamsView.setUpSkillsColumn();
   TeamsView.setUpTeamViewFilters();
   TeamsView.setUpWorkerDirectoryFilters();
 
--- a/plugin-hrm-form/src/components/teamsView/Filters.ts
+++ b/plugin-hrm-form/src/components/teamsView/Filters.ts
@@ -58,33 +58,24 @@ const skillsFilterDefinition: FilterDefinitionFactory = (_appState, _teamFilters
   };
 };
 
+/**
+ * This function sets up filters for the TeamsView component
+ * The activity filter omits the default since we already include offline in the above, ie Flex.TeamsView.activitiesFilter
+ * The skills filter is included if the feature flag is enabled.
+ */
 export const setUpTeamViewFilters = () => {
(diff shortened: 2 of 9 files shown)


plugin-hrm-form/src/types/types.ts
```diff
@@ -293,6 +293,7 @@ export type FeatureFlags = {
   enable_twilio_transcripts: boolean; // Enables Viewing Transcripts Stored at Twilio
   enable_upload_documents: boolean; // Enables Case Documents
   enable_voice_recordings: boolean; // Enables Loading Voice Recordings
+  enable_teams_view: boolean; // Enables custom Teams View UI
```
nickhurlburt, line 296: Alphabetical please!

plugin-hrm-form/src/HrmFormPlugin.tsx
```diff
@@ -140,9 +140,9 @@ const setUpComponents = (
     if (featureFlags.enable_canned_responses) Components.setupCannedResponses();
   }
 
-  Components.setupTeamViewFilters();
-  setUpSkillsColumn();
-  Components.setupWorkerDirectoryFilters();
+  if (featureFlags.enable_teams_view) TeamsView.setUpSkillsColumn();
+  TeamsView.setUpTeamViewFilters();
```
nickhurlburt, line 144: Seems strange to me to check the feature flag here for `setUpSkillsColumn()` but check it inside of `setUpTeamViewFilters()`. Is it necessary to do it that way?

twilio-iac/helplines/as/configs/service-configuration/development.json
```diff
@@ -22,7 +22,8 @@
             "enable_transcripts": true,
             "backend_handled_chat_janitor": false,
             "enable_voice_recordings": false,
-            "enable_post_survey": false
+            "enable_post_survey": false,
+            "enable_teams_view": true
```
nickhurlburt, line 26: please add these alphabetically, even if it's out of order to start. It makes it much easier to track these after they are added.

twilio-iac/helplines/as/configs/service-configuration/staging.json
```diff
@@ -17,7 +17,8 @@
             "enable_csam_report": true,
             "enable_lex": true,
             "enable_save_insights": true,
-            "enable_voice_recordings": true
+            "enable_voice_recordings": true,
+            "enable_teams_view": true
```
nickhurlburt, line 21: same

nickhurlburt:
gave a few quick comments, though someone else should do a more complete code review

plugin-hrm-form/src/HrmFormPlugin.tsx
mythilytm: There was a feature introduced in teams view where activity of `Offline` checked off, while the rest are on, if you look for `activityNoOfflineByDefault`. This affects all helplines, hence I am assuming that we still want to show this by default. 

A cleaner/consistent way could be to have the check with in the `setUpSkillsColumn` module

nickhurlburt: Hmm... I see. We can see if another review has thoughts, based on other ways we've been handling feature flags. As is, it feels misleading when reading the code at this level, though I'm not sure how to best resolve that.

stephenhand: You could pass in the flag as an 'enableSkillFilters' parameter from here rather than reading it directly in the method, this would keep all the FF conditional code here

stephenhand: But also, the FF code *should* be temporary, so it *shouldn't* matter that it's not beautifully readable & maintainable for the long term.

twilio-iac/helplines/as/configs/service-configuration/development.json
```diff
@@ -9,20 +9,21 @@
...
             "enable_csam_clc_report": true,
             "enable_csam_report": true,
             "enable_client_profiles": true,
             "enable_fullstory_monitoring": true,
             "enable_lex": true,
+            "enable_post_survey": false,
             "enable_resources": true,
             "enable_resources_elastic_search": true,
             "enable_save_insights": true,
+            "enable_teams_view": true,
```
stephenhand, line 24: Perhaps we could call this something a bit more descriptive of the change, like `enable_teams_view_enhancements` or `enable_teams_view_skills`?

plugin-hrm-form/src/components/teamsView/Filters.ts
```diff
@@ -0,0 +1,90 @@
+/**
```
stephenhand, line 1: Thanks for moving this

plugin-hrm-form/src/components/teamsView/Filters.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  // eslint-disable-next-line camelcase
+  if (enable_teams_view) {
+    TeamsView.defaultProps.filters = [
+      activityNoOfflineByDefault,
+      /*
+       * Omit the default since we already include offline in the above
+       * Flex.TeamsView.activitiesFilter
+       */
+      skillsFilterDefinition,
```
stephenhand, line 73: FYI you can optionally include things in an array using the spread operator like so:

```suggestion
      ...(enable_teams_view ? [skillsFilterDefinition] : []),
```

What you have might be more readable in this case so not asking you to change it, just showing you a useful technique in case you weren't aware. 

You can also use a similar trick to optionally include properties in objects, for when you specifically want the property to be *absent* from the object in the negative case, rather than present and set to undefined / null

stephenhand (approved):
Nothing blocking

plugin-hrm-form/src/components/teamsView/Filters.ts
mythilytm: Thanks this is v clean.

twilio-iac/helplines/as/configs/service-configuration/development.json
mythilytm: I like enable_teams_view_enhancements

PR #2177, opened 2024-03-21
CHI-2190: Contact Detail: Add to Existing Case

acedeywin:
## Description

- New feature: Contact Detail: Add to Existing Case
- Ability to select between add to new case or existing case
- The search cases screen should show up as a modal over the contact details page.
- After the contact is added to the case, it should close the Search Cases Modal and be back to the Contact Details Page, showing the blue banner.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

[CHI-2190](https://tech-matters.atlassian.net/browse/CHI-2190)

### Other Related Issues

None

### Verification steps

- Search for a contact
- Select a contact - there should be a blue banner if the contact has been added to a case or the "Add to case" button if vice versa.
- Click on the "Add to case" button and select the "Existing case" button
- Verify that you are routed to the search page with "Search for cases" as the header
- Search for your case
- Add the contact to a case
- Verify that you can see the blue banner after adding the contact to a case
- Click on "Remove from Case"
- Close the orange banner
- Verify that the "Add to case" button appears after closing the orange banner

[CHI-2190]: https://tech-matters.atlassian.net/browse/CHI-2190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugin-hrm-form/src/components/caseMergingBanners/AddToCaseBanner.tsx
+++ b/plugin-hrm-form/src/components/caseMergingBanners/AddToCaseBanner.tsx
@@ -41,7 +41,7 @@ type MyProps = {
 const mapStateToProps = (state: RootState, { task }: MyProps) => {
   const route = selectCurrentTopmostRouteForTask(state, task.taskSid);
   const { connectedCase } = selectCurrentRouteCaseState(state, task.taskSid) ?? {};
-  const contactId = selectContextContactId(state, task.taskSid, 'search', 'case-results');
+  const contactId = selectContextContactId(state, task.taskSid, 'case', 'home');
   const contact = selectContactStateByContactId(state, contactId)?.savedContact;
   return {
     connectedCase,
--- a/plugin-hrm-form/src/___tests__/components/PreviousContactsBanner.test.tsx
+++ b/plugin-hrm-form/src/___tests__/components/PreviousContactsBanner.test.tsx
@@ -180,12 +180,12 @@ test('calls the correct function when the cases link is clicked', () => {
       searchContacts={searchContacts}
       searchCases={searchCases}
       openContactSearchResults={openContactSearchResults}
-      openCaseSearchResults={openCaseSearchResults(('10001'))}
+      openCaseSearchResults={openCaseSearchResults('10001')}
     />,
   );
 
   fireEvent.click(getByTestId('banner-link-cases'));
-  expect(openCaseSearchResults).toHaveBeenCalledWith(('10001'));
+  expect(openCaseSearchResults).toHaveBeenCalledWith('10001');
 });
 
 test('a11y', async () => {


plugin-hrm-form/src/components/tabbedForms/AddCaseButton.tsx
```diff
@@ -61,7 +67,7 @@ const AddCaseButton: React.FC<Props> = ({ handleNewCaseType, handleExistingCaseT
       data-testid="BottomBar-SaveAndAddToCase-Button"
       disabled={!enabled}
     >
-      <StyledAddNewCaseDropdown style={{ display: dropdown ? 'block' : 'none' }}>
+      <StyledAddNewCaseDropdown style={conditionalStyles}>
```
mythilytm, line 70: `dropdown` and `position` can be passed as optional props like this: 
<StyledAddNewCaseDropdown dropdown top (or bottom) />

In styles, you can define the CSS properties instead. 
type StyledAddNewCaseDropdownProps = {
  dropdown: boolean;
  position: 'top' | 'bottom';
}

const StyledAddNewCaseDropdown = styled.div<StyledAddNewCaseDropdownProps>`
  display: ${props => (props.dropdown ? 'block' : 'none')};
  ${props => props.position === 'top' ? 'bottom: 100%;' : 'top: 100%;'}
`;

// Usage
<StyledAddNewCaseDropdown dropdown={dropdown} position="top" />

plugin-hrm-form/src/states/contacts/types.ts
```diff
@@ -73,6 +74,12 @@ export type ContactsState = {
   existingContacts: ExistingContactsState;
   contactsBeingCreated: Set<string>;
   contactDetails: ContactDetailsState;
+  standAloneContactId?: string;
```
stephenhand, line 77: Again, adding global state to redux is not an acceptable way to implement these sorts of features in most circumstances. We need to pass the contact ID down from the contact details page within the context of the operation.

I would suggest adding a `contextContactId` to the route for search and case routes, this can be set opening the modal and passed down tot the case search and the case page for the purposes of adding a case.

This would also be specified when opening the search or the new case modal from the tabbed form, so we would always use the contextContactId and can stop looking up the contact from the task directly in these components.

A refinement of this approach might be to add a generic `contextVars` property to routes, which always gets copied in full in the routing reducer when new routes or modals are added to the stack. You can see the same approach used for the `context` string var.

stephenhand (changes requested):

plugin-hrm-form/src/states/contacts/types.ts
acedeywin: @stephenhand - `contextContactId` has been added to routes.

plugin-hrm-form/src/states/routing/types.ts
```diff
@@ -37,6 +37,7 @@ export type Contexts = typeof CONTEXTS[number];
 
 export type RouteWithContext = {
   context?: Contexts;
+  contextContactId?: string;
```
stephenhand, line 40: Could we make this a only property for routes that it applies to i.e. the Case routes and the case search results / search form, rather than every route?

plugin-hrm-form/src/components/tabbedForms/AddCaseButton.tsx
```diff
@@ -26,11 +26,12 @@ import { StyledNextStepButton } from '../../styles/buttons';
 type OwnProps = {
```
stephenhand, line 26: Maybe we should move this button somewhere else, since we it's no longer only used in tabbed forms

plugin-hrm-form/src/components/search/CasePreview/index.tsx
```diff
@@ -46,7 +48,17 @@ type OwnProps = {
...
-  const taskContact = isStandaloneITask(task) ? undefined : selectContactByTaskSid(state, task.taskSid)?.savedContact;
+  let contactId: string;
+
+  const currentRoute = getCurrentTopmostRouteForTask(state[namespace].routing, task.taskSid);
+
+  if (currentRoute.route === 'search' && currentRoute.subroute === 'case-results') {
+    contactId = currentRoute.contextContactId;
+  }
+
+  const taskContact = isStandaloneITask(task)
```
stephenhand, line 59: We should pass the task contact ID as the route's contextContactId when we launch the search form from the tabbed forms, so that here we just look up the contectContextId here and on the case form, rather than having to check in 2 places

stephenhand (commented):
Couple of comments - also, do we not need to update the AddToCaseBanner.tsx to update us adding the contact to the case from the Case home page

acedeywin:
> Couple of comments - also, do we not need to update the AddToCaseBanner.tsx to update us adding the contact to the case from the Case home page

@stephenhand  - Can you throw more light on this? I not clear what you mean.

plugin-hrm-form/src/components/search/CasePreview/index.tsx
acedeywin: @stephenhand  - Can you throw more light on this?

plugin-hrm-form/src/components/tabbedForms/AddCaseButton.tsx
acedeywin: @stephenhand - Can you throw more light on this? If I am to move it, where should I move it?

plugin-hrm-form/src/components/search/CasePreview/index.tsx
stephenhand: When we open a case search or a case from the tabbed forms when adding a new contact, we should pass the ID for the new contact as the contextContactId.

Then in here we only need to check for the contextContactId, we don't need to fall back to the task contact

stephenhand:
> > Couple of comments - also, do we not need to update the AddToCaseBanner.tsx to update us adding the contact to the case from the Case home page
> 
> @stephenhand - Can you throw more light on this? It's not clear what you mean.

If we want to be able to add an existing contact to an existing case from the Case home page, rather than the search preview (see https://marvelapp.com/prototype/8afa1ae/screen/92299919), the current code on master is not correct and will need updating. We will need to pass the contextContactId down on the route from the search page and change the button logic to use it

plugin-hrm-form/src/states/contacts/getContextContactId.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import { RootState } from '..';
+import { getCurrentTopmostRouteForTask } from '../routing/getRoute';
+import { namespace } from '../storeNamespaces';
+
+const getContextContactId = (state: RootState, taskSid: string) => {
+  let contactId: string;
+
+  const currentRoute = getCurrentTopmostRouteForTask(state[namespace].routing, taskSid);
```
stephenhand, line 24: You could just use `selectCurrentTopmostRouteForTask` here, `getCurrentTopmostRouteForTask` is kinda deprecated for use outside the module, we should try to remove the 'export' from it

stephenhand (commented):
Comments aside, the changes you make cover most of the points raised in the PR. There is still the piece of work to cover connecting existing cases to existing contacts from the case home page using the AddToCaseBanner.tsx component

plugin-hrm-form/src/components/caseMergingBanners/AddToCaseBanner.tsx
```diff
@@ -30,21 +30,26 @@ import asyncDispatch from '../../states/asyncDispatch';
...
 
 type MyProps = {
   task: CustomITask | StandaloneITask;
 };
 
 const mapStateToProps = (state: RootState, { task }: MyProps) => {
   const { connectedCase } = selectCurrentRouteCaseState(state, task.taskSid) ?? {};
-  const taskContact = isStandaloneITask(task) ? undefined : selectContactByTaskSid(state, task.taskSid)?.savedContact;
+  const contactId = selectContextContactId(state, task.taskSid, 'case', 'home');
+  const taskContact = selectContactStateByContactId(state, contactId)?.savedContact;
```
stephenhand, line 43: This is not the 'taskContact' any longer, because it isn't necessarily the context associated with a task. `contextContact`, or suimply `contact`, since this is the only contact this banner interacts with, might be a better name for this variable now

plugin-hrm-form/src/components/tabbedForms/BottomBar.tsx
```diff
@@ -113,7 +113,7 @@ const BottomBar: React.FC<
   if (!showBottomBar) return null;
 
   const openSearchModal = () => {
-    openModal({ route: 'search', subroute: 'form', action: 'select-case' });
+    openModal({ contextContactId: savedContact.id, route: 'search', subroute: 'form', action: 'select-case' });
```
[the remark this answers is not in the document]
stephenhand: @acedeywin has this been addressed?

stephenhand (commented):
Almost there, couple of small comments

plugin-hrm-form/src/components/profile/IdentifierBanner/PreviousContactsBanner.tsx
```diff
@@ -126,10 +128,12 @@ PreviousContactsBanner.displayName = 'PreviousContactsBanner';
 
 const mapStateToProps = (state: RootState, { task }: OwnProps) => {
   const { taskSid } = task;
+  const contextContactId = selectContextContactId(state, taskSid, 'search', 'case-results');
```
stephenhand, line 131: It's my understanding that `selectContextContactId` only works when there is a contextContactId already on the route? Since PreviousContactsBanner is part of a tabbed form, this won't be the case, and the ID of the contact would need to be lookked up from the task using `selectContactFromTaskSid` or whatever the function is

plugin-hrm-form/src/components/tabbedForms/TabbedFormsTabs.tsx
```diff
@@ -56,14 +56,15 @@ import { TabbedFormsCommonProps } from './types';
...
+import selectContextContactId from '../../states/contacts/selectContextContactId';
 
 type OwnProps = TabbedFormsCommonProps;
 
 const mapStateToProps = (state: RootState, { task: { taskSid } }: OwnProps) => {
   const {
     [namespace]: { configuration, routing: routingState },
   } = state;
-
+  const contextContactId = selectContextContactId(state, taskSid, 'search', 'case-results');
```
stephenhand, line 67: Same comment as previous contacts - the contextContactId is not on the route here & needs to be looked up from the task. In fact here we already map the correct contact Id as `contactId` here so you don't need to expose another one

stephenhand (commented):
Some logic issues in the new code

plugin-hrm-form/src/components/tabbedForms/TabbedFormsTabs.tsx
acedeywin: @stephenhand  - It's done.

stephenhand (approved):
Looks good now, thanks @acedeywin

PR #2178, opened 2024-03-21
CHI-2559 use minimal case data

stephenhand:
## Description

This PR changes how case data is queried from HRM

Instead of everything being pulled at search / initial GET time, now only the data on the case record itself is pulled then, and the associated data is pulled separately

Cases now have a notion of 'timelines'. In Redux, each 'timeline' represents an ordered set of 'activities', which can either be a Case Section, like a household or a document or a note, or a contact associated with the case.

Data for these timelines is queried from HRM via the timeline API, where you can specify whether you want to include contacts and which case section types you would like to include, as well as pagination info to query the correct 'slice'. Order logic is hardcoded for now but will probably be configurable in the future.

When moved into redux, the payloads are not stored within the timeline structures themselves. Instead, all that is stored in the 'timelines' entry in redux is an array of objects that contain a timestamp, a type (case-section vs contact) and identifiers to find the actual object elsewhere in redux. Contacts are added to the activeContyacts.existingContacts map used for all contact data, whereas case sections are added to a new 'sections' map on the CaseStateEntry item in connectedCases.cases, alongside connectedCase, timelines, caseWorkingCopy etc.

This means that the 'Case' type no longer has any associations stored under it directly, they are stored in separate structures and referenced via timelines. I think this is a good model for other entities, like Contacts, we can move CSAM reports, referred resources and conversation media out of the savedContact in existingContact entries into their own top level -properties. This means we can deal with a single 'Case' and 'Contact' type internally in flex, without having to create a bunch of different overlapping types depending on which bits have been loaded.

**Note** - because we still rely on the 'first contact' of a case for various things like the name in some instances, cases still come from the API with a single contact if they have one, the first one that was connected. In Flex this is now stored under a property called 'firstContact'. There is a helper redux 'select' method that should be used to read the first contact in preference to reading this property directly, because a contact with the same ID in the existingContacts store should always be used in preference to one in the firstContact property. In fact, the firstContact is added to existingContacts on load, so we should always be able to use a version in there in preference 

**Timelines currently in use:**

Notes / referrals / contacts timeline
Other sections on the case home page are each a 'timeline', so there is a 'documents' timeline, a 'perpetrators' timeline etc.
Contacts on the print view

The case sections have been refactored to manage their own timelines & redux data, similar to how the main timeline was already doing it.

**Force id properties to string** - In order to ensure some specific issues in this PR were addressed and to clear up some longstanding confusion

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### Verification steps

Case search / connect / disconnect / update regression

### AFTER YOU MERGE


murilovmachado (approved):
Steve, I've skimmed through it and I'm already pre-approving this PR, everything seems ok to me. I don't want to have this PR hanging here for too long.

I'll continue doing a more in-depth review just to make sure I'm not missing some of the details. but feel free to merge this.

PR #2182, opened 2024-03-22
CHI-2660: Remove Datadog RUM monitoring

nickhurlburt:
## Description
Removes Datadog's SDK from our frontend and stops sending real-user monitoring traffic to Datadog. We have not been using this since adding Full Story, and we are about to start getting charged for it. There is no reason to continue.

I **think** this should be a patch release next week. But there's a cost vs time tradeoff here and it's a valid question.

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Verification steps
- Find an account where datadog RUM monitoring is in use (should be all accounts). Double-check by running Flex and checking the network tab for calls to datadog
- Deploy this change
- Note that datadog calls are no longer being sent

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/package.json
+++ b/plugin-hrm-form/package.json
@@ -41,7 +41,6 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.16.3",
-    "@datadog/browser-rum": "^1.20.0",
     "@emoji-mart/react": "^1.1.1",
     "@emotion/css": "^11.9.3",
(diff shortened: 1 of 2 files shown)


nickhurlburt:
@stephenhand @murilovmachado @GPaoloni could any of you provide a quick review please. It's almost trivial enough to merge without one, but I might be missing something.

murilovmachado (approved):
LGTM

PR #2185, opened 2024-03-25
CHI-2566 Teams view skills filtering proposal

mythilytm:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/teamsView/Filters.ts
+++ b/plugin-hrm-form/src/components/teamsView/Filters.ts
@@ -38,20 +38,32 @@ const activityNoOfflineByDefault: FilterDefinitionFactory = (appState, _teamFilt
   };
 };
 
+const skillsArray = Manager.getInstance().serviceConfiguration.taskrouter_skills?.map(skill => {
+  return {
+    value: skill.name,
+    label: skill.name,
+  };
+});
+
 /**
  * This function returns a list of skills defined in the taskrouter_skills configuration
  */
-const skillsFilterDefinition: FilterDefinitionFactory = (_appState, _teamFiltersPanelProps) => {
-  const skillsArray = Manager.getInstance().serviceConfiguration.taskrouter_skills?.map(skill => ({
-    value: skill.name,
-    label: skill.name,
-    default: true,
-  }));
-
+const skillsFilterDefinition: FilterDefinitionFactory = () => {
   return {
     id: 'data.attributes.routing.skills',
-    title: 'Skills',
+    title: 'Available Skills',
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/components/teamsView/Filters.ts
```diff
@@ -38,26 +38,38 @@ const activityNoOfflineByDefault: FilterDefinitionFactory = (appState, _teamFilt
   };
 };
 
-/**
- * This function returns a list of skills defined in the taskrouter_skills configuration
- */
-const skillsFilterDefinition: FilterDefinitionFactory = (_appState, _teamFiltersPanelProps) => {
-  const skillsArray = Manager.getInstance().serviceConfiguration.taskrouter_skills?.map(skill => ({
+const skillsArray = Manager.getInstance().serviceConfiguration.taskrouter_skills?.map(skill => {
```
nickhurlburt, line 41: I could be wrong, but I would guess that if `skillsArray` is defined here, then this code will only get executed once when Flex starts up. Thus if someone adds or removes skills via the Flex Admin page, someone will have to reload Flex to get them. (Again I could be wrong -- you could do a quick test to see).

To allow skills to update more frequently and also have DRY code, you could make this a function, eg.
```
const getSkillsArray = () => {
 return Manager.getInstance().serviceConfiguration.taskrouter_skills?.map(skill => {
    value: skill.name,
    label: skill.name,
  });
}
```
and then call that function in the `FilterDefinitionFactory`s below.

nickhurlburt (approved):
LGTM given one comment below, and pending QA acceptance

plugin-hrm-form/src/components/teamsView/Filters.ts
mythilytm: you are right in suspecting this

nickhurlburt:
@mythilytm I'm not sure your most recent change here actually changed it to a function that will be executed every time... I think it's still just executing once on startup. Unless you've tested this and it worked as intended?

In addition to the change in my previous comment (`const getSkillsArray = () =>` etc), you'd need to change the factory functions to be more like:
```
  const skillsArray = getSkillsArray();
  return {
    id: 'data.attributes.routing.skills',
    title: 'Enabled Skills',
    fieldName: 'skills',
    type: FiltersListItemType.multiValue,
    options: skillsArray ? sortBy(skillsArray, ['label']) : [],
    condition: 'IN',
  };
```

PR #2203, opened 2024-04-01
CHI-2670: Read text for notes on CasePrintView correctly

stephenhand:
## Description

Correct the CassePrintNotes.tsx component to infer the printed text from a case section using the same method that was previously used in the NoteActivity type

BUG CAUSE ANALYSIS:

As part of the rework to use case section specific endpoints (CHI-2422), there was an effort to standardise the types that components displaying case section data used as a data source, which included moving rendering case notes in a PDF away from using the NoteActivity type, a secondary type that was used in some parts of the UI but not in others to using the ApiCaseSection type, the standard type that is stored in Redux.

As part of this move, the part of the component rendering the text itself omitted some of the logic required to render the desired text. This logic was previously in the conversion to NoteActivity, but when the use of that type was removed, so was the mapping of the displayable note text, and it needed to be moved to the component. This was missed in the conversion

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/case/casePrint/CasePrintNotes.tsx
+++ b/plugin-hrm-form/src/components/case/casePrint/CasePrintNotes.tsx
@@ -18,20 +18,23 @@
 /* eslint-disable dot-notation */
 import React from 'react';
 import { Text, View } from '@react-pdf/renderer';
+import { DefinitionVersion } from 'hrm-form-definitions';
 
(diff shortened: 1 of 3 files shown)


murilovmachado (approved):
Looks good to me!

PR #2208, opened 2024-04-02
CHI-2574 Enhance teams view task card with queue name

mythilytm:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/teamsView/TaskCardEnhancement.ts
+++ b/plugin-hrm-form/src/components/teamsView/TaskCardEnhancement.ts
@@ -1,0 +1,57 @@
+/**
+ * Copyright (C) 2021-2023 Technology Matters
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+import { DefaultTaskChannels, TaskChannels } from '@twilio/flex-ui';
+
+import { TwitterChatChannel, InstagramChatChannel, LineChatChannel } from '../../channels/setUpChannels';
+import { getAseloFeatureFlags, getTemplateStrings } from '../../hrmConfig';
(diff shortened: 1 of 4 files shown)


plugin-hrm-form/src/components/teamsView/TaskCardEnhancement.ts
```diff
@@ -0,0 +1,61 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+import { DefaultTaskChannels } from '@twilio/flex-ui';
+
+import { setupTwitterChatChannel, setupInstagramChatChannel, setupLineChatChannel } from '../../channels/setUpChannels';
```
GPaoloni, line 18: :bangbang: I think calling this functions will execute all it's logic again. While may not be causing bugs at the moment, I don't think we want to do that still (we'll have to guarantee impotence).
Maybe a better approach is to export
```
const setTaskCardInfo = channel => {
    channel.templates.TaskCard.firstLine = task => {
      const identifier = maskIdentifiers ? strings.MaskIdentifiers : '@';
      return `${task.queueName} | ${identifier}`;
    };
  };

```  
and call that function only once when the channels are initialized?

GPaoloni: Another approach might be to export
```
const TaskChannels = [
    DefaultTaskChannels.Chat,
    DefaultTaskChannels.ChatSms,
    DefaultTaskChannels.Default,
    DefaultTaskChannels.ChatMessenger,
    DefaultTaskChannels.ChatWhatsApp,
]
```
from `plugin-hrm-form/src/channels/setUpChannels.tsx`
and then push into the task channels array the custom channels as they are initialized? Also notice that Modica is missing in the list, I don't remember where that is initialzied but should be included here.

GPaoloni (commented):
Leaving some notes around the approach you are taking here. I may be wrong, I'm happy to talk more about this one!

plugin-hrm-form/src/components/teamsView/TaskCardEnhancement.ts
mythilytm: I understand that Modica is part of the SMS channel. Is that right @murilovmachado ?

mythilytm: I will explore the best way to only call this function only when the channels are initialized.

GPaoloni (approved):
LGTM!

PR #2228, opened 2024-04-09
CHI-2672: Ensure case sections are correctly sorted for display

stephenhand:
## Description

* Contrary to prior assumptions, the case sections were coming from the API unsorted. This is fixed with the move to timelines, but needs a temporary client side fix here

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Other Related Issues

None

### Verification steps

See ticket

### AFTER YOU MERGE


murilovmachado (approved):
Looks good to me.

When the backend gets fixed this won't be needed anymore. Will we merge this into `master` in the future? Or the plan is for this commit to live at the RC-branch only?

stephenhand:
> Looks good to me.
> 
> When the backend gets fixed this won't be needed anymore. Will we merge this into `master` in the future? Or the plan is for this commit to live at the RC-branch only?

I'll probably 'merge' it, even if I don't take the actual changes, just to keep the branches clean. It will be a PITA if this always has commits on it we never merge to master

PR #2229, opened 2024-04-10
CHI-2661: Extra recording check

stephenhand:
## Description

Because FS data revealed that often, when the conference callSid is not available, the conversations.segment_link is, we check for this first and infer the recording location from there if it's available

Because the opposite situation is also true, the original check needs to remain in place

### Checklist
- [X] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Other Related Issues

None

### Verification steps

General regression. Check voice media is saving for voice calls consistently

### AFTER YOU MERGE


murilovmachado (approved):
LGMT.

I've noticed the sorting sections commit is present in this PR as well. Make sure to do the necessary steps, so this PR contains only the new code when merging this.

stephenhand:
> LGMT.
> 
> I've noticed the sorting sections commit is present in this PR as well. Make sure to do the necessary steps, so this PR contains only the new code when merging this.

Sorry, it was branched off this other branch initially, fixed now

PR #2244, opened 2024-04-11
CHI-2190: Add existing case to contact fixes

acedeywin:
## Description

- This PR addresses the QA feedback from the product team as seen [here](https://www.figma.com/file/3m7IJjYbussNcTs2VRhfYz/case-merging---QA?type=design&node-id=261-409&mode=design&t=4bGYTrp9CwtivH6N-0).

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### Verification steps

[CHI-2190](https://tech-matters.atlassian.net/browse/CHI-2190)

### AFTER YOU MERGE

[CHI-2190]: https://tech-matters.atlassian.net/browse/CHI-2190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugin-hrm-form/src/components/search/SearchForm/index.tsx
```diff
@@ -130,7 +134,10 @@ const SearchForm: React.FC<Props> = ({
     (helpline && helpline.value) ||
     contactNumber;
 
-  const submitSearch = () => handleSearch(searchParams);
+  const submitSearch = () => {
```
stephenhand, line 137: Is clearing the search form really a requirement now? This doesn't seem great as a UX. The idea of keeping the settings is that the user can refine their search iteratively without all having to re-enter all the same settingds each time. Keeping the settings is consistent with the Resource Search experience, which hads an explicit clear form button. 

Also this seems to be out of scope functionality changes for this story

stephenhand: If this is to address an issue where the form to search for a case has the same settings that you originally searched for contacts on, then this is an artefact of the fact that we are only tracking the state of 1 search per task, and we need to be able to track multiple. 

I suspect this underlying issue will cause a lot of user facing bugs and needs to be addressed at the root cause, by changing our redux store to be able to track multiple search states per task

acedeywin: @stephenhand - This was part of the feedback from the Product team, specifically Jana, as seen [here](https://www.figma.com/file/3m7IJjYbussNcTs2VRhfYz/case-merging---QA?type=design&node-id=261-409&mode=design&t=wgneHwOO82RCReAs-0).

stephenhand: From that feedback, I'm assuming that the new behaviour of the original search form being blanked when you close the search results is undesirable and would be considered a bug

plugin-hrm-form/src/components/search/index.tsx
```diff
@@ -47,41 +53,48 @@ type OwnProps = {
...
     changeSearchPage: (subroute: SearchResultRoute['subroute'], action?: SearchRoute['action'], contactId?: string) =>
       dispatch(
         changeRoute(
           { route: 'search', subroute, action, casesPage: 0, contactsPage: 0, contextContactId: contactId },
           taskId,
         ),
       ),
-    searchContacts: searchContacts(dispatch)(taskId),
-    searchCases: searchCases(dispatch)(taskId),
+    searchContacts: (context: string) => searchContacts(dispatch)(taskId, context),
```
stephenhand, line 94: Can we make the searchContacts and searchCases APIs consistent?

plugin-hrm-form/src/components/search/index.tsx
```diff
@@ -47,41 +53,48 @@ type OwnProps = {
...
+const mapStateToProps = (state: RootState, { task, contactId }: OwnProps) => {
   const {
     [namespace]: { searchContacts, activeContacts, routing },
   } = state;
   const taskId = task.taskSid;
-  const taskSearchState = searchContacts.tasks[taskId];
   const currentRoute = getCurrentTopmostRouteForTask(routing, taskId);
+  const contextContactId = selectContextContactId(state, taskId, 'search', 'form');
+  const searchRootId =
+    currentRoute.route === 'search' && currentRoute.action === 'select-case' ? contextContactId : contactId || 'root';
```
stephenhand, line 64:
```suggestion
    contextContactId ? `contact-${contextContactId} : 'root';
```

plugin-hrm-form/src/components/search/index.tsx
```diff
@@ -47,41 +53,48 @@ type OwnProps = {
...
-const mapStateToProps = (state: RootState, { task }: OwnProps) => {
+const mapStateToProps = (state: RootState, { task, contactId }: OwnProps) => {
   const {
     [namespace]: { searchContacts, activeContacts, routing },
   } = state;
   const taskId = task.taskSid;
-  const taskSearchState = searchContacts.tasks[taskId];
   const currentRoute = getCurrentTopmostRouteForTask(routing, taskId);
+  const contextContactId = selectContextContactId(state, taskId, 'search', 'form');
+  const searchRootId =
```
stephenhand, line 63: Confusing variable name, change to `searchContext`

plugin-hrm-form/src/components/search/index.tsx
```diff
@@ -100,14 +113,23 @@ const Search: React.FC<Props> = ({
...
+  contextContactId,
+  handleCreateNewSearch,
+  searchRootId,
 }) => {
   const [mockedMessage, setMockedMessage] = useState('');
   const [searchParams, setSearchParams] = useState<any>({});
 
+  useEffect(() => {
+    if (!form) {
+      handleCreateNewSearch(contextContactId);
```
stephenhand, line 125: Use searchContext, not ID

plugin-hrm-form/src/components/tabbedForms/TabbedFormsSearch.tsx
```diff
@@ -54,7 +53,7 @@ const TabbedFormsSearch: React.FC<Props> = ({ task, draftContact, savedContact,
     <Search
       task={task}
       currentIsCaller={savedContact?.rawJson?.callType === callTypes.caller}
-      contactId={savedContact?.id}
```
stephenhand, line 57: This doesn't see like it should be changed

plugin-hrm-form/src/states/contacts/types.ts
```diff
@@ -112,6 +112,7 @@ export type ContactUpdatingAction = {
     | typeof LOAD_CONTACT_FROM_HRM_BY_ID_ACTION_FULFILLED
     | typeof LOAD_CONTACT_FROM_HRM_BY_TASK_ID_ACTION_FULFILLED;
   payload: { contact: Contact; contactCase?: Case; previousContact?: Contact };
+  contactId?: string;
```
stephenhand, line 115: Not needed

stephenhand (changes requested):
Address comments, primarily use create and use contexts as discussed rathger than trying to just use the contact ID - also build is failing

plugin-hrm-form/src/components/search/index.tsx
acedeywin: @stephenhand - `contactId || 'root' `is needed. `contactId` is for active contact.

plugin-hrm-form/src/___tests__/states/search/reducer.test.ts
```diff
@@ -62,7 +62,7 @@ describe('search reducer', () => {
 
     const { tasks } = result;
     expect(tasks[task.taskSid]).not.toBeUndefined();
-    expect(tasks[task.taskSid][context]).toStrictEqual(newTaskEntry);
+    expect(tasks[task.taskSid]['']).toStrictEqual(newTaskEntry);
```
stephenhand, line 65: This seems strange - can we update the test to simulate passing a context?

plugin-hrm-form/src/components/search/index.tsx
stephenhand: I meant `searchRootId` should change to `searchContext`, not `contextContactId`

plugin-hrm-form/src/states/search/reducer.ts
```diff
@@ -99,7 +103,10 @@ const contactUpdatingReducer = (state: SearchState, action: ContactUpdatingActio
     ...updatedState,
     tasks: {
       ...updatedState.tasks,
-      [contact.taskId]: newTaskEntry,
+      [contact.taskId]: {
+        ...updatedState.tasks[contact.taskId],
+        [contact.id]: newTaskEntry,
```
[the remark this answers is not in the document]
stephenhand: It was line 108 I was thinking was unnecessary, not 107

plugin-hrm-form/src/states/search/reducer.ts
```diff
@@ -140,7 +139,7 @@ export function reduce(
       tasks: Object.fromEntries(
         Object.entries(state.tasks).map(([key, value]) => [
           key,
-          { ...value, root: { ...value.root, caseRefreshRequired: true } },
+          { ...value, context: { ...value.context, caseRefreshRequired: true } },
```
stephenhand, line 142: This will set the `caseRefreshRequired: true` for a context called `context` - whereas what we want to do is set it for all the contexts that currently exist under that task.

plugin-hrm-form/src/states/search/reducer.ts
```diff
@@ -128,7 +127,7 @@ export function reduce(
       tasks: Object.fromEntries(
         Object.entries(state.tasks).map(([key, value]) => [
           key,
-          { ...value, root: { ...value.root, contactRefreshRequired: true } },
+          { ...value, context: { ...value.context, contactRefreshRequired: true } },
```
stephenhand, line 130: This will set the `caseRefreshRequired: true` for a context called `context` - whereas what we want to do is set it for all the contexts that currently exist under that task.

plugin-hrm-form/src/components/search/index.tsx
```diff
@@ -59,21 +54,24 @@ const mapStateToProps = (state: RootState, { task, contactId }: OwnProps) => {
   } = state;
   const taskId = task.taskSid;
   const currentRoute = getCurrentTopmostRouteForTask(routing, taskId);
-  const contextContactId = selectContextContactId(state, taskId, 'search', 'form');
+  const searchContext =
+    selectContextContactId(state, taskId, 'search', 'form') ||
```
stephenhand, line 58: `selectContextContactId` shouldn't really need to take route and subroute parameters should it? The function itself can check if the current route is one of those which is valid to have a contextContactId, rather than the caller having to pass in the values to check against?

stephenhand (changes requested):

plugin-hrm-form/src/components/search/index.tsx
acedeywin: @stephenhand - Passing in the route and subroute parameters makes the function dynamic so we can use it to determine what the current route is in real time.

plugin-hrm-form/src/states/search/reducer.ts
acedeywin: @stephenhand - The problem is how do I set it for all contexts? 

Value must be of type (parameter) value: { [context: string]: SearchStateTaskEntry}. How do I dynamically pass in context?

plugin-hrm-form/src/components/search/index.tsx
stephenhand: The function already determines what the current route is in real time, it calls selectCurrentTopmostRouteForTask which always gives the current route for the task

The routes the support a `contextContactId` are static and don't change in real time, so can be checked by the function witrhout passing anything in I think?

plugin-hrm-form/src/states/search/reducer.ts
stephenhand: You can see a technique already being used in the layer above to ensure it is set on every task entry. where it converts the map to entries using `Object.entries`, updates each value using `.map` and then recreating the map using `Object.fromEntries`, thereby processing all the contexts in the map the same way

plugin-hrm-form/src/components/profile/IdentifierBanner/PreviousContactsBanner.tsx
```diff
@@ -129,12 +132,18 @@ PreviousContactsBanner.displayName = 'PreviousContactsBanner';
 
 const mapStateToProps = (state: RootState, { task }: OwnProps) => {
   const { taskSid } = task;
+  const { routing } = state[namespace];
   const contact = selectContactByTaskSid(state, task.taskSid);
+  const currentRoute = getCurrentTopmostRouteForTask(routing, task.taskSid);
+  const contextContactId =
```
stephenhand, line 138: You could use `selectContextContactId` here possibly?

stephenhand (approved):
One optional comment - otherwise looks good once the tests pass

PR #2245, opened 2024-04-11
fix: Frontend spams backend with failed profiles requests [CHI-2657]

GPaoloni:
## Description
This PR fixes an existing bug, where Flex will constantly try to fetch for profiles after a failed fetch. The failure can occur for several reasons, like temporary error or lack of permissions.
To fix this issue, we consider the case where there's an `error` after loading a profile, then act accordingly:
- For 5xx errors, we add a retry mechanisms (with back-off).
- For 4xx errors, since the problem is on the client, we don't retry at all.
This behavior is abstracted in a new hook, `useLoadWithRetry`.
This PR also refactors all other profiles related "loader hooks" to use the above mentioned hook.
The idea of introducing this hook comes from an attempt to standardize how we do async operations at a state level.

This PR also fixes an existing (and mostly inoffensive) bug with the profiles list, where each time page or filters are updated, Flex performs two parallel calls to HRM, requesting for the same list of profiles. The cause of this bug is that "load profiles list hook" was being called twice in the same component tree. To fix this issue, I'm moving the components to only rely on `useProfileList` hook, and passing down an `autoload: false` property on each component that's not the actual profiles list component, in order for them to consume the state but not trigger a re-load.

A followup PR will reintroduce the changes from https://github.com/techmatters/flex-plugins/pull/2116, but reusing this same pattern. The long term ideal state is one where we defer all the "async loading" to this hook, abstracting and unifying all the retry and back-off in a single place, plus standardizing at a state (Redux) level the way we handle async-related things like `loading` and `error` states.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2657)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Mock `enable_client_profiles` flag and start a local Flex server. Confirm that profile load is not attempted at all on Contact Details view.
- Start an HRM local server, changing `viewProfile` permissions to `[]` (no-one).
- Start Flex server and confirm that the client stops attempting to load a profile after the first `401` response (because of lack of permissions).
- Change HRM `viewProfile` permissions to `[["everyone"]]` and restart the server, and add an intentional 5xx error in the "get profile endpoint".
- Confirm that this time, Flex will try to reload the case after each failed attempt, with an initial back-off of 1 second, doubled after each attempt, up to a maximum of 10 attempts (can reduce the max retries count to make it easier to test).
- Change HRM to use the actual logic for "get profile endpoint", in order for the whole thing to succeed next time.
- Confirm that this time, the profile is loaded and Flex works as intended.
- **PLEASE:** Regression test around all the profiles features (profiles list, profile details view, adding/updating/removing flags and sections on a profile, etc) to confirm I'm not breaking stuff.

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/states/hooks/useLoadWithRetry.ts
+++ b/plugin-hrm-form/src/states/hooks/useLoadWithRetry.ts
@@ -1,0 +1,85 @@
+/**
+ * Copyright (C) 2021-2023 Technology Matters
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+import { useEffect, useRef } from 'react';
+
+import { ParseFetchErrorResult } from '../parseFetchError';
+
+/**
+ * @param params - Parameters for the hook
+ * @param params.loadFunction - Callback that performs the actual work
+ * @param params.error - A ParseFetchErrorResult object representing the error, if any
+ * @param params.loading - Indicates if we are in the middle of loadFunction execution
+ * @param params.safeToLoad - Indicates if the load function is safe to call
+ * @param params.shouldLoad - Indicates if the requirements are met to trigger a load
+ * @param params.retry - Indicates if the retry mechanism should be enabled
+ */
+export const useLoadWithRetry = ({
+  loadFunction,
+  error,
+  loading,
+  safeToLoad,
+  shouldLoad,
+  retry,
+}: {
+  loadFunction: () => void;
+  error: ParseFetchErrorResult;
+  loading: boolean;
+  safeToLoad: boolean;
+  shouldLoad: boolean;
+  retry: boolean;
+}) => {
+  // effect to trigger loading, if conditions are met
+  useEffect(() => {
+    if (!safeToLoad) return;
+
+    if (shouldLoad) {
+      loadFunction();
+    }
+  }, [loadFunction, safeToLoad, shouldLoad]);
+
+  // vars used to handle the retry and backoff logic
+  const retryCount = useRef(0);
+  const backoff = useRef(1);
+  const retrying = useRef(false);
+  const timerId = useRef(null);
+
+  // effect to trigger retries, in case of errors
+  useEffect(() => {
+    if (!retry) {
+      return;
+    }
+
+    if (error && !loading && !retrying.current && retryCount.current < 10) {
+      // don't retry 4xx, the problem is on the client
+      if (error.status >= 400 && error.status < 500) {
+        return;
+      }
+
+      retrying.current = true;
+      timerId.current = setTimeout(() => {
+        loadFunction();
+        retrying.current = false;
+        retryCount.current += 1;
+        backoff.current *= 2;
+      }, 1000 * backoff.current);
+    }
+  }, [error, loadFunction, loading, retry]);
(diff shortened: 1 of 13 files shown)


plugin-hrm-form/src/hooks/useFeatureFlags.ts
```diff
@@ -0,0 +1,33 @@
...
+import { getAseloFeatureFlags } from '../hrmConfig';
+
+/**
+ * This hook is calling on every re-render.
+ * Ideally we'll move FF to live as part of a context provider, so updates will "trigger a re-render"
+ * and not the other way around
+ */
+export default function useFeatureFlags() {
+  return getAseloFeatureFlags();
+}
```
GPaoloni, line 33: This is a silly hook but it made sense to start consuming from here in the other hooks, as one day (I think) we'll want to have the FF as part of a shared state (Redux, Context, wareva). Wrapping that in a hook means we'll only have to change this one rather than all the other ones. Happy to remove and go the `getAseloFeatureFlags` if you think is better.

stephenhand: Is there any benefit to having FF in state, given it is read only? I suppose they can be update 'live' in the future, although I'm not sure if that's a good thing or not, as it creates a whole new class of potential test scenarios, i.e. 'In addition to Feature flag set to this, feature flag set to that' we now have 'feature flag started a this but then changed to that'.

But still, I guess there is something to be said for keeping all our shared state in the same place, so I'm ok with it

plugin-hrm-form/src/states/hooks/useLoadWithRetry.ts
```diff
@@ -0,0 +1,85 @@
...
+  const timerId = useRef(null);
+
+  // effect to trigger retries, in case of errors
+  useEffect(() => {
+    if (!retry) {
+      return;
+    }
+
+    if (error && !loading && !retrying.current && retryCount.current < 10) {
+      // don't retry 4xx, the problem is on the client
```
stephenhand, line 66: Should we be logging out errors here?

plugin-hrm-form/src/states/hooks/useLoadWithRetry.ts
```diff
@@ -0,0 +1,85 @@
...
+        return;
+      }
+
+      retrying.current = true;
+      timerId.current = setTimeout(() => {
+        loadFunction();
+        retrying.current = false;
+        retryCount.current += 1;
+        backoff.current *= 2;
+      }, 1000 * backoff.current);
```
stephenhand, line 77:
```suggestion
      }, 1000 * Math.pow(2, retryCount.current + 1));
```

Saves us tracking another reference?

plugin-hrm-form/src/states/hooks/useLoadWithRetry.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (error && !loading && !retrying.current && retryCount.current < 10) {
+      // don't retry 4xx, the problem is on the client
+      if (error.status >= 400 && error.status < 500) {
+        return;
+      }
+
+      retrying.current = true;
+      timerId.current = setTimeout(() => {
+        loadFunction();
```
stephenhand, line 73: Do we need to check 'safeToLoad' or 'retrying' or any of the other flags again here?

stephenhand (approved):
Few minor comments / questions but overall it looks good

plugin-hrm-form/src/states/hooks/useLoadWithRetry.ts
GPaoloni: Ha, why was I re-implementing pow :laughing:. Thanks for the suggestion!

plugin-hrm-form/src/states/hooks/useLoadWithRetry.ts
GPaoloni: Hm, is not really needed, as `loadFunction` will only be called in those scenarios, and the retry will be triggered only after the above ended with error. But I don't see any harm including the two "is safe" and "should" checks here too :p
For the "is retrying" (`retrying.current`), that one is checked before placing the timeout, and only that timeout can set it back to false, so there should never happen that we span multiple retries from the same hook.

PR #2247, opened 2024-04-11
refactor: trim spaces for phone number kind of identifiers [CHI-2673]

GPaoloni:
## Description
This PR is part of https://github.com/techmatters/hrm/pull/620, where we are moving this per-channel identifier sanitization on the client side rather than on the server.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2673)
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- Regression test to see identifiers / previous contacts works as expected for the different channels. 
 
### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/utils/task.ts
+++ b/plugin-hrm-form/src/utils/task.ts
@@ -15,7 +15,7 @@
  */
 import { ITask } from '@twilio/flex-ui';
 
-import { channelTypes } from '../states/DomainConstants';
+import { ChannelTypes, channelTypes } from '../states/DomainConstants';
 import { getHrmConfig } from '../hrmConfig';
 import { CustomITask, isTwilioTask } from '../types/types';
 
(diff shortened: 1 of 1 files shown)


plugin-hrm-form/src/utils/task.ts
```diff
@@ -15,7 +15,7 @@
  */
```
GPaoloni, line 15: Maybe we could/should the logic being modified here to another file (`utils/task.ts` feels weird?), but I'm not sure where this belongs. Any suggestions?

plugin-hrm-form/src/utils/task.ts
```diff
@@ -25,18 +25,33 @@ const getContactValueFromWebchat = task => {
...
+  voice: [trimSpaces],
+  sms: [trimSpaces],
+  whatsapp: [s => s.replace('whatsapp:', ''), trimSpaces],
+  modica: [s => s.replace('modica:', ''), trimSpaces],
+  facebook: [s => s.replace('messenger:', '')],
+  instagram: [],
+  line: [],
+  twitter: [],
+  web: [],
+};
```
GPaoloni, line 41: To avoid having to specify `trimSpaces` for each "phone number related identifier", as we'll also add a `trimDashes` soon probably, we could have a mapping that says "if channel is phone number related apply this transformations", then the channel specific ones etc. Didn't wanted to spend more than 30 minutes on this for now so I'll defer that to a next :p

plugin-hrm-form/src/utils/task.ts
mythilytm: `src/channels` ?

mythilytm (approved):
Looks great!

stephenhand (approved):

PR #2250, opened 2024-04-12
fix: fetch case when not loaded into state [CHI-2456] [CHI-2657]

GPaoloni:
Before you review:
- This PR can be reviewed by commits.
- It may also help if you first focus on the changes introduced in `/states/case/` directory, then see how this changes are used in the components.

TODO:
- [x] Fix CI
- [x] Test that in case of failure the retry and back-off play nicely with cases (may need UI adjustments but that's a separate ticket since we are not doing that now anyways :smile_cat:)

## Description
This PR is a reopen of https://github.com/techmatters/flex-plugins/pull/2116, using the `useLoadWithRetry` hook introduced in https://github.com/techmatters/flex-plugins/pull/2245 for the loading logic. Refer to that PR to gather context on the broader discussion that lead to this.

**NOTES**: 
- I'm NOT handling error states at components in this PR, I'm just unifying the way this is done for case loads as it for profiles loads. We definitively should start doing so tho.
- I'm NOT unifying other case async operations with the approach (like those in `src/states/case/saveCase.ts`) a as that will increase the scope of this PR. I think it will better to do so as soon as possible tho, since there are scenarios like `handlePendingAction(handleAction, updateCaseOverviewAsyncAction.pending)` and `handleRejectedAction(handleAction, updateCaseOverviewAsyncAction.rejected)` where `loading` and `error` portions of the case state are not being properly updated. I think we should. That could also be done as part of "moving other state interactions to hooks".

### Checklist
- [x] Corresponding issue has been opened (originally [1](https://tech-matters.atlassian.net/browse/CHI-2456) but also [2](https://tech-matters.atlassian.net/browse/CHI-2657))
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Verification steps
- [ ] TODO

### AFTER YOU MERGE


stephenhand (approved):
Looks great, especially with the retry management integrated

I'd like to handle retrying down in the redux layer rather than in hooks longer term, but we need better middleware to do that well, so this is nice in the meantime

PR #2252, opened 2024-04-16
Group and Sort counselors based on activity

mythilytm:
## Description
- This PR alters the default order of the counsellors on the Teams View page which is an automatic grouping of agents based on their activity/status. For instance, agents with the status of "available" will be listed first, followed by those who are "unavailable" followed by "break" agents, and finally for "offline". Two way toggle is supported
- The activities / statuses and the order are helpline specific. 

### Checklist
- [x] Corresponding issue has been [opened](https://tech-matters.atlassian.net/browse/CHI-2570)
- [n/a] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### Verification steps
- In Teams view, assign a few counsellors with different statuses.
- Check the order when you enter the page and also by sorting(click on the carat) on the Agent header
Screenshot 2024-04-16 at 1 37 08 PM

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/teamsView/teamsViewSorting.ts
+++ b/plugin-hrm-form/src/teamsView/teamsViewSorting.ts
@@ -14,7 +14,7 @@
  * along with this program.  If not, see https://www.gnu.org/licenses/.
  */
 
-import { AgentsDataTable, TaskHelper } from '@twilio/flex-ui';
+import { AgentsDataTable, TaskHelper, Manager } from '@twilio/flex-ui';
 import { SupervisorWorkerState } from '@twilio/flex-ui/src/state/State.definition';
 
 import { getAseloFeatureFlags } from '../hrmConfig';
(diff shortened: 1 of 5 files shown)


plugin-hrm-form/src/teamsView/teamsViewSorting.ts
```diff
@@ -85,10 +85,47 @@ const sortWorkersByChatDuration = (a: SupervisorWorkerState, b: SupervisorWorker
   return 0;
 };
 
+const ACTIVITIES = Array.from(Manager.getInstance().store.getState().flex.worker.activities.values()).reduce(
+  (accum, activity, currIndex) => {
+    accum[activity.name] = currIndex;
+    return accum;
+  },
+  {},
+);
```
GPaoloni, line 94: I suggest moving this within the same function where   `AgentsDataTable.defaultProps.sortWorkers = sortWorkersByActivity;` is called (`setUpSortingCallsAndChats`), just so we make sure `Manager.getInstance().store.getState().flex.worker.activities` is initialized before this const is defined. Maybe there's no risk here, but because of how `import`s work this may cause a race conditions (I think).

plugin-hrm-form/src/teamsView/teamsViewSorting.ts
```diff
@@ -85,10 +85,47 @@ const sortWorkersByChatDuration = (a: SupervisorWorkerState, b: SupervisorWorker
...
+  // Sort by activity value/index
+  if (aActivityValue > bActivityValue) {
+    return 1;
+  } else if (aActivityValue < bActivityValue) {
+    return -1;
+  }
+
+  // If activities are the same, sort by worker name
+  return a.worker.name > b.worker.name ? 1 : -1;
+};
```
GPaoloni, line 121: The whole thing makes sense to me! However I'd like to point to some things:
- This assumes `Offline` is the `0` index, which may not be the case for helplines using other languages. I don't know if it's better to hard-code the number zero here. Is fine for now tho, just have this in mind in case we receive some feedback :P
- As this is, the reason why "availables" are placed on top is because `Available` activity is position `1`, but consider the case where a helpline has `Ready` as position 5: we want it to be bubbled to the top too, but it will go to the bottom instead. I think the first thing you want to do in this sort function is "if worker is available, move it up", and that can be done with the `worker.isAvailable` boolean (this means that [the worker is ready to accept tasks in TaskRouter](https://www.twilio.com/docs/taskrouter/api/worker#worker-properties)).

GPaoloni (approved):
Looks good to me.
Just a small observation that looks like a potential bug for helplines with "custom available activities".

PR #2263, opened 2024-04-22
Make contact identifier checkbox marked when there're previous contacts

murilovmachado:
## Description
**Previous behavior:**
When the user clicked on the previous contacts banner and navigate to the search form, the checkbox "Only show records for <contact-identifier>" was not marked by default.

**Behaviour after fix:**
Now the checkbox "Only show records for <contact-identifier>" is marked by default on the search form when there are previous contacts.

### Other Related Issues
https://tech-matters.atlassian.net/browse/CHI-2669

### Verification steps
When `enable_client_profiles` is false, and there're previous contacts, the checkbox "Only show records for <contact-identifier>"should be marked by default.

### AFTER YOU MERGE


stephenhand (approved):
LGTM once the tests are passing

PR #2277, opened 2024-04-30
CHI-2726: Support the hasContacts hack for driving the display of the previous …

stephenhand:
…contacts banner

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### Verification steps

Requires a HRM build with https://github.com/techmatters/hrm/pull/632

Shows profile when contacts you have no permission to view are attached:

1. Create a contact request from a new IP or new client profile.
2. Take the contact with a supervisor account and save it.
3. Create a new contact request from the same IP.
4. Take the contact with a supervisor account, you should see the yellow banner appear and saying that there is 1 client and 1 previous contact. Save the contact.
5. Create a new contact request from the same IP.
6. Take the contact with an agent account. No yellow banner appears. There should be a yellow banner saying that there is 1 client.

Does not show profile when no contacts at all are attached:

1. Log in with a user with 'view profile' permissions
2. Start creating a new chat contact from a 'clean' IP with no existing contacts
3. Verify that the banner is NOT present

Also regression test other banner functionality like when viewable contacts and cases are present

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/profile/IdentifierBanner/ProfileIdentifierBanner.tsx
+++ b/plugin-hrm-form/src/components/profile/IdentifierBanner/ProfileIdentifierBanner.tsx
@@ -67,7 +67,8 @@ const ProfileIdentifierBanner: React.FC<Props> = ({ task, openProfileModal, open
    */
   const profileId = identifier?.profiles?.[0]?.id;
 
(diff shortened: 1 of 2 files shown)


mythilytm (approved):
LGTM

PR #2281, opened 2024-05-01
CHI-2726: Fix issue where identifier reducer was packing another copy of the profile state into its own data property

stephenhand:
## Description

This was causing the profile not to load with a new contact being accepted, because the profile loader was detecting data on the data property, ebven though that data was BS

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Other Related Issues

None

### Verification steps

See ticket

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/states/profile/profiles.ts
+++ b/plugin-hrm-form/src/states/profile/profiles.ts
@@ -161,11 +161,8 @@ const handleLoadIdentifierFulfilledAction = (state: t.ProfilesState, action: any
   let newState = { ...state };
   for (const profile of profiles) {
     const profileUpdate = {
-      data: {
-        ...t.newProfileEntry,
-        ...newState[profile.id]?.data,
-        ...profile,
-      },
+      ...t.newProfileEntry,
+      ...profile,
     };
 
     newState = loadProfileEntryIntoRedux(newState, profile.id, profileUpdate);


plugin-hrm-form/src/states/profile/profiles.ts
```diff
@@ -161,11 +161,8 @@ const handleLoadIdentifierFulfilledAction = (state: t.ProfilesState, action: any
   let newState = { ...state };
   for (const profile of profiles) {
     const profileUpdate = {
-      data: {
-        ...t.newProfileEntry,
-        ...newState[profile.id]?.data,
-        ...profile,
-      },
+      ...t.newProfileEntry,
```
stephenhand, line 164: Copying the existing state causes issues with stale profile data not being refreshed

plugin-hrm-form/src/states/profile/profiles.ts
```diff
@@ -161,11 +161,8 @@ const handleLoadIdentifierFulfilledAction = (state: t.ProfilesState, action: any
   let newState = { ...state };
   for (const profile of profiles) {
     const profileUpdate = {
-      data: {
-        ...t.newProfileEntry,
-        ...newState[profile.id]?.data,
```
stephenhand, line 166: Removed this because copying the existing state causes issues with stale profile data not being refreshed if you take anothe contact for a already profile loaded into state

GPaoloni (approved):
Thanks for the fix, LGTM!

PR #2291, opened 2024-05-08
CHI-2749: Fix request spam

stephenhand:
## Description

Fix a number of small issues that contribute to getting the Flex plugin into a request loop

* Disable case summary and case save buttons when a request is in flight
* Fix the routing redux logic so routes cannot be completely emptied
* Fix logic issue in TaskView.tsx where it always tries to retrieve a contact if there is no route - this makes no sense because retrieving a contact won't create a route

### Checklist
- [ ] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Other Related Issues

None

### Verification steps

1. Log into Aselo
2. Start creating an offline contact
3. Add a new case to the offline contact
4. Open Developer Tools and go to the 'Network' tab
5. Set Network Throttling to 'Slow 3G'
6. Click 'Edit on the Case Summary and make a change to the case summary text
7. Click 'Save' as many times as you can before the button disappears

Expected Behaviour:

The button should disable after the initial click, and only 1 PUT request to ../case/###/overview should be logged on the nework tab
Repeated requests to the GET 'byTaskSid/offline-task-WK-###' and GET cases/### should NOT be being loggged in the network tab

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/case/EditCaseSummary.tsx
+++ b/plugin-hrm-form/src/components/case/EditCaseSummary.tsx
@@ -66,8 +66,9 @@ const mapStateToProps = (state: RootState, { task }: EditCaseSummaryProps) => {
   const connectedCaseState = selectCurrentRouteCaseState(state, task.taskSid);
   const historyDetails = selectCaseHistoryDetails(state, connectedCaseState?.connectedCase);
   const workingCopy = connectedCaseState?.caseWorkingCopy.caseSummary;
+  const isUpdating = (connectedCaseState?.outstandingUpdateCount ?? 0) > 0;
   const definitionVersion = selectDefinitionVersionForCase(state, connectedCaseState?.connectedCase);
-  return { connectedCaseState, workingCopy, definitionVersion, historyDetails };
+  return { connectedCaseState, workingCopy, definitionVersion, historyDetails, isUpdating };
 };
 
 const mapDispatchToProps = (dispatch, { task }: EditCaseSummaryProps) => {
@@ -103,6 +104,7 @@ const EditCaseSummary: React.FC<Props> = ({
   closeActions,
   can,
   updateCaseAsyncAction,
+  isUpdating,
 }) => {
   const { connectedCase, availableStatusTransitions } = connectedCaseState ?? {};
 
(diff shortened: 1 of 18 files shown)


plugin-hrm-form/src/components/case/EditCaseSummary.tsx
```diff
@@ -221,6 +223,7 @@ const EditCaseSummary: React.FC<Props> = ({
         <div style={{ width: '100%', height: 5, backgroundColor: '#ffffff' }} />
         <BottomButtonBar>
           <StyledNextStepButton
+            disabled={isUpdating}
```
stephenhand, line 226: @mmythily @murilovmachado - is this the correct approach to momentarily disable a button when a request is in flight, now we have the buttons mapped to a design system?

plugin-hrm-form/src/states/case/saveCase.ts
```diff
@@ -83,6 +84,7 @@ const updateConnectedCase = (state: HrmState, connectedCase: Case): HrmState =>
   const caseDefinitionVersion = state.configuration.definitionVersions[connectedCase?.info?.definitionVersion];
   const stateCase = state.connectedCase.cases[connectedCase?.id]?.connectedCase;
   const stateInfo = stateCase?.info;
+  const outstandingUpdateCount = (state.connectedCase.cases[connectedCase.id]?.outstandingUpdateCount ?? 1) - 1;
```
murilovmachado, line 87: Can you comment on why we decrease 1 here?

I know we do this when we finish updating a case section. So I'm supposing we're doing the same when connecting a contact, but I didn't really see where it was increased before in this scenario. I might have missed it.

murilovmachado (approved):
Looks good to me. 
I just didn't fully understand on change related to connecting a contact.

plugin-hrm-form/src/states/case/saveCase.ts
murilovmachado: Oh, it's related to **connectedCase**, actually (not **connectedContact**). Okay, I think I understand it better now. Not 100% clear, though.

stephenhand: Yes so this function updates a connected case, rather than connecting a case. The 'outstandingUpdateCount' semaphore is only used when changing case contents, either sections or the overview. It is not currently set for connecting / disconnecting contacts. It could be in the future but it wasn't needed for these changes

PR #2320, opened 2024-05-28
CHI-2685: Use customChannelType in preference to channelType if it is set on the task attributes

stephenhand:
## Description
Required for custom channels on conversations, complements https://github.com/techmatters/serverless/pull/618

Updates Flex to check for `customChannelType` and use it in preference to `channelType` in task attributes and use it in preference where available. This is to work around an issue we are seeing where our attempt to set the channelType in our studio flow is being ignored for conversations.

### Checklist
- N/A New tests added
- N/A Feature flags added
- N/A Strings are localized

### Other Related Issues

None

### Verification steps

Ensure Line contacts are correctly identified as Line Contacts everywhere in the UI, HRM and Insights when using 'Line on Conversations'

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/services/ContactService.ts
+++ b/plugin-hrm-form/src/services/ContactService.ts
@@ -173,7 +173,7 @@ export const createContact = async (
   const { definitionVersion } = getHrmConfig();
   const contactForApi: Contact = {
     ...contact,
-    channel: task.channelType as Contact['channel'],
+    channel: (task.attributes?.channelType || task.channelType) as Contact['channel'],
     rawJson: {
       definitionVersion,
       ...contact.rawJson,
--- a/plugin-hrm-form/src/services/ContactService.ts
+++ b/plugin-hrm-form/src/services/ContactService.ts
@@ -173,7 +173,7 @@ export const createContact = async (
   const { definitionVersion } = getHrmConfig();
   const contactForApi: Contact = {
     ...contact,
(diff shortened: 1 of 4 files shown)


plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -173,7 +173,7 @@ export const createContact = async (
   const { definitionVersion } = getHrmConfig();
   const contactForApi: Contact = {
     ...contact,
-    channel: task.channelType as Contact['channel'],
+    channel: (task.attributes?.channelType || task.channelType) as Contact['channel'],
```
GPaoloni, line 176: Is `task.attributes?.channelType` what we want here? Aren't we looking for `customChannelType` just in case? :thinking:

GPaoloni (approved):
Looks good to me!
As we chatted, is sad that we need this workaround but it makes total sense :P

plugin-hrm-form/src/services/ContactService.ts
murilovmachado: I was confused here, also

murilovmachado (approved):
Looks good to me!
There's just one question that's Gian's already raised.

plugin-hrm-form/src/services/ContactService.ts
stephenhand: Yes you are correct, good catch thx

PR #2330, opened 2024-06-03
L3: conversation fails to load

acedeywin:
## Description

- Bug fix
The final resolution is to display the error message `"Error loading messages. Refresh your browser"` since it’s a poor internet connection problem. The conversation will not load properly if the user’s internet connection is poor. The solution of adding a `Refresh` button will not work because the conversation props don’t have the right data to reload the `MessageList` component so the user will have to refresh the browser to refresh the `MessageList`.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


plugin-hrm-form/src/translations/en-CA/flexUI.json
```diff
@@ -154,7 +154,7 @@
   "SharedStateSaveContactError": "Could not save pending contact at the Shared State.",
   "Transfer-TransferButton": "Transfer",
   "Transfer-RefreshButton": "Refresh",
-  "Transfer-ErrorLoadingMessages": "Error loading messages.",
+  "Transfer-ErrorLoadingMessages": "Error loading messages. Refresh your browser",
```
stephenhand, line 157: I would make the message a bit more informative like - 'Error loading messages. To retry loading them, refresh your browser'

plugin-hrm-form/src/components/transfer/RefreshButton.tsx
```diff
@@ -16,9 +16,9 @@
 
 /* eslint-disable react/prop-types */
 import React from 'react';
-import { withTaskContext, Manager, ConversationState, Actions, Template } from '@twilio/flex-ui';
+import { withTaskContext, Template } from '@twilio/flex-ui';
```
stephenhand, line 19: I think this component needs a more appropriate name than 'RefreshButton' since the button was removed

plugin-hrm-form/src/components/transfer/RefreshButton.tsx
```diff
@@ -16,9 +16,9 @@
...
 
-import { RefreshStyledSpan, RefreshStyledButton } from '../../styles/buttons';
+import { RefreshStyledSpan } from '../../styles/buttons';
 
 // Checks if there was an error while loading conversation
 export const isConversationNotLoaded = conversationState => {
```
stephenhand, line 24: If this function is used elsewhere we should move it to a separate file. Is isn't good practice to have anything other than the component itself exported from the component files, all shared functions should live in .ts files rather than .tsx files

Also the name of the function is a little confusing. Is it intended to specifically detect when there was an error loading the conversation? In which case it should be called something like `isErrorLoadingConversation`. Or is it simply intended to check if the conversation is not loaded for any reason - in which case the logic should be flipped so it can be `isConversationLoaded`

stephenhand: Also on line 26 there is no defensive `?.` check when referring to a property on `conversationState`, where it is present on all the other property references, is this function supposed to be able to handle an undefined `conversationState` or not? If it is we need to look at the logic flow because simply adding a .? to the above reference will make this function return `false` - is that the intention?

acedeywin: @stephenhand - This has been updated. I added a check to see if `conversationState` is undefined, which will return true if it does.

I also created an `isErrorLoadingConversation.ts` file which now houses the `isErrorLoadingConversation` function.

plugin-hrm-form/src/components/transfer/RefreshButton.tsx
```diff
@@ -16,46 +16,22 @@
...
-  const messageError = conversationState?.messages.some(item => item.error);
-  const errorLoadingConversations = conversationState.errorWhileLoadingConversation || messageError;
-  const messagesAndParticapantStateEmpty =
-    conversationState?.messages.length === 0 && conversationState?.participants.size === 0;
-
-  return Boolean(messagesAndParticapantStateEmpty || errorLoadingConversations);
-};
-
-const RefreshButton: React.FC<TaskContextProps> = ({ task }) => {
+const RefreshMessage: React.FC<TaskContextProps> = ({ task }) => {
```
stephenhand, line 23: The file also needs renaming

stephenhand (approved):
1 minor comment, approving assuming this will be done

plugin-hrm-form/src/components/transfer/RefreshButton.tsx
acedeywin: @stephenhand - It's done.

PR #2361, opened 2024-06-20
NZ: Alejandro/nz/direct transfers fix

janorivera:
## Description

Direct transfer fix via Task Router filter
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


stephenhand (approved):
So I understand - this approach routes ALL transfers to the Youthline Helpline queue.

I initially tried something where the target only matched emergency chats and used the Priority Youthline Queue and left the other chats using the existing behaviour, but that was the one I couldn't get working, so if this behaviour is ok and it works, great

PR #2366, opened 2024-06-25
Pre-emptively check if token is valid before using it to make API calls

stephenhand:
## Description

A minor update to pre-emptively check if a token is still valid before calling an API with it. Prevents the back end being spammed from expired flex sessions

### Verification steps

Monitor the '4xx from ELB target' alarm to see if it stops spiking

### AFTER YOU MERGE


murilovmachado (approved):
🚀

PR #2371, opened 2024-06-27
CHI-2778: Add Telegram Logo to Yellow Banner

acedeywin:
## Description

- Bug fix: Add Telegram Logo to Yellow Banner

- Problem: For some reason, the `channelType` in the task attribute is returned as `web` instead of `telegram`. This might also be the case for other channels not displaying the right logo.

- Solution: We get the correct `channelType` from the `from` object in the task attribute in situations where the task `channelType` is not returning the correct `channelType` value.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


plugin-hrm-form/src/components/profile/IdentifierBanner/selectChannelType.ts
```diff
@@ -0,0 +1,30 @@
...
+import { CustomITask } from '../../../types/types';
+
+/* Get the value of the channel from task attribute and return it when 
+   task channelType is not returning the correct channelType value 
+*/
+const selectChannelType = (task: CustomITask) => {
+  const channelTypeValue = task.attributes.from?.split(':')[0];
+  return channelTypeValue && channelTypeValue !== task.channelType
+    ? iconsFromTask[channelTypeValue]
+    : iconsFromTask[task.channelType];
```
murilovmachado, line 27: I think it's risky to rely on the array split to get the channel type (for many reasons).

I think you can achieve the same with the following code:
```
return task.attributes.customChannelType || task.attributes.channelType;
```

murilovmachado (changes requested):
I think this can be more simple. Take a look at my comment below

plugin-hrm-form/src/components/profile/IdentifierBanner/selectChannelType.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+import { iconsFromTask } from './iconsFromTask';
+import { CustomITask } from '../../../types/types';
+
+/* Get the value of the channel from task attribute and return it when 
+   task channelType is not returning the correct channelType value 
+*/
+const selectChannelType = (task: CustomITask) => {
+  const channelTypeValue = task.attributes.from?.split(':')[0];
```
stephenhand, line 24: You should be using 'customChannelType' attribute rather than trying to extract it from the 'from' attribute that doesn't always follow this format. If you are going to wrap the logic in a function, it should be somewhere like `utils/task.ts` as it in not limited to banners for use cases

plugin-hrm-form/src/components/profile/IdentifierBanner/selectChannelType.ts
acedeywin: @murilovmachado - It's done. I didn't know there was a customChannelType in the attribute.

plugin-hrm-form/src/components/profile/IdentifierBanner/selectChannelType.ts
acedeywin: @stephenhand -  It's done. Also moved selectChannelType file to utils.

plugin-hrm-form/src/utils/selectChannelType.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import { iconsFromTask } from '../components/profile/IdentifierBanner/iconsFromTask';
+import { CustomITask } from '../types/types';
+
+/* Get the value of the channel from task attribute and return it when 
+   task channelType is not returning the correct channelType value 
+*/
+const selectChannelType = (task: CustomITask) => {
+  const channelTypeValue = task.attributes.customChannelType || task.attributes.channelType;
```
stephenhand, line 24:
```suggestion
  const channelTypeValue = task.attributes.customChannelType || task.channelType;
```

channelType is a top level property, not an attribute

plugin-hrm-form/src/utils/selectChannelType.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  return iconsFromTask[channelTypeValue];
```
stephenhand, line 26: This function should just look up the channel type string, the icon should be looked up in the component

acedeywin: @stephenhand - It's done.

stephenhand (approved):

stephenhand:
LGTM - @murilovmachado requested changes so he needs to approve too :-)

murilovmachado (approved):
🚀

PR #2377, opened 2024-07-05
CHI-2811: Support media in transcripts

acedeywin:
## Description

- New feature: This PR is to enable media display as link on Transcript. The Twilio conversations media resource API is called (another [PR in Serverless](https://github.com/techmatters/serverless/pull/637)), and the message object in the custom MessageList gets updated with the necessary data for media.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

[CHI-2811](https://tech-matters.atlassian.net/browse/CHI-2811)

### Verification steps

- Search for a contact with that has uploaded media content (pdf, image files, etc)
- Load the transcript
- Verify that you can see the media file on the transcript
- Click on it to view

### AFTER YOU MERGE

[CHI-2788]: https://tech-matters.atlassian.net/browse/CHI-2788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CHI-2811]: https://tech-matters.atlassian.net/browse/CHI-2811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugin-hrm-form/src/components/contact/MediaSection/TranscriptSection.tsx
+++ b/plugin-hrm-form/src/components/contact/MediaSection/TranscriptSection.tsx
@@ -29,7 +29,6 @@ import { GroupedMessage } from '../../Messaging/MessageItem';
 import { MessageList } from '../../Messaging/MessageList';
 import { ErrorFont, ItalicFont, LoadMediaButton, LoadMediaButtonText } from './styles';
 import { contactFormsBase, namespace } from '../../../states/storeNamespaces';
-import { getMediaUrl } from '../../../services/ServerlessService';
 
 type OwnProps = {
   contactId: string;
--- a/plugin-hrm-form/src/components/contact/MediaSection/TranscriptSection.tsx
+++ b/plugin-hrm-form/src/components/contact/MediaSection/TranscriptSection.tsx
@@ -178,36 +178,26 @@ const TranscriptSection: React.FC<Props> = ({
   }
 
   if (transcript) {
-    const fetchUpdatedGroupMessages = async () => {
-      const groupedMessages = groupMessagesAndAddSenderInfo(transcript);
+    const groupedMessages = groupMessagesAndAddSenderInfo(transcript);
+    const updatedMessages = groupedMessages.map(message => {
+      if (message.media) {
+        // This updates the message object with the media data to be used on transcript
+        const contentType = message.media.content_type?.split('/')[1];
+        message.serviceSid = transcript.serviceSid;
+        message.contentType = contentType;
 
-      const updatedMessages = await Promise.all(
-        groupedMessages.map(async message => {
-          if (message.media) {
-            // This updates the message object with the media data to be used on transcript
-            const mediaUrl = await getMediaUrl(transcript.serviceSid, message.media.sid);
-            const contentType = message.media.content_type?.split('/')[1];
-            message.mediaUrl = mediaUrl.links.content_direct_temporary;
-            message.contentType = contentType;
-
-            if (message.media.filename) {
-              message.body = message.media.filename;
-            } else {
-              message.body = `untitled.${contentType}`;
-            }
-          }
-          return message;
-        }),
-      );
-
-      setUpdatedGroupMessages(updatedMessages);
-    };
-
-    fetchUpdatedGroupMessages();
+        if (message.media.filename) {
+          message.body = message.media.filename;
+        } else {
+          message.body = `untitled.${contentType}`;
+        }
+      }
+      return message;
+    });
 
     return (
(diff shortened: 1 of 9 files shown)


plugin-hrm-form/src/components/contact/MediaSection/TranscriptSection.tsx
```diff
@@ -175,13 +177,37 @@ const TranscriptSection: React.FC<Props> = ({
...
   if (transcript) {
-    const groupedMessages = groupMessagesAndAddSenderInfo(transcript);
+    const fetchUpdatedGroupMessages = async () => {
+      const groupedMessages = groupMessagesAndAddSenderInfo(transcript);
+
+      const updatedMessages = await Promise.all(
+        groupedMessages.map(async message => {
+          if (message.media) {
+            // This updates the message object with the media data to be used on transcript
+            const mediaUrl = await getMediaUrl(transcript.serviceSid, message.media.sid);
```
stephenhand, line 188: I don't think we should be generating temporary URLs for every media message in the transcript every time we open it.

Could we change the flow so that instead of a simple link for the media, we have a button that looks like a link, that calls the serverless service, gets the URL, then opens the media in a single operation, similar to how we do it when opening a case document?

stephenhand: Another approach could be to have the serverless service actually download the content once it has the URL and send it back to the client as the response - then simple links to the serverless service could be used in the transcript. However I think the client side button approach is best, it is consistent with what we do elsewhere and has less scope for having to fiddle around getting HTTP content headers right etc.

acedeywin: @murilovmachado  - I have implemented the feedback from @stephenhand. I went with the first approach, if you would please take a look at it since @stephenhand will be out this week.

plugin-hrm-form/src/utils/selectMediaIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+/* eslint-disable react/no-multi-comp */
+import React from 'react';
+
+import ImageIcon from '../components/common/icons/ImageIcon';
+import PdfIcon from '../components/common/icons/PdfIcon';
+import UnsupportedFileIcon from '../components/common/icons/UnsupportedFileIcon';
+
+const contentType = {
+  pdf: 'pdf',
+  image: 'jpeg' || 'jpg' || 'png',
```
murilovmachado, line 25: Shouldn't it be using the single `|` in here instead of double `||`?

plugin-hrm-form/src/utils/selectMediaIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+} as const;
+
+export type ContentType = typeof contentType[keyof typeof contentType];
+
+export const displayMediaSize = (bytes: number) => {
+  if (bytes === 0) return '0 B';
+  const k = 1000;
+  const sizes = ['B', 'KB', 'MB'];
+  const i = Math.floor(Math.log(bytes) / Math.log(k));
+  return `${parseFloat((bytes / Math.pow(k, i)).toFixed(2))} ${sizes[i]}`;
```
murilovmachado, line 35: I think `k` should be `1024` instead of `1000`.
Can you double-check that?

murilovmachado (changes requested):
It mostly looks good Stephen, but there's one thing that I think should be fixed:
The math to calculate the size of the media, I believe you're using the wrong multiplicator.

plugin-hrm-form/src/utils/selectMediaIcon.tsx
acedeywin: @murilovmachado - Yes, K should be 1024 but I made it 1000 to tidy up the display. So let's say the file is 1018 KB, that's what will be displayed. So I want it to accommodate only 999KB and when it gets to 1000, it's shortened to 1 MB. It may not be accurate but I notice that's how Twilio also handles their file size display.

plugin-hrm-form/src/utils/selectMediaIcon.tsx
acedeywin: @murilovmachado - I have made the necessary changes.

murilovmachado (approved):
Looks good to me, thanks!

PR #2385, opened 2024-07-10
Refactor media support code to accommodate other supported media types

acedeywin:
## Description

- Chore: This is a refactor of the media [support in transcript](https://github.com/techmatters/flex-plugins/pull/2377) task. I discovered that Twilio media also supports audio and video files. So this refactor accommodates the different types of audio, video and image files.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/utils/selectMediaIcon.tsx
+++ b/plugin-hrm-form/src/utils/selectMediaIcon.tsx
@@ -18,16 +18,9 @@ import React from 'react';
 
 import ImageIcon from '../components/common/icons/ImageIcon';
 import PdfIcon from '../components/common/icons/PdfIcon';
+import AudioIcon from '../components/common/icons/AudioIcon';
+import VideoIcon from '../components/common/icons/VideoIcon';
 import UnsupportedFileIcon from '../components/common/icons/UnsupportedFileIcon';
-
-const contentType = {
-  pdf: 'pdf',
-  jpg: 'jpg',
-  png: 'png',
-  jpeg: 'jpeg',
-} as const;
-
-export type ContentType = typeof contentType[keyof typeof contentType];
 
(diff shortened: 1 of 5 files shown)


plugin-hrm-form/src/utils/selectMediaIcon.tsx
```diff
@@ -37,14 +30,25 @@ export const displayMediaSize = (bytes: number) => {
...
+export const abbreviateMediaFilename = (filename: string): string => {
+  const text = filename.split('');
+  if (text.length > 20) {
+    // Create the abbreviated string
+    return `${text.slice(0, 14).join('')}...${text.slice(-6).join('')}`;
+  }
+  return filename;
+};
+
+export const selectMediaIcon = (contentType: string, mediaType: string) => {
```
murilovmachado, line 42: Can't we just use the `mediaType` attribute in here?
I believe we don't need to split it: `const mediaType = message.media.content_type?.split('/')[1];`

murilovmachado (approved):
Looks good, just on comment on the `selectMediaIcon` function. I think we can use only one parameter

plugin-hrm-form/src/utils/selectMediaIcon.tsx
acedeywin: @murilovmachado - Rather than have an object that contains all the various media types in the different media, what we are doing here is to dynamically check for series of media types. E.g, for the audio media, we are checking when it's an audio/mpeg or audio/aac or audio/ogg, etc.

murilovmachado:
@acedeywin LGTM, you can go on and merge this ✅

PR #2396, opened 2024-07-11
Production ready Search UI 

mythilytm:
## Description
- This PR seeks to match the search form styling based on [mocks](https://www.figma.com/design/KfilUxXBZZ9R1fjo72M2hU/Search?node-id=82-2043&t=1NzR1069lz8TZUEz-1)

Screenshot 2024-07-11 at 2 30 57 PM
Screenshot 2024-07-11 at 2 31 03 PM

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2728)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


plugin-hrm-form/src/components/search/index.tsx
```diff
@@ -123,8 +123,8 @@ const Search: React.FC<Props> = ({
   handleNewCreateSearch,
   searchContext,
 }) => {
-  // const enableGeneralizedSearch = getAseloFeatureFlags().enable_generalized_search;
-  const enableGeneralizedSearch = true;
+  const enableGeneralizedSearch = getAseloFeatureFlags().enable_generalized_search;
+  // const enableGeneralizedSearch = true;
```
GPaoloni, line 127: Remove comment?

plugin-hrm-form/src/components/search/index.tsx
```diff
@@ -153,6 +153,7 @@ const Search: React.FC<Props> = ({
     return searchCases(searchContext)({ ...form, ...newSearchParams }, CASES_PER_PAGE, newOffset);
   };
   const setSearchParamsAndHandleSearch = async newSearchParams => {
+    console.log('>>> newSearchParams', newSearchParams);
```
GPaoloni, line 156: Remove this one?

plugin-hrm-form/src/components/common/forms/formGenerators.tsx
```diff
@@ -222,6 +225,51 @@ export const getInputType = (parents: string[], updateCallback: () => void, cust
           }}
         </ConnectForm>
       );
+    case FormInputType.SearchInput:
```
GPaoloni, line 228: I am of the idea that "form generators" should be more abstract than the very specific case scenario of "search input". I would advocate for changing this for an `IconInput`, that accepts one extra parameter on top of the input's ones: `iconName`. Then you can define what are the valid "icon names" - for now it will only be `searchIcon`.

This allows that the next time we need to use an "icon input", we don't have to define the entire thing in here, but instead we just add `newIcon` to the list of valid icons and is ready to go. 

If you can and want to make that change, it will improve the quality of this code and we'll benefit from it in the long run. But please, if you do so, do it in a separate PR as I think this one is good enough for it's purpose :)

GPaoloni (approved):
Code LGTM!

PR #2441, opened 2024-07-23
ZW Case Permissions

sinehome:
Change case permissions to allow other counsellors to add notes to all cases


dee-luo (approved):
LGTM - please make the same changes in the v2.16-rc branch as well!

PR #2479, opened 2024-08-09
CHI-1734: Remove hardcoded locations

stephenhand:
## Description

Removes the lists of Canadian location hardcoded in lists and adds a client to load them from the back end

Currently just loads the full lists from the backend at the start and lets the client filter them like they were alreardy doing. This seems fine for the size of the lists we are loading, but the backend supports filtering so we can do something a bit fancier in the future if need be

### Checklist
- [ ] New tests added
- N/A Feature flags added
- N/A Strings are localized

### Other Related Issues

None

### Verification steps

Ensure HRM v1.18.0-qa.2 or later is deployed
Regression test locations

### AFTER YOU MERGE


mythilytm (approved):
LGTM

PR #2482, opened 2024-08-13
CHI 2889 Implement Status column in Teams View with sorting

mythilytm:
## Description
This PR is for implementing a Status column (move from Agent column) in the Teams View with sorting functionality with `sortStatusColumn`. 
- Note that a useEffect was created to calculate the dynamic display of time (based on interval or 1s or 1 min)
- Implemented truncating and tooltip for activities with long (12+) characters
- For sorting, created `convertDurationToSeconds` to handle special Twilio's display (Convert duration string in the format "59s", "59:59", "23h" or "23h 59min", "59d" or "5d 3h"  to seconds)
- Minor refactors to reuse styled-components
- I moved the teams view work to the `components` directory.

### Checklist
- [x] Corresponding issue has been [opened](https://tech-matters.atlassian.net/browse/CHI-2889)

### Other Related Issues

None

### Verification steps
- Check sorting logic (Sorted available workers first, offline workers last, then by activity index, then within each activity by duration, with longest to shortest)
- Test and check styling details against [mocks](https://www.figma.com/design/fMuQjby2d6FEqGNMW1qXIH/Teams-View?node-id=581-14507&t=FVke1VhOUG6YDRbN-0) , notably truncating and tooltip behavior
Screenshot 2024-08-13 at 2 40 55 PM

### AFTER YOU MERGE


plugin-hrm-form/src/components/teamsView/teamsViewSorting.ts
```diff
@@ -125,6 +131,101 @@ const sortWorkersByActivity = (a: SupervisorWorkerState, b: SupervisorWorkerStat
...
+
+  return aSkills - bSkills;
+};
+
+/**
+ * Converts a duration string in the format "59s", "59:59", "23h" or "23h 59min", "59d" or "5d 3h"  to seconds.
+ */
+const convertDurationToSeconds = (duration: string): number => {
+  // Handle the "59:59" format (minutes and seconds)
+  if (duration.includes(':')) {
```
stephenhand, line 152: I notice we're doing a lot of manual functions around parsing and processing durations. Have you looked at  the support for handling durations that comes with date-fns? It comes with a `Duration` type and a `formatDuration` function. If you haven't already it might be worth looking to see if we can use those instead of rolling our own functions? https://date-fns.org/v3.6.0/docs/formatDuration

If they aren't suitable it might be worth unit testing non trivial logic like this

stephenhand (approved):
Looks good overall, just one question / suggestion - not blocking to change but would like to hear what you think

plugin-hrm-form/src/components/teamsView/teamsViewSorting.ts
mythilytm: Thank you for this. I eventually got rid of this function and used a date-fns helper method to calculate it more accurately from the worker attribute's "dateUpdated" value.

mythilytm: `differenceInSeconds(new Date(), new Date(x.worker.dateUpdated));`

PR #2506, opened 2024-08-29
fix: flush generalized search form values before triggering search

GPaoloni:
## Description
This PR fixes the bug [General Search: redux form not updated](https://tech-matters.atlassian.net/browse/CHI-2924).
The bug is caused because our form generators are designed to reduce the amount of updates it sends to redux, hence only calling the `updateCallback` when the input is "un-focused" (`onBlur` event) (this is not true for all inputs, but it is for free text ones). The way the generalized search form is implemented supports submitting the search when pressing enter. While this is a better UX, it does not triggers the `updateCallback` if the usear does not "un-focus" the input first.

The fix involves two parts
- Make sure that `updateCallback` function is called before calling `onSubmit` handler, so the values are flushed before performing the search and make they way into redux.
- The call to `generalizedSearchCases` is the merge between whatever the redux state is plus the "new values", which are the latest values that the form hold (even if they are not in redux yet). This is because we don't have access to the redux values in the action itself, so we pass them from the component.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2924)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Verification steps
See the Slack thread in the linked Jira ticket and confirm that you can't replicate the bug, or what is the same
- Start development server (`npm run dev`).
- Perform a search (using generalized search) by typing something in the "search input" and immediately pressing enter without changing the input focus.
- Confirm that the values are correctly picked by the search and they show up in the "search terms bar" as in the image.
  image

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/search/GeneralizedSearchForm/index.tsx
+++ b/plugin-hrm-form/src/components/search/GeneralizedSearchForm/index.tsx
@@ -15,7 +15,7 @@
  * along with this program.  If not, see https://www.gnu.org/licenses/.
  */
 
-import React, { useEffect, useMemo, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useForm, FormProvider } from 'react-hook-form';
 import { useDispatch, useSelector } from 'react-redux';
 import type { FormDefinition } from 'hrm-form-definitions';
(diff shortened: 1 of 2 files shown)


mythilytm (approved):
Thank you. LGTM

stephenhand (approved):
LGTM

PR #2553, opened 2024-10-08
CHI-2994: Update workflows to self host plugins

stephenhand:
## Description

- Ensure versions of the plugins are uploaded to `/plugins/branch/<branch-name>` every push
- Ensure versions of the plugins are copied to `/plugins/tag/<tag-name>` from their origin every release
- Update QA tag E2E deploys to use self hosted plugin
- Allow deploys to reconfigure flex to point to a tag or branch plugin in the bucket rather than uploading to Twilio

This means that the new flow for deploying Flex will be as follows

1. Every push uploads a version of the bundle to the assets S3 buckets under `plugins/hrm-form/branch/<branch-name>`
2. Releasing a version will copy the bundle uploaded from `plugins/hrm-form/branch/<branch-name>` to `plugins/hrm-form/tag/<tagged-version>`
3. 'Deploying' a new version to the helpline will simply involve running a cURL command to update the service configuration to change the folder that it looks for the plugin js in. The HRM forms will need to be uploaded to s3

For now, all user initiated deployments can select between deploying to Twilio and deploying to point at our s3 version. The only deploys that always use the new method are the deploys to the E2E environment we do for running tests as part of a QA release and the weekly deploy of AS DEV to master

Implications for this change:

* We are, or at least can be, self hosting plugins, meaning we can go over 10MB, serve compressed bundles and deliver via a CDN
* Some jobs should be faster because they don't need to build flex - QA release tags and deploys
* When we deploy a branch, that accounts code will now get updated when the code on the branch i9s updated, without the need for explicit deploys. This is handy for dev, but means that branch deploys should be used with great caution on staging environments, and if deploying from branches wasn't a terrible idea before, it certainly is now!
* Greater 'fidelity' of builds we deploy. Because we don't rebuild code for prod, we know what we tested is exactly what gets deployed
* We might need to replace some build time env vars with runtime checks?

I was going to add bo as a reviewer but given he has much more important things to be doing I hesitate to even put him as optional on stuff

TODO: For some reason the JS file isn't being compressed, It's weird, the webchat js is compressed, the demo version Bo did is served compressed but the bundles uploaded via these jobs are not served compressed from the same bucket / cloudfront distribution, and I'm struggling to figure out why.

### Other Related Issues

None

### AFTER YOU MERGE


.github/actions/build-and-upload-plugin/action.yml
```diff
@@ -0,0 +1,159 @@
...
+    - name: Set Twilio account SID for AS DEV
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
+        ssm_parameter: "/development/twilio/AS/account_sid"
+        env_variable_name: "TWILIO_ACCOUNT_SID"
+    - name: Set Twilio Auth Token for AS DEV
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
+        ssm_parameter: "/development/twilio/${{env.TWILIO_ACCOUNT_SID}}/auth_token"
+        env_variable_name: "TWILIO_AUTH_TOKEN"
```
GPaoloni, line 53: Is this needed? :thinking:

.github/actions/build-and-upload-plugin/action.yml
```diff
@@ -0,0 +1,159 @@
...
+        touch ./src/private/secret.js
+      working-directory: ./plugin-hrm-form
+      shell: bash
+    - name: Fill secret.js
+      run: |
+        cat <<EOT >> ./src/private/secret.js
+        export const datadogAccessToken = '$DATADOG_ACCESS_TOKEN';
+        export const datadogApplicationID = '$DATADOG_APP_ID';
+        export const fullStoryId = '$FULLSTORY_ID';
+        EOT
```
GPaoloni, line 91: This probably means those "secrets" can be seen by everyone? Maybe we want to move them to service config?

.github/actions/copy-plugin-to-release-location/action.yml
```diff
@@ -0,0 +1,78 @@
...
+
+    - name: Copy plugin to Staging S3 release folder
+      shell: bash
+      run: |
+        aws s3api copy-object --copy-source assets-staging.tl.techmatters.org/plugins/hrm-form/${{ inputs.source-git-reference-type }}/${{ inputs.source-git-reference }}/plugin-hrm-form.js --bucket assets-staging.tl.techmatters.org --key plugins/hrm-form/tag/${{ inputs.target-git-tag }}/plugin-hrm-form.js
+        aws s3api copy-object --copy-source assets-staging.tl.techmatters.org/plugins/hrm-form/${{ inputs.source-git-reference-type }}/${{ inputs.source-git-reference }}/plugin-hrm-form.js.map --bucket assets-staging.tl.techmatters.org --key plugins/hrm-form/tag/${{ inputs.target-git-tag }}/plugin-hrm-form.js.map
+        aws s3api copy-object --copy-source assets-staging.tl.techmatters.org/plugins/hrm-form/${{ inputs.source-git-reference-type }}/${{ inputs.source-git-reference }}/plugin-hrm-form.js.LICENSE.txt --bucket assets-staging.tl.techmatters.org --key plugins/hrm-form/tag/${{ inputs.target-git-tag }}/plugin-hrm-form.js.LICENSE.txt
+
+    - name: Copy plugin to Development S3 release folder
```
GPaoloni, line 61:
```suggestion
    - name: Copy plugin to Production S3 release folder
```

.github/workflows/flex-ci.yml
```diff
@@ -157,4 +157,29 @@ jobs:
...
+        uses: actions/checkout@v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18.x'
+      - name: Install Flex Plugin
+        uses: ./.github/actions/install-flex-plugin
+      - name: Run hrm-form-definitions Tests
+        run: npm run test
+        working-directory: ./hrm-form-definitions
```
GPaoloni, line 176: Maybe this can be factored out into a prior job to avoid installing the same thing 3 times? Not sure if possible tho, and certainly not necessary as part of this PR.

.github/workflows/global-qa-release-tag.yml
```diff
@@ -58,15 +59,13 @@ jobs:
         with:
           node-version: '18.x'
 
-      - name: Install Flex Plugin
-        uses: ./.github/actions/install-flex-plugin
-
```
GPaoloni, line 63: Isn't this a breaking change? Or are we assuming the tag was already tested when the code was uploaded to begin with?

GPaoloni (approved):
Awesome, thank you!! :rocket: :shipit: :rocket: :shipit: :rocket: :shipit: 


>TODO: For some reason the JS file isn't being compressed, It's weird, the webchat js is compressed, the demo version Bo did is served compressed but the bundles uploaded via these jobs are not served compressed from the same bucket / cloudfront distribution, and I'm struggling to figure out why.

Probably not true anymore?

.github/workflows/global-qa-release-tag.yml
stephenhand: We actually run the tests against the version already deployed to the E2E account in the `deploy-flex`, so I'm not entirely sure why we were building it again locally in the first place? Just a C&P fail? Or maybe we needed the source built locally in the past because the test project depended on it? We don't need it any more though

.github/actions/build-and-upload-plugin/action.yml
stephenhand: The bundle uploaded to Twilio is publicly downloadable as well, so these 'secrets' are no less secure on the self hosted plugin than they are hosted by Twilio.

But yea maybe? :-)

.github/actions/build-and-upload-plugin/action.yml
stephenhand: Yes, the Twiliop CLI commands we call later need these environment variables set

.github/workflows/flex-ci.yml
stephenhand: I thought this would be difficult when I looked at it initially, but I suppose it wouldn't be difficult to change it to upload to development and then copy from there to the other 2 buckets, similar to how we do the releases

PR #2563, opened 2024-10-15
CHI-2994: Add deployment matrix update to self hosted deploy

stephenhand:
## Description

Also add steps to automatically disable Twilio hosted plugins for self hosted deploys

Most of the deploy speed benefit has been lost because the flex plugins CLI doesn't seem to be able to disable a plugin without running from the plugins source directory with all the dependencies installed, for some reason - I opened a ticket with Twilio about this because it seems like a but

But we will regain most of that when the support for Twilio hosted plugins is dropped

### Checklist
- N/A New tests added
- [X] Feature flags added
- N/A Strings are localized

### Other Related Issues

None

### Verification steps

Tested workflows

### AFTER YOU MERGE

### DIFF
--- a/.github/workflows/flex-deploy-all-production-helplines.yml
+++ b/.github/workflows/flex-deploy-all-production-helplines.yml
@@ -36,7 +36,7 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
       - id: determine-target-helplines
-        name: Executing main-action
+        name: Determining helplines to deploy to
         uses: ./.github/actions/determine-target-helplines-action
         with:
           environment: production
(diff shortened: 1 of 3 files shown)


mythilytm (approved):
LGTM.
Something I was wondering. Would i make more sense to have this logic in the "main-action" action (which is mostly twilio-hosted logic) or an adjacent action like "self-host action"?

stephenhand:
> LGTM. Something I was wondering. Would i make more sense to have this logic in the "main-action" action (which is mostly twilio-hosted logic) or an adjacent action like "self-host action"?

The action seems redundant at this point, since the only place that calls it is the Flex Deploy workflow, so the logic may as well be inline there. I think that action used to be called directly from several workflows, but now those all call Flex Deploy

PR #2568, opened 2024-10-18
chore: upgrade flex-ui to v2.9.0 [CHI-3003]

GPaoloni:
## Description
This PR:
- Upgrades `@twilio/flex-ui` package to version 2.9.0.
- Fixes broken imports - `flex-ui-core` is not a separate dependency from `flex-ui` in this new version, but instead bundled with it (symlink dependency).
  - Import type from where in the dependency tree.
  - Replace utility function to add hover behavior for it's analogous (hovering diminishes the background opacity to 80%).

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-3003)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Verification steps
Regression test.

### AFTER YOU MERGE


plugin-hrm-form/src/components/AseloMessaging/AseloMessageList.tsx
```diff
@@ -15,7 +15,7 @@
  */
 
 import React from 'react';
-import type { MessageListChildrenProps } from '@twilio/flex-ui-core/src/components/channel/MessageList/MessageList.definitions';
+import type { MessageListChildrenProps } from '@twilio/flex-ui/src/flex-ui-core/src/components/channel/MessageList/MessageList.definitions';
```
stephenhand, line 18: Can we not just remove this and the other Aselo messaging components now instead of maintaining them?

stephenhand (approved):
Apart from fixing the tests LGTM

PR #2569, opened 2024-10-21
Update Support for 'facebook' and 'messenger' as one channel in queue.

mythilytm:
## Description
- 'messenger' labeled channelTypes for channels migrated to Conversations are not being supported within contact saving logic. This PR is a fix for this.
- Twilio now uses the `channelType` of `messenger` rather than `facebook`. This PR reflects this change that is being used in our codebase
- 'facebook' and 'messenger'  are now grouped as one channel > see `utils/groupedChannels.ts`

- This change is reflected in serverless in this PR https://github.com/techmatters/serverless/pull/712
- Once programmable chat is deprecated, we can remove the use of 'facebook' to only support 'messenger' https://tech-matters.slack.com/archives/CV4QM23UJ/p1729793503052129

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/___tests__/components/queuesStatus/QueuesStatus.test.tsx
+++ b/plugin-hrm-form/src/___tests__/components/queuesStatus/QueuesStatus.test.tsx
@@ -47,7 +47,7 @@ function createState(queuesStatusState) {
 const colors = {
   voice: '#000000',
   web: '#000000',
-  messenger: '#000000',
+  facebook: '#000000',
   sms: '#000000',
   whatsapp: '#000000',
   telegram: '#000000',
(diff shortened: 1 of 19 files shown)


stephenhand (commented):
We should probably keep compatibility for the Facebook channel - leave it in the channel types enum and handle it, since that is still used for Programmable Chat, and that isn't fully deprecated yet. We should put them both under FB on the queue status though, since we should never be using both

stephenhand (approved):
LGTM, thanks Mythily

PR #2575, opened 2024-10-24
Revert self host plugin on deploy.

stephenhand:
Just in case we are rolling back from a v2.18+ self host


mythilytm (approved):
I am not sure I understand the situation where we need to roll back from a v2.18+ self host. The workflow step makes sense. 

Looks good

stephenhand:
> I am not sure I understand the situation where we need to roll back from a v2.18+ self host. The workflow step makes sense.
> 
> Looks good

2.17 only supports Twilio hosting, so we have to revert back to Twilio hosting to go back to that version

Rollbacks are stressful enough, the last thing we need is it not working because the plugin configuration was left misaligned

PR #2623, opened 2024-11-28
CHI-3099: Create task for offline contact when saving offline contact from contact details

stephenhand:
## Description

* Reworked 'Save and End' button to reuse the 'submit form' flow used in the tabbed forms. This will mean the following additional work is done: 
  - A job to retrieve the transcript is created. 
  - A task is created and assigned to an offline contact
  - The details are saved to insights (if the task still exists)
* The ContactDetailsHome.tsx component is refactored to use redux hooks
* The Banner component gets data from redux not prop drilling
* Setting the conversation duration is handled up in the redux layer rather than down in the service. This is a more appropriate place for such business logic and it simplifies the service layer call signatures
* Added fallback code that adds a unique placeholder task sid for offline contacts in the event we fail to create a task in Twilio. This means that the contact can be finalised without messing up the 'in progress' offline task

### Checklist
- [X] New tests added
- [ ] Feature flags added
- [N/A] Strings are localized

### Other Related Issues

None

### Verification steps

See ticket
Also, general Save & End regression

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
+++ b/plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
@@ -18,7 +18,7 @@
 import React, { useEffect, useMemo } from 'react';
 import { format } from 'date-fns';
 import { Actions, Icon, Insights, Template } from '@twilio/flex-ui';
-import { connect } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { callTypes, DataCallTypes, isNonSaveable } from 'hrm-form-definitions';
 import { Edit } from '@material-ui/icons';
 import { Grid } from '@material-ui/core';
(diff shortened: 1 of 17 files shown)


plugin-hrm-form/src/services/twilioTaskService.ts
```diff
@@ -69,3 +68,30 @@ export const wrapupConversationTask = async (taskSid: TaskSID) =>
  */
 export const completeConversationTask = async (taskSid: TaskSID) =>
   fetchProtectedApi('/interaction/transitionAgentParticipants', { taskSid, targetStatus: 'closed' });
+
```
mythilytm, line 71: Good to know about this module

plugin-hrm-form/src/components/contact/ContactDetailsHome.tsx
```diff
@@ -72,33 +74,52 @@ type OwnProps = {
...
-  canViewTwilioTranscript,
-  createDraftCsamReport,
   task,
-  showRemovedFromCaseBanner,
-  openModal,
-  createNewCase,
 }) {
-  const version = savedContact?.rawJson.definitionVersion;
-  const definitionVersion = definitionVersions[version];
+  // Redux selectors
```
mythilytm, line 90: Thanks for this big redux refactor

PR #2656, opened 2024-12-17
CHI-3069: Handle contacts being created from task router events

stephenhand:
## Description

* feature flag enable_backend_hrm_contact_creation has been added
* When this is set, the plugin expects a fully prepopulated contact to be created for it and will not try to create one itself, or proepopulate it from the task data
* Original behaviour should be in effect when the flag is off

### Checklist
- [ ] New tests added
- [X] Feature flags added
- N/A Strings are localized

### Other Related Issues

None

### Verification steps

Deploy with https://github.com/techmatters/flex-plugins/pull/2656 and 

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/___tests__/states/contacts/saveContact.test.ts
+++ b/plugin-hrm-form/src/___tests__/states/contacts/saveContact.test.ts
@@ -121,7 +121,11 @@ describe('actions', () => {
       };
       mockGetContactByTaskSid.mockResolvedValue(taskContact);
       const actionPromiseResult = (dispatch(
-        newLoadContactFromHrmForTaskAsyncAction({ taskSid: 'WT-load-me' } as CustomITask, 'WK', 'mock-ref'),
+        newLoadContactFromHrmForTaskAsyncAction(
+          { taskSid: 'WT-load-me', attributes: {} } as CustomITask,
+          'WK',
+          'mock-ref',
+        ),
       ) as unknown) as Promise<void>;
       const pendingState = getState();
       expect(pendingState.contactsBeingCreated.has('WT-load-me')).toBe(true);


plugin-hrm-form/src/components/TaskView.tsx
```diff
@@ -43,37 +47,66 @@ import asyncDispatch from '../states/asyncDispatch';
...
-    updateHelpline,
-    loadContactFromHrmByTaskSid,
-    createContact,
-    isModalOpen,
-    contactIsCreating,
-  } = props;
-
+const TaskView: React.FC<Props> = ({ task }) => {
+  const { enable_backend_hrm_contact_creation: enableBackendHrmContactCreation } = getAseloFeatureFlags();
+  const taskContactId = (task?.attributes as any)?.contactId;
```
GPaoloni, line 60: Is this added to the task attributes upon contact creation? I thought it may be easier to use "get HRM contact by task id", is there any reason why are you preferring the task attributes instead? (This is a question not a change requirement).

plugin-hrm-form/src/states/contacts/saveContact.ts
```diff
@@ -277,8 +277,15 @@ export const loadContactFromHrmByTaskSidAsyncAction = createAsyncAction(
...
   async (contactId: string, reference: string = contactId) => {
+    const contact = await getContactById(contactId);
+
+    let contactCase: Case | undefined;
+    if (contact?.caseId) {
+      contactCase = await getCase(contact.caseId);
+    }
     return {
       contact: await getContactById(contactId),
+      contactCase,
```
GPaoloni, line 288: - Why do we need the case from the contact here?
- Why returning `contact: await getContactById(contactId)`, effectively getting the same contact twice in a single call to this function?

GPaoloni (approved):
This LGTM, just leaving some questions.

plugin-hrm-form/src/states/contacts/saveContact.ts
stephenhand: This is being tidied up in a follow up commit

plugin-hrm-form/src/components/TaskView.tsx
stephenhand: I just want to try to reduce the size of our HRM API overall, do as much as possible via the 'id' endpoint and hopefully remove the /byTaskSid endpoint in future

stephenhand: Also I want to be less reliant on 'twilio stuff' in our core HRM domain business logic, we should be using contact IDs to reference contacts where possible

PR #2673, opened 2024-12-27
chore: define permission_config SSM parameter (configure stage)

GPaoloni:
## Description
This PR
- Creates a new SSM parameter per account, called `/permission_config`.
- Configures `as` (Aselo helplines) to create the correct parameters.

This work is part of https://tech-matters.atlassian.net/browse/CHI-2897, in order to provision the needed parameters for HRM consumption (moving them away from the `.env` file).

TODO: do the same for the rest of the helplines.
@janorivera let me know if this looks correct to you, and let me know if you would be able to help me fill the rest of the helplines (I can do that, I just want to make sure I'm not messing nor missing any of them :p). I would also appreciate your help applying this changes, since they do not look "clean" when I `make apply`.

NOTE:
- Since `as/development` account is not managed by terraform, I've manually created the corresponding parameter.
- For `e2e/development`, trying to apply the changes result in an error. Maybe this account hasn't been applied in long time @janorivera?

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-2897)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### AFTER YOU MERGE


janorivera:
what @stephenhand is suggesting would be cool! that way we'll only need to manually change helplines that for some reason are using a permission configuration different than the helpline short code.

janorivera:
and yes I can help with the applies!

GPaoloni:
> I assume permission_config defaults to the helpline short code somehow? I just can't see how from this PR

@stephenhand There's no code to do that, the `permission_config` parameter is explicit here. I just added for development environments to see if what I was doing was correct. 

@janorivera I can make the changes in the rest of the helplines, but I will ask you to do the applies if you don't mind.

PS: the implicit configuration seems like a bad idea to me. We can do so but let's keep that for a separate if you don't mind,

janorivera:
@GPaoloni no worries, I can do the applies!

janorivera:
@GPaoloni I'm curious, why having a default config is a bad idea? Is it a security concern or something else? It's OK to do it separately btw

PR #2690, opened 2025-01-06
CHI-3053 Add contact id and queue name as a header for tabbed forms during an active call.

mythilytm:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/tabbedForms/TabbedFormsTabs.tsx
+++ b/plugin-hrm-form/src/components/tabbedForms/TabbedFormsTabs.tsx
@@ -38,7 +38,7 @@ import { newCSAMReportActionForContact } from '../../states/csam-report/actions'
 import { CSAMReportType, CSAMReportTypes } from '../../states/csam-report/types';
 import { changeRoute, newOpenModalAction } from '../../states/routing/actions';
 import { AppRoutes, ChangeRouteMode, TabbedFormRoute, TabbedFormSubroutes } from '../../states/routing/types';
-import { Box, Row, StyledTabs } from '../../styles';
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/types/types.ts
```diff
@@ -342,6 +342,7 @@ export const standaloneTaskSid = 'standalone-task-sid';
 
 export type StandaloneITask = {
   taskSid: typeof standaloneTaskSid;
+  queueName?: string;
```
stephenhand, line 345: StandaloneITask doesn't need a queue name property

plugin-hrm-form/src/components/tabbedForms/TabbedFormsTabs.tsx
```diff
@@ -261,6 +261,12 @@ const TabbedFormsTabs: React.FC<Props> = ({
           </Box>
         )}
       </Row>
+      <Row style={{ paddingLeft: '9px', paddingTop: '3px' }}>
+        <FontOpenSans>
+          <Bold>#{contactId}</Bold>
+          {!isOfflineContactTask(task) && <OpaqueText style={{ fontStyle: 'italic' }}> ({task.queueName})</OpaqueText>}
```
stephenhand, line 267: You should eliminate StandaloneITask as a possible type with additional checks here if you want TS to stop complaining that StandaloneITask doesn't have a queueName property

plugin-hrm-form/src/types/types.ts
mythilytm: Sound good. I removed the queueName and added an `isStandaloneTask` helper

PR #2715, opened 2025-01-16
feat: add isPII flag to form definitions [CHI-3107]

GPaoloni:
## Description
This PR
- Adds a new flag to form definitions, `isPII`, which can be used to flag the fields as PII. This flag can be added to
  - Any customizable form input (contact, case, case section, etc).
  - Case overview. For this a new file, `caseForms/CaseOverview.json`, needs to be added.
  - Profile root. For this a new file, `profileForms/ProfileOverview.json`, needs to be added.
  - Profile sections.
- Adds examples to `AS` form definitions.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-3107)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Verification steps
- CI passes.
- `npm ci && npm run dev` to confirm local server compiles and runs correctly.

### AFTER YOU MERGE

### DIFF
--- a/hrm-form-definitions/src/formDefinition/loadDefinition.ts
+++ b/hrm-form-definitions/src/formDefinition/loadDefinition.ts
@@ -20,6 +20,7 @@ import path from 'path';
 import {
   CallTypeButtonsDefinitions,
   CannedResponsesDefinitions,
+  CaseOverviewDefinition,
   CategoriesDefinition,
   DefinitionVersion,
   FormDefinition,
@@ -31,6 +32,7 @@ import {
   isSelectDefinitionWithReferenceOptions,
   LayoutVersion,
   ProfileFlagDurationDefinition,
+  ProfileOverviewDefinition,
   ProfileSectionDefinition,
 } from './types';
 import { OneToManyConfigSpecs, OneToOneConfigSpec } from './insightsConfig';
@@ -146,6 +148,7 @@ export async function loadDefinition(baseUrl: string): Promise<DefinitionVersion
    */
   const [
     layoutVersion,
+    caseOverview,
     householdForm,
     incidentForm,
     noteForm,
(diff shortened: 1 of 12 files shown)


hrm-form-definitions/src/formDefinition/types.ts
```diff
@@ -49,6 +49,10 @@ export enum FormInputType {
   CustomContactComponent = 'custom-contact-component',
 }
 
+type IsPIIFlag = {
```
stephenhand, line 52: I don't think we need to define this in the Flex type, since Flex doesn't care about it. Similarly we don't want to define all the UI related stuff in the data warehouse type. Types should only define what your code cares about, especially compile only type systems that don't even validate at runtime

hrm-form-definitions/src/formDefinition/types.ts
```diff
@@ -307,11 +312,23 @@ export type StatusInfo = {
   transitions: string[]; // possible statuses this one can transition to (further update may be to include who can make such a transition for a more granular control)
 };
 
+export type CaseOverviewDefinition = {
```
stephenhand, line 315: Similarly I don't think we should capture these extra form definitions as types here. It will just cause more confusion sine the 'real' form definitions for these forms are hardcoded in Flex.

We should probably just put a comment explaining why they aren't defined here

hrm-form-definitions/src/formDefinition/loadDefinition.ts
```diff
@@ -166,10 +169,12 @@ export async function loadDefinition(baseUrl: string): Promise<DefinitionVersion
     prepopulateKeys,
     referenceData,
     blockedEmojis,
+    profileOverview,
     profileSections,
     profileFlagDurations,
   ] = await Promise.all([
     fetchDefinition<LayoutVersion>('LayoutDefinitions.json'),
+    fetchDefinition<CaseOverviewDefinition>('caseForms/CaseOverview.json', {}),
```
stephenhand, line 177: We shouldn't be loading definitions into Flex that aren't going to be used

stephenhand (commented):
Just some thoughts about the types

stephenhand (approved):

PR #2752, opened 2025-01-30
chore: add definitionVersion root field in create contact payload [CHI-3109]

GPaoloni:
This PR is related to https://github.com/techmatters/hrm/pull/821

## Description
This PR
- Modifies "create contact" service usage, by including definition version as a top level property of contact.
- Does same as above for "create contact" lambda.

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### AFTER YOU MERGE


stephenhand (approved):
LGTM

Maybe add a few comments to the legacy property on the type and where it gets set to indicate it's deprecated. This will discourage people from referencing it and remind us to remove it in future

stephenhand:
Also, are there more references in the Flex code we should update so we can safely remove the old property? Doesn't look to be a lot in here.

If we move them in this PR we can remove the property after next deploy. It will draw things out longer if we leave it

PR #2753, opened 2025-01-30
CHI-3190 Support transcripts for contacts from (inactive) search contact view

mythilytm:
## Description
- This PR allows for conversation media to be added to the contact for in progress contacts from the search contact view.
-  The root cause of the bug was with having so many types of tasks in the app. In this case, even though we receive a twilio task from twilio api, the helpers like TaskHelper.isChatBasedTask are not working, hence handleTwilioTask was not returning useful data. I implemented a check on channelType as a workaround. Updated logic in handleTwilioTask to better handle different media types. Tasks coming from twilio in getTaskAndReservations are not being supported by 'TaskHelper' method.
- This relies on `getTaskAndReservations` serverless function to get the reservationSid on top of task. The service level call fetch both task object and reservations
- Enhanced UI logic for displaying success banner after contact has finalized

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### Verification steps
1. Ensure[ this branch in serverless is deployed](https://github.com/techmatters/serverless/pull/748)
2. Start a task of any channel and accept the task in flex
3. Go to the contact in the search contact view. Try to Save and End from the contact banner. Ensure that the transcript or recording is added to the contact. The change should be right away
4. Also, try to Save and End from the contact banner for a really only contact with an expired task. This should gracefully finanlize the contact even though there is not media added.

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/contact/ContactInProgressBanners.tsx
+++ b/plugin-hrm-form/src/components/contact/ContactInProgressBanners.tsx
@@ -36,6 +36,7 @@ import { getAseloFeatureFlags } from '../../hrmConfig';
 import { RootState } from '../../states';
 import selectContactStateByContactId from '../../states/contacts/selectContactStateByContactId';
 import { checkTaskAssignment } from '../../services/twilioTaskService';
+import { getContactById } from '../../services/ContactService';
 
 type ContactBannersProps = {
   contactId: string;
@@ -57,6 +58,21 @@ const ContactInProgressBanners: React.FC<ContactBannersProps> = ({ contactId }) 
     workerRoles,
   ]);
 
+  const checkContactFinalization = async (contactId: string, retries = 3, delay = 1000) => {
+    try {
+      const contact = await getContactById(contactId);
+      if (contact?.finalizedAt !== null) {
+        setShowResolvedBanner(true);
+        return;
+      }
+      if (retries > 0) {
+        setTimeout(() => checkContactFinalization(contactId, retries - 1, delay), delay);
+      }
+    } catch (error) {
+      console.error('Failed to check contact finalization status:', error);
+    }
+  };
+
   const updateAndSaveContact = async () => {
     const updatedContact = {
       ...savedContact,
@@ -65,8 +81,12 @@ const ContactInProgressBanners: React.FC<ContactBannersProps> = ({ contactId }) 
         ...savedContact.rawJson,
       },
     };
-    dispatch(newSubmitAndFinalizeContactFromOutsideTaskContextAsyncAction(updatedContact));
-    setShowResolvedBanner(true);
+    try {
+      await dispatch(newSubmitAndFinalizeContactFromOutsideTaskContextAsyncAction(updatedContact)).payload;
+      checkContactFinalization(savedContact.id);
+    } catch (error) {
+      console.error('Failed to save and finalize contact:', error);
+    }
   };
 
   const handleSaveAndEnd = async () => {
--- a/plugin-hrm-form/src/components/contact/ContactInProgressBanners.tsx
+++ b/plugin-hrm-form/src/components/contact/ContactInProgressBanners.tsx
@@ -31,15 +31,11 @@ import {
   SaveAndEndButton,
 } from '../../styles';
 import getCanEditInProgressContact from '../../permissions/canEditInProgressContact';
-import {
-  newSubmitAndFinalizeContactFromOutsideTaskContextAsyncAction,
-  loadContactFromHrmByIdAsyncAction,
-} from '../../states/contacts/saveContact';
+import { newSubmitAndFinalizeContactFromOutsideTaskContextAsyncAction } from '../../states/contacts/saveContact';
 import { getAseloFeatureFlags } from '../../hrmConfig';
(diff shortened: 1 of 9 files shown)


plugin-hrm-form/src/services/fetchProtectedApi.ts
```diff
@@ -53,8 +53,18 @@ const fetchProtectedApi = async (endPoint, body: Record<string, string> = {}) =>
...
+      } = error;
+
+      const message = body?.message || `Error response: ${status} (${statusText})`;
+      const protectedError = new ProtectedApiError(message, { response: error.response, body: error.body }, error);
+      if (status === 404 && options.method === 'GET') {
+        console.warn(`Resource not found at ${endPoint}:`, protectedError);
+        return null;
+      }
+      throw protectedError;
     } else throw error;
```
mythilytm, line 68: These changes are to support better fallback when we receive 404. Currently, this is causing runtime errors, and undesirable UX

plugin-hrm-form/src/services/ContactService.ts
```diff
@@ -222,6 +222,7 @@ export const updateContactInHrm = async (
   const options = {
     method: 'PATCH',
     body: JSON.stringify(body),
+    returnNullFor404: true,
```
stephenhand, line 225: Is this needed here? We should expect a contact we are trying to patch to existing, this isn't an expected case

plugin-hrm-form/src/services/fetchApi.ts
```diff
@@ -39,6 +39,7 @@ export const generateUrl = (baseUrl: URL, endpointPath: string): URL => {
 
 export type FetchOptions = RequestInit & {
   returnNullFor404?: boolean;
+  returnUndefinedFor404?: boolean;
```
stephenhand, line 42: Why is a null not sufficient?

plugin-hrm-form/src/components/contact/ContactInProgressBanners.tsx
```diff
@@ -57,6 +58,30 @@ const ContactInProgressBanners: React.FC<ContactBannersProps> = ({ contactId })
     workerRoles,
   ]);
 
+  const checkContactFinalization = async (contactId: string, retries = 3, delay = 1000) => {
```
stephenhand, line 61: Making service calls directly from the component isn't the right approach here. We already update redux with the `finalizedAt` date when the contact is successfully saved, you can just drive this display logic off that. You can use a component state property like `saveAndEndClicked` if you need to in order to prevent the banner appearing at inappropriate times

stephenhand (changes requested):
Looks ok mostly, just the fetchApi property and the banner logic need looking at

plugin-hrm-form/src/components/contact/ContactInProgressBanners.tsx
stephenhand: I didn't mean to just use the action.

You shouldn't have to kick off new requests to the backend at all. The update action should get the updated contact as a response from the backend and update the version of the contact in redux

You should just be able to have a `useEffect` hook that has the contact.finalizedAt date as one of it's checked values

plugin-hrm-form/src/components/contact/ContactInProgressBanners.tsx
```diff
@@ -65,8 +94,12 @@ const ContactInProgressBanners: React.FC<ContactBannersProps> = ({ contactId })
         ...savedContact.rawJson,
       },
     };
-    dispatch(newSubmitAndFinalizeContactFromOutsideTaskContextAsyncAction(updatedContact));
-    setShowResolvedBanner(true);
+    try {
+      await dispatch(newSubmitAndFinalizeContactFromOutsideTaskContextAsyncAction(updatedContact)).payload;
+      checkContactFinalization(savedContact.id);
```
stephenhand, line 99: You can set a 'finalizeRequested' component state variable here, and use that in the `useEffect` hook that determines whether to show the banner. Basically if finalizeRequested and contact.finalizedAt are both set, show the banner

stephenhand (changes requested):

plugin-hrm-form/src/components/contact/ContactInProgressBanners.tsx
```diff
@@ -57,6 +62,21 @@ const ContactInProgressBanners: React.FC<ContactBannersProps> = ({ contactId })
...
+        Notifications.registerNotification({
+          id: 'NoConversationMediaNotification',
+          closeButton: true,
+          content: <Template code="ContactDetails-NoConversationMediaNotification" />,
+          type: NotificationType.warning,
+        });
+        Notifications.showNotification('NoConversationMediaNotification');
+      }
+    }
+  }, [finalizeRequested, savedContact]);
```
stephenhand, line 78: One minor thing here, it might be betterto assign `finalizedAt` to its own variable outside the `useEffect`, then use that as a check rather than `savedContact` - there might be edge cases where it doesn't run the effect as expected if you use the parent contact

stephenhand (approved):
One minor thing but approving assuming it's not a big deal to change

PR #2761, opened 2025-02-05
CHI-3207: Replace section specific permissions with generic ones that cover all section types

stephenhand:
## Description

* Replaced all the add / edit permissions for specific case sections with addCaseSection and editCaseSection
* Refactored ViewCaseItem to use redux hooks rather than mappings

### Checklist
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


GPaoloni (approved):
LGTM!

I assume there's an HRM PR this one depends on (hrm needs to be in place before this is deployed I think)?

stephenhand:
> LGTM!
> 
> I assume there's an HRM PR this one depends on (hrm needs to be in place before this is deployed I think)?

Yes, this version will be dependent on the new HRM version being deployed (but not vice versa) - I think you already approved that PR

PR #2764, opened 2025-02-05
CHI-3174: Custom case section support

stephenhand:
## Description

This PR aims to make it possible to add custom case section types and remove existing ones, rather than always using the hardcoded 6 types we have currently.

- Added a 'CaseSections.json' definition file to define the case section types used by this helpline
- Updated definition loader to support dynamically looking up form definiitons for case sections based on the information in CaseSections.json
- Removed all internal functions & types specific to a hardcoded case section type
- Updated form definitions 
- Update insights to scan all case section types (but still only use the ones specified in the mappings
- Updated translations to have keys in a common format if they are referenced dynamically depending on case section type

### Checklist
- [x] New tests added
- [n/a] Feature flags added (configuration is used to keep things the same as previously)
- [x] Strings are localized

### Other Related Issues

None

### Verification steps

See JIRA ticket, detailed instructions about setting up new case sections are available

### AFTER YOU MERGE


hrm-form-definitions/form-definitions/as/v1/CaseSections.json
```diff
@@ -0,0 +1,30 @@
...
+    "formPath": "caseForms/IncidentForm.json"
+  },
+  "document": {
+    "label": "Document",
+    "formPath": "caseForms/DocumentForm.json"
+  },
+  "chicken": {
+    "label": "Chicken",
+    "formPath": "caseForms/ChickenForm.json"
+  }
```
GPaoloni, line 29: :chicken: :laughing:

hrm-form-definitions/form-definitions/as/v1/LayoutDefinitions.json
```diff
@@ -10,34 +10,67 @@
...
-      "splitFormAt": 7
-    },
-    "perpetrators": {
-      "splitFormAt": 7
-    },
-    "incidents": {
-      "previewFields": ["date", "duration", "location"],
-      "layout": {
-        "date":  {
+    "sectionTypes": {
```
GPaoloni, line 13: I'm wondering, since is getting complex to define the shape of this stuff, if we should dedicate some time to document it in a README close to where this are defined?

hrm-form-definitions/form-definitions/as/v1/caseForms/ChickenForm.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "name": "boc",
+    "label": "Boc",
+    "type": "select",
+    "options": [
+      { "value": "", "label": "" },
+      { "value": "Boc", "label": "Boc" },
+      { "value": "Bwaaaaak", "label": "Bwaaaaak" },
```
GPaoloni, line 9: LOL sorry but this is art, reminded me of https://www.youtube.com/watch?v=7-9-vLufOvg

hrm-form-definitions/src/formDefinition/loadDefinition.ts
```diff
@@ -198,21 +214,27 @@ export async function loadDefinition(baseUrl: string): Promise<DefinitionVersion
...
+    caseSections,
+    referenceData,
+    fetchDefinition,
+  );
   const { helplines } = helplineInformation;
   const defaultHelpline =
     helplineInformation.helplines.find((helpline: HelplineEntry) => helpline.default)?.value ||
     helplines[0].value;
   return {
+    // Legacy hardcoded case forms
```
GPaoloni, line 228: Why do we keep this? Is for compatibility?

hrm-form-definitions/src/formDefinition/types.ts
```diff
@@ -255,6 +246,18 @@ export type CallTypeButtonsEntry = {
...
 
+export type CaseSectionTypeJsonEntry = {
+  label: string;
+  formPath: string;
+};
+
+type CaseSectionTypeEntry = {
+  label: string;
+  form: FormDefinition;
+};
```
GPaoloni, line 257: Aren't this two the same?

plugin-hrm-form/src/components/case/CaseHome.tsx
```diff
@@ -198,79 +205,69 @@ const CaseHome: React.FC<Props> = ({
...
                 />
-              )}
-            />
-          </Box>
-        )}
+              );
+              break;
+            }
+            case 'household':
+            case 'perpetrator': {
```
GPaoloni, line 232: Are we aiming to move this to the generalized component eventually?

GPaoloni (approved):
Haven't executed the code, but changes LGTM! A big yet a fun one :rocket:

hrm-form-definitions/src/formDefinition/types.ts
stephenhand: No, one is the definition as it is in the json file - the second is when it has been 'expanded' with the actual form definition

plugin-hrm-form/src/components/case/CaseHome.tsx
stephenhand: Yes, this will be removed in the PR for CHI-3202

hrm-form-definitions/form-definitions/as/v1/LayoutDefinitions.json
stephenhand: I put detailed instructions in the ticket, but this should probably live somewhere more permanent, just wondering if somewhere more accessible by the product team is better, since those are the main users?

hrm-form-definitions/src/formDefinition/loadDefinition.ts
stephenhand: This can actually go now, it was just to keep stuff building whilst I was halfway through the changes

hrm-form-definitions/form-definitions/as/v1/LayoutDefinitions.json
GPaoloni: `README` is pretty accessible (at least Ale is used to them) and I think it will be kept updated, whereas a box document will fall behind, that's why I think is the better place.. But whatever really

PR #2774, opened 2025-02-10
CHI-3202-customise case section display

stephenhand:
## Description

* Replaces the 3 different list '[row' components we had with a single one
* Added extra configuration properties to control the display of rows in lists
* Added configuration to allow to to specify whether you want a section displayed in it's own list or in the timeline
* Added configuration to allow to to specify whether you want a section printed in it's own tabular block or as a compact list
* Added support in the code for these extra options
* Removed some redundant helper / formatter functions

TODO: Move the components I've touched over from mapped redux to redux hooks like a good boy. Will do in separate PR to keep this one cleaner

One last thing to make sections fully customisable is to be able to specify the 'timeline sorting date' used in queries on the backend. At the moment that's hardcoded and might not be so straightforward to generalise, so might put it off until we know we need it

### Checklist
- [x] New tests added
- [x] Feature flags added
- [x] Strings are localized

### Other Related Issues
CHI-3174

### Verification steps

Detailed instructions will be added to the ticket, similar to CHI-3174

### AFTER YOU MERGE


mythilytm (approved):
I mostly tested the functionality of the case section and they all work as expected (add and edit sections)

PR #2776, opened 2025-02-11
add firstContact when updating case

mythilytm:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/states/case/saveCase.ts
+++ b/plugin-hrm-form/src/states/case/saveCase.ts
@@ -95,6 +95,7 @@ const updateConnectedCase = (state: HrmState, connectedCase: Case): HrmState => 
         [connectedCase.id]: {
           connectedCase: {
             ...connectedCase,
+            firstContact: stateCase?.firstContact,
             ...((stateInfo || connectedCase.info) && {
               info: { ...stateInfo, ...connectedCase.info },
             }),


plugin-hrm-form/src/states/case/saveCase.ts
```diff
@@ -95,6 +95,7 @@ const updateConnectedCase = (state: HrmState, connectedCase: Case): HrmState =>
         [connectedCase.id]: {
           connectedCase: {
             ...connectedCase,
+            firstContact: stateCase?.firstContact,
```
stephenhand, line 98: Have you tested this change in scenarios where you disconnect and connect contacts to a case? I think you might get stale data issues in some parts of the UI?

stephenhand:
We get `firstContact` from the backend when we search to populate some fields on case search results that actually come from contacts. However, once we get the results in the service we put the `firstContact` into the contacts redux store, and we should always be using it from there rather than continuing to use the property value - in fact we should remove it from the case object when we load it into redux

I know there was something stopping me doing this in the original redux refactor, which is why there are still some lingering `firstContact` references in the code. I can't remember what it was or if it's still an issue, so I'd test my suggested fix thoroughly, especially around connecting / disconnecting contacts from cases and the states of the various banners

stephenhand (approved):

PR #2781, opened 2025-02-12
CHI-3221-permission_read_fix

stephenhand:
## Description

Counselor filters were being hidden because their permissions check function had been made async but was still being called synchronously - the resulting unawaited Promise<boolean> rather than a boolean was always evaluating true

The only reason these calls were async at all was because there was an async call to `fetchRules` in the case where the feature flag was off - if it was on it just reads the cached rules via `getRules`.

Since wa already call `fetchRules` on startup to populate the cached version, and `fetchRules` looks in the appropriate place depending on the flag, it seems like the cached version can be looked up regardless of whether the flag is set or not?

So I've removed the flag check form `checkRule` and now it always uses the cached version and seems to work

### Checklist
- [n/a] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### Verification steps

See ticket

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/permissions/search-permissions.ts
+++ b/plugin-hrm-form/src/permissions/search-permissions.ts
@@ -19,7 +19,7 @@ import sortBy from 'lodash/sortBy';
 
 import { getRules } from '.';
 import { fetchRules } from './fetchRules';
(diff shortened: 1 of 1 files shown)


mythilytm (approved):
Looks good. Thank you

PR #2786, opened 2025-02-14
CHI-3184-correct_llm_button

stephenhand:
## Description

- Adds some extra logic to the new FormTextArea component to allow it to properly update when it's value changes in redux independently of user input
- Add 'additionalActions' to textarea form items. This allows you to add custom components that get displayed to the right of the label
- Add a GenerateSummaryButton custom form control, that can be pointed at any contact form item and will request a summary based on the current transcript
- Reconfigure AS forms to include a GenerateSummaryButton on their callSummary textarea
- Update redux action to have a configurable target property, append the summary to the existing value, and mark the field as LLM supported

### Checklist
- [ ] New tests added
- [x] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


plugin-hrm-form/src/states/contacts/llmAssistant.ts
```diff
@@ -17,55 +17,108 @@
...
+} as const;
+
+type Role = typeof roles[keyof typeof roles];
+
+const inferRole = (author: string): Role => {
+  if (author.toLowerCase() === 'bot') {
+    return roles.Chatbot;
+  }
+  // super dirty check for role - ok until we know we need this long term?
+  return author.includes('_40') ? roles.Counselor : roles.Caller;
```
mythilytm, line 40: I did not know that you can infer a counselor with '_40'. Interesting. 
I am wondering if this will be still work if there is a transfer.

mythilytm (approved):
LGTM

plugin-hrm-form/src/states/contacts/llmAssistant.ts
stephenhand: See the comment, it's a quick and dirty check that will work for the trial, it's not something we can rely on for production

PR #2803, opened 2025-02-18
CHI-3150 helpline based translations refactor

mythilytm:
## Description

This PR refactors the translation handling to enable a new structure that supports **base language, locale overrides, and helpline overrides**.
- `enable_hierarchical_translations` feature flag is introduced to support legacy translations until the the migrations of the strings happen.
- Note that this refactor also supports messages. It is now possible to remove duplication of [translations and messages in serverless](https://github.com/techmatters/serverless/tree/master/assets/translations) 
- This work is preliminary and does not include any UI interactivity to change languages.

### Checklist
- [x] New tests added
- [x] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### Verification steps
1. Hardcode enable_translations_v2 to true
2. Check that translations are working as expected

### AFTER YOU MERGE

3. Comment on the ticket with the release tag version AND any additional instructions required to configure an environment to test the changes.
4. Only then move the ticket into the QA column in JIRA


plugin-hrm-form/src/translations/helplineOverrides/as.json
```diff
@@ -0,0 +1,12 @@
+{
```
stephenhand, line 1: I think helpline specific overrides should be part of their form definitions, lets try to keep all helpline specific customisations together as far as possible

stephenhand: Also, I think it might be best for helpline specific overrides to have a file per language to be consistent with the base translations. So the form definitions would have a `substitutions` directory, with `en.json`, and then in KHP's case they would have an `fr.json` etc.

plugin-hrm-form/src/hrmConfig.ts
```diff
@@ -66,7 +66,8 @@ const readConfig = () => {
 
   const chatServiceSid = manager.serviceConfiguration.chat_service_instance_sid;
   const workerSid = manager.workerClient.sid as WorkerSID;
-  const { helpline, counselorLanguage, full_name: counselorName, roles } = manager.workerClient.attributes as any;
+  const workerClientAttributes = manager.workerClient.attributes as any;
```
stephenhand, line 69: I'm not sure this helps readability any

plugin-hrm-form/src/translations/translations-README.md
```diff
@@ -0,0 +1,24 @@
+| HL   | helplineLanguage | translation use case        |
```
stephenhand, line 1: Maybe this should explain how the system works rather than listing every helpline? I think this will quickly go out of date

plugin-hrm-form/src/translations/index.ts
```diff
@@ -0,0 +1,219 @@
...
+  'es-CO': esCOMessages,
+  'es-ES': esESMessages,
+  'hu-HU': huHUMessages,
+  'pt-BR': ptBRMessages,
+  'th-TH': thTHMessages,
+};
+
+// TODO(mythily): potentially move Zambia messages to flex from serverless https://github.com/techmatters/serverless/pull/51/files
+/* eslint-disable import/no-unused-modules */
+export const loadV2Translations = (language: string): Record<string, string> => {
```
stephenhand, line 102: Lets remove all references to 'v2', nobody will know what that means in a few months, use names that describe what things are doing instead

plugin-hrm-form/src/translations/baseLanguage/es.json
```diff
@@ -0,0 +1,23 @@
+{
```
stephenhand, line 1: Shouldn't these base languages have full sets of translations? I know some like French aren't fully populated yet but Spanish should be?

plugin-hrm-form/src/translations/index.ts
```diff
@@ -0,0 +1,219 @@
...
+  const [baseLanguage] = language.split('-');
+
+  try {
+    const baseTranslation = require(`./baseLanguage/${baseLanguage}.json`);
```
stephenhand, line 106: I don't think the base language and the locale overrides need to be in separate folders, if they were all in the same folder, the files would sort nicely so that the locale overrides appeared alongside the base set

plugin-hrm-form/src/translations/index.ts
```diff
@@ -0,0 +1,219 @@
...
+    const localeOverrides = require(`./localeOverrides/${language}.json`);
```
stephenhand, line 107: We should be able to handle not having a translation file for a given locale

plugin-hrm-form/src/translations/index.ts
```diff
@@ -0,0 +1,219 @@
...
+    const { helplineCode } = getHrmConfig();
+    const helplineModule = require(`./helplineOverrides/${helplineCode}.json`);
+    const helplineOverrides = helplineModule.translations[baseLanguage];
```
stephenhand, line 110: We should also be able to handle not having a helpline overrides file for a helpline

plugin-hrm-form/src/___tests__/translations/index.test.ts
```diff
@@ -127,6 +140,12 @@ describe('Test initLocalization', () => {
     afterNewStrings.mockClear();
   });
 
+  test('Default language with translations v2 enabled', () => {
```
stephenhand, line 143: We should test this more comprehensively. We should test edge cases like 
- base file present, locale file present, helpline specific missing,  
- base file present, locale file missing, helpline specific present,  
- base file present, locale file missing, helpline specific missing

We should also validate by looking up specific translations and comparing them to what is expected. At the moment your test code looks an awful lot like your implementation code, which you should try to avoid. It means you risk replicating the same mistake in your test as you have in your code

stephenhand (commented):
A few comments but they are related to testing, naming, organisation etc. The core logic looks solid

plugin-hrm-form/src/translations/translations-README.md
mythilytm: https://github.com/techmatters/flex-plugins/blob/164bf1c9892afad5db4c2af588689f6248c8c547/plugin-hrm-form/src/translations/README.md

plugin-hrm-form/src/translations/README.md
```diff
@@ -0,0 +1,28 @@
+# Translations / Internationalization Setup
```
stephenhand, line 1: 🙏

plugin-hrm-form/src/translations/index.ts
```diff
@@ -0,0 +1,237 @@
...
+      const localeOverrides = require(`./locales/${language}.json`);
+      translations = { ...translations, ...localeOverrides };
+    } catch (error) {
+      console.error(`Locale-specific translations not found for ${language}`);
+    }
+  }
+
+  const { helplineCode } = getHrmConfig();
+  try {
+    const helplineTranslations = require(`../../../hrm-form-definitions/form-definitions/${helplineCode}/v1/translations/Substitutions.json`);
```
stephenhand, line 122: This isn't going to work when we deploy flex because the form definitions are no longer deployed as part of the bundle, they are loaded separately from a different s3 location. You need to load the translations along with the rest of the form definitions and load them from redux here

plugin-hrm-form/src/translations/index.ts
```diff
@@ -0,0 +1,237 @@
...
+  } catch (error) {
+    console.error(`Base language file not found for ${baseLanguage}`);
+  }
+
+  if (language !== baseLanguage) {
+    try {
+      const localeOverrides = require(`./locales/${language}.json`);
+      translations = { ...translations, ...localeOverrides };
+    } catch (error) {
+      console.error(`Locale-specific translations not found for ${language}`);
```
stephenhand, line 116: I don't think missing locales should be treated as an error, we should only have to specify a locale overrides file if there are locale overrides we need to apply - we should be able to omit the file without having errors otherwise

plugin-hrm-form/src/translations/index.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  const { helplineCode } = getHrmConfig();
+  try {
+    const helplineTranslations = require(`../../../hrm-form-definitions/form-definitions/${helplineCode}/v1/translations/Substitutions.json`);
+
+    if (helplineTranslations && helplineTranslations[baseLanguage]) {
+      translations = { ...translations, ...helplineTranslations[baseLanguage] };
+    }
+  } catch (error) {
+    console.warn(`Helpline translations not found in hrm-form-definitions for helpline: ${helplineCode}`);
```
stephenhand, line 128: Again, not a warning, not helpline translations should be considered normal, so this could be a debug or an info at most - but once you are loading them from the definitions it shouldn't ne3ed to be a try / catch anyway

stephenhand (changes requested):
There's one comment about the way that the translations are loaded that is critical - otherwise it's just minor stuff

Whilst they are needed for compatibility now, we should get rid of locale files that don't have any translations and empty helpline substitutions files when we can

plugin-hrm-form/src/translations/index.ts
stephenhand: Also, would prefer the folder we store these files under to be something like `customText` rather than `translations`

plugin-hrm-form/src/translations/index.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  try {
+    const baseTranslations = require(`./locales/${baseLanguage}.json`);
+    translations = { ...baseTranslations };
+  } catch (error) {
+    console.error(`Base language file not found for ${baseLanguage}`);
+  }
+
+  if (language !== baseLanguage) {
+    const localeOverrides = require(`./locales/${language}.json`);
```
stephenhand, line 111: Do we need another try ... catch in case this file is absent?

stephenhand (approved):
One minor q, approving though

Does this include the changes to move the majority of the translations up to the base languages or is that still TODO?

plugin-hrm-form/src/translations/index.ts
mythilytm: >Does this include the changes to move the majority of the translations up to the base languages or is that still TODO?

I addressed this in this PR

PR #2805, opened 2025-02-19
CHI-3170: Upgraded to node 20 where building flex, 22 elsewhere

stephenhand:
## Description

Upgrades the following to node 20:
- .nvmrc
- workflows that build flex

Upgrades the following to node 22:
- Twilio Lambda Docker Container
- Any node workflows that don't build flex


mythilytm (approved):
Looks like even though node 20 is used widely as in 20.18, in some github actions and docker, you have v22. I wonder if this might cause any incompatibility issues down the line.
otherwise LGTM

PR #2831, opened 2025-02-27
Invalidate Cloudfront caches in separate step after full webchat deploy

stephenhand:
## Description

Currently 2 or 3 webchat deployments usually 'fail' when doing a full staging or prod deploy. This is because we do a cloudfront cache invalidation for each helpline, to ensure users get the new version next time they load the client

A separate invalidation per helpline trips an AWS quota limit so this starts failing

To address this, this PR introduces a flag to disable the per helpline invalidations, and a separate step at the end of the all staging / prod deploys to invalidate every helpline in a single invalidation command

We've used the same strategy for Flex forms for a while now ands it fixed the same issue there


mythilytm (approved):
LGTM

PR #2845, opened 2025-03-06
CHI-3251 Case Overview Configurability

mythilytm:
## Description
- This PR introduces the ability to configure fields within the Case Overview.
  - The `cases/[id]/overview` api call has been updated in [this hrm PR](https://github.com/techmatters/hrm/pull/845) to saves dynamic case overview properties in the info object
  - Set common and optional type definitions for case overview fields, specifically to include the required fields (status, createdAt, updatedAt, summary) for backward compatibility. Added definitions to match old hardcoded fields in all helplines. Added tests to ensure use and validation of the CaseOverview definition
  - Added tests for updating case overviews and handling optional fields.

### Checklist
- [x] Corresponding issue has been [opened](https://tech-matters.atlassian.net/browse/CHI-3251)
- [x] New tests added
- [n/a] Feature flags added
- [x] Strings are localized

### Other Related Issues

None

### Verification steps
- 

### AFTER YOU MERGE


hrm-form-definitions/form-definitions/uscr/v1/caseForms/CaseOverview.json
```diff
@@ -0,0 +1,35 @@
...
+    "name": "updatedAt",
+    "label": "Updated Date",
+    "type": "date-input"
+  },
+  {
+    "name": "followUpDate",
+    "label": "Follow Up Date",
+    "type": "date-input"
+  },
+  {
```
stephenhand, line 27: Would this not need the customisations in AS_DEV too?

hrm-form-definitions/src/__tests__/specification/validateDefinitions.test.ts
```diff
@@ -175,6 +175,37 @@ describe('Validate form definitions', () => {
           });
         }
       });
+
```
stephenhand, line 178: 🚀

plugin-hrm-form/src/states/case/caseWorkingCopy.ts
```diff
@@ -219,20 +219,30 @@ const initialiseCaseSummaryWorkingCopyReducer = (
 ): CaseState => {
   const caseState = state.cases[action.caseId];
   if (!caseState) return state;
-  const { childIsAtRisk, summary, followUpDate } = caseState.connectedCase.info;
+
+  const caseInfo = caseState.connectedCase.info || {};
+
+  const caseSummary = {
+    status: caseState.connectedCase.status ?? action.defaults.status ?? 'open',
+    summary: caseInfo.summary ?? action.defaults.summary ?? '',
```
stephenhand, line 227: Do summary, childIsAtRisk and followUpDate need to be treated specially here? can they just be used in the loop below?

plugin-hrm-form/src/states/case/types.ts
```diff
@@ -105,6 +105,7 @@ export type CaseSummaryWorkingCopy = {
   followUpDate: string;
```
stephenhand, line 105: I'm not sure why the other properties that are just part of the case info need to be called out specially, rather than just being part of the generic map? Helplines might well want to get rid of these in future and I don't see why we need to prevent that?

plugin-hrm-form/src/components/case/caseOverview/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+          {caseStatusField && (
+            <CaseOverviewItem
+              labelId="CaseDetailsStatusLabel"
+              templateCode="Case-CaseDetailsStatusLabel"
+              inputId="Details_CaseStatus"
+              value={statusLabel}
+            />
+          )}
+          {childIsAtRiskField && (
+            <CaseOverviewItem
```
stephenhand, line 140: Is this treated as a special field for layout reasons? Or does it need to be treated differently for logic reasons too?

stephenhand (commented):
My comments are really around whether we need to keep treating the fields we've previously stored in the info blob as special cases in the code, or whether we just treat them as 'additional fields' - Apart from layout, I can't see why we would need to apply logic to these fields that we wouldn't to other fields that might be present in the definition

plugin-hrm-form/src/components/case/caseOverview/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const formattedFollowUpDate = parseISO(followUpDate).toLocaleDateString();
+
+  const renderDateValue = (fieldName: string) => {
+    switch (fieldName) {
+      case CASE_OVERVIEW_FIELDS.CREATED_AT:
+        return formattedCreatedAt;
+      case CASE_OVERVIEW_FIELDS.UPDATED_AT:
+        return formattedUpdatedAt;
+      case CASE_OVERVIEW_FIELDS.FOLLOW_UP_DATE:
+        return formattedFollowUpDate === 'Invalid Date' ? '—' : formattedFollowUpDate;
```
stephenhand, line 77: I would argue that we need to start putting follow up date in the form as a standard additional field from now on, but that's a discussion we need to have with Jana & Annalise

plugin-hrm-form/src/components/case/caseOverview/index.tsx
mythilytm: Changed this to use checkbox fields

stephenhand (approved):
Looks good, thanks for the changes

plugin-hrm-form/src/components/case/caseOverview/EditCaseOverview.tsx
```diff
@@ -105,11 +104,32 @@ const EditCaseOverview: React.FC<Props> = ({
...
   isUpdating,
+  definitionVersion,
 }) => {
   const { connectedCase, availableStatusTransitions } = connectedCaseState ?? {};
 
+  const caseOverviewFields = definitionVersion?.caseOverview;
+
   const formDefinition: FormDefinition = useMemo(() => {
     try {
+      if (caseOverviewFields && Array.isArray(caseOverviewFields)) {
```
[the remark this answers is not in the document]
mythilytm: Sounds good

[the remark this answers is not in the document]
mythilytm: thanks

PR #2853, opened 2025-03-10
CHI-3089 Ensure categories are part of case state

mythilytm:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/states/case/saveCase.ts
+++ b/plugin-hrm-form/src/states/case/saveCase.ts
@@ -92,13 +92,15 @@ const updateConnectedCase = (state: HrmState, connectedCase: Case): HrmState => 
       cases: {
         ...state.connectedCase.cases,
         [connectedCase.id]: {
+          ...stateCase,
           connectedCase: {
+            ...stateCase?.connectedCase,
+            ...connectedCase,
             categories: stateCase?.connectedCase?.categories ?? connectedCase.categories,
             info: {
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/states/case/saveCase.ts
```diff
@@ -92,13 +92,15 @@ const updateConnectedCase = (state: HrmState, connectedCase: Case): HrmState =>
       cases: {
         ...state.connectedCase.cases,
         [connectedCase.id]: {
+          ...stateCase,
           connectedCase: {
+            ...stateCase?.connectedCase,
+            ...connectedCase,
             categories: stateCase?.connectedCase?.categories ?? connectedCase.categories,
```
stephenhand, line 99: Do we need to swap this condition around then? Otherwise the categories will never be updated if there are pre-existing ones set?

stephenhand (commented):
'Nice to have' = wait until next deploy, patches onto RC should be for urgent changes only

Just one question around the state update logic

stephenhand:
I changed the base to master

plugin-hrm-form/src/states/case/saveCase.ts
mythilytm: Currently, changes to categories are being recorded accurately

PR #2881, opened 2025-03-21
Update CaseSections.json

dee-luo:
Add response teams case section in timeline


hrm-form-definitions/form-definitions/uscr/v1/CaseSections.json
```diff
@@ -7,6 +7,10 @@
     "label": "Incident Report",
     "formPath": "caseForms/IncidentReportForm.json"
   },
+  "assignedResponders": {
```
stephenhand, line 10: Should be singular case

stephenhand, line 10:
```suggestion
  "assignedResponder": {
```
We use singular case for section types as a convention

stephenhand (approved):
LGTM!

PR #2889, opened 2025-03-26
Chi 3227 case filters registered filters

mythilytm:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/caseList/filters/FilterComponentsRegistry.ts
+++ b/plugin-hrm-form/src/components/caseList/filters/FilterComponentsRegistry.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright (C) 2021-2023 Technology Matters
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+
 import React from 'react';
 
 import MultiSelectFilter, { Item } from './MultiSelectFilter';
--- a/plugin-hrm-form/src/components/caseList/filters/FilterComponentsRegistry.ts
+++ b/plugin-hrm-form/src/components/caseList/filters/FilterComponentsRegistry.ts
@@ -1,0 +1,169 @@
+import React from 'react';
+
+import MultiSelectFilter, { Item } from './MultiSelectFilter';
+import CategoriesFilter, { Category } from './CategoriesFilter';
+import DateRangeFilter from './DateRangeFilter';
+import { DateFilterValue } from '../../../states/caseList/dateFilters';
+import { DateFilter } from './dateFilters';
+
+// Define types for filter component props
+type FilterComponentBaseProps = {
+  name: string;
+  openedFilter: string | undefined;
+  setOpenedFilter: (filter: string | undefined) => void;
+};
+
+type FilterComponentConfig = FilterComponentBaseProps & {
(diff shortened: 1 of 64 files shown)


plugin-hrm-form/src/components/caseList/filters/FilterComponentsRegistry.ts
```diff
@@ -0,0 +1,175 @@
...
+import { DateFilter } from './dateFilters';
+
+type FilterComponentBaseProps = {
+  name: string;
+  openedFilter: string | undefined;
+  setOpenedFilter: (filter: string | undefined) => void;
+};
+
+type FilterComponentConfig = FilterComponentBaseProps & {
+  [key: string]: any;
```
stephenhand, line 32: This will overwrite FilterComponentBaseProps  IIRC, so name will be any and not string

The approach works for other types but I think any is special

plugin-hrm-form/src/components/caseList/filters/FilterComponentsRegistry.ts
```diff
@@ -0,0 +1,175 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+
+import React from 'react';
```
stephenhand, line 17: This feels a little bit over complicated - do we have to get a component and it's properties separately and put them together after? Seems like we could just generate fully formed components here with the properties already set

Also, the 'component registry' pattern might not be adding much here. 

It's handy for forms, where custom components could be almost anything, living anywhere in the code, and being able to register them independently form where they are used helps us keep things organised

In this case, all the components we are going to be using are different variations of filters and they are all kept with the case list components, so just having a big switch statement might work as well and be easier to follow

plugin-hrm-form/src/components/caseList/filters/FilterComponentsRegistry.ts
stephenhand: Another advantage of generating the component here is that we don't have to create a 'universal props' type, we just set the props each type of filter needs here

plugin-hrm-form/src/components/caseList/filters/dateFilters.ts
```diff
@@ -86,8 +88,11 @@ const nextXDays = (days: number): RelativeDateRange => ({
   to: referenceDate => endOfDay(addDays(referenceDate, days)),
 });
 
-const withoutDate = (): ExistsDateFilter => ({
-  titleKey: 'CaseList-Filters-DateFilterOptions-WithoutDate',
+const withoutDate = (titleKey: string): ExistsDateFilter => ({
```
stephenhand, line 91: can we not just set `titleParameters` to `{ labelKey: titleKey }` instead of putting a react element into `titleKey`? That's what `titleParameters` is intended for.

(also, just a personal opinion but I think stuffing react elements into component properties is a bit of an ugly pattern that stores up maintenance headaches for the future and we should try to avoid if possible)

mythilytm: thank you. i implemented both of these

plugin-hrm-form/src/components/caseList/filters/FilterComponentsRegistry.ts
mythilytm: i removed the registry pattern

plugin-hrm-form/src/components/caseList/filters/FilterComponentsRegistry.ts
mythilytm: not sure if there is something actionable here. Can you clarify if you are asking for changes

stephenhand: If you address the other comments you shouldn't need this type at all, just ensuring you understand how the type system treats this sort of type for work you do in the futyure

stephenhand (approved):
Thanks for making those changes, this looks good!

PR #2908, opened 2025-04-07
CHI-3291: Ensure the sidWithTaskControl transferMeta attribute is set when accepting a transfer reservation.

stephenhand:
## Description

- ensure that we set the sidWithTaskControl when accepting a reservation for a transfer if we are handling contact creation in task router
- do not set sidWithTaskControl in Flex if we are handling task creation on the backend. This is not strictly necessary but it makes redundant code easier to remove later and saves a call to Twilio
- Ensure we reload a contact if it's task ID doesn't match the task sid with it#'s contact ID, it means it's stale in redux following a transfer

### Checklist
- [x] New tests added
- [x] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


mythilytm (approved):
Looks good to me.

PR #2916, opened 2025-04-11
CHI-3286-record_taskSids_on_convos_all_channels

stephenhand:
## Description

- Change the logic for putting task sids in the channel attributes to apply to all tasks EXCEPT 'survey' tasks (i.e. chatbot control tasks)
- Add an example of how a studio flow would be updated to use the task sids in the channel attributes to prevent the initial studio flow being rerun when the conversation is already ongoing
- Unit tests for the listener to add the task sids to the channel attributes on task creation

### Checklist
- [x] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/lambdas/account-scoped/src/conversation/addTaskSidToChannelAttributesTaskRouterListener.ts
+++ b/lambdas/account-scoped/src/conversation/addTaskSidToChannelAttributesTaskRouterListener.ts
@@ -26,18 +26,17 @@ import { getChatServiceSid, getWorkspaceSid } from '../configuration/twilioConfi
 
 type RelevantTaskAttributes = {
   isContactlessTask?: boolean;
-  channelType?: string;
-  customChannelType?: string;
 };
 
 const isTaskThatNeedsToBeAddedToChannelAttributes = async (
   taskSid: TaskSID,
-  { isContactlessTask, channelType, customChannelType }: RelevantTaskAttributes,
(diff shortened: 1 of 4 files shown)


GPaoloni (approved):
Code looks good to me, however I have a question:
The reason of this bug is that the flow is triggering twice for `survey` tasks? That should not happen, since the "channel capture" process removes the studio trigger webhooks as part of the capture. Sorry to bother I'm just trying to understand exactly what the bug is here :stuck_out_tongue:

stephenhand:
> Code looks good to me, however I have a question: The reason of this bug is that the flow is triggering twice for `survey` tasks? That should not happen, since the "channel capture" process removes the studio trigger webhooks as part of the capture. Sorry to bother I'm just trying to understand exactly what the bug is here 😛

This is a scenario where an error message comes in from Lex 20 minutes after it returned a failure in the initial flow, when the chat with the counselor has already started. I think the channel capture re-added the hooks after it failed? It would have to since the MT workflow has multiple chatbots that each need to trigger the flow. Full details are in the ticket anyway

This seems like a sensible 'catch all' check to ensure workflows intended for initial routing only never get called once a task is in progress, because I think a few things cause this 'duplicate tasks' issue and I bet they are all around incorrectly run studio flows

PR #2918, opened 2025-04-14
CHI-3283 Fix to load the missing contact state fields for outbound calls

mythilytm:
## Description
- There are two different fixes:
  - outbound voice calls were getting any hrm contacts created because they were not labeled 'voice' calls with `enable_backend_hrm_contact_creation` flag on. There is now an explicit check to ensure backend contacts were being created.
  - The other fix is potentially an much older one where conversationDuration is not being saved because `recreated` flag is setting start time to `null` for calculating conversationDuration

### Checklist
- [n/a] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/lambdas/account-scoped/src/hrm/createHrmContactTaskRouterListener.ts
+++ b/lambdas/account-scoped/src/hrm/createHrmContactTaskRouterListener.ts
@@ -171,6 +171,7 @@ export const handleEvent = async (
       definitionVersion,
       ...BLANK_CONTACT.rawJson,
     },
+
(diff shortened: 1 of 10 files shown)


stephenhand:
Why was the `recreated` flag being used to stop recording conversationDuration in the first place? Feels like just removing that logic without knowing the original reasoning behind it risks recreating the original issue it was intended to fix?

mythilytm:
closing in favour of https://github.com/techmatters/flex-plugins/pull/2921

PR #2933, opened 2025-04-23
CHI-3291: Update contact owner when accepting a reservation for its associated task

stephenhand:
## Description

If a 'reservation accepted' comes in and the task already has a contactId attribute set, ensure the taskSid is set to match the task and the owning twilioWorkerId is set to match the accepting worker

This should fix issues with transfers by replicating behaviour in Flex that is bypassed when the backend HRM contact creation flag is set

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


lambdas/account-scoped/src/hrm/internalHrmRequest.ts
```diff
@@ -42,39 +43,87 @@ const requestFromInternalHrmEndpoint = async <TRequest, TResponse>(
...
+      });
+    }
+    console.warn(
+      `HRM request ${method} ${endpointUrl} attempt ${attempt + 1} failed`,
+      errorResult.message,
+      errorResult.error,
+    );
+    if (attempt + 1 < retryLimit) {
+      // eslint-disable-next-line @typescript-eslint/no-loop-func
+      await new Promise(resolve => setTimeout(resolve, 1000));
```
GPaoloni, line 78: Maybe we want some back-off here? Not requiring a change just asking if it makes sense.
```suggestion
      await new Promise(resolve => setTimeout(resolve, 1000 * attempt));
```

GPaoloni (approved):

lambdas/account-scoped/src/hrm/internalHrmRequest.ts
stephenhand: I just wanted to keep it simple and not delay the caller too long

PR #2935, opened 2025-04-24
Update oneToOneConfigSpec.json

dee-luo:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/hrm-form-definitions/form-definitions/uscr/v1/tabbedForms/CallerInformationTab.json
+++ b/hrm-form-definitions/form-definitions/uscr/v1/tabbedForms/CallerInformationTab.json
@@ -2421,67 +2421,67 @@
       },
       {
         "label": "CAO",
-        "value": "cao_office_dept"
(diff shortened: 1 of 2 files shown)


hrm-form-definitions/form-definitions/uscr/v1/tabbedForms/CallerInformationTab.json
mythilytm, line 1: Replacing internal value fields with labels (e.g., "CD1" vs "cd1_office_dept") risks duplicate values and could break search or backend logic since we reply on unique identifiers. The label is what the user sees.

 I am wondering if there is a reason for this... it might be safer to keep the original `value` unless we’ve verified are sure there are no other dependencies... or it doesn't matter since LA has only been live for less than a month.

mythilytm (approved):
Good to go

PR #2946, opened 2025-04-29
CHI-3279 Remove categories & connected contacts from typesz

stephenhand:
## Description

Remove connectedContacts, firstContact and categories from the case type. Instead, use categories from a timeline to provide this data where needed

In case lists and case search results, load a contact timeline per result as a separate, follow up operation

### Checklist
- [x] New tests added
- [ ] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### Verification steps

Regression test case lists, case search, case creation, editing, printing and adding / removing contacts

### AFTER YOU MERGE


plugin-hrm-form/src/components/search/CasePreview/index.tsx
```diff
@@ -27,66 +27,60 @@ import { getDefinitionVersion } from '../../../services/ServerlessService';
...
-  const taskContact = selectContactStateByContactId(state, contactId)?.savedContact;
-  const firstConnectedContact = selectFirstCaseContact(state, currentCase);
-  return {
-    definitionVersions: state[namespace].configuration.definitionVersions,
-    taskContact,
-    firstConnectedContact,
-    isConnectedToTaskContact: Boolean(taskContact?.caseId && taskContact.caseId === currentCase?.id),
-  };
-};
+const CONTACTS_TIMELINE_ID = 'print-contacts';
```
GPaoloni, line 51: What's this constant for? :thinking:

plugin-hrm-form/src/states/case/timeline.ts
```diff
@@ -168,5 +170,46 @@ export const selectTimeline = (
...
-  state[namespace].connectedCase.cases[caseId]?.timelines?.[timelineId]?.length;
+export const selectTimelineCount = (
+  state: RootState,
+  caseId: string,
+  timelineId: string,
+  filter: (activity: ContactIdentifierTimelineActivity | CaseSectionIdentifierTimelineActivity) => boolean = () => true,
+): number | undefined => state[namespace].connectedCase.cases[caseId]?.timelines?.[timelineId]?.filter(filter)?.length;
+
+export const selectTimelineContactCategories = (state: RootState, caseId: string, timelineId: string) => {
+  const timeline = selectTimeline(state, caseId, timelineId, { offset: 0, limit: 10000 });
```
GPaoloni, line 181: Is the `limit: 10000` intentional?

GPaoloni (approved):
LGTM

plugin-hrm-form/src/states/case/timeline.ts
stephenhand: Yes, could probably be lower, we just want all the contacts ideally

plugin-hrm-form/src/components/search/CasePreview/index.tsx
stephenhand: That is the identifier for the timeline that contains all the contacts. Each case can have multiple timelines associated with it, and each of these are to populate a different part of the UI, and can contain contacts and / or different case section types depending on how you configure the backend call. The id is used to refer back to an individual timeline

PR #2988, opened 2025-05-20
TZ - Map Insights

sinehome:
Map Insights for TZ


dee-luo (changes requested):
Looks good! Can we also add mapping Contact ID to a OneToManyConfigSpec.json file too? You can see how this looks here: https://github.com/techmatters/flex-plugins/blob/master/hrm-form-definitions/form-definitions/as/v1/insights/oneToManyConfigSpecs.json

dee-luo (approved):

PR #2990, opened 2025-05-20
Dee/tz/fixes

dee-luo:
Change ongoing case status to Open to fix Case List 
Change surname to lastName so we can populate surnames in Case List and Search


dee-luo:
@sinehome can you merge this in the morning and cut a new QA tag? Want to make sure this makes it into the deploy

sinehome (approved):

PR #3042, opened 2025-06-09
KHP Switchboarding Configs

dee-luo:
Add "Switchboard Queue" via TF
Save original target queue as part of Task Attributes in Studio

### DIFF
--- a/twilio-iac/helplines/ca/common.hcl
+++ b/twilio-iac/helplines/ca/common.hcl
@@ -98,6 +98,10 @@ locals {
         "target_workers" = "routing.skills HAS '988NSFr'",
         "friendly_name"  = "988 NS French"
       },
+      "switchboard" : {
+        "target_workers" = "routing.skills HAS 'Supervisor'",
+        "friendly_name"  = "Switchboard Queue"
+      },
       e2e_test : {
         "target_workers" = "email=='aselo-alerts+production@techmatters.org'",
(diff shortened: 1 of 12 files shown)


twilio-iac/helplines/ca/common.hcl
```diff
@@ -98,6 +98,10 @@ locals {
         "target_workers" = "routing.skills HAS '988NSFr'",
         "friendly_name"  = "988 NS French"
       },
+      "switchboard" : {
+        "target_workers" = "routing.skills HAS 'Supervisor'",
+        "friendly_name"  = "Switchboard Queue"
+      },
```
gpaoloni, line 104: Could the SID of this queue be added as an SSM parameter named like `/< env >/twilio/< accountSid >/switchboard_queue_sid)`? E.g. `/staging/twilio/ACxxxxxxxxxxxxxxxx/switchboard_queue_sid)`.
This change would allow us to rely on SSM parameters rather name queue names, which would be more reliable.

gpaoloni (approved):
Looks good to me, but would be great if @janorivera gives a look at it too.

twilio-iac/helplines/ca/common.hcl
janorivera: @gpaoloni done!

gpaoloni: Awesome Ale, thank you!

PR #3089, opened 2025-06-20
chore: move pending tasks from target to switchboard queue [CHI-3353]

gpaoloni:
This PR is related to https://github.com/techmatters/serverless/pull/774.

## Description
This PR implements "moving tasks to switchboarding queue" when switchboard is enabled.
To achieve this, we use the same idea as we use in transfers: we clone them with a different target queue. More in detail:
- When switchboarding is enabled, we look for all the `pending` tasks in the "target queue" (refer to them as original tasks).
- Update original tasks' attributes to include a new attribute `switchboardInProgress` set to `true`. This attribute is important to avoid assigning the pending tasks while they are being moved (very unlikely but potential race condition).
- Attempt to clone each of those tasks, but specifying the switchboarding queue as the routing target.
  - If successful, we then try to delete the original task that corresponds to each new task.
  - If failed, we roll back the `switchboardInProgress` update, setting the flag back to `false`. This means that the task couldn't be moved to switchboarding queue so it will be kept in the original one.

For this to work, there are four components to update:
- Serverless functions, with the above linked branch.
- Twilio lambdas, with this branch.
- Update Twilio account configs:
  -  Master Workflow expressions
    - Wrap whole expression in parenthesis.
    - Add `AND (task.switchboardInProgress != true)` at the end of it.
  - Modify Studio Flows to include a new task attribute` "switchboardInProgress": false` in all the `SendToFlex` widgets (all studio flows).
- Enable `enable_switchboarding_move_tasks` feature flag.
  All of this has to be done to avoid assigning tasks that are being moved when switchboard is enabled. This means we have to change our "default expression", and every workflow filter expression for the accounts that will use switchboard. @janorivera @dee-luo @annaliseirby090 FYI, I'm not sure where's the best place to capture this information.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-3353)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Verification steps
- Deploy serverless with the branch linked above.
- Deploy this branch to Twilio lambdas.
- Test chat contacts:
  - Make sure switchboard is OFF.
  - Start a new chat contact - do not accept it.
  - Confirm that the task is routed through the usual queue (`Admin` is development). This can be done via Twilio console, or going available and checking the "info" page of the task WITHOUT accepting it.
  - Turn switchboard ON.
  - Confirm that the task has been moved to the switchboard queue.
  - Accept it and confirm you are able to interact with the chat - i.e. chat channel has been preserved.
- Test voice contacts:
  - Make sure switchboard is OFF.
  - Start a new voice contact - do not accept it.
  - Confirm that the task is routed through the usual queue (`Admin` is development). This can be done via Twilio console, or going available and checking the "info" page of the task WITHOUT accepting it.
  - Turn switchboard ON.
  - Confirm that the task has been moved to the switchboard queue.
  - Accept it and confirm you are able to interact with the call - i.e. voice call has been preserved.

### AFTER YOU MERGE


stephenhand (approved):
Less changes than I thought would be needed!

PR #3508, opened 2025-08-21
CHI-3343: Add support for pulling 'about self' answer from service user choices beyond the hardcoded survey question. Update AS Dev configs to test

stephenhand:
### Checklist
- [ ] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


lambdas/packages/hrm-form-definitions/src/formDefinition/formSelector/selectFormsFromServiceUserChoice.ts
```diff
@@ -0,0 +1,70 @@
...
+        return ['CaseInformationTab', 'CallerInformationTab'];
+      }
+      return ['CaseInformationTab'];
+    } else {
+      if (serviceUserChoice === FormAbout.Other) {
+        return ['CaseInformationTab', 'CallerInformationTab'];
+      }
+      return ['CaseInformationTab', 'ChildInformationTab'];
+    }
+  },
```
gpaoloni, line 59: I might be wrong here, so please take this with a grain of salt, but can't we express the same logic in a simpler way like?
```suggestion
    switch (serviceUserChoice) {
      case FormAbout.Self: {
        return ['CaseInformationTab', 'ChildInformationTab'];
      }
      case FormAbout.Other: {
        return ['CaseInformationTab', 'CallerInformationTab'];
      }
      default: {
        return source === 'survey' ? ['CaseInformationTab'] : ['CaseInformationTab', 'ChildInformationTab'];
      }
    },
```

PR #3571, opened 2025-09-11
feat: update search UX to support USCH [CHI-3448]

gpaoloni:
This PR is related to https://github.com/techmatters/hrm/pull/934.

## Description
This PR
- Extends src/components/resources/mappingComponents/
  - Each account that supports resources must expose the customizable parts of the UI, depending on how they interact with the custom resource mappings.
  - `FilterSelectionState` defines behavior in different parts of the Redux resources search state:
    - `loadReferenceActionFunction`: specifies how to lookup the reference resources.
    - `handleLoadReferenceLocationsAsyncActionFulfilled`: specifies how to handle the above action.
    - `handlerUpdateSearchFormAction`: specifies how to handle the form updates upong changing search filters.
  - `ResourceSearchFilters` defines the custom component that adds filters to the search page (below the omnisearch bar).
  - `ResourcesSearchResultsDescriptionDetails` defines the custom component that informs the user of the selected filters after a search.
- As part of the above change, parts of the search resource state are refactored to support dynamic shapes, as well as dynamic ways to update it (defined as part of the mapping components, under `FilterSelectionState`).
- Refactors `SearchResourcesForm` and `SearchResourcesResults` components to use the above mentioned custom components (`ResourceSearchFilters` and `ResourcesSearchResultsDescriptionDetails` respectively).
- Refactors KHP mapping components accordingly.
- Introduces USCH mapping components.
  - Filters are only location based (`country`, `province`, `city`).
- Previously `TaxonomyLevelNameCompletion` type has been made more abstracted into `SuggestSearch`, which supports different "keys" corresponding to the suggest fields of the EalasticSearch mappings.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-3448)
- [ ] New tests added
- [ ] Feature flags added
- [x] Strings are localized

### Verification steps
- Deploy HRM and Lambdas using the above linked PR.
  - Delete resources from DB, and corresponding index from ElasticSearch.
  - Trigger a reindex using USCH resources.
- `npm run dev` to start local server.
- Confirm the resources search page and results now display the appropriate filters, and that the onmisearch suggester appropriately uses the categories field from ElasticSearch resources.
- KHP QA to make sure no bug is being introduced.

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/components/resources/mappingComponents/khpMappings/filterSelectionState.ts
+++ b/plugin-hrm-form/src/components/resources/mappingComponents/khpMappings/filterSelectionState.ts
@@ -62,7 +62,8 @@ export type KHPFilterOptions = {
   maxEligibleAge: FilterOption<number>[];
 };
 
-const initialFilterOptions: KHPFilterOptions = {
+// Exported for testing
+export const initialFilterOptions: KHPFilterOptions = {
   feeStructure: [
     { value: 'Free' },
     { value: 'Cost Unknown' },
--- a/plugin-hrm-form/src/components/resources/mappingComponents/khpMappings/filterSelectionState.ts
+++ b/plugin-hrm-form/src/components/resources/mappingComponents/khpMappings/filterSelectionState.ts
@@ -1,0 +1,210 @@
+/* eslint-disable import/no-unused-modules */
+/**
+ * Copyright (C) 2021-2023 Technology Matters
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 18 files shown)


plugin-hrm-form/src/components/resources/mappingComponents/khpMappings/filterSelectionState.ts
stephenhand, line 1: State code should go under the 'state' directory, not components

plugin-hrm-form/src/components/resources/mappingComponents/uschMappings/filterSelectionState.ts
```diff
@@ -0,0 +1,166 @@
...
+  state: ReferrableResourceSearchState,
+  { payload }: { payload: SearchSettings },
+) => {
+  const updatedFilterOptions = getFilterOptionsBasedOnSelections(state.referenceLocations);
+  const validatedFilterSelections = ensureFilterSelectionsAreValid(
+    { ...state.parameters.filterSelections, ...payload.filterSelections },
+    updatedFilterOptions,
+  );
+
+  console.log('>>>>>>>>>>> payload', payload);
```
stephenhand, line 119: Remove? Also, never use `console.log`. Always use a level specific method

stephenhand:
@gpaoloni Should the province filter options be filtered by the selected country, and the city options be filtered by the selected province? Looking at the code and the supporting HRM endpoint, we only support returning complete lists all the time.

Has this been descoped?

PR #3620, opened 2025-10-08
Move FeatureFlags type to separate file

stephenhand:
## Description

- Move FeatureFlags type to its own file
- Add comments for FF only referenced outside flex
- Remove legacy FieldText and associated RequiredAsterisk and type

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### Verification steps

Regression

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/types/FeatureFlags.ts
+++ b/plugin-hrm-form/src/types/FeatureFlags.ts
@@ -16,13 +16,10 @@
 /* eslint-disable camelcase */
 export type FeatureFlags = {
   // Please keep this in alphabetical order!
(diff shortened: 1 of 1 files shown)


stephenhand:
@annaliseirby090 @janorivera @sinehome 

This PR will move the feature flags into a single separate file, which should be considered the source of truth for feature flags currently in use. This is better than a document outside the source tree because there will be accurate up to date copies for each version we deploy, so there shouldn't be confusion about what flags are used in what version

@gpaoloni - we should move this to a shared package once we figure out how to put everything under a workspace, until then ,the commented out FFs workaround is needed

PR #3652, opened 2025-10-24
refactor: move conversationJoined subscriptions - afterAcceptTask > beforeAcceptTask [CHI-3543]

gpaoloni:
## Description
This PR is a cherry pick of the latest commit of https://github.com/techmatters/flex-plugins/pull/3644, which fixes the bug reported by USCH (https://tech-matters.atlassian.net/browse/CHI-3543).

The fix consist of moving the `conversationJoined` event subscriptions to be done during `beforeAcceptTask` event, rather than `afterAcceptTask`, to avoid a race condition on events ordering.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-3543)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/utils/setUpActions.ts
+++ b/plugin-hrm-form/src/utils/setUpActions.ts
@@ -140,6 +140,21 @@ const sendWelcomeMessageOnConversationJoined = (
   manager.conversationsClient.once('conversationJoined', (c: Conversation) => trySendWelcomeMessage(c, 0, 0));
 };
 
+export const beforeAcceptTask = (setupObject: SetupObject, getMessage: GetMessage) => async (
+  payload: ActionPayload,
+) => {
+  const { task } = payload;
+
+  if (TaskHelper.isChatBasedTask(task)) {
+    subscribeAlertOnConversationJoined(task);
+  }
+
+  // If this is the first counsellor that gets the task, say hi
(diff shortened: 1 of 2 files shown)


plugin-hrm-form/src/utils/setUpActions.ts
```diff
@@ -140,6 +140,21 @@ const sendWelcomeMessageOnConversationJoined = (
   manager.conversationsClient.once('conversationJoined', (c: Conversation) => trySendWelcomeMessage(c, 0, 0));
 };
 
+export const beforeAcceptTask = (setupObject: SetupObject, getMessage: GetMessage) => async (
```
stephenhand: You should be aware that `beforeAcceptTask` fires whether the counselor accepts the task or not. It seems like this would be unintended behaviour if the user didn't accept the task they were offered

I'm pretty sure this is the case because we've had production outages caused by this very problem before. I wouls at least very thoroughly test the scenario where a task is refused and accepted by someone else before merging this

stephenhand: Disregard, tested and confirmed it happens just before accepting

PR #3679, opened 2025-11-05
CHI-3349: Add translations for queue name

stephenhand:
## Description

- Add translations for queue name to call canvas, incoming task canvas and task info. 
- Translate queue name on contacts waiting and tabbed form
- Change queue name translation not to include colon
- Remove duplicated queue name translations

### Checklist
- [n/a] New tests added
- [n/a] Feature flags added
- [x] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.tsx
+++ b/plugin-hrm-form/src/HrmFormPlugin.tsx
@@ -91,7 +91,7 @@ const setUpComponents = (featureFlags: FeatureFlags, setupObject: ReturnType<typ
   Components.setUpAddButtons(featureFlags);
   Components.setUpNoTasksUI(featureFlags, setupObject);
   Components.setUpCustomCRMContainer();
-  Components.setUpExtraTranslations(setupObject);
+  Components.setUpExtraTranslations();
 
   // set up default and custom channels
   Channels.setupDefaultChannels();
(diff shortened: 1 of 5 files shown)


plugin-hrm-form/src/HrmFormPlugin.tsx
```diff
@@ -48,6 +48,7 @@ import { validateAndSetPermissionRules } from './permissions/rules';
 import { setupLlmNotifications } from './components/contact/GenerateSummaryButton/setUpLlmNotifications';
 import { FeatureFlags } from './types/FeatureFlags';
 import { setUpFullStory } from './fullStory/setUp';
+import { setUpExtraTranslations } from './utils/setUpComponents';
```
gpaoloni: Why having an "extra translations" rather than doing all in one place? Even if the setup translation function calls another function, I think it's easier if all happens at the same place?

stephenhand: I see `setUpLocalization` as setting up the localisation system itself, rather than adding some extra labels to the UI, which I thought belonged as part of the 'setUpComponents' phase with the rest of the UI

gpaoloni (approved):

PR #3724, opened 2025-11-28
Chi 3523 aselo webchat poc backend

stephenhand:
## Description

Follows on from #3726 by implementing our own auth backend in the account scoped lambda.

It adds an auth mechanism for endpoints restricted to 'guest' webchat users, and allows routes using the helpline short code (because new webchat hides account SIDs from users)

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/aselo-webchat-react-app/tsconfig.json
+++ b/aselo-webchat-react-app/tsconfig.json
@@ -10,7 +10,8 @@
     "esModuleInterop": true,
     "lib": [
       "dom",
-      "es2015"
+      "es2015",
+      "es2020"
     ],
     "jsx": "react-jsx",
     "target": "es5",


lambdas/account-scoped/src/Result.ts
```diff
@@ -55,6 +55,20 @@ export const newErr = <TError>({
   };
 };
 
+export const newJsErrorResult = <TError extends Error>(
```
gpaoloni: Isn't this the same as `newErr({ message: error.message, error })`?

lambdas/account-scoped/src/router.ts
```diff
@@ -105,6 +110,14 @@ const ROUTES: Record<string, FunctionRoute> = {
...
     handler: flexToInstagramHandler,
   },
+  'webchatAuth/initWebchat': {
+    requestPipeline: [],
+    handler: initWebchatHandler,
+  },
+  'webchatAuth/refreshToken': {
+    requestPipeline: [],
+    handler: refreshTokenHandler,
+  },
```
gpaoloni: Are these needed in both. here and in `ENV_SHORTCODE_ROUTES`?

PR #3761, opened 2025-12-19
CHI-3575: Enhance 'Save and End' button for existing contacts

stephenhand:
## Description

- Adds a 'cancel / remove task' endpoint, which will try to cancel a task, then failing that, try to cancel the reservations, then failing that, remove the task
- Enhances logic for completing tasks to try completing reservations if completing the overall task fails
- Enhances logic for completing tasks to explicitly call the channel janitor (it might get called by task router event, but this thing gets used in a broken state, so no harm to try explicitly anyway)
- Allows non supervisor workers who have the task assigned access to the task assignment endpoints (but only their own reservations)
- Updates UI to allow the user to try to remove the task if completing it fails

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/lambdas/account-scoped/src/task/cancelOrRemoveTask.ts
+++ b/lambdas/account-scoped/src/task/cancelOrRemoveTask.ts
@@ -40,12 +40,10 @@ export const cancelOrRemoveTaskHandler: FlexValidatedHandler = async (
   const byLogClause = `by ${isSupervisor(tokenResult) ? 'supervisor' : 'agent'} ${tokenResult.worker_sid}`;
   const lookupResult = await getTaskAndReservations(accountSid, taskSid, tokenResult);
   if (isErr(lookupResult)) {
-    if (isErr(lookupResult)) {
-      return newHttpErrorResult(
-        lookupResult.error.cause,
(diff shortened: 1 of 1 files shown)


lambdas/account-scoped/src/task/cancelOrRemoveTask.ts
```diff
@@ -0,0 +1,106 @@
...
+      );
+    } else {
+      if (cancelReservationsResult.data.updatedReservationSids?.length) {
+        console.info(
+          `Cancelled reservations for task ${taskSid}: ${cancelReservationsResult.data.updatedReservationSids}.`,
+        );
+      }
+      cancelReservationsResult.data.updatedReservationSids.map(cre =>
+        console.warn(`Error cancelling reservation ${byLogClause}:`, cre),
+      );
```
gpaoloni: I think I'm missing something here. If this is part of the `else` statement after, so "NOT `isErr(cancelReservationsResult)`", how can we have reservations that errored being cancelled?

Also out of ignorance, why are we cancelling the reservations if we are about the remove the task anyways?

gpaoloni (approved):
LGTM

PR #4177, opened 2026-04-13
fix: expandable component leading to excessive set state calls [CHI-3817]

gpaoloni:
## Description
This PR fixes the bug linked below, where search page would result in an unhandled error. The bug was introduced in https://github.com/techmatters/flex-plugins/pull/4143 and it caused because the "trigger compute overflow" function was invoked too many times, causing excessive set state function calls.

To fix the issue:
- `useLayoutEffect` is removed from `plugin-hrm-form/src/components/ExpandableBlocks/ExpandableText.tsx`. This fixes the bug introduced in the above linked PR.
- `MutationObserver`s are set for components using the `useIsOverflow` hook, if they are a descendant of a component that is "hidden" (`display: none` CSS attribute) when it is mounted. The observers trigger a compute overflow when the hidden ancestors become visible, fixing the bug attempted in the above linked PR.
- `useExpandableOnOverflow` triggers a compute overflow when the component is collapsed, to fix a potential bug that occurs when a component is hidden after it was expanded, but never unmounted (e.g. tabbing in contact forms).
- Some minor UI adjustments for case and contact preview (search results) to make the rest of the component relying on above hooks behave as expected.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-3817)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Verification steps
Confirm that the behavior is the expected in the following components:
- Field descriptions.
- Search results
  - Case preview when summary text is too long to fit the card.
  - Contact preview when summary text is too long to fit the card.
  - Case tags when they are too long to fit the card.

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/hooks/useIsOverflowing.ts
+++ b/plugin-hrm-form/src/hooks/useIsOverflowing.ts
@@ -20,7 +20,8 @@ import React from 'react';
 // Walks up the DOM tree and returns all ancestors that have display:none
 const getHiddenAncestors = (node: Element): Element[] => {
   const hiddenAncestors: Element[] = [];
-  let current = node.parentElement;
+
+  let current = node?.parentElement;
 
   while (current) {
     if (getComputedStyle(current).display === 'none') {
--- a/plugin-hrm-form/src/hooks/useIsOverflowing.ts
+++ b/plugin-hrm-form/src/hooks/useIsOverflowing.ts
@@ -15,11 +15,31 @@
  */
 
 // https://www.robinwieruch.de/react-custom-hook-check-if-overflow/
-import { debounce } from 'lodash';
 import React from 'react';
 
+// Walks up the DOM tree and returns all ancestors that have display:none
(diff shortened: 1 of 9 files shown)


plugin-hrm-form/src/hooks/useIsOverflowing.ts
```diff
@@ -37,9 +58,46 @@ export const useIsOverflowing = ({ ref, callback }: { ref: any; callback?: (isOv
     }
   }, [callback, ref]);
 
+  const cleanupMutationObservers = React.useCallback(() => {
+    mutationObserversRef.current.forEach(obs => obs.disconnect());
+    mutationObserversRef.current = [];
+  }, []);
+
+  const setupObservers = React.useCallback(() => {
```
stephenhand, line 66: I can see this got a bit hairy :-/

stephenhand (approved):
I think I see what you did, thanks for some of the helpful comments

PR #4501, opened 2026-07-01
CHI-3708: add recordingCompleteCallback scafold

gpaoloni:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


lambdas/account-scoped/src/hrm/createHrmContactTaskRouterListener.ts
```diff
@@ -245,6 +248,44 @@ export const handleEvent = async (
   const savedTimeOfContactDate = parseISO(savedTimeOfContactString);
   console.info(`Created HRM contact with id ${id} for task ${taskSid}`);
 
+  if (channel === ('voicemail' as any)) {
```
stephenhand, line 251: Should we attempt to consolidate setting recordings on the backend instead of having this as a special case for the productionising step?

PR #4547, opened 2026-07-22
CHI-3711: POC for voice and sms E2E tests - CI only

stephenhand:
## Description

This creates a POC for testing SMS and Voice calls with E2E tests

- They are only tested as CI tests running locally, not for production in lambdas
- There's probably a few hardcoded bits that need tidying up when productionising
- We send all the test tasks to the same queue - webchat, SMS and voice, is this an issue?
- We need a task to clean up conversations for the SMS tests to run after a failure. Should this be necessary?

### Checklist
- [x] New tests added
- [n/a] Feature flags added
- [n/a] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


stephenhand:
@gpaoloni ?

e2e-tests/twilio/channels.ts
```diff
@@ -30,50 +30,103 @@ const encodeEmailToUnicode = (email: string) => {
...
+    serviceSmsNumber,
+  );
+
+  // Delete conversations from sender Twilio account
+  await deleteSmsConversationFromOneEnd(
+    senderAccountSid,
+    senderAuthToken,
+    serviceSmsNumber,
+    senderSmsNumber,
+  );
```
gpaoloni, line 89: About this, I understand the "client account" is the one we use to send SMS messages. I see that is defined via config but in practice what account will be used for this?

gpaoloni (approved):
LGTM 🚀 
Do we have documentation on how to setup/configure the e2e tests for accounts somewhere? If so, could that be linked somewhere in the 2e README?

e2e-tests/twilio/channels.ts
stephenhand: I'm just using AS_DEV for now

PR #4549, opened 2026-07-23
Alejandro/usnm/migrating function to lambda

janorivera:
## Description

@stephenhand here is the migrated function! 
I've already tested it and it works! 
Let me know if you would like to change anything. 

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


lambdas/account-scoped/src/worker/isSkilledWorkerAvailable.ts
```diff
@@ -0,0 +1,284 @@
...
+ * along with this program.  If not, see https://www.gnu.org/licenses/.
+ */
+
+import { AccountScopedHandler } from '../httpTypes';
+import { newOk } from '../Result';
+import { newHttpErrorResult, newMissingParameterResult } from '../httpErrors';
+
+import { getTwilioClient, getWorkspaceSid } from '@tech-matters/twilio-configuration';
+
+interface WorkerAttributes {
```
stephenhand: We use type aliases rather than interfaces as a coding convention. I think they are practically the same in later versions of TS but we prefer a consistent convention to make onbiarding new engineers, text searches etc. easier
```suggestion
type WorkerAttributes = {
```

lambdas/account-scoped/src/worker/isSkilledWorkerAvailable.ts
```diff
@@ -0,0 +1,284 @@
...
+   * We iterate only once over the workers instead of using
+   * map().filter().filter(), which creates intermediate arrays.
+   */
+  const skilledWorkers: Array<{
+    worker: (typeof availableWorkers)[number];
+    attributes: WorkerAttributes;
+  }> = [];
+
+  for (const worker of availableWorkers) {
+    const attributes = parseWorkerAttributes(worker.attributes);
```
stephenhand: Since Twilio won't let you save worker attributes with invalid JSON i don't think we need defensive checks here. A straightforward `JSON.parse(...)` and assuming it works would more readable without sacrificing robustness

lambdas/account-scoped/src/worker/isSkilledWorkerAvailable.ts
```diff
@@ -0,0 +1,284 @@
...
+    targetChannel,
+    requiredSkills,
+    totalAvailableWorkers: availableWorkersForChannel.length,
+    availableWorkerSids,
+  });
+
+  return newOk({
+    totalAvailableWorkers: availableWorkersForChannel.length,
+    availableWorkerSids,
+    availableSkilledWorkersData: availableWorkersForChannel,
```
stephenhand: Very nitpicky but I don't like the term 'data' in names except for rare exceptions. Everything is data!
```suggestion
    availableSkilledWorkers: availableWorkersForChannel,
```

stephenhand (approved):
A few minor naming and coding convention nits

PR #4587, opened 2026-08-05
Add ability to remove message file attachments

janorivera:
## Description

NAMI requested this and they shared the line we needed to add so I took a stab a it. 
The attachment button is there by default to in order to make the ff to remain as an "enable" I had to do the if(!feature_flag)

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/plugin-hrm-form/src/HrmFormPlugin.tsx
+++ b/plugin-hrm-form/src/HrmFormPlugin.tsx
@@ -117,6 +117,7 @@ const setUpComponents = (featureFlags: FeatureFlags, setupObject: ReturnType<typ
   setUpReferrableResources();
 
   if (featureFlags.enable_emoji_picker) Components.setupEmojiPicker();
+  if (!featureFlags.enable_message_file_attachments) Components.removeMessageFileAttachments();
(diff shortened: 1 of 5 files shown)


stephenhand (approved):
LGTM. One minor nit, could you update the config property name to something that makes it clear the flag controls file attaching in flex rather for service users. Like `allowOutgoingAttachments` or `allowFlexToSendAttachments`

Or you might consider making the flag negative. `preventSendingAttachmentsFromFlex`. This has 2 benefits. One, it indicates more clearly this is a standard flex feature we are turning off, rather than one of our features we are turning on. Two, it means that current behaviour will be preserved with no config being applied to a helpline. If I understand correctly, as it currently stands we would have to apply configuration to every helpline before deploying this code to prevent attachments being disabled for them

gpaoloni (approved):
🚀 

(agree with Steve's comment on the naming)

janorivera:
thanks guys! I made the updates so I'll merge it now

PR #4604, opened 2026-08-12
Post Survey: Adding task attributes to new studio flow

janorivera:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/lambdas/account-scoped/src/studioFlow/postStudioFlowTaskRouterListener.ts
+++ b/lambdas/account-scoped/src/studioFlow/postStudioFlowTaskRouterListener.ts
@@ -146,6 +146,7 @@ const triggerPostStudioFlow = async ({
               contactId,
               contactTaskSid: taskSid,
               taskQueueSid,
+              taskAttributes,
             },
             to: taskAttributes.from,
           });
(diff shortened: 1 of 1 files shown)


janorivera:
It works!
{41ACB5C0-CA97-45CA-A0B0-60468EAD8630}

lambdas/account-scoped/src/studioFlow/postStudioFlowTaskRouterListener.ts
```diff
@@ -146,6 +146,7 @@ const triggerPostStudioFlow = async ({
               contactId,
               contactTaskSid: taskSid,
               taskQueueSid,
+              taskAttributes,
```
gpaoloni: Be aware that this will include all task attributes into the post survey workflow call. Is there something that could be sensible that we should not include? Probably not, just making sure you checked that this stuff is not carried around forever (e.g. the task could have insights-ready attributes that we might not want to go in there?)

gpaoloni (approved):

lambdas/account-scoped/src/studioFlow/postStudioFlowTaskRouterListener.ts
janorivera: good point, I had an idea that the more data we sent, the more we can use in the future to change the behavior of the studio flow. Maybe in the future we can add a postSurveyVars or something like that if needed, but maybe that something that we'll need to do if the need actually arises.
OK I've changed to the to now sent only what is required and nothing more so in this case is the taskLanguage.
`taskLanguage: taskAttributes.language ?? '',`

PR #4611, opened 2026-08-13
NZBA Insights Mapping

sinehome:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


dee-luo:
@sinehome a couple things: 

- I replaced customer attributes with conversation attributes, as conversation attributes are generally more reliable than customer attributes so I wanted to exhaust those first (we've had issues in the past with customer attributes being overwritten). I believe all the attributes I picked are customizable but we may need to replace them with customer attributes if they don't work. 
- There was one case field ("Further Action Needed") that I took out because it's a free-text field, and we don't want to risk PII getting sent, which is our general rule of thumb for Insights. 
- I noticed the post-survey file is still empty, is that intentional? Noting that conversation measures sadly don't work for post-survey, we may need to use customer attributes here.

dee-luo (approved):

PR #4615, opened 2026-08-13
Chatbot bug fix

janorivera:
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized
- [ ] 

This is a pretty small bug fix for the post survey chatbot when the user responded with an unrecognized response (slot value)

Here is an example where instead of Si  or No, I've responded "blah"
{DD296EEB-22CE-479D-9E60-467D260FF0C6}

Basically the bot broke with error:
`026-08-13T18:38:57.402Z 170e032e-ec04-4d6c-a28d-cc5cd58037fb ERROR handler for path /lambda/twilio/account-scoped/AC5c77a3f308fe1361bc12b3a48b98f68d/channelCapture/chatbotCallback resulted in error Cannot read properties of null (reading 'value') { statusCode: 500, cause: TypeError: Cannot read properties of null (reading 'value') at /var/task/channelCapture/lexClient.js:109:55 at Array.reduce (<anonymous>) at Object.getBotMemory (/var/task/channelCapture/lexClient.js:109:34) at Object.handleChatbotCallback [as handler] (/var/task/channelCapture/chatbotCallback.js:127:51) at process.processTicksAndRejections (node:internal/process/task_queues:103:5) at async Runtime.handler (/var/task/index.js:70:28)`

The problem was this line (accum, [q, { value }]) => ({ ...accum, [q]: value?.interpretedValue || '' }), which was using {value} but in this case was null.

So I've changed it to the below (using AI help)
(accum, [q, slot]) => ({
      ...accum,
      [q]: slot?.value?.interpretedValue || '',
    }),

ChatGPT also suggested processing the messages first and then call the cleanup end point which I thought it was a good suggestion so I've added that as well.

I've tried this out and now it works:
{B3D1AFD8-2E1D-4DAA-90EE-9F5F2E3D6D19}

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/lambdas/account-scoped/src/channelCapture/chatbotCallback.ts
+++ b/lambdas/account-scoped/src/channelCapture/chatbotCallback.ts
@@ -150,18 +150,6 @@ export const handleChatbotCallback: AccountScopedHandler = async (
       const twilioWorkspaceSid = await getWorkspaceSid(accountSid);
 
       const { lexResponse } = lexResult.data;
-      // If the session ended, we should unlock the channel to continue the Studio Flow
-      if (LexClient.isEndOfDialog(lexResponse)) {
-        await chatbotCallbackCleanup({
-          accountSid,
-          twilioClient,
-          conversation,
-          channel,
-          channelAttributes,
-          memory: LexClient.getBotMemory({ lexResponse }),
-          twilioWorkspaceSid,
-        });
-      }
 
       // TODO: unify with functions/channelCapture/channelCaptureHandlers.ts
       const messages = (lexResponse.messages || []).map(m => m.content || '');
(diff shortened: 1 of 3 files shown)


lambdas/account-scoped/src/channelCapture/chatbotCallback.ts
```diff
@@ -185,6 +173,19 @@ export const handleChatbotCallback: AccountScopedHandler = async (
         }
       }
 
+      // If the session ended, we should unlock the channel to continue the Studio Flow
```
gpaoloni: Is this change part of the fix? Can't we keep it where it was if no?

gpaoloni (approved):
Thanks Ale 🚀

lambdas/account-scoped/src/channelCapture/chatbotCallback.ts
janorivera: sure sure I've rolled it back!

PR #4624, opened 2026-08-19
Adding blocking reason

janorivera:
## Description

I've added the blockReason, since I need to know the reason in order to play different recordings/messages
### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/lambdas/account-scoped/src/voice/filterCountryOrVoIP.ts
+++ b/lambdas/account-scoped/src/voice/filterCountryOrVoIP.ts
@@ -19,7 +19,13 @@ import { getTwilioClient } from '@tech-matters/twilio-configuration';
 import { newErr, newOk } from '../Result';
 import type { AccountScopedHandler, HttpRequest } from '../httpTypes';
 
-const BLOCKED_CARRIERS = ['HD Carrier LLC'];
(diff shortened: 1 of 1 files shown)


lambdas/account-scoped/src/voice/filterCountryOrVoIP.ts
```diff
@@ -19,7 +19,13 @@ import { getTwilioClient } from '@tech-matters/twilio-configuration';
 import { newErr, newOk } from '../Result';
 import type { AccountScopedHandler, HttpRequest } from '../httpTypes';
 
-const BLOCKED_CARRIERS = ['HD Carrier LLC'];
+const BLOCKED_CARRIERS = ['HD Carrier LLC', 'Onvoy Spectrum', 'Onvoy Spectrum, LLC'];
```
stephenhand, line 22: We should make this a parameter we pass in from the Studio Flow in future. Not for this PR but in future

stephenhand (approved):
LGTM!

lambdas/account-scoped/src/voice/filterCountryOrVoIP.ts
janorivera: aaah you are right, so it works for other helplines as well.

PR #4627, opened 2026-08-20
CHI-3945: UI update for voicemails, translation key name change, couple of minor fixes

stephenhand:
## Description

- Update Voicemail UI as specified in ticket
- Rename a translation key to fit with our 3 part convention
- Fix voicemail channel type display
- Add fallbacks to lookupTranslation function
- Fix issue that causes startup console error in ContactAddedToCaseBanner

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE


stephenhand:
@copilot - add unit test cases for scenarios when multiple keys are specified in the lookupTranslation method

plugin-hrm-form/src/components/callTypeButtons/CallTypeButtons.tsx
```diff
@@ -135,6 +139,16 @@ const CallTypeButtons: React.FC<Props> = ({ task, localization }) => {
...
 
+  let taskType: 'default' | 'call' | 'chat' = 'default';
+
+  if (!isOfflineContactTask(task)) {
+    if (isCallTask(task)) {
+      taskType = 'call';
+    } else if (isChatChannel(task.attributes.customChannelType || task.attributes.channelType || task.channelType)) {
+      taskType = 'chat';
+    }
+  }
```
gpaoloni, line 150: Won't this result in `voicemail` falling in the `chat` `taskType`?

plugin-hrm-form/src/translations/index.ts
```diff
@@ -120,7 +120,13 @@ export const initLocalization = (localizationConfig: LocalizationConfig, helplin
...
+export const lookupTranslation = (code: string | string[], parameters: Record<string, string> = {}): string => {
   const strings = getTemplateStrings();
-  return Handlebars.compile(strings[code] ?? code)(parameters);
+  const codes = typeof code === 'string' ? [code] : code;
+  for (const code of codes) {
+    if (strings[code]) {
+      return Handlebars.compile(strings[code])(parameters);
+    }
+  }
+  return codes[0];
```
gpaoloni, line 131: This change returns the first `code` that exists in `strings`, right? The idea is to pass an array of codes then? How would that be used?

plugin-hrm-form/src/voicemail/VoicemailTaskPanel.tsx
```diff
@@ -17,51 +17,120 @@
...
+
 const VoicemailTaskPanel: React.FC<Props> = ({ task }) => {
   const contact = useSelector((state: RootState) => selectContactByTaskSid(state, task.taskSid));
 
   if (!task || !contact?.savedContact) {
     return null;
   }
 
+  const { receivedTime, from, maxCallbackAttempts } = task.attributes;
+  const callbackAttempts = task.attributes.callbackAttempts ?? [];
```
gpaoloni, line 48: I was not using this `task.attributes.callbackAttempts` in my PRs chain, so is something I will have to fix once this is merged. I'm happy to do so, just so you know that things won't immediately after PRs are merged.
What should we record in the `task.attributes.callbackAttempts` by the way? I mean what propery, a task sid?

gpaoloni (approved):
Looks good to me! 🚀

plugin-hrm-form/src/voicemail/VoicemailTaskPanel.tsx
stephenhand: I only look for the timestamp, but a task SID would be good too, yes

plugin-hrm-form/src/components/callTypeButtons/CallTypeButtons.tsx
stephenhand: Only if 'voicemail' returns true for `isChatChannel` - which it shouldn't?

plugin-hrm-form/src/translations/index.ts
stephenhand: It means you can pass a fallback key if the first one doesn't exist. It's for dynamic key cases mostly

For example, you could have a generic message for a task list item

"TaskListItem-Voicemail-Secondline"

but then want to override it for a specific status

"TaskListItem-Voicemail-Secondline/pending"

but you don't want to override all the statuses. In this case, you could set the 2 keys above, and then for the translation lookup provide [\`TaskListItem-Voicemail-Secondline/${task.status}\`, "TaskListItem-Voicemail-Secondline"]  as the key

Then all the other statuses would use "TaskListItem-Voicemail-Secondline" whereas pending would use "TaskListItem-Voicemail-Secondline/pending". You could even only override certain statuses for specific helplines if you want

It's demonstrated how it works in the unit tests but sorry, yes should have talked you through it

plugin-hrm-form/src/voicemail/VoicemailTaskPanel.tsx
stephenhand: I was planning to update your PR after you merged it but if you could that would be great

PR #4634, opened 2026-08-25
fix: voicemail - race condition between contact creation and media attached [CHI-3992]

gpaoloni:
## Description
This PR fixes a bug that causes the voicemail audio fetch to fail.
The reason for this bug is that there is a race condition between the contact creation and the conversation media attachement in the backend.
Behavior before this PR:
- Contact is created.
- Task is updated with the contact id.
- Conversation media is attached.
2nd step causes the client to "fetch the contact from HRM". If the conversation media attachement is still in progress by the time the contact is loaded by Flex, the user can't access the voicemail recording.

The fix is to swap steps 2 and 3 so that conversation media is always attached before the task is updated. 

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-3992)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/lambdas/account-scoped/src/hrm/createHrmContactTaskRouterListener.ts
+++ b/lambdas/account-scoped/src/hrm/createHrmContactTaskRouterListener.ts
@@ -28,7 +28,7 @@ import {
   postToInternalHrmEndpoint,
 } from './internalHrmRequest';
 import { isErr, isOk } from '@tech-matters/result-type';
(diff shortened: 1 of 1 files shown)


stephenhand (approved):
Good catch

PR #4647, opened 2026-08-28
filterCountryOrVoIP : adding hidden caller id

janorivera:
## Description

It seems that in the US, people can hide their phone numbers by using *67

The  object that we receive in those cases starts like this:
```
"Called": "+18444271995",
      "ToState": "",
      "CallerCountry": "TH",
      "Direction": "inbound",
      "CallerState": "",
      "ToZip": "",
      "CallSid": "CA6c1bf2457532aadeac93c4617da9fca4",
      "To": "+18444271995",
      "CallerZip": "",
      "ToCountry": "US",
      "StirVerstat": "TN-Validation-Passed-A",
      "CallToken": "%7B%22parentCallInfoToken%22%3A%22eyJhbGciOiJFUzI1NiJ9.eyJjYWxsU2lkIjoiQ0E2YzFiZjI0NTc1MzJhYWRlYWM5M2M0NjE3ZGE5ZmNhNCIsImZyb20iOiJBbm9ueW1vdXMiLCJ0byI6IisxODQ0NDI3MTk5NSIsImlhdCI6IjE3ODc2ODQwODIifQ.zc-DIzA9bq7D7_iSkgrnViz1ENnJoQ96d6qA9lz_xmnqOjj8rT__v68vcbV44QCRFqhX_L4jjl9QkAQvZ56WcQ%22%2C%22identityHeaderTokens%22%3A%5B%5D%7D",
      "CalledZip": "",
      "ApiVersion": "2010-04-01",
      "CalledCity": "",
      "CallStatus": "ringing",
      "From": "Anonymous",
```

Previously our function filterCountryOrVoIP was detecting that as coming from another country ("TH") which is not actually true, so we want to handle this type of cases differently. 
So I've created a new block reason "HIDDEN_CALLER_ID" and the function now checks for that first, if "from" is a valid e164 number (instead of Anonymous, Restricted, etc).

### Checklist
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/lambdas/account-scoped/src/voice/filterCountryOrVoIP.ts
+++ b/lambdas/account-scoped/src/voice/filterCountryOrVoIP.ts
@@ -25,6 +25,7 @@ const BLOCK_REASONS = {
   NON_US_COUNTRY: 'non_us_country',
   VOIP: 'voip',
   BLOCKED_CARRIER: 'blocked_carrier',
+  HIDDEN_CALLER_ID: 'hidden_caller_id',
 } as const;
 
 export const filterCountryOrVoIPHandler: AccountScopedHandler = async (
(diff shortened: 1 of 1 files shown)


gpaoloni (approved):
Looks good Ale, thanks for handling it! 🚀

PR #4655, opened 2026-09-01
chore: implement twilio-external-task Lambda [CHI-3955]

gpaoloni:
This PR implements the Lambda introduced by https://github.com/techmatters/infrastructure-config/pull/588.

## Description
This PR implements twilio-external-task:
- The implementation follows the same idea as account-scoped Lambda, but instead of account sid it relies on account short code.
- Auth is done via static API key.
- If valid, creates a new `external_task` channel type task in the target Twilio account.

### Example request:
```
❯ curl -X POST 'https://hrm-development.tl.techmatters.org/lambda/twilio/external-task/as' -H "Authorization: Basic < external_task_api_key >" -H "Content-Type:
application/json" -H "Accept: application/json" -d '{"key": "test-external-id" }'
```
where `external_task_api_key` is the SSM parameter that corresponds to the AS development account sid.

### Configuration
To test the integration, other than applying the linked PR in infra and deploying this code
- Add SSM `/<environment>/twilio/<account_sid>/external_task_api_key` parameter. Make sure is only number (lower and upper case) and numbers. E.g. `/development/twilio/ACd8a2e89748318adf6ddff7df6948deaf/external_task_api_key`
- In Twilio
  - Create `external` channel type.
  - Edit workflow expressions to match external tasks.
  - [OPTIONAL] Add custom task queue.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-3955)
- [ ] New tests added
- [ ] Feature flags added
- [ ] Strings are localized

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/lambdas/external-task/src/requestValidator.ts
+++ b/lambdas/external-task/src/requestValidator.ts
@@ -92,9 +92,15 @@ export const authenticateWithExternalApiKey = async ({
   const externalTaskApiKeyParam = `/${process.env.NODE_ENV}/twilio/${accountSid}/external_task_api_key`;
 
   try {
-    const requestSecret = authorization.replace('Basic ', '');
     console.debug(`Authenticating against key ${externalTaskApiKeyParam} `);
+    const requestSecret = authorization.replace('Basic ', '');
     const externalTaskApiKey = await getSsmParameter(externalTaskApiKeyParam);
+
+    // TODO: delete!!!
+    console.debug(
+      `requestSecret ${requestSecret}`,
+      `externalTaskApiKey ${externalTaskApiKey}`,
+    );
 
     const isStaticSecretValid =
       externalTaskApiKey &&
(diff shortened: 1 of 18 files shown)


lambdas/external-task/src/requestValidator.ts
```diff
@@ -0,0 +1,113 @@
...
+    return result;
+  }
+
+  const { accountSid } = result.data;
+  const externalTaskApiKeyParam = `/${process.env.NODE_ENV}/twilio/${accountSid}/external_task_api_key`;
+
+  try {
+    console.debug(`Authenticating against key ${externalTaskApiKeyParam} `);
+    const requestSecret = authorization.replace('Basic ', '');
+    const externalTaskApiKey = await getSsmParameter(externalTaskApiKeyParam);
```
stephenhand: Might be better to move this call out of the try / catch block so we don't get misleading 403s for a configuration issue?

stephenhand (approved):

